PR #10, opened 2021-05-26
Limit tree size for DC tiles

eustas:
Image "size" for DC tile is 2048x2048. If there is e.g. 13 channels,
then tree is allowed to grow to 50M nodes. But when tree reaches 32M nodes,
for resizing it would temporarlily require ~4GiB...

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -430,8 +430,25 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   const ANSCode *code = &code_storage;
   const std::vector<uint8_t> *context_map = &context_map_storage;
   if (!header.use_global_tree) {
-    size_t tree_size_limit = 1024 + image.w * image.h * nb_channels;
-    JXL_RETURN_IF_ERROR(DecodeTree(br, &tree_storage, tree_size_limit));
+    size_t max_tree_size = 1024;
+    for (size_t i = 0; i < nb_channels; i++) {
+      Channel &channel = image.channel[i];
+      if (!channel.w || !channel.h) {
+        continue;  // skip empty channels
+      }
+      if (i >= image.nb_meta_channels && (channel.w > options->max_chan_size ||
+                                          channel.h > options->max_chan_size)) {
+        break;
+      }
+      size_t pixels = channel.w * channel.h;
+      if (pixels / channel.w != channel.h) {
+        return JXL_FAILURE("Tree size overflow");
+      }
+      max_tree_size += pixels;
+      if (max_tree_size < pixels) return JXL_FAILURE("Tree size overflow");
+    }
+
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/encoding/encoding.cc
```diff
@@ -403,7 +403,8 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   const ANSCode *code = &code_storage;
   const std::vector<uint8_t> *context_map = &context_map_storage;
   if (!header.use_global_tree) {
-    size_t tree_size_limit = 1024 + image.w * image.h * nb_channels;
+    size_t tree_size_limit = std::min(options->tree_size_limit,
+                                      1024 + image.w * image.h * nb_channels);
```
jonsneyers, line 407: Instead of using that formula here, it would be better to do it based on total number of samples that this local tree will cover. In the squeeze case, there can be a large number of small channels, so `image.w * image.h * nb_channels` is a big overestimation of the actual number of samples since `image.w`/`image.h` refer to the unsqueezed image dimensions while `nb_channels` is taken after doing the meta-transforms.

jonsneyers (changes requested):
The bound on the local tree size is too loose, better to tighten it.

Also I think you need to update the email address attached to the commit (`git commit --amend --reset-author`).

eustas:
Yup, going to tighten the limit soon.

lib/jxl/modular/encoding/encoding.cc
eustas: Pulled calculation closer to the place where it is used (after MetaApply). Now it corresponds to planes that are passed to `DecodeModularChannelMAANS`.

eustas:
@jonsneyers friendly ping

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -444,8 +444,25 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   const ANSCode *code = &code_storage;
   const std::vector<uint8_t> *context_map = &context_map_storage;
   if (!header.use_global_tree) {
-    size_t tree_size_limit = 1024 + image.w * image.h * nb_channels;
-    JXL_RETURN_IF_ERROR(DecodeTree(br, &tree_storage, tree_size_limit));
+    size_t total_pixels = 1024;
```
jonsneyers, line 447: Maybe nicer to call this variable `max_tree_size`? It might be slightly confusing this way.

jonsneyers (dismissed):

lib/jxl/modular/encoding/encoding.cc
eustas: Sure, one sec.

eustas:
@deymo @jonsneyers please rubber-stamp. approval is outdated after variable rename.

jonsneyers (approved):

PR #11, opened 2021-05-26
Fuzz: accessing undefined values

eustas:
If group_rect gets squeezed to 0, it causes the situation when non-empty channel
is not decoded.

Validate channel configuration after MetaApply.

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -358,6 +358,30 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
 
 GroupHeader::GroupHeader() { Bundle::Init(this); }
 
+Status ValidateChannelDimensions(const Image &image,
+                                 const ModularOptions &options) {
+  size_t nb_channels = image.channel.size();
+  for (bool is_dc : {true, false}) {
+    size_t group_dim = options.group_dim * (is_dc ? kBlockDim : 1);
+    size_t c = image.nb_meta_channels;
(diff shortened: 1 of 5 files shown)


lib/jxl/modular/encoding/encoding.cc
```diff
@@ -358,6 +358,30 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
 
+Status ValidateChannelDimensions(const Image &image,
+                                 const ModularOptions &options) {
+  size_t nb_channels = image.channel.size();
+  for (bool is_dc : {true, false}) {
+    size_t group_dim = options.group_dim * (is_dc ? kBlockDim : 1);
+    size_t c = image.nb_meta_channels;
+    for (; c < nb_channels; c++) {
+      const Channel& ch = image.channel[c];
+      if (ch.w > group_dim || ch.h > group_dim) break;
```
jonsneyers, line 369: This `group_dim` should be `options.group_dim`, also in the DC case, unless I'm missing something.

eustas: Right! Fixed.

jonsneyers (approved):

jonsneyers:
@deymo can you take a look? This one seems to be somehow stuck in the Ubuntu Builds...

deymo:
@eustas Can you rebase this change? We added new required checks that are not in this tree's workflows, so you need to rebase in order to get them.

eustas:
Rebasing. Sorry for the late response, was OOO last days.

jonsneyers (approved):

PR #12, opened 2021-05-26
Add AppVeyor configuration file

EwoutH:
Add an AppVeyor configuration file that builds libjxl statically for Windows x64 and Arm64 using CMake and MSVC.

Arm64 builds are allowed to fail now due to dependencies not compiling correctly.

An example build can be [found here](https://ci.appveyor.com/project/EwoutH/libjxl/builds/39328673).

AppVeyor can be enabled via the [GitHub Marketplace](https://github.com/marketplace/appveyor).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,38 @@
+image: Visual Studio 2019
+configuration: Release
+clone_depth: 1
+
+environment:
+  VCPKG: C:/Tools/vcpkg
+  matrix:
+  - arch: x64
+  - arch: ARM64
+
+matrix:
+  allow_failures:
+    - arch: ARM64
+
+cache:
+- c:\tools\vcpkg\installed\
+- c:\Users\appveyor\AppData\Local\vcpkg\
+
+install:
+  - git submodule update --init --recursive
+  - vcpkg install gtest:%arch%-windows
+  - vcpkg install giflib:%arch%-windows
+  - vcpkg install libjpeg-turbo:%arch%-windows
+  - vcpkg install libpng:%arch%-windows
+  - vcpkg install zlib:%arch%-windows
+
+
+before_build:
+  - mkdir build
+  - cd build
+  - cmake .. -A %arch% -DBUILD_TESTING=OFF -DJPEGXL_STATIC=ON
+
+build:
+  project: build/JPEGXL.sln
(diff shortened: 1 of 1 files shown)


brunoais:
May it also make a build for linux? Maybe as an appimage?

deymo:
I would like to have more context on this. Why AppVeyor?

We currently don't have Windows builds and it would be great to add them; but I'm concerned about the proliferation of multiple CI/CD tools in the same project and having to maintain them when they break because of tooling versions for example (we have hit many compiler bugs). Is there a significant advantage to have this in appveyor as opposed to having it directly in github actions running on windows-latest?

eustas:
I agree, github-actions looks like a perfect fit for Windows builds.

EwoutH:
In the past I have encountered quite some complications configuring MSVC builds using GitHub Actions. Multiple third-party actions were needed to set it up correctly, mainly due to how environment variables (including PATH) are handled by GitHub.

So for Windows MSVC builds, I strongly prefer AppVeyor. For Ubuntu and macOS builds GitHub Actions is fine.

deymo:
> So for Windows MSVC builds, I strongly prefer AppVeyor. For Ubuntu and macOS builds GitHub Actions is fine.

Having it integrated into GitHub Actions has some advantages: output artifacts from actions are here, common infrastructure. Let's say that you want to update some flag we pass to cmake, or update some version of a dependency we compile; for ubuntu/mac you do it one way but for windows you do it in a different place with different config file and CI/CD concepts. I'd prefer to try to get it to work in github actions first before we consider moving windows to a new CI/CD provider.

jonsneyers:
Any update on getting Windows builds to work with GitHub Actions? We are getting more requests from people who just want the latest `cjxl.exe` and it would be nice to automate that and have it easily found here (as opposed to finding and trusting third parties).

ziemek99:
Is this PR now relevant at all since #419 has been merged almost 2 months ago and GitHub Actions builds for Windows run fine and are available for non-logged in users from [Luca's website](https://artifacts.lucaversari.it/libjxl/libjxl/latest/)?

jonsneyers:
I think this one has become obsolete indeed. Closing. Please open an issue or a new PR if I missed something.

PR #13, opened 2021-05-26
Allow alternative to linear srgb when xyb with ICC

lvandeve:
For decoding a XYB-encoded image with ICC profile as original color
profile, allow configuring the decoder to decode to a color encoding of
your choosing, rather than the decoder always converting xyb to srgb.

Reason for this: the decoder is able to convert xyb to all supported
color encoding values (srgb, pq, hlg, linear, p3, ...). It does this
when it knows the color encoding, but not when an ICC profile is used
(since the libjxl decoder does not have color management system to parse
it, nor would all icc profiles be representable as a jxl color encoding)

If the decoder does not know the color encoding, it default to linear
srgb.

However, if you have a system, like the chrome browser, that can
determine form the ICC profile what the color encoding is, then it is
better to be able to let the decoder convert the xyb immediately to
that, to avoid needing two passes over the data.

Therefore this option is added to the API and implemented in the
decoder.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -550,6 +550,47 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
     const JxlDecoder* dec, const JxlPixelFormat* format,
     JxlColorProfileTarget target, uint8_t* icc_profile, size_t size);
 
+/** Sets the color profile to use for JXL_COLOR_PROFILE_TARGET_DATA for the
+ * special case when the decoder has a choice. This only has effect for a JXL
+ * image where uses_original_profile is false, and the original color profile is
+ * encoded as an ICC color profile rather than a JxlColorEncoding with known
+ * enum values. In most other cases (uses uses_original_profile is true, or the
+ * color profile is already given as a JxlColorEncoding), this setting is
+ * ignored and the decoder uses a profile related to the image.
+ * No matter what, the JXL_COLOR_PROFILE_TARGET_DATA must still be queried to
+ * know the actual data format of the decoded pixels after decoding.
+ *
+ * The intended use case of this function is for cases where you are using
+ * a color management system to parse the original ICC color profile
+ * (JXL_COLOR_PROFILE_TARGET_ORIGINAL), from this you know that the ICC
+ * profile represents one of the color profiles supported by JxlColorEncoding
+ * (such as sRGB, PQ or HLG): in that case it is beneficial (but not necessary)
+ * to use JxlDecoderSetPreferredColorProfile to match the parsed profile. The
+ * JXL decoder has no color management system built in, but can convert XYB
+ * color to any of the ones supported by JxlColorEncoding.
+ *
+ * Can only be set after the JXL_DEC_COLOR_ENCODING event occured and before
+ * any other event occured, and can affect the result of
+ * JXL_COLOR_PROFILE_TARGET_DATA (but not of JXL_COLOR_PROFILE_TARGET_ORIGINAL),
+ * so should be used after getting JXL_COLOR_PROFILE_TARGET_ORIGINAL but before
+ * getting JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding must be grayscale
+ * if num_color_channels from the basic info is 1, RGB if num_color_channels
+ * from the basic info is 3.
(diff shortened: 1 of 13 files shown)


lib/jxl/decode_test.cc
```diff
@@ -1236,8 +1236,8 @@ TEST(DecodeTest, IccProfileTestXybEncoded) {
             JxlDecoderGetColorAsEncodedProfile(
                 dec, &format, JXL_COLOR_PROFILE_TARGET_DATA, &pixel_encoding));
   EXPECT_EQ(JXL_PRIMARIES_SRGB, pixel_encoding.primaries);
-  // The API returns LINEAR because the colorspace cannot be represented by enum
-  // values.
+  // The API returns LINEAR by defautl when the colorspace cannot be represented
```
veluca93, line 1239: default

veluca93 (approved):

lib/jxl/decode_test.cc
lvandeve: fixed

lib/include/jxl/decode.h
```diff
@@ -550,6 +550,44 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
+ * know the actual data format of the decoded pixels after decoding.
+ *
+ * The intended use case of this function is for cases where you are using
+ * a color management system to parse the original ICC color profile
+ * (JXL_COLOR_PROFILE_TARGET_ORIGINAL), from this you know that the ICC
+ * profile represents one of the color profiles supported by JxlColorEncoding
+ * (such as sRGB, PQ or HLG): in that case it is beneficial (but not necessary)
+ * to use JxlDecoderSetDefaultColorProfile to match the parsed profile. The
+ * JXL decoder has no color management system built in, but can convert XYB
+ * color to any of the ones supported by JxlColorEncoding.
```
jonsneyers, line 570: Another potential use case of this would be for viewers to ignore the original color profile (enum or ICC) and try to get the image decoded immediately to the display space when possible, right?

E.g. when the display is Display P3 and the image has whatever original ICC profile (sRGB, P3, ProPhoto, whatever), then you may want to call `JxlDecoderSetPreferredColorProfile` with a Display P3 `color_encoding` so if it's an XYB image you directly get pixels in Display P3 (in other cases you still have to do your own color management).

Might be worth adding that to the comment.

lvandeve: This setting doesn't *force* it however, if the decoder understands the color type (because it's an enum instead of ICC for XYB color) or cannot convert the color (because it's not XYB color but uses original profile), it ignores your setting.

The reason why it will ignore your setting if it's XYB color with enums (even though it theoretically could convert the XYB to your preferred choice) is that for blending (and some other operations, as far as I know), the specification requires the decoder to do it in that color space, so it would have to convert twice then. But in case of ICC profile with XYB, this does not apply, and decoder can decode to your preferred one.

A force option in theory is possible, but not for this MR :)

lib/include/jxl/decode.h
```diff
@@ -550,6 +550,44 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
+ *
+ * Can only be set after the JXL_DEC_COLOR_ENCODING event occured and before
+ * any other event occured, and can affect the result of
+ * JXL_COLOR_PROFILE_TARGET_DATA (but not of JXL_COLOR_PROFILE_TARGET_ORIGINAL),
+ * so should be used after getting JXL_COLOR_PROFILE_TARGET_ORIGINAL but before
+ * getting JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding must be grayscale
+ * if num_color_channels from the basic info is 1, RGB if num_color_channels
+ * from the basic info is 3.
+ *
+ * If JxlDecoderSetDefaultColorProfile is not used, the default is linear sRGB
```
deymo, line 580: "default" --> maybe make it clear that this is the default *preference*, not the default JXL_COLOR_PROFILE_TARGET_DATA

deymo (approved):

lib/include/jxl/decode.h
lvandeve: Thanks, fixed the function name and expanded comment a bit

deymo (approved):

PR #20, opened 2021-05-26
Improve fuzzing

eustas:
* generate files with preview images
 * toggle "keep orientation" Decoder API switch

### DIFF
--- a/tools/fuzzer_corpus.cc
+++ b/tools/fuzzer_corpus.cc
@@ -80,7 +80,7 @@ struct ImageSpec {
       << ", butteraugli=" << spec.params.butteraugli_distance
       << ", modular_mode=" << spec.params.modular_mode
       << ", lossy_palette=" << spec.params.lossy_palette
-      << ", noise=" << spec.params.noise
+      << ", noise=" << spec.params.noise << ", preview=" << spec.params.preview
       << ", fuzzer_friendly=" << spec.fuzzer_friendly
       << ", is_reconstructible_jpeg=" << spec.is_reconstructible_jpeg
       << ", orientation=" << spec.orientation << ">";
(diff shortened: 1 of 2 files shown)


tools/fuzzer_corpus.cc
```diff
@@ -122,7 +122,8 @@ struct ImageSpec {
     bool modular_mode = false;
     bool lossy_palette = false;
     bool noise = false;
-    uint8_t padding_[1] = {};
+    bool preview = false;
+    // uint8_t padding_[0] = {};
```
deymo, line 126: maybe add a comment on why is this commented here; like that we need to add padding because "packed".

tools/fuzzer_corpus.cc
```diff
@@ -386,36 +388,38 @@ int main(int argc, const char** argv) {
             }
             for (uint32_t num_frames : {1, 3}) {
               spec.num_frames = num_frames;
-
+              for (uint32_t preview : {0, 1}) {
```
deymo, line 391: we are already generating many many inputs, which takes time and might not improve the coverage. Given that keep orientation is pretty independent from other features, would it make more sense to set this as a random value rather than generating every input with and without this? It is actually not an encoder settings so it won't change the testcase (only the last bit).

tools/fuzzer_corpus.cc
eustas: Done.

tools/fuzzer_corpus.cc
eustas: There are two changes in this PR:
 - test "keep orientation"; it uses one bit of flags;
 - add samples with preview images; preview images are already tested, but those are difficult to be created from scratch by fuzzer; we employ encoder param (see line 251) to generate it; orientation was here before this PR; not sure how to setup github to "ignore" spaces to make that visible...

eustas: There is actually setting to hide whitespace changes (under the cog button).

deymo: Sorry I missed the preview part I was focused on the keep orientation.

deymo (approved):

PR #22, opened 2021-05-26
Fix occasional artifacts in the comparison tool

sboukortt:


veluca93 (approved):
Ah, good old missing clamping...

PR #23, opened 2021-05-26
Share LCMS transform across threads

kleisauke:
`cmsFLAGS_NOCACHE` disables the 1-pixel cache and makes calling `cmsDoTransform()` thread-safe. So instead of creating an LCMS transform for each thread, one can pass this flag and share a single transform across all threads.

I made an attempt to benchmark this, but the benefits (in terms of performance) are minimal, and I'm not sure whether it can be properly measured. Therefore, this should be considered as a general improvement.

<details>
  <summary>Benchmark results</summary>

Benchmarked with a [photo of Wat Arun](https://commons.wikimedia.org/wiki/File:The_sculptures_of_two_mythical_giant_demons,_Thotsakan_and_Sahatsadecha,_guarding_the_eastern_gate_of_the_main_chapel_of_Wat_Arun,_Bangkok.jpg) (JPEG, 7360x4912, 29MB) containing the common sRGB IEC61966-2.1 profile.

Run with:
```bash
$ SKIP_TEST=1 ./ci.sh release -DJPEGXL_ENABLE_SKCMS=OFF
$ build/tools/benchmark_xl --input "~/wat-arun.jpg" --codec jxl:fast:d1
```

Before:
```
Compr               Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method             Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
---------------------------------------------------------------------------------------------------------------------------------
jxl:fast:d1      36152320  6101574    1.35019251877   1  45.831  36.947   1.68089497   0.58562449946  0.7907058179756201        0
Aggregate:       36152320  6101574    1.35019251877 ---  45.831  36.947   1.68089497   0.58562449946  0.7907058179756201        0
```

After:
```
Compr               Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method             Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
---------------------------------------------------------------------------------------------------------------------------------
jxl:fast:d1      36152320  6101574    1.35019251877   1  45.905  37.082   1.68089497   0.58562449946  0.7907058179756201        0
Aggregate:       36152320  6101574    1.35019251877 ---  45.905  37.082   1.68089497   0.58562449946  0.7907058179756201        0
```

</details>

### DIFF
--- a/lib/jxl/enc_color_management.cc
+++ b/lib/jxl/enc_color_management.cc
@@ -201,9 +201,7 @@ void DoColorSpaceTransform(ColorSpaceTransform* t, const size_t thread,
         &t->skcms_icc_->profile_src_, buf_dst, skcms_PixelFormat_RGB_fff,
         skcms_AlphaFormat_Opaque, &t->skcms_icc_->profile_dst_, t->xsize_));
 #else   // JPEGXL_ENABLE_SKCMS
-    JXL_DASSERT(thread < t->transforms_.size());
-    cmsHTRANSFORM xform = t->transforms_[thread];
-    cmsDoTransform(xform, xform_src, buf_dst,
+    cmsDoTransform(t->lcms_transform_, xform_src, buf_dst,
                    static_cast<cmsUInt32Number>(t->xsize_));
 #endif  // JPEGXL_ENABLE_SKCMS
   }
@@ -735,9 +733,7 @@ void ColorEncoding::DecideIfWantICC() {
 ColorSpaceTransform::~ColorSpaceTransform() {
 #if !JPEGXL_ENABLE_SKCMS
   std::lock_guard<std::mutex> guard(LcmsMutex());
-  for (void* p : transforms_) {
-    TransformDeleter()(p);
-  }
+  TransformDeleter()(lcms_transform_);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


veluca93:
I'm not entirely sure that benchmark_xl even measures CMS speed, there might be other ways to check that - but I also think that this is a Good Thing (tm) independently of the potential speed benefits :)

lib/jxl/enc_color_management.cc
```diff
@@ -735,9 +733,7 @@ void ColorEncoding::DecideIfWantICC() {
 ColorSpaceTransform::~ColorSpaceTransform() {
 #if !JPEGXL_ENABLE_SKCMS
   std::lock_guard<std::mutex> guard(LcmsMutex());
```
veluca93, line 735: I think this wouldn't be needed anymore, either.

sboukortt: I am not entirely sure but it seems to me that it might, based on the comment next to the definition of `LcmsMutex`.

sboukortt (approved):

lib/jxl/enc_color_management.cc
veluca93: Ah, that's disappointing - we should probably not create and/or use it when using skcms though... But that's better left to a separate MR.

deymo (approved):

jonsneyers (approved):

lib/jxl/enc_color_management.cc
kleisauke: fwiw, I think that mutex can be removed entirely after commit https://github.com/mm2/Little-CMS/commit/a35bacde52e06aac4704db3e35ad434822938dc2. See https://github.com/mm2/Little-CMS/issues/170#issuecomment-640677221.

PR #24, opened 2021-05-26
Check for AUTHORS file updates and re-enable PR checks.

deymo:
Pull Request checks should only run on pull_request events, and not
needed to run on pull_request_review. This gets the feedback to the PR
as soon as possible. This was also causing issues in the PR with the
way we get the SHA values from the current environment.

This patch fixes the pull request workflow which was needed for the
AUTHORS check and adds this new check to update the AUTHORS file when
new contributors send patches.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -4,8 +4,6 @@ name: PR
 on:
   pull_request:
     types: [opened, reopened, labeled, unlabeled, synchronize]
-  pull_request_review:
-    types: [submitted, dismissed]
 
 jobs:
   build:
@@ -47,10 +45,17 @@ jobs:
       run:
         '[[ ${{ matrix.mode }} != "release" ]] || ./ci.sh test'
 
+  # Checks that the AUTHORS files is updated with new contributors.
+  authors:
+    runs-on: [ubuntu-latest]
+    steps:
+    - name: Checkout the source
+      uses: actions/checkout@v2
+    - name: Check AUTHORS file
+      run:
+        ./ci.sh authors
+
   format:
-    # TODO(deymo): Currently disabled because the git structure where the PR
-    # runs is different.
-    if: false
     runs-on: [ubuntu-latest]
     steps:
     - name: Install build deps
--- a/tools/check_author.py
+++ b/tools/check_author.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+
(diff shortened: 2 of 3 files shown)


deymo:
hmm.. something is still wrong with the sha because of the merge commit generated by GitHub, I'll take a look.

tools/check_author.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+import re
+import sys
+
+
+def IsAuthorInFile(email, name, filename):
+  """Return whether we find the name/email in the authors filename"""
+  # Organization emails have emails listed as <*@domain.com>. This matches those
+  # patterns.
+  email_pattern_regex = re.compile(r'.*(<[^>]+>)')
```
sboukortt, line 21: Wouldn’t it be simpler (e.g. on line 41) to have `<` / `>` outside of the group?

tools/check_author.py
```diff
@@ -0,0 +1,76 @@
...
+
+def CheckAuthor(args):
+  ret = IsAuthorInFile(
+      args.email, args.name, os.path.join(args.source_dir, 'AUTHORS'))
+  if not ret:
+    print("User %s <%s> not found, please add youself to the AUTHORS file" % (
+              args.name, args.email),
+          file=sys.stderr)
+  if not ret and not args.dry_run:
+    sys.exit(1)
```
sboukortt, line 56: How about merging this with the previous branch:

```suggestion
    if not args.dry_run:
      sys.exit(1)
```

sboukortt (approved):

tools/check_author.py
deymo: Done.

tools/check_author.py
deymo: Done.

.github/workflows/pull_request.yml
```diff
@@ -47,10 +45,23 @@ jobs:
...
+    runs-on: [ubuntu-latest]
+    steps:
+    - name: Checkout the source
+      uses: actions/checkout@v2
+      with:
+        # Needs to check out the HEAD of the pull request instead of the merge
+        # and enough history from the pull request to reach the common ancestor.
+        # Normally pull-requests have only 1 commit, so 10 should be enough.
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 10
```
deymo, line 59: @sboukortt  this is not pretty but it will do for now. I think we can do something like always get the merge commit with depth: 2 and compare HEAD with HEAD^; but that's not how we had this set-up and the commit author of the merge is the default email address which is not the one in the commit being merged/rebased.

deymo: moved this to the ci.sh.

sboukortt (approved):

PR #30, opened 2021-05-27
 Fuzz: fix heap-buffer-overflow in codec_psd

eustas:
decode_layer assumes that chan_id.size() == invert.size().
On one callsite (where invert is actually used) this is true.
On the other callsite invert is just an "all-false" stub, but
its size is set to be `real_nb_channels` which might be not enough.


jonsneyers:
Good catch!

jonsneyers (approved):

PR #36, opened 2021-05-27
Update issue templates

jonsneyers:


deymo:
it looks like you are using your gmail.com account. If you created this from the UI there's a setting to select the email from the UI when creating the commit.

jonsneyers:
Yes except for this one I don't think the UI actually let me choose an email address. I suppose it's hard to change it now. Ugh.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
...
+Steps to reproduce the behavior:
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots or example input/output images to help explain your problem.
+
+**Additional context**
+Add any other context about the problem here (e.g. OS).
```
deymo, line 23: I think a lot of bugs are related to the environment (compiler version, compile flags used, CPU where you run it) specially related to the SIMD code. We should explicitly ask for those.

jonsneyers: RIght. PTAL if this is better.

deymo (approved):

PR #47, opened 2021-05-31
Faster fast lossless

jonsneyers:
Makes modular falcon, thunder and lightning faster. Slower speeds should also get slightly faster. Lightning is now consistently at least slightly faster than thunder.

Changes:
- Specialized encode path for channel palette. Especially for high bit depth this is a lot faster.
- Skip channel palette in lightning mode, except in case of > 8-bit.
- No tree in lightning mode, specialized code path for no tree Gradient.
- Use fast ANS histogram computation in lightning mode.
- Use same predefined WP and Gradient trees as for DC (is mostly the same but slightly better compression)

Some examples:

Before:
```
ClassA_8bit_p06.ppm.png
Compr                   Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method                 Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
-------------------------------------------------------------------------------------------------------------------------------------
jxl:m:falcon         10989056 14237220   10.36465370638   1  12.067   7.465   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:thunder        10989056 15871570   11.55445563295   1  16.936  16.151   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:lightning      10989056 16132257   11.74423499161   1  16.996  18.441   0.00000000   0.00000000000  0.0000000000000000        0
```

After:
```
ClassA_8bit_p06.ppm.png
Compr                   Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method                 Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
-------------------------------------------------------------------------------------------------------------------------------------
jxl:m:falcon         10989056 14234495   10.36266991450   1  14.687   7.651   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:thunder        10989056 15871630   11.55449931277   1  21.662  17.482   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:lightning      10989056 16788658   12.22209296231   1  26.845  17.152   0.00000000   0.00000000000  0.0000000000000000        0
```

On a 10-bit image: (stored as 16-bit PNG, so channel palette is very useful)

Before:
```
ClassC_BLENDER_Sintel2_4096x1744p_24_10b_sRGB_444_00004606.ppm.png
Compr                   Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method                 Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
-------------------------------------------------------------------------------------------------------------------------------------
jxl:m:falcon          7143424 13724411   15.37012054723   1   3.596   6.815   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:thunder         7143424 14989468   16.78687195384   1   3.999  14.577   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:lightning       7143424 15148674   16.96516852423   1   3.879  15.114   0.00000000   0.00000000000  0.0000000000000000        0
```

After:
```
ClassC_BLENDER_Sintel2_4096x1744p_24_10b_sRGB_444_00004606.ppm.png
Compr                   Input    Compr            Compr       Compr  Decomp  Butteraugli                                             
Method                 Pixels     Size              BPP   #    MP/s    MP/s     Distance    Error p norm           BPP*pnorm   Errors
-------------------------------------------------------------------------------------------------------------------------------------
jxl:m:falcon          7143424 13656686   15.29427456637   1  11.603   6.819   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:thunder         7143424 14937761   16.72896470936   1  16.527  14.843   0.00000000   0.00000000000  0.0000000000000000        0
jxl:m:lightning       7143424 15759170   17.64886978569   1  16.968  14.769   0.00000000   0.00000000000  0.0000000000000000        0
```


lib/jxl/modular/transform/palette.h
```diff
@@ -443,23 +446,68 @@ static Status FwdPalette(Image &input, uint32_t begin_c, uint32_t end_c,
...
+        }
+        if (candidate_palette.size() > nb_colors) {
+          return false;  // too many colors
+        }
       }
-      const bool new_color = candidate_palette.insert(color).second;
-      if (new_color) {
-        candidate_palette_imageorder.push_back(color);
+    }
+  } else if (!lossy) {
```
veluca93, line 468: maybe we should move this up and make it an early return before we even create the std::set...

lib/jxl/modular/encoding/enc_encoding.cc
```diff
@@ -237,6 +237,25 @@ Status EncodeModularChannelMAANS(const Image &image, pixel_type chan,
         wp_state.UpdateErrors(r[x], x, y, channel.w);
       }
     }
+  } else if (tree.size() == 1 && tree[0].predictor == Predictor::Gradient &&
```
veluca93, line 240: Can we add a test that exercises this path and one that does but uses skip_encoder_fast_path?

veluca93 (changes requested):

lib/jxl/modular/encoding/enc_encoding.cc
jonsneyers: There actually isn't a specific decoder fast path for this (yet) — it goes to the slightly more generic decoder path of "no tree, not WP". Do you think a test would be useful here?

lib/jxl/modular/transform/palette.h
jonsneyers: Makes sense.

lib/jxl/modular/encoding/enc_encoding.cc
veluca93: Are you sure there isn't? I remember adding one...

A test would be useful anyway :)

jonsneyers: There is a decoder path for `is_gradient_only`, actually that's the one that will be used here, with an unnecessary ctx lookup because everything has the same ctx. Maybe I should add a decoder path for this no-tree gradient-only no-multiply/offset case, it might be slightly faster.

There's no `skip_decoder_fast_path`, right? Maybe it would be useful to also test that — encode with fast paths, decode with generic path...

veluca93: Oh, this is single-tree gradient only... I think that's OK as it is on the decoder side. Still needs a test though :)

As for testing that - likely not needed if we test the encoder fast paths...

jonsneyers: Changing the decoder fast paths to first see if it's a no-tree case before it tries to do the lookup table thing for WP/Gradient.

Adding a specific decode fast path for no-tree gradient-only helps a bit.

before:
4064 x 2704, geomean: 47.08 MP/s [37.02, 49.18], 30 reps, 4 threads.
after:
4064 x 2704, geomean: 52.10 MP/s [42.82, 54.56], 30 reps, 4 threads.

Added a test a for the encoder skip_fast_path for lightning mode.

lib/jxl/modular/transform/palette.h
jonsneyers: Moved it up, PTAL.

lib/jxl/modular/encoding/enc_encoding.cc
jonsneyers: Made the above changes, PTAL.

lib/jxl/jxl_test.cc
```diff
@@ -949,6 +949,25 @@ TEST(JxlTest, JXL_SLOW_TEST(RoundtripLosslessNoEncoderFastPathGradient)) {
...
+TEST(JxlTest, JXL_SLOW_TEST(RoundtripLosslessNoEncoderVeryFastPathGradient)) {
+  ThreadPoolInternal pool(8);
+  const PaddedBytes orig =
+      ReadTestData("wesaturate/500px/tmshre_riaphotographs_srgb8.png");
+  CodecInOut io;
+  ASSERT_TRUE(SetFromBytes(Span<const uint8_t>(orig), &io, &pool));
+
+  CompressParams cparams = CParamsForLossless();
+  cparams.speed_tier = SpeedTier::kLightning;
+  cparams.options.skip_encoder_fast_path = true;
```
veluca93, line 961: Can we do both?

veluca93 (approved):

lib/jxl/jxl_test.cc
jonsneyers: Both with and without skip_encoder_fast_path? Sure.

jonsneyers: Done.

PR #48, opened 2021-05-31
fix: unused CPU_[ZERO|SET] on HaikuOS

ahwayakchih:
Building was failing because `CPU_ZERO` and `CPU_SET` in `os_specific.cc` were unused on HaikuOS.

This change makes HaikuOS follow example of MacOS: call them even though in the end thread will not be pinned anyway.

HaikuOS build could use `JXL_DISABLE_PINNING`, but then it would have to add some defines to exclude declaring `CPU_SET` and `CPU_ZERO`, which would lessen code readability for everyone.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,3 +18,4 @@ Google LLC <*@google.com>
 # Individuals:
 Jon Sneyers <jon@cloudinary.com>
 Pieter Wuille
+Marcin Konicki <ahwayakchih@gmail.com>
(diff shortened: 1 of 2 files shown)


ahwayakchih:
Added myself to AUTHORS to pass the CI check, but i don't think this change is enough to require that, so let me know if i can revert that commit.

deymo (approved):

jan-wassenberg (approved):

deymo:
> Added myself to AUTHORS to pass the CI check, but i don't think this change is enough to require that, so let me know if i can revert that commit.

AUTHORS is for Copyright purposes and it is better to err on the safe side.

PR #54, opened 2021-05-31
Fuzz: Undefined-shift in jxl::ModularFrameDecoder::FinalizeDecoding

eustas:
MetaPalette erases channels with non-zero h-/v-shift; when this transform is undone,
channels with zero h-/v-shift are added; un-squeeze of those will create channels
with negative h-/v-shift; repeated un-squeeze causes assertion failure.

Check h-/v-shift in MetaPalette.

### DIFF
--- a/lib/jxl/modular/transform/palette.h
+++ b/lib/jxl/modular/transform/palette.h
@@ -383,9 +383,11 @@ static Status CheckPaletteParams(const Image &image, uint32_t begin_c,
   if (c1 > image.channel.size() || c2 >= image.channel.size() || c2 < c1) {
     return JXL_FAILURE("Invalid channel range");
   }
+  const auto &ch1 = image.channel[begin_c];
   for (size_t c = begin_c + 1; c <= end_c; c++) {
-    if (image.channel[c].w != image.channel[begin_c].w ||
-        image.channel[c].h != image.channel[begin_c].h) {
+    const auto &ch2 = image.channel[c];
+    if (ch1.w != ch2.w || ch1.h != ch2.h || ch1.hshift != ch2.hshift ||
+        ch1.vshift != ch2.vshift) {
       return false;
     }
   }


lib/jxl/modular/transform/palette.h
```diff
@@ -389,6 +389,11 @@ static Status CheckPaletteParams(const Image &image, uint32_t begin_c,
       return false;
     }
   }
+  for (size_t c = begin_c; c <= end_c; c++) {
+    if (image.channel[c].hshift != 0 || image.channel[c].vshift != 0) {
+      return false;
```
jonsneyers, line 394: See also https://github.com/libjxl/libjxl/pull/35

It's not incorrect to do palette on channels that have nonzero shift; this could in fact be useful, and the spec does not disallow it.

It is however incorrect to do palette on channels of different dimensions (which we already test elsewhere).

In our code we assume that when the dimensions are the same, the hshift/vshift is also the same, which I think is an invariant that should hold in all sane cases but there are some weird cases where it isn't necessarily the case: for example when you have a 3x100 3-channel image on which you do a 100-color palette and then do a 2-channel palette on the palette itself (which will also happen to be a 3x100 channel with hshift -1) plus the index image itself (which is a 3x100 channel with hshift 0).

So it is good to check that the hshift/vshifts are **the same** (so all equal to those of `begin_c`), because that's the actual invariant we assume to hold. In the reconstruction, the hshift/vshift of the channels [`begin_c`+1, `end_c`] will be restored to those of channel `begin_c`, so if this invariant is violated, we get a problematic case where MetaApply + undo_transforms does not produce something with the original channel metadata.

eustas: #35 looks promising, but it does not fix the problem =(
Going to check if the invariant is held.

eustas: PTAL

jonsneyers (approved):

PR #56, opened 2021-05-31
Add NGINX plugins

veluca93:


jonsneyers:
This is cool and can be put here for now, but eventually I'd like to have these "applications using libjxl" in separate repos, either all in one place (`github.com/libjxl/plugins` or something) or each one with their own repo (`github.com/libjxl/nginx-jxl` etc).

It is unnecessary to run libjxl CI every time a plugin gets improved, it adds noise to the pull requests here, and it is likely better to have separate issues for the plugins. Also, we may want to give other people write access to specific plugins, e.g. people who want to maintain that plugin and who may know more about the thing the plugin is made for than we do.

Since I think repos under an organization are just flat (we cannot have e.g. `github.com/libjxl/plugins/nginx`, afaik), it might be a good idea to have a naming scheme that makes it easier to see what's a plugin and what's something else. Maybe `jxl-plugin-nginx` or something like that would work.

deymo:
There's an argument for both. If we make changes to libjxl that break the plugins, or that break the usage of libjxl from the plugin it is useful for know at the time you are making the change to libjxl. With all the plugins/ using the C API it is less likely for that to be the case; but eventually we want to test together the plugins with new libjxl. Having separate repos and separate releases for plugins and library means that we also need to maintain those dependencies and add requirements to the plugin build instructions (like, this plugin requires libjxl >= 0.4). This is not zero work, so we should look into the overhead of multiple repos (multiple GitHub CI/CD configurations, multiple release instructions, multiple repo setup CONTRIBUTING / AUTHORS / LICENSE, etc). We can move them out when we need to, as some of these plugins might not make sense on their own repo.

veluca93:
Yes, I think we can consider the question of splitting out the plugins independently of whether we should add the NGINX plugin now.

plugins/ngx_jxl/README.md
```diff
@@ -0,0 +1,6 @@
+# NGINX JPEG XL modules
+This folder provides two modules for serving losslessly recompressed JPEG
+files in a transparent way.
+
+The `static` module serves `x.jpg.jxl` files instead of `x.jpg` files if they
```
deymo, line 5: Could you expand on this documentation? what's the status of these plugins? compile instructions?
(how were these tested should be in the commit message).

For `static`, do we need to keep both the `x.jpg.jxl` and the `x.jpg`?

For the `filter`, I wonder why the choice of storing the .jpg in the server and compress to jxl on the fly instead of doing the opposite: storing the .jxl on the server and decompress to .jpg on the fly. The latter would save space on the server side and provide bandwidth savings (at the cost of CPU) but the former just costs more CPU for the server for only the bandwidth reduction so there's less incentive to use it.

Would it make sense to have a single plugin? It seems that having a single plugin would make it easier for people to use:
* if you request "foo.jpg" but accept image/jxl we can serve the cached foo.jpg.jxl; or transcode the foo.jpg to jxl (if for example an option in the plugin config enables that).
* if you request "foo.jxl" but don't accept image/jxl; we can serve the decoded jxl-to-jpg file as a image/jpg.

For reviewing it will be easier if we start with a smaller commit with just one module.

veluca93: I expanded the README.

The idea of the filter module is that you can enable it without thinking about it on existing webservers. Moreover, likely in a world where jxl support is not common it is preferable to spend more CPU for the JXL case (the CPU expenditure is remarkably similar either way).

As for a single plugin - I agree, but I don't know if that's at all possible. I assume that ngx_brotli has two plugins for a reason :)

For reviewing: the two plugins are entirely separate, so feel free to review only one of the plugins to begin with.

jonsneyers (approved):

PR #71, opened 2021-06-01
Use a sentinel value instead of 0 for memory sanitizer.

veluca93:
This should make it more evident when actually using "uninitialized"
padding memory in sanitizer builds.

### DIFF
--- a/lib/jxl/common.h
+++ b/lib/jxl/common.h
@@ -25,6 +25,12 @@
 namespace jxl {
 // Some enums and typedefs used by more than one header file.
 
+#ifdef MEMORY_SANITIZER
+// Chosen so that kSanitizerSentinel is four copies of kSanitizerSentinelByte.
+constexpr uint8_t kSanitizerSentinelByte = 0x48;
+constexpr float kSanitizerSentinel = 205089.125f;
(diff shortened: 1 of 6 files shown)


jonsneyers (approved):

lib/jxl/common.h
```diff
@@ -25,6 +25,12 @@
 namespace jxl {
 // Some enums and typedefs used by more than one header file.
 
+#ifdef MEMORY_SANITIZER
+// Chosen so that kSanitizerSentinel is four copies of kSanitizerSentinelByte.
+constexpr uint8_t kSanitizerSentinelByte = 0x48;
+constexpr float kSanitizerSentinel = 205089.125f;
```
deymo, line 31: what do you think about using a NaN instead?

veluca93: I thought about it, but I am not sure it is a good idea - doing SIMD operations on NaNs (and subnormals) can have weird performance problems, and it is possible that sooner or later we'll end up using the sentinel in places where we shouldn't.

Also, sanitizers are likely to complain more about NaN values in padding, and possibly create false positives...

PR #73, opened 2021-06-01
Progressive steps

veluca93:
Disable progressive_dc with -p, and remove a 4x downsampled pass that
can look worse than just upsampling DC.

### DIFF
--- a/lib/jxl/enc_file.cc
+++ b/lib/jxl/enc_file.cc
@@ -58,8 +58,6 @@ PassDefinition progressive_passes_dc_lf_salient_ac_other_ac[] = {
      /*suitable_for_downsampling_of_at_least=*/0}};
 
 PassDefinition progressive_passes_dc_quant_ac_full_ac[] = {
-    {/*num_coefficients=*/8, /*shift=*/2, /*salient_only=*/false,
-     /*suitable_for_downsampling_of_at_least=*/4},
     {/*num_coefficients=*/8, /*shift=*/1, /*salient_only=*/false,
      /*suitable_for_downsampling_of_at_least=*/2},
     {/*num_coefficients=*/8, /*shift=*/0, /*salient_only=*/false,
--- a/tools/cjxl.cc
+++ b/tools/cjxl.cc
@@ -339,7 +339,8 @@ void CompressArgs::AddCommandLineOptions(CommandLineParser* cmdline) {
   cmdline->AddOptionValue(
       's', "speed", "ANIMAL",
       "Deprecated synonym for --effort. Valid values are:\n"
-      "    lightning (1), thunder, falcon, cheetah, hare, wombat, squirrel, kitten, tortoise (9)\n"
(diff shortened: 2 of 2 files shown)


jonsneyers:
Can you rebase without the msan commit?

Also a bit of a pull request description would be nice :)

I guess this is making the default progressive skip the 1:4 pass to avoid the ugly regression from smooth-upsampled 1:8 to ugly-blocky 1:4 ? And also to skip progressive-dc, for which we don't have a good way yet to show it anyway (it's currently broken, I think, and to show it nicely it'll require some kind of blur/fancy upsampling).

veluca93:
Ah, sorry, the push ate the description.

lib/jxl/enc_file.cc
```diff
@@ -58,8 +58,6 @@ PassDefinition progressive_passes_dc_lf_salient_ac_other_ac[] = {
      /*suitable_for_downsampling_of_at_least=*/0}};
 
 PassDefinition progressive_passes_dc_quant_ac_full_ac[] = {
-    {/*num_coefficients=*/8, /*shift=*/2, /*salient_only=*/false,
-     /*suitable_for_downsampling_of_at_least=*/4},
     {/*num_coefficients=*/8, /*shift=*/1, /*salient_only=*/false,
      /*suitable_for_downsampling_of_at_least=*/2},
```
jyrkialakuijala, line 62: reflecting the LCP experiments by Moritz there wouldn't be any downsampling here -- just a shift by 1

what is the impact of 'suitable for downsampling of at least' ?

jonsneyers: There's no downsampling done in encoder or decoder here, that field is just a decoder hint that gets signalled in the headers, basically saying that if after decoding, you're going to downsample it 1:2  (e.g. because it's a 1000x1000 image that will be rendered in a screen area of 500x500), then maybe just decoding up to this pass will be 'good enough'.

tools/cjxl.cc
```diff
@@ -578,7 +579,6 @@ jxl::Status CompressArgs::ValidateArgs(const CommandLineParser& cmdline) {
 
   if (progressive) {
     params.qprogressive_mode = true;
-    params.progressive_dc = 1;
```
jonsneyers, line 581: I know we currently don't render DC progressively anyway, but that's a current decoder limitation. Even if we only render something when all the DC is available, it might still be worthwhile to do the DC frame thing though, just because it will have the DC earlier (the DC frame _only_ does DC, while default in-frame DC is mixed with ACmeta). So wouldn't it be better to keep this?

veluca93: I'd prefer it to be explicitly enabled, as it can have surprising effects on quality - `-p` doesn't.

jonsneyers: OK for me.

jonsneyers (approved):

Issue #81, opened 2021-06-02
Premultiplied alpha not rendered correctly


jonsneyers:
In the case of premultiplied (associated) alpha, `djxl` and other applications ignore that image header field, and treat everything as if it is the usual non-premultiplied (unassociated) alpha like in PNG.

It would make sense to do the same thing here as with the Orientation: by default, the decoder should return pixels in a default way (orientation corrected, unassociated alpha), so applications that want to interpret the `alpha_associated` info themselves and get the raw data would have to explicitly ask the decoder to not do the normalization.

Issue #76, opened 2021-06-01
Add option to cjxl do clear invisible pixels and/or do premultiplied alpha


jonsneyers:
The bitstream can do premultiplied alpha, but currently cjxl has no way to select whether to do that or not.

Current behavior is that it does whatever the input format does, e.g. non-premultiplied alpha on PNG input and premultiplied alpha on EXR input.

It would be nice to have a flag to control this.

Also, in the non-premultiplied case, it would be good to have an option in the lossless mode to say "I don't care about the invisible pixels", so they can be made black or whatever else is good for compression (in lossy mode this is already done, but in lossless mode there is currently no way to do it, i.e. we now do what `cwebp -exact` does).

ziemek99:
Shouldn't this issue be closed since #123 has been merged? (Tip: you can link multiple PRs to one issue.)

jonsneyers:
Right, closing...

PR #82, opened 2021-06-02
Add --premultiply option to cjxl

jonsneyers:
Addresses one of the two feature requests in https://github.com/libjxl/libjxl/issues/76 :
Adds an option `--premultiply` to cjxl which will store the image in a premultiplied-alpha way (converting non-premultiplied to premultiplied).

Note: until https://github.com/libjxl/libjxl/issues/81 is resolved, this option will cause image to be rendered incorrectly.

### DIFF
--- a/lib/jxl/codec_in_out.h
+++ b/lib/jxl/codec_in_out.h
@@ -183,6 +183,19 @@ class CodecInOut {
     }
     return true;
   }
+  // Calls PremultiplyAlpha for each ImageBundle (preview/frames).
+  void PremultiplyAlpha() {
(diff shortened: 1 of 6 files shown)


veluca93:
Can we add a test?

jonsneyers:
Added a test, also tightened the test for non-premultiplied alpha a bit (I think it was from a different time, when BA didn't take alpha into account yet, alpha was stored losslessly and invisible pixels were not touched).

veluca93 (dismissed):

veluca93 (approved):

PR #83, opened 2021-06-02
do unpremultiply if needed in ConvertToExternal

jonsneyers:
Partially fixes https://github.com/libjxl/libjxl/issues/81

In the cases where ConvertToExternal is called to get an RGBA buffer, it will now always be unassociated alpha (which is what most applications will expect, including e.g. the png encoder in djxl).

If there is a need to also have the option of not doing the unpremultiply and getting the original, premultiplied values, this can be added later; for now it is important to get things right in decoders/viewers that are not looking at the premultiplied field and assuming things to be non-premultiplied.

This doesn't yet fix code paths that do not involve ConvertToExternal.

### DIFF
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -270,7 +270,7 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
   const size_t bytes_per_pixel = num_channels * bytes_per_channel;
 
   const Image3F* color = &ib.color();
-  Image3F temp_color;
+  Image3F temp_color, unpremul;
   const ImageF* alpha = ib.HasAlpha() ? &ib.alpha() : nullptr;
   ImageF temp_alpha;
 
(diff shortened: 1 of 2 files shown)


veluca93:
Can we instead do this in dec_reconstruct.cc?

jonsneyers:
Can do it in dec_reconstruct, but then there's a bit of a complication: the unpremultiply can only happen after all blending has been done... I mean of all frames. Otherwise you can get different results with those light-emitting pixels (A=0, RGB>0) depending on whether you first blend frames and then unpremultiply them, or first unpremultiply decoded frames and then blend them.

Also for the case of decoding to EXR, unpremultiply should be skipped (which does happen with this PR, since the EXR encoder does not use ConvertToExternal).

veluca93:
Ok, then in dec_frame we should make it so that the early output path is
skipped if we need to unpremultiply, and once blending is done rect-by-rect
we'll take care of it.

As for EXR, the proper solution for that is specifying a flag to disable
un-premultiplying.

jonsneyers:
The problem is deeper I think, cannot unpremultiply a frame until all future frames that will depend on it (blend/patch with it as a source) don't need it anymore... Or is that ok, can the unpremultiply be done after the save to reference frame? (also in the save_before_ct==false case, I mean)

veluca93:
Well, for now we can do it in dec_external_image and then do it post
blending in dec_reconstruct.cc once all the pieces are there :)

jonsneyers:
Sounds good to me. Let's merge this and then reopen #81 to remember to move it to dec_reconstruct later.

veluca93:
Ok, but please do add the check in dec_frame.h :)

jonsneyers:
Added the check in dec_frame.h.

veluca93 (approved):

PR #88, opened 2021-06-02
Add ccache to the pipeline to speed it up

deymo:
In Pull Requests, this uses the cache from the parent commit from main the first time, and re-uses its own cache in case of a rebuild of the same commit.

In main this uses the previous commit as a starting point as well. There's a 200MB ccache limit so old objects are evicted before we use up too much space in the per-repo GitHub cache, so multiple build targets can have space for their caches.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -37,11 +37,15 @@ jobs:
               -DJPEGXL_ENABLE_DEVTOOLS=OFF -DJPEGXL_ENABLE_PLUGINS=OFF
               -DJPEGXL_ENABLE_VIEWERS=OFF
 
+    env:
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
+
     steps:
     - name: Install build deps
       run: |
         sudo apt update
         sudo apt install -y \
+          ccache \
           clang-7 \
           cmake \
           doxygen \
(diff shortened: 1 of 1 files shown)


jonsneyers:
I don't know how to review this but I do welcome this improvement!

.github/workflows/build_test.yml
```diff
@@ -61,14 +65,38 @@ jobs:
...
       with:
         submodules: true
+        fetch-depth: 2
+    - name: Git environment
+      id: git-env
+      run: |
+        echo "::set-output name=parent::$(git rev-parse ${{ github.sha }}^)"
+      shell: bash
+    - name: ccache
+      uses: actions/cache@v2
```
veluca93, line 75: I assume that's "cache of previous commit"?

veluca93 (approved):

.github/workflows/build_test.yml
deymo: actions/cache@v2 is the GitHub action that lets you retrieve and store "a cache", up to 5GB per repo. A "cache" has a key and it is created only once (until it is garbage collected when you run out of space). So this get the cache of key "$os_name-$my_commit-sha-$build_name" and uses it. If it doesn't exist (if the key: doesn't exist) it then tries to retrieve the one from restore-keys (you can pass multiple). If there was no cache for key: it will create one at the end of the pipeline.

deymo: @v2 is just the release v2 of the actions/cache repo https://github.com/actions/cache which is github's own actions org.

PR #94, opened 2021-06-03
don't try to do lossless jpeg transcoding and resampling at the same time

jonsneyers:
When doing `cjxl input.jpg --resampling 2 output.jxl`, decoding the input to pixels has to be forced (can't resample _and_ still do lossless jpeg recompression).


veluca93 (dismissed):

veluca93:
We should also fix it in the decoder I think (refuse to decode to JPEG if you have resampling)

jonsneyers:
The decoder will already fail just because the `jbrd` doesn't match the image data and leads to failure.
I don't know if it makes sense to have a special check for resampling - there are many ways in which you could make a bitstream where the image doesn't match the `jbrd`, but that will just fail when trying to decode to JPEG.

It might make sense to still allow `--resampling 2 --already_downsampled` — that way you get a fancy upsampling of an existing JPEG, and it can still be decoded to the original (not upsampled) JPEG.
Maybe the condition in cjxl should be updated to allow that when doing lossless jpeg transcode, and just not allow resampling where the encoder is downsampling first?

veluca93:
> The decoder will already fail just because the `jbrd` doesn't match the image data and leads to failure.
> I don't know if it makes sense to have a special check for resampling - there are many ways in which you could make a bitstream where the image doesn't match the `jbrd`, but that will just fail when trying to decode to JPEG.
> 
> It might make sense to still allow `--resampling 2 --already_downsampled` — that way you get a fancy upsampling of an existing JPEG, and it can still be decoded to the original (not upsampled) JPEG.
> Maybe the condition in cjxl should be updated to allow that when doing lossless jpeg transcode, and just not allow resampling where the encoder is downsampling first?

That might make sense.

veluca93 (approved):

PR #95, opened 2021-06-03
fix check in MaybeSetFloatCallback, factor out duplicate code

jonsneyers:
Blending check in MaybeSetFloatCallback was not quite correct, possibly causing segfaults.

Since the conditions are / should be the same between the two MaybeSet* functions, better to factor them out and avoid the duplication.

### DIFF
--- a/lib/jxl/dec_frame.h
+++ b/lib/jxl/dec_frame.h
@@ -122,26 +122,11 @@ class FrameDecoder {
     return frame_header_.encoding == FrameEncoding::kVarDCT && finalized_dc_;
   }
 
-  // If the image has default exif orientation and no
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_frame.h
```diff
@@ -217,6 +192,18 @@ class FrameDecoder {
     if (use_task_id_) return task;
     return thread;
   }
+  bool CanDoRgbOutOrPixelCallback() const {
```
veluca93, line 195: Could also call it "CanDoLowMemoryPath()" or something like that :)

Can you please move/update the comment on MaybeSet...?

jonsneyers: Done.

veluca93 (approved):

PR #97, opened 2021-06-03
New JPEGXL_ENABLE_TRANSCODE_JPEG build flag to disable decoding to JPEG

deymo:
JPEG reconstruction in the API is not needed if we are only interested
in decoding to pixels. Disabling support for decoding to JPEG in the API
saves about 95 kB in libjxl_dec.so in an arm build, and about 50 kB in a
size-optimized arm build.


deymo:
@erikandre FYI, maybe you want to give this a try if you are not using the decode to JPEG bytes functionality from the API (`JXL_DEC_JPEG_RECONSTRUCTION` event).

lib/jxl/jpeg/jpeg_data.h
```diff
@@ -217,9 +217,15 @@ struct JPEGData : public Fields {
         has_zero_padding_bit(false) {}
 
   const char* Name() const override { return "JPEGData"; }
+#if JPEGXL_ENABLE_TRANSCODE_JPEG
   // Doesn't serialize everything - skips brotli-encoded data and what is
   // already encoded in the codestream.
   Status VisitFields(Visitor* visitor) override;
+#else
+  Status VisitFields(Visitor* /* visitor */) override {
+    return JXL_ABORT("JPEG transcoding support not enabled");
```
veluca93, line 226: either no return, or JXL_FAILURE?

lib/jxl/jpeg/jpeg_data.h
```diff
@@ -245,8 +251,15 @@ struct JPEGData : public Fields {
   std::vector<uint8_t> padding_bits;
 };
 
+#if JPEGXL_ENABLE_TRANSCODE_JPEG
 // Set ICC profile in jpeg_data.
 Status SetJPEGDataFromICC(const PaddedBytes& icc, jpeg::JPEGData* jpeg_data);
+#else
+static JXL_INLINE Status SetJPEGDataFromICC(const PaddedBytes& /* icc */,
+                                            jpeg::JPEGData* /* jpeg_data */) {
+  return JXL_ABORT("JPEG transcoding support not enabled");
```
veluca93, line 260: Also here

veluca93 (approved):

lib/jxl/jpeg/jpeg_data.h
deymo: If there's no JPEG support built in the idea is that we shouldn't be calling these functions from the API, so I'd prefer an abort here. We are certainly not calling them from the decode.cc.

jonsneyers: Wouldn't it be better to not have those API functions at all then? If someone would accidentally link an application that does need these functions against a libjxl_dec that doesn't have them, I'd rather have a link time error than a runtime abort, I think...

veluca93: Sure, but why the return ABORT then, instead of just ABORT?

deymo: clearly I should have used just JXL_ABORT without return..

In no case you will have a runtime abort when using the public API. The JXL_ABORT() here if reached means programming error. If you request the `JXL_DEC_JPEG_RECONSTRUCTION` it will simply never happen, and passing the output buffer will return error.

We can't really remove the functions from the decode.h in the API; unless we make this macro public in the cflags in the .pc file, for systems that use pkg-config, which is not a great solution. Having the functions in the API but not calling them at runtime (or calling them and reacting to the lack of `JXL_DEC_JPEG_RECONSTRUCTION` support) sounds like a valid use-case to me; otherwise you have no way to have a client code that can handle either case (you'd get linker errors).

The question in this patch is whether we want to return error if requesting JXL_DEC_JPEG_RECONSTRUCTION event when not supported or not.

lib/jxl/decode_to_jpeg.h
```diff
@@ -18,12 +18,17 @@
 
 #include "jxl/decode.h"
 #include "lib/jxl/base/status.h"
+#include "lib/jxl/common.h"  // JPEGXL_ENABLE_TRANSCODE_JPEG
 #include "lib/jxl/image_bundle.h"
 #include "lib/jxl/jpeg/dec_jpeg_data.h"
+#if JPEGXL_ENABLE_TRANSCODE_JPEG
 #include "lib/jxl/jpeg/dec_jpeg_data_writer.h"
```
lvandeve, line 25: It might be simpler to do the implementation of all the functions in the .cc file, and this inclusion in the cc file as well, so only in the cc file you have the macro toggle and this implementation related header inclusion.

But up to you

lvandeve (approved):

lib/jxl/decode_to_jpeg.h
deymo: I definitively want the disabled-case version to be inlined into decode.cc so all the code behind `if (functionThatReturnsFalse())` is removed by the compiler and we save on that size. For the other ones I kept the inline decisions in decode.cc and only moved the code here.

veluca93 (approved):

lib/jxl/jpeg/jpeg_data.h
jonsneyers: Oh, I understand now, thanks for the explanation. I misunderstood and thought it would do a runtime abort when an application tries to do a jpeg reconstruction.

I wouldn't return an error when subscribing to `JXL_DEC_JPEG_RECONSTRUCTION` with a libjxl_dec that doesn't support it. Applications cannot / should not expect all jxl input to be reconstructible to jpeg anyway, so they should not expect that event to certainly happen. By not returning an error, an application will simply consider all inputs to be only decodable to pixels. Returning an error is like more inconvenient for most applications (since they are subscribing to all events at once, and if that fails, they're likely going to give up instead of assuming it's missing jpeg reconstruction support).

That said, it would be good if there is some simple way for applications to know whether this feature is supported or not. In general, as we may add more events to subscribe to later, it would be good for an application to know which events it is actually subscribed to (in the sense that the event can actually happen with the libjxl it is using), instead of only getting a 'SUCCESS' or 'ERROR' after subscribing. Ideally, the SubscribeEvents function would return a bitmask describing the events that were actually subscribed to (which would be a subset of those requested, depending on capabilities of that build/version of the library). Since we can't just change that function now, maybe adding an inspector function that returns the bitmask of currently subscribed-to events would be best?
That way an application can detect that `JXL_DEC_JPEG_RECONSTRUCTION` was not _actually_ subscribed to, if it wants to know that.

deymo: Yeah, that makes sense. There's the question of forwards/backwards compatibility too. If you run your code against an older version of libjxl (same ABI) some events are not know to that libjxl (you are still able to pass them because for example you compiled your code against a newer header.h, but running against an old one which is legit unless you require symbols not in the other version or when the elf symbol has a different version number). Right now we don't return error for unknown bits in the subscribe-to-events calls, which is why I decided not to return error in this PR. But a new function with the mask of supported events makes sense.

PR #103, opened 2021-06-04
Update Highway to 0.12.2 +1

jan-wassenberg:
A fix for Windows.h filename case landed after that release.
This update is required for GCC builds on ARM, and Wasm.


deymo (approved):

eustas (approved):
LGTM

PR #106, opened 2021-06-04
Add a resizable parallel runner.

veluca93:
This parallel runner is also tuned for running on systems with
non-uniform computation capabilities, such as arm's big.LITTLE CPUs that
are often found in phones.

### DIFF
--- a/examples/decode_oneshot.cc
+++ b/examples/decode_oneshot.cc
@@ -16,8 +16,8 @@
 
 #include "jxl/decode.h"
 #include "jxl/decode_cxx.h"
-#include "jxl/thread_parallel_runner.h"
-#include "jxl/thread_parallel_runner_cxx.h"
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/resizable_parallel_runner_cxx.h"
 
 /** Decodes JPEG XL image to floating point pixels and ICC Profile. Pixel are
  * stored as floating point, as interleaved RGBA (4 floating point values per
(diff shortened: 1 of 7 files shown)


veluca93:
@erikandre FYI :)

examples/decode_oneshot.cc
```diff
@@ -16,8 +16,9 @@
 
 #include "jxl/decode.h"
 #include "jxl/decode_cxx.h"
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/resizable_parallel_runner_cxx.h"
 #include "jxl/thread_parallel_runner.h"
```
deymo, line 21: do you still need this include? I have the impression that we should use either one or the other.

lib/threads/resizable_parallel_runner.cc
```diff
@@ -0,0 +1,171 @@
...
+      num_running_workers_++;
+    }
+    while (true) {
+      uint32_t task = next_task_++;
+      if (task >= end_task_) {
+        std::unique_lock<std::mutex> l(m_);
+        num_running_workers_--;
+        work_available_ = false;
+        if (num_running_workers_ == 0) {
+          cv_.notify_all();
```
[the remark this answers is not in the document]
veluca93: Made this into two CVs.

lib/threads/resizable_parallel_runner.cc
```diff
@@ -0,0 +1,191 @@
...
+  // max_running_workers_)
+  std::condition_variable workers_can_proceed_;
+
+  // Workers are done, and the main thread can proceed (num_running_workers_ ==
+  // 0)
+  std::condition_variable work_done_;
+
+  std::vector<std::thread> workers_;
+
+  std::mutex state_mutex_;  // protects all the remaining variables.
```
deymo, line 135: func_, jxl_opaque_ and end_task_ are read outside the mutex in the workerBody so these are not protected by the mutex_ (which is fine). next_task_ is an atomic and also not protected by the mutex since it is accessed outside (also fine).

Only the last four are protected by the mutex.

deymo (approved):
LGTM

lib/threads/resizable_parallel_runner.cc
veluca93: Updated the comment.

deymo (approved):

PR #109, opened 2021-06-04
Split fuzzer builds to "asanfuzz" and "msanfuzz".

deymo:
The new ci.sh commands "asanfuzz" and "msanfuzz" are like "asan" and
"msan" but building all the code with -fsanitize=fuzzer-no-link.

This also allows to use a different set of flags when building the
fuzzers since the new JPEGXL_FUZZER_LINK_FLAGS cmake variable can be
changed to something else to use a different fuzzer driver.

Manually tested asanfuzz and msanfuzz.


veluca93 (approved):

eustas:
LGTM

PR #113, opened 2021-06-05
Do not try LZ77ing DC in VarDCT mode.

veluca93:
This fixes a speed regression introduced in bdde644b9.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -1053,11 +1053,12 @@ Status ModularFrameEncoder::EncodeGlobalInfo(BitWriter* writer,
                             : HistogramParams::ClusteringType::kFast;
     params.ans_histogram_strategy =
         HistogramParams::ANSHistogramStrategy::kApproximate;
-    params.lz77_method = cparams.decoding_speed_tier >= 3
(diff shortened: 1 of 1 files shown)


lib/jxl/enc_modular.cc
```diff
@@ -1053,11 +1053,12 @@ Status ModularFrameEncoder::EncodeGlobalInfo(BitWriter* writer,
                             : HistogramParams::ClusteringType::kFast;
     params.ans_histogram_strategy =
         HistogramParams::ANSHistogramStrategy::kApproximate;
-    params.lz77_method = cparams.decoding_speed_tier >= 3
-                             ? (cparams.speed_tier >= SpeedTier::kFalcon
-                                    ? HistogramParams::LZ77Method::kRLE
-                                    : HistogramParams::LZ77Method::kLZ77)
-                             : HistogramParams::LZ77Method::kNone;
+    params.lz77_method =
+        cparams.decoding_speed_tier >= 3 && cparams.modular_mode
```
jonsneyers, line 1057: Might make sense to at some point have different choices depending on what gets encoded. For DC likely lz77 is a bad idea, but for alpha it might make sense, for example. Anyway, that would be for a future improvement.

jonsneyers (approved):

PR #114, opened 2021-06-05
Avoid extra computation for empty histograms in clustering.

veluca93:
Some speedup, especially for small images with faster_decoding=4.

16.63 MP/s -> 17.01 MP/s

### DIFF
--- a/lib/jxl/enc_cluster.cc
+++ b/lib/jxl/enc_cluster.cc
@@ -77,38 +77,60 @@ float HistogramDistance(const Histogram& a, const Histogram& b) {
 
 // First step of a k-means clustering with a fancy distance metric.
 void FastClusterHistograms(const std::vector<Histogram>& in,
-                           const size_t num_contexts, size_t max_histograms,
+                           const size_t num_contexts_in, size_t max_histograms,
                            float min_distance, std::vector<Histogram>* out,
                            std::vector<uint32_t>* histogram_symbols) {
   PROFILER_FUNC;
   size_t largest_idx = 0;
-  for (size_t i = 0; i < num_contexts; i++) {
+  std::vector<uint32_t> nonempty_histograms;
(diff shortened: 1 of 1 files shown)


lib/jxl/enc_cluster.cc
```diff
@@ -77,38 +77,53 @@ float HistogramDistance(const Histogram& a, const Histogram& b) {
 
 // First step of a k-means clustering with a fancy distance metric.
 void FastClusterHistograms(const std::vector<Histogram>& in,
-                           const size_t num_contexts, size_t max_histograms,
+                           const size_t /*num_contexts*/, size_t max_histograms,
```
jonsneyers, line 80: Get rid of the argument instead of commenting it out?

lib/jxl/enc_cluster.cc
```diff
@@ -77,38 +77,53 @@ float HistogramDistance(const Histogram& a, const Histogram& b) {
...
   PROFILER_FUNC;
   size_t largest_idx = 0;
-  for (size_t i = 0; i < num_contexts; i++) {
+  std::vector<uint32_t> nonempty_histograms;
+  nonempty_histograms.reserve(in.size());
+  for (size_t i = 0; i < in.size(); i++) {
+    if (in[i].total_count_ == 0) continue;
     HistogramEntropy(in[i]);
     if (in[i].total_count_ > in[largest_idx].total_count_) {
       largest_idx = i;
```
jonsneyers, line 91: Set it to nonempty_histograms.size() so it immediately has the index for the right vector? Have to maintain the maxcount too then but I don't like that std::find below so much...

veluca93: Eh, I had this originally but it messed up things since I couldn't really do in[largest_idx] reliably... I prefer this version :)

jonsneyers (approved):

PR #118, opened 2021-06-07
Fix counterclockwise order for progressive

mo271:
Before this fix, the groups in the boundary of the concentric squares
when using middleout progression didn't appear consistently in
counterclockwise nor counterclockwise order, but somewhat random
depending on the position of the center of the image in relation to the
center of group which contains the center. This was due to the fact that
for images thati weren't with dimenions not a multiple of the group size,
the center wasi not computed correctly.
Now it works as intendent. Currently counterclockwise fills the boundary
of the next concentric square from a square on the left side of the
boundary.


veluca93:
I always knew I couldn't do maths :P

veluca93 (dismissed):

mo271:
The current behavior is kind of peculiar: for example for an image of the size 2848 x 4272, the order of the groups starts like this:

```
    17  9 10 11 12
    16  4  1  2 13
    19  5  0  3 14
    18  8  7  6 15
    24 23 22 21 20 
```
After this fix, the tile order will like this:
```
    10 11 12 13 14
     9  1  2  3 15
    24  8  0  4 16
    23  7  6  5 17
    22 21 20 19 18
``` 
I think this is what was intended. However this order can be improved further: Now the boundary of the next square starts on the square that has the smallest atan2 compared to the middle. This has two disadvantages: 

- The first two groups are like two diagonally adjacent squares (see places of `0` and `1` in the matrix above) and in general starting near the middle of the left hand side of the boundary of the next concentric squares, which yields weird shapes during filling up the boundary.
- It doesn't take into account the "true" center (as supposed to the center of the central group) of the image. It might be reasonable to start the progression there. 

I plan to come up with a strategy to fix these problems. Ideally, I think the progression should start in a group of the boundary of the new square which is next to the corner of the boundary and end at the corner. For larger squares, that gives 8 possible candidates to start the next square, taking into account both counter-clockwise and clockwise orders. Out of these 8 it might be optimal to take the one closest to the true center.

mo271:
Changed back to counterclockwise order and made sure that it always start with the next square exactly in the middle of the left side. So now the progression starts like this
```
    23 22 21 20 19 
    24  8  7  6 18
     9  1  0  5 17
    10  2  3  4 16
    11 12 13 14 15
```
This makes sure it looks good when only two groups are present. I plan to improve this further in an upcoming pull request.

veluca93 (approved):

jonsneyers:
> Changed back to counterclockwise order and made sure that it always start with the next square exactly in the middle of the left side. So now the progression starts like this
> 
> ```
>     23 22 21 20 19 
>     24  8  7  6 18
>      9  1  0  5 17
>     10  2  3  4 16
>     11 12 13 14 15
> ```
> 
> This makes sure it looks good when only two groups are present. I plan to improve this further in an upcoming pull request.

This seems better than the one that started diagonally, but wouldn't it make more sense to start the spiral upwards (so if there's a person/face, it's more likely to get the face/eyes earlier)? Also it might make sense to avoid going to diagonally adjacent groups also later on, e.g. in the example above to have 9 where 24 now is?

mo271:
> > Changed back to counterclockwise order and made sure that it always start with the next square exactly in the middle of the left side. So now the progression starts like this
> > ```
> >     23 22 21 20 19 
> >     24  8  7  6 18
> >      9  1  0  5 17
> >     10  2  3  4 16
> >     11 12 13 14 15
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This makes sure it looks good when only two groups are present. I plan to improve this further in an upcoming pull request.
> 
> This seems better than the one that started diagonally, but wouldn't it make more sense to start the spiral upwards (so if there's a person/face, it's more likely to get the face/eyes earlier)? Also it might make sense to avoid going to diagonally adjacent groups also later on, e.g. in the example above to have 9 where 24 now is?

Yes, I agree (and tried to explain that in the first comment.) I think the best starting places would be the groups `24`, `22`, `20`, `18`,`16`, `14`, `12`, or `10`  in the matrix above and then going in the direction such that it ends with a corner. Out of those groups one could choose the one closest to the true center. This would most often give a starting point near the bottom right for images of random dimensions. 

One might possibly think it is problematic going counter clockwise some of the time and clockwise some of the time, like a microwave, because it might be irritating or distractive. However, unless loading at extremely slow speeds, we do not expect to render all of the progression steps and hence the clockwise or counterclockwise motion will hardly be visible. 

In this pull request I mainly want to restore intended behavior. In addition to finding the best group to start with in each new square, I plan to provide a command line argument for an arbitrary starting point instead of the center for the central group in the future. When not providing such a starting point manually, we can later come up with more sophisticated methods of saliency detection to provide a high quality and non-disturbing group order during encoding instead of center-middleout.

PR #120, opened 2021-06-08
Remove overzealous chromaticity check

sboukortt:


lib/jxl/color_encoding_internal.cc
```diff
@@ -714,11 +714,8 @@ static const float kBradfordInv[9] = {
 
 // Adapts whitepoint x, y to D50
 Status AdaptToXYZD50(float wx, float wy, float matrix[9]) {
-  if (wx < 0 || wx > 1 || wy < 0 || wy > 1) {
-    return JXL_FAILURE("xy color out of range");
-  }
-
   float w[3] = {wx / wy, 1.0f, (1.0f - wx - wy) / wy};
+  JXL_RETURN_IF_ERROR(w[0] == w[0] && w[2] == w[2]);
```
jonsneyers, line 718: This is to check they're not NaN? Shouldn't it also be checked that they're not +-inf?

sboukortt: You are right, it’s likely best.

(`isfinite(NaN)` is false.)

veluca93 (approved):

PR #121, opened 2021-06-08
mark DC API functions as deprecated

lvandeve:


veluca93 (approved):

deymo (approved):

lib/include/jxl/decode.h
```diff
@@ -214,6 +215,7 @@ typedef enum {
    * status only indicates we're past this point in the codestream. This event
    * occurs max once per frame and always later than JXL_DEC_FRAME_HEADER
    * and other header events and earlier than full resolution pixel data.
+   * DEPRECATED: the DC feature in this form will be removed.
```
jonsneyers, line 218: Maybe add a comment here or elsewhere to indicate in favor of what it gets deprecated?
(I assume you now get an upsampled DC in the main image buffer instead of using a separate DC image?)

lvandeve: Done, added JxlDecoderFlushImage as alternative. Note that we need to add functionality to allow flushing exactly at DC later

lvandeve:
Had to also update for djxl_fuzzer's warning, PTAL

veluca93 (approved):

PR #123, opened 2021-06-08
add cjxl option --keep_invisible

jonsneyers:
Allows preserving the color of invisible pixels in lossy mode (`--keep_invisible=1`) or forgetting the color of invisible pixels in lossless mode (`--keep_invisible=0`). For now just sets invisible pixels to black in lossless mode (probably something better is possible, but the smearing we're doing in lossy mode is not it).

### DIFF
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -399,7 +399,7 @@ Status MakeFrameHeader(const CompressParams& cparams,
 // edge duplication but not more. It would probably be better to smear in all
 // directions. That requires an alpha-weighed convolution with a large enough
 // kernel though, which might be overkill...
-void SimplifyInvisible(Image3F* image, const ImageF& alpha) {
+void SimplifyInvisible(Image3F* image, const ImageF& alpha, bool lossless) {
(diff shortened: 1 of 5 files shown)


jonsneyers:
I'll rebase once the other one lands, trying to avoid merge conflicts.

lib/jxl/enc_frame.cc
```diff
@@ -1161,15 +1165,15 @@ Status EncodeFrame(const CompressParams& cparams_orig,
       CopyImageTo(ib.color(), &opsin);
     }
     if (ib.HasAlpha() && !ib.AlphaIsPremultiplied() &&
-        (frame_header->encoding == FrameEncoding::kVarDCT ||
-         cparams.quality_pair.first < 100) &&
-        !cparams.keep_invisible &&
+        cparams.keep_invisible == Override::kOff &&
```
veluca93, line 1168: there's an ApplyOverride method that might be better

jonsneyers: Right, that way the default logic can be put in enc_frame instead of in cjxl. Updated.

veluca93 (approved):

PR #125, opened 2021-06-08
Add instructions to revert unintended submodule changes.

deymo:
When bash_test fails in the test_deps_version() print instructions on
how to revert the change if it was unintended.


jonsneyers:
That's useful! :)

jonsneyers (approved):

PR #127, opened 2021-06-08
Silence deprecation warnings when building with gcc

deymo:
This fixes the post-submit checks for the v0.4.x branch.


jonsneyers (approved):

deymo:
checked that the v0.4.x branch passes post-submit tests.

PR #128, opened 2021-06-09
Fix use of undefined value in 2x upsampling.

eustas:
Multiplication by 0 does not save from mixing-in undefined values,
because NaN is toxic. Added special "unlikely" case for the tail
of input in case X-repeat is in action (only 2x upsampling, and
vector size >4).


eustas:
Woo-hoo PR number 10000000

veluca93 (approved):

PR #133, opened 2021-06-09
Run CIFuzz continuous integration on PRs.

deymo:
oss-fuzz provides a GitHub CI integration that would run the fuzzer on a
given pull request for a period of time. The minimum recommended time is
600 seconds, which is the value included in this commit.

See the following link for details:

  https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,52 @@
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# CI on pull-requests to run the fuzzer from oss-fuzz. See:
+#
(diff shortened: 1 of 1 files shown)


veluca93 (dismissed):

eustas (dismissed):

jonsneyers:
Would be cool if somehow the fuzzer could be configured to focus on code paths that involve code touched by the PR, but I guess that's not an easy thing to do.

jonsneyers:
It might be worthwhile to only do this (and also the build/test CI) when changes are made to the code (`**.cc`, `**.h`), as described here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#branches-and-paths

Currently this also gets run when only making changes to documentation or fixing a typo in the README.md, which is a bit unnecessary...

veluca93 (approved):

veluca93 (approved):

PR #137, opened 2021-06-09
Fix clang-tidy issue: alpha can be nullptr

lvandeve:


jonsneyers (approved):

jonsneyers:
Shouldn't really be necessary, since `AlphaIsPremultiplied()` returns false if there is no alpha, but doesn't hurt to check explicitly in case the implementation of `AlphaIsPremultiplied()` should change at some point...

lvandeve:
> Shouldn't really be necessary, since `AlphaIsPremultiplied()` returns false if there is no alpha, but doesn't hurt to check explicitly in case the implementation of `AlphaIsPremultiplied()` should change at some point...

Indeed, AlphaIsPremultiplied() returns false if there's no alpha channel, so this was indeed not an actual bug, but indeed not a bad thing to fix this, and this should make clang-tidy's status green again

jan-wassenberg (approved):

jan-wassenberg:
Thanks for looking into this :)

deymo:
In cases like this it would be nice if you can add that context to the commit message so when we are looking at whether we need to cherry-pick commits we don't have to dig into the discussions from the PR to get the context. Thanks.

PR #142, opened 2021-06-10
don't use ClampToRange in modular

jonsneyers:
Don't do unnecessary clamping and use of wider types in modular transforms and in the weighted predictor.

Small speedup of 1-4%, for example:

Modular falcon:
before: 4064 x 2704, geomean: 24.53 MP/s [21.31, 24.99], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 24.77 MP/s [21.54, 25.62], 30 reps, 4 threads.

Modular default speed:
before: 4064 x 2704, geomean: 14.51 MP/s [13.62, 14.76], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 14.61 MP/s [13.73, 14.99], 30 reps, 4 threads.

Modular default speed -Q 80:
before: 4064 x 2704, geomean: 17.86 MP/s [15.21, 18.50], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 18.13 MP/s [14.79, 18.53], 30 reps, 4 threads.

Modular thunder:
before: 4064 x 2704, geomean: 49.24 MP/s [40.27, 51.04], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 51.29 MP/s [42.49, 53.30], 30 reps, 4 threads.

### DIFF
--- a/lib/jxl/modular/transform/squeeze.h
+++ b/lib/jxl/modular/transform/squeeze.h
@@ -111,8 +111,8 @@ void InvHSqueeze(Image &input, int c, int rc, ThreadPool *pool) {
         pixel_type_w A =
             ((avg * 2) + diff + (diff > 0 ? -(diff & 1) : (diff & 1))) >> 1;
         pixel_type_w B = A - diff;
-        p_out[0] = ClampToRange<pixel_type>(A);
-        p_out[1] = ClampToRange<pixel_type>(B);
+        p_out[0] = A;
+        p_out[1] = B;
 
         for (size_t x = 1; x < chin_residual.w; x++) {
           pixel_type_w diff_minus_tendency = p_residual[x];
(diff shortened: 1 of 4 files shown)


lib/jxl/modular/transform/squeeze.h
```diff
@@ -104,28 +103,28 @@ void InvHSqueeze(Image &input, int c, int rc, ThreadPool *pool) {
...
         // special case for x=0 so we don't have to check x>0
-        pixel_type_w avg = p_avg[0];
-        pixel_type_w next_avg = (1 < chin.w ? p_avg[1] : avg);
-        pixel_type_w tendency = SmoothTendency(avg, avg, next_avg);
-        pixel_type_w diff = p_residual[0] + tendency;
-        pixel_type_w A =
+        pixel_type avg = p_avg[0];
+        pixel_type next_avg = (1 < chin.w ? p_avg[1] : avg);
+        pixel_type tendency = SmoothTendency(avg, avg, next_avg);
+        pixel_type diff = p_residual[0] + tendency;
```
veluca93, line 109: this could overflow in theory, no? We probably should cast one of the two to a wider type...

veluca93: (also below)

veluca93:
Can you also remove ClampToRange from common.h?

lib/jxl/modular/transform/squeeze.h
jonsneyers: The calculation in SmoothTendency() could certainly overflow for >29-bit values, and probably things here too.

With > 24-bit integers we have a problem anyway though (converting them to float is lossy), so I'm not sure if it really makes sense to prevent overflow here if we can't really handle it anyway. On older 32-bit devices the widening probably does have a more significant cost...

veluca93: Well, signed integer overflow is UB and sanitizers/fuzzers complain about it, so doing something else there would be useful... perhaps just casting to u32, doing the arithmetic and then going back to i32 is enough.

jonsneyers: Doing that for RCT, reverting to wider arithmetic in Squeeze...

veluca93 (approved):

PR #149, opened 2021-06-10
Refactor Modular to be able to output rects.

veluca93:


veluca93:
This now works, but disables some parallelization - shouldn't be submitted until a follow-up is ready.

jonsneyers (approved):

jonsneyers:
Now https://github.com/libjxl/libjxl/pull/205 has merged, what is still needed from this one?

The functions `GetColorRect()` and `GetECRect()` are basically the same thing as `ModularImageToDecodedRect()` from https://github.com/libjxl/libjxl/pull/205, except they split things up in color and EC - but I don't think that matters much. 

There are two flows now: the old flow where modular uses full image (and for now it still does a full int-to-float copy in FinalizeDecoding, but that could be replaced by rect copies), and the new flow (only if there are no transforms besides RCT) where there is no modular full image and the int-to-float copy happens per rect immediately after decoding the rect. I suppose the new flow can only be used if there is a decoded full-image buffer (not with the pixel callback). The old flow can output rects now but I guess FinalizeDecode() should be changed to only undo transforms, right? And then I suppose there should be some way to let dec_frame.cc know whether we're in the new flow or in the old flow (because in the new flow, it shouldn't try to do the int-to-float copy, since it is already done). Does that make sense?

veluca93:
> Now #205 has merged, what is still needed from this one?
> 
> The functions `GetColorRect()` and `GetECRect()` are basically the same thing as `ModularImageToDecodedRect()` from #205, except they split things up in color and EC - but I don't think that matters much.
> 
> There are two flows now: the old flow where modular uses full image (and for now it still does a full int-to-float copy in FinalizeDecoding, but that could be replaced by rect copies), and the new flow (only if there are no transforms besides RCT) where there is no modular full image and the int-to-float copy happens per rect immediately after decoding the rect. I suppose the new flow can only be used if there is a decoded full-image buffer (not with the pixel callback). The old flow can output rects now but I guess FinalizeDecode() should be changed to only undo transforms, right? And then I suppose there should be some way to let dec_frame.cc know whether we're in the new flow or in the old flow (because in the new flow, it shouldn't try to do the int-to-float copy, since it is already done). Does that make sense?

This seems about right, although we should eventually (unless some transforms are enabled) output to a float *rect* :)

jonsneyers:
Not sure what the status of this one is, is it still relevant?

Issue #150, opened 2021-06-10
fuzzer_corpus hits an assert


deymo:
**Describe the bug**
fuzzer_corpus fails on an assert. This failure was introduced by https://github.com/libjxl/libjxl/pull/25

**To Reproduce**
```bash
mkdir corpusdir
build/tools/fuzzer_corpus -j 0 corpusdir
```

Result:
```
Generating ImageSpec<size=8x8 * chan=3 depth=8 alpha=0 (premult=0) x frames=1 seed=407987, speed=1, butteraugli=1.5, modular_mode=0, lossy_palette=0, noise=0, preview=0, fuzzer_friendly=0, is_reconstructible_jpeg=1, orientation=4> as a6abe1aab9970bb0771cf78b376749e2
../lib/jxl/enc_frame.cc:1066: JXL_ASSERT: metadata->m.xyb_encoded == (cparams.color_transform == ColorTransform::kXYB)
Illegal instruction
```

**Environment**
ci.sh release build; Linux build.

lvandeve:
The cause is the addition of the following in #25:

```
  if (ib.color_transform != ColorTransform::kNone) {
    cparams.color_transform = ib.color_transform;
  }
```

in enc_frame.cc: this overwrites the xyb color transform for xyb_encoded

lvandeve:
Fixed with https://github.com/libjxl/libjxl/pull/151

PR #151, opened 2021-06-10
Fix for https://github.com/libjxl/libjxl/issues/150

lvandeve:
Overriding the color transform in enc_frame can an xyb_encoded <->
xyb mismatch assert to fail.

However, overriding color transform is still necessary to encode other
images correctly with correct color, such as y4m. Do this earlier, in
enc_file, now.

### DIFF
--- a/lib/jxl/enc_file.cc
+++ b/lib/jxl/enc_file.cc
@@ -185,11 +185,17 @@ Status WriteHeaders(CodecMetadata* metadata, BitWriter* writer,
   return true;
 }
 
-Status EncodeFile(const CompressParams& cparams, const CodecInOut* io,
+Status EncodeFile(const CompressParams& cparams_orig, const CodecInOut* io,
                   PassesEncoderState* passes_enc_state, PaddedBytes* compressed,
                   AuxOut* aux_out, ThreadPool* pool) {
   io->CheckMetadata();
   BitWriter writer;
+
+  CompressParams cparams = cparams_orig;
+  if (io->Main().color_transform != ColorTransform::kNone) {
+    // Set the color transform to YCbCr or XYB if the original image is such.
+    cparams.color_transform = io->Main().color_transform;
+  }
 
   std::unique_ptr<CodecMetadata> metadata = jxl::make_unique<CodecMetadata>();
   JXL_RETURN_IF_ERROR(PrepareCodecMetadataFromIO(cparams, io, metadata.get()));
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -1016,9 +1016,6 @@ Status EncodeFrame(const CompressParams& cparams_orig,
   passes_enc_state->special_frames.clear();
 
   CompressParams cparams = cparams_orig;
-  if (ib.color_transform != ColorTransform::kNone) {
-    cparams.color_transform = ib.color_transform;
-  }
 
   if (cparams.progressive_dc < 0) {
     if (cparams.progressive_dc != -1) {


lvandeve:
Hi Jon, assigning this to you since you likely remember what the original reason for adding this code was and what alternative fix needs to be done for y4m. Please take a look, thank you :)

lib/jxl/enc_frame.cc
```diff
@@ -1016,9 +1016,6 @@ Status EncodeFrame(const CompressParams& cparams_orig,
   passes_enc_state->special_frames.clear();
 
   CompressParams cparams = cparams_orig;
-  if (ib.color_transform != ColorTransform::kNone) {
```
jonsneyers, line 1019: Basically the idea is that if the input image is not RGB but it is already something else (YCbCr or XYB), then we need to override the color_transform choice in `cparams` to match what we already have.
The y4m input relies on this.

The problem with this is that it is done too late - at the frame level, while it should have been at the image level, because it is too late here to change xyb_encoded. This is not an issue for y4m or RGB input, but would be a problem if e.g. the original cparams says kNone and the input imagebundle is (somehow) in XYB already.

lvandeve: Do you know of a case that reproduces the issue where the value in cparams doesn't match the input image's YCbCr or XYB? I could try to move this to, for example, enc_file.cc, but would be good to have a way to check that it works

jonsneyers: `ffmpeg -i some_image.png input.y4m`

Then try `cjxl input.y4m -q 100`.

lvandeve: That worked (with -q 100 indeed, without it doesn't but that is an unrelated issue). I verified that without overriding color transform, the colors after decoding look wrong, and that moving this assignment to enc_file, fixes that, and also fixes fuzzer_corpus.

lvandeve:
Actual fix implemented now, PTAL

lib/jxl/dec_frame.h
```diff
@@ -104,6 +104,11 @@ class FrameDecoder {
   // Must be called exactly once per frame, after all calls to ProcessSections.
   Status FinalizeFrame();
 
+  // Returns dependencies of this frame as a bit mask: bits 0-3 indicate
+  // reference frame 0-3 for patches and blending, bits 4-7 indicate DC
+  // frames this frame depends on. Must be called
+  int GetDependencies() const;
```
jonsneyers, line 110: this is likely unintentional / from another branch, right?

lib/jxl/enc_file.cc
```diff
@@ -185,12 +185,18 @@ Status WriteHeaders(CodecMetadata* metadata, BitWriter* writer,
...
+Status EncodeFile(const CompressParams& cparams_orig, const CodecInOut* io,
                   PassesEncoderState* passes_enc_state, PaddedBytes* compressed,
                   AuxOut* aux_out, ThreadPool* pool) {
   io->CheckMetadata();
   BitWriter writer;
 
+  CompressParams cparams = cparams_orig;
+  if (io->Main().color_transform != ColorTransform::kNone) {
+    // Set the color transform to YCbCr or XYB if the original image is such.
+    cparams.color_transform = io->Main().color_transform;
```
jonsneyers, line 197: For now this is good enough, but in theory different frames could have a different color transform (not with anything cjxl will pass to EncodeFile() so it's only in theory).

For example, at some point we might want to do something like inputting both a JPEG (YCbCr, recompressed using VarDCT) and a PNG (RGBA, recompressed using Modular) to produce a JXL that has the JPEG photo as the bottom layer and the PNG overlay blended on top of it...

jonsneyers:
LGTM, just got something unrelated in dec_frame.h that ended up in the commit somehow.

lib/jxl/dec_frame.h
lvandeve: oops, removed from here

lib/jxl/enc_file.cc
lvandeve: EncodeFile is not actually used by the encoding API, for encoder API the user should be able to provide the format a frame has, and if this feature is implemented in the cjxl tool it can set it correctly for the frames. Since we have only 1 color transform field in cparams, something is definitely wrong currently for supporting multiple frames with different color transforms. (my guess is that cparams shouldn't have a color_transform field if this is a per-frame property). But that's unrelated to this MR, which is about fixing fuzzer_corpus :)

lvandeve:
All fixed now

jonsneyers (approved):

PR #157, opened 2021-06-11
add test for modular encoding of ycbcr 4:2:0

jonsneyers:
Adds a test for modular encoding of YCbCr 4:2:0 input.


jonsneyers:
The test does currently fail because there is in fact currently a bug.

jonsneyers:
Test should succeed now. It was not just a bug in the upsampling, but also I accidentally made a "tv range" y4m file while the y4m reader assumes things to be "full range", so that didn't quite match up :)

(for the record: if you use ffmpeg to make a y4m, use `-pix_fmt yuvj420p`, with the `j` for "jpeg's full range", I guess)

veluca93 (approved):

veluca93 (approved):

PR #161, opened 2021-06-11
fix *san complaining about 1 << 32

jonsneyers:
Asan/msan tests fail because `1u << 32` has a too large shift. This only happens iff the result is not used anyway, but let's keep asan/msan happy.


veluca93 (approved):

deymo:
@jonsneyers  asan still complaining:
```
[ RUN      ] RoundtripTest.FloatFrameRoundtripTest
../lib/jxl/dec_modular.cc:452:33: runtime error: shift exponent 32 is too large for 32-bit type 'unsigned int'
```

PR #162, opened 2021-06-11
pull modular fwd transforms out of libjxl_dec

jonsneyers:
The decoder-only library contained the Modular forward transforms. This should fix that.


veluca93 (approved):
Does this remove the symbols from libjxl_dec?

jonsneyers:
> Does this remove the symbols from libjxl_dec?

It does for me. Maybe @deymo can check what the effect is on Android chrome binary size?

deymo:
I tested a chrome android-binary-size build and I see no difference (+0.22kb total compared to a ~week ago). It is still useful to do this since it removes those lines from the decoder coverage stats (they are compiled but never reached by the decoder tests/fuzzer).

lib/jxl/modular/transform/enc_transform.cc
```diff
@@ -0,0 +1,38 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "lib/jxl/fields.h"
+#include "lib/jxl/modular/modular_image.h"
+#include "lib/jxl/modular/transform/transform.h"
+
+namespace jxl {
+
+Status FwdRCT(Image &input, size_t begin_c, size_t rct_type);
```
deymo, line 12: this normally goes in a enc_rct.h header. Elsewhere we have the enc_foo.h header include the dec_foo.h header or just foo.h header if you want.

jonsneyers: Yes, that would be the logical thing to do, but I felt it was a bit overkill to add 3 more .h files each with just 1 function prototype in them (and that being unlikely to change). I'll do it if you insist though :)

veluca93: Could also do enc_transforms.h and dec_transforms.h and put everything in those :)

jonsneyers: I prefer keeping the 3 transforms in separate files, it's a bit much code to put it all together, easier to edit this way.

veluca93:
I meant having a single .h file for the declarations, but splitting the .cc
files as they are now :)

lib/jxl/modular/transform/enc_transform.cc
deymo: It helps with understanding where are things if you have the .h file, specially if you are unfamiliar with the project. I'd find it surprising that enc_transform.cc depends on enc_rct.cc but there's no include of enc_rct.h; when dealing with stripping out dependencies I move around code and you use these explicit dependencies to understand what depends on what and what can be removed or moved to another library.

jonsneyers: OK, makes sense.

jonsneyers: Better now?

deymo (dismissed):

lib/jxl/modular/transform/enc_transform.cc
deymo: thanks

jonsneyers:
Please approve again, @deymo or @veluca93 (needed clang-format on some files).

veluca93 (approved):

PR #167, opened 2021-06-14
Move changelog to a CHANGELOG.md

deymo:
The debian/changelog file is for the Debian package, which should list
changes made to the debian package, like changed flags or dependencies,
added patches to fix existing versions, etc. In our case, we normally
only just bump the version of the library and make no changes to the
package, so "Bump JPEG XL version to ..." is an appropriate message
there.

This patch moves all the available Changelog information to a new
CHANGELOG.md file following the guidelines from
https://keepachangelog.com/en/1.0.0/ . This will allow us to keep track
of notable changes when they land and make the release process easier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,107 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
+ - API: New function to decode the image using a callback outputting a part of a
+   row per call.
+ - API: 16-bit float output support.
+
+### Changed
+ - Full JPEG XL part 1 specification support: Implemented all the spec required
+   to decode files to pixels, including cases that are not used by the encoder
+   yet. Part 2 of the spec (container format) is final but not fully implemented
+   here.
+ - Butteraugli metric improvements. Exact numbers are different from previous
+   versions.
+ - Memory reductions during decoding.
+ - Reduce the size of the jxl_dec library by removing dependencies.
+ - A few encoding speedups.
+ - Clarify the security policy.
+
+### Removed
+- API for returning a downsampled DC was deprecated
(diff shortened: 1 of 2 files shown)


sboukortt (dismissed):

CHANGELOG.md
```diff
@@ -0,0 +1,105 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
+ - API: New function to decode the image using a callback outputting a part of a
+   row per call.
+ - API: 16-bit float output support.
+
+### Changed
+ - Full specification support: Implemented all the spec, including cases that
```
jonsneyers, line 15: Might want to clarify this is about the codestream spec (part 1). There are some things in part 2 (in particular, the `brob`) that are not yet implemented. Not needed to decode to pixels, but can be needed to reconstruct a jpeg file.

deymo: wdyt now?

jonsneyers: LGTM.

CHANGELOG.md
```diff
@@ -0,0 +1,107 @@
...
+### Changed
+ - Full JPEG XL part 1 specification support: Implemented all the spec required
+   to decode files to pixels, including cases that are not used by the encoder
+   yet. Part 2 of the spec (container format) is final but not fully implemented
+   here.
+ - Butteraugli metric improvements. Exact numbers are different from previous
+   versions.
+ - Memory reductions during decoding.
+ - Reduce the size of the jxl_dec library by removing dependencies.
+ - A few encoding speedups.
```
jonsneyers, line 23: Also I guess some (minor) decode speedups?

Could also mention the new speeds `thunder` and `lightning`, and the deprecation of `cjxl --speed` in favor of `cjxl --effort` (where it makes more sense that 3 is fast and 9 is slow).

deymo: I'll add this in a follow up. We should have done this changelog.md before; and so far all of this goes to v0.4.x which I need to cherrypick; I'll work on adding the new stuff in main but not in v0.4.x (like the rewind api) on a follow up.

jonsneyers (approved):

PR #169, opened 2021-06-14
Backport fixes to v0.4.x

deymo:
* Cherry-pick fixes to fuzzer-found bugs and other critical failures.
* Cherry-pick build-time flag to disable JPEG decoding and ASSERT messages.


jonsneyers:
@googlebot I consent.

deymo:
@eustas for some reason the bot wants your consent to cherry-pick from main.

deymo:
One more tiny fix to make the CI happy.

veluca93 (approved):

veluca93 (approved):

eustas:
@googlebot I consent.

eustas (approved):

eustas:
Fuzzer bug fixes:

1. "Fuzz: Undefined-shift in jxl::ModularFrameDecoder::FinalizeDecoding" - inverse palette transform creates channels with same h/v shift values; check that meta-transform is applied to channels with same h/v shift values, otherwise, at some moment negative shift value might occur; security risk: low
2. "Fuzz: Integer-overflow in jxl::palette_internal::GetPaletteValue" - another edge case with value scaling (replacement for division); security risk: very low
3. "Fuzz: fix DASSERT in FinalizeFrameDecoding" - fix "src" region size; though upsampling is driven by destination region size, source region size is also checked; old one didn't took into account that upsampler takes care of boundary mirroring; security risk: very low
4. "Remove NaN check" - ideally NaN should not appear in output; though there are legal inputs that produce outrageously large output values; it is not yet clear if / when the outrageous values should be clamped; library security risk: none; client security risk is still in place, e.g. if client does not tolerate NaN values, handle them unsafely
5. "Fuzz: fix heap-buffer-overflow in codec_psd" - *encoder* some inputs could cause encoder to touch regions outside of vector; security risk: medium
6. "Fuzz: set_from_bytes timeout" - *encoder* disable support of reading arithmetic-encoded JPEGs; those might contain CPU-zip-bomb; security risk: high

PR #170, opened 2021-06-14
Properly consider orientation in *OutBufferSize.

veluca93:
Otherwise, mismatches in computed strides can result in memory
corruption and/or OOB writes.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -1820,9 +1820,8 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderPreviewOutBufferSize(
   JxlDecoderStatus status = PrepareSizeCheck(dec, format, &bits);
   if (status != JXL_DEC_SUCCESS) return status;
 
-  const auto& metadata = dec->metadata.m;
-  size_t xsize = metadata.preview_size.xsize();
-  size_t ysize = metadata.preview_size.ysize();
+  size_t xsize = dec->metadata.oriented_preview_xsize(dec->keep_orientation);
+  size_t ysize = dec->metadata.oriented_preview_ysize(dec->keep_orientation);
 
   size_t row_size =
       jxl::DivCeil(xsize * format->num_channels * bits, jxl::kBitsPerByte);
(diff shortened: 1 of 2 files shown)


lib/jxl/decode.cc
```diff
@@ -1887,12 +1887,11 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderImageOutBufferSize(
   JxlDecoderStatus status = PrepareSizeCheck(dec, format, &bits);
   if (status != JXL_DEC_SUCCESS) return status;
 
-  size_t row_size =
-      jxl::DivCeil(dec->metadata.size.xsize() * format->num_channels * bits,
-                   jxl::kBitsPerByte);
+  size_t row_size = dec->metadata.size.xsize();
   if (format->align > 1) {
-    row_size = jxl::DivCeil(row_size, format->align) * format->align;
+    row_size = jxl::RoundUpTo(row_size, format->align);
```
lvandeve, line 1892: the user documentation at `JxlPixelFormat` in types.h says that the alignment is in bytes. It seems to perform the alignment before multiplying with num_channels here. Does this match the API documentation now?

lib/jxl/decode.cc
```diff
@@ -1887,12 +1887,11 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderImageOutBufferSize(
   JxlDecoderStatus status = PrepareSizeCheck(dec, format, &bits);
   if (status != JXL_DEC_SUCCESS) return status;
 
-  size_t row_size =
-      jxl::DivCeil(dec->metadata.size.xsize() * format->num_channels * bits,
-                   jxl::kBitsPerByte);
```
lvandeve, line 1892: there's similar code in JxlDecoderPreviewOutBufferSize and JxlDecoderDCOutBufferSize, it should probably be changed the same way, if this change is needed (see other comment)

veluca93:
PTAL, updated code.

lib/jxl/image_metadata.h
```diff
@@ -377,6 +377,34 @@ struct CodecMetadata {
 
   size_t xsize() const { return size.xsize(); }
   size_t ysize() const { return size.ysize(); }
+  size_t oriented_xsize(bool keep_orientation) const {
```
lvandeve, line 380: not a must, but if you want this new function could also be used in GetBasicInfo

lvandeve (approved):

lvandeve:
LGTM, thanks for the fix!

PR #171, opened 2021-06-14
Disable JNI test on xSAN builds

deymo:
For proper xSAN run JVM needs to be driven by special wrapper.
There are no out-of-the box solutions, so currently test
is disabled to avoid unnecessary complexity.


deymo:
This will fix the jni error in *san in v0.4.x.

eustas:
@googlebot I consent

eustas (approved):

PR #175, opened 2021-06-15
simplify modular channel resize()

jonsneyers:
The modular channel resize() function was copying and/or initializing pixels, for no good reason really. In the encoder there was one spot where copying was needed, the other instances of resize() were either just doing a shrinking of the dimensions with no need to initialize things, or a way to make sure the channel plane is actually allocated (in case of Squeeze, that wasn't always the case).

Reorganized things so this complicated resize() function is no longer needed. Channel planes are now always allocated immediately (change in MetaSqueeze), and explicit zeroing is only done when needed (when a modular decode is interrupted because of progressive: the not-yet-decoded channels all need to be zeroed then, not just the ones that weren't allocated yet — this fixes a potential uninitialized memory bug in that case).

This slightly speeds up modular decoding, in particular in the cases where squeeze was used:
before: 4064 x 2704, geomean: 23.36 MP/s [19.60, 24.32], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 24.53 MP/s [20.48, 25.28], 30 reps, 4 threads.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -522,7 +522,7 @@ Status ModularFrameEncoder::ComputeEncodingData(
             enc_state->shared.frame_header.chroma_subsampling.VShift(c);
         size_t xsize_shifted = DivCeil(xsize, 1 << gi.channel[c_out].hshift);
         size_t ysize_shifted = DivCeil(ysize, 1 << gi.channel[c_out].vshift);
-        gi.channel[c_out].resize(xsize_shifted, ysize_shifted);
+        gi.channel[c_out].shrink(xsize_shifted, ysize_shifted);
         for (size_t y = 0; y < ysize_shifted; ++y) {
           const float* const JXL_RESTRICT row_in = color->PlaneRow(c, y);
           pixel_type* const JXL_RESTRICT row_out = gi.channel[c_out].Row(y);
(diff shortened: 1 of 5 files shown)


lib/jxl/enc_modular.cc
```diff
@@ -1584,7 +1584,6 @@ void ModularFrameEncoder::AddACMetadata(size_t group_index, bool jpeg_transcode,
     }
   }
   image.channel[2].w = num;
-  image.channel[2].resize();
```
jonsneyers, line 1587: This is the one spot where the values are already in the plane and need to be kept — just keeping the larger plane and changing its width also does the job (doesn't release the unused acs/qf values, but that likely doesn't really make any difference for encode memory usage).

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -479,6 +478,7 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
     if (!br->AllReadsWithinBounds()) {
       if (!allow_truncated_group) return JXL_FAILURE("Truncated input");
       ZeroFillImage(&channel.plane);
+      while (++i < nb_channels) ZeroFillImage(&image.channel[i].plane);
```
jonsneyers, line 481: This is a bugfix I think — there could be cases where channels were not initialized nor decoded, yet their values would be used when showing a progressive preview in the `allow_truncated_group == true` case.

veluca93 (approved):

PR #177, opened 2021-06-15
Add reference getters to dec_frame

lvandeve:
Add getters for which frame storage locations are referenced, and
referencable by, frames. This is for use for more optimal rewinding in
the API, but it is not implemented yet there.

Also added a test for a jxl image with patches to decode_test, to
prepare for when this feature is added to the API.

### DIFF
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -827,6 +827,67 @@ Status FrameDecoder::Flush() {
   return true;
 }
 
+int FrameDecoder::SavedAs() const {
+  if (is_finalized_) {
+    // header not parsed
+    return 0;
+  }
+  if (frame_header_.frame_type == FrameType::kDCFrame) {
+    // bits 16, 32, 64, 128 for DC level
+    return 16 << (frame_header_.dc_level - 1);
+  } else if (frame_header_.CanBeReferenced()) {
+    // bits 1, 2, 4 and 8 for the references
+    return 1 << frame_header_.save_as_reference;
+  }
+
+  return 0;
+}
+
+int FrameDecoder::References() const {
+  if (is_finalized_) {
+    return 0;
+  }
+  if ((!decoded_dc_global_ || !decoded_ac_global_ ||
+       *std::min_element(decoded_dc_groups_.begin(),
+                         decoded_dc_groups_.end()) != 1 ||
+       *std::min_element(decoded_passes_per_ac_group_.begin(),
+                         decoded_passes_per_ac_group_.end()) < max_passes_)) {
+    return 0;
+  }
+
+  int result = 0;
+
+  // Blending
+  if (frame_header_.frame_type == FrameType::kRegularFrame ||
+      frame_header_.frame_type == FrameType::kSkipProgressive) {
+    bool cropped = frame_header_.custom_size_or_origin;
+    if (cropped || frame_header_.blending_info.mode != BlendMode::kReplace) {
+      result |= (1 << frame_header_.blending_info.source);
+    }
+    const auto& extra = frame_header_.extra_channel_blending_info;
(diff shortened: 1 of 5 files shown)


lib/jxl/dec_frame.cc
```diff
@@ -827,6 +827,63 @@ Status FrameDecoder::Flush() {
...
+                         decoded_dc_groups_.end()) != 1 ||
+       *std::min_element(decoded_passes_per_ac_group_.begin(),
+                         decoded_passes_per_ac_group_.end()) < max_passes_)) {
+    return 0;
+  }
+
+  int result = 0;
+
+  // Blending
+  if (frame_header_.blending_info.mode != BlendMode::kReplace) {
```
jonsneyers, line 861: Also if it is kReplace, there can be a dependency if the frame is not a full frame but a cropped one.

lvandeve: Is this the case as soon as `custom_size_or_origin` is true, or only if `custom_size_or_origin` is true and the frame_size and frame_origin are an actual crop?

veluca93: Second one - but first one likely works too...

There's also a NeedsBlending() method somewhere.

lib/jxl/dec_frame.h
```diff
@@ -104,6 +104,18 @@ class FrameDecoder {
   // Must be called exactly once per frame, after all calls to ProcessSections.
   Status FinalizeFrame();
 
+  // Returns dependencies of this frame on reference ids as a bit mask: bits 0-3
+  // indicate reference frame 0-3 for patches and blending, bits 4-7 indicate DC
+  // frames this frame depends on. Only returns a valid rsult after all calls to
```
veluca93, line 109: result

lib/jxl/dec_frame.h
```diff
@@ -104,6 +104,18 @@ class FrameDecoder {
...
+  // ProcessSections are finished and before FinalizeFrame.
+  int References() const;
+
+  // Returns reference id of storage location where this frame is stored as a
+  // bit flag, or 0 if not stored.
+  // Matches the bit mask used for GetReferences: bits 0-3 indicate it is stored
+  // for patching or blending, bits 4-7 indicate DC frame.
+  int ReferencedBy() const;
```
veluca93, line 117: maybe SavedAs? ReferencedBy means something a bit different.

lib/jxl/decode_test.cc
```diff
@@ -2212,6 +2213,83 @@ TEST(DecodeTest, AnimationTest) {
...
+    std::ostringstream filename;
+    filename << "jxl/blending/cropped_traffic_light_frame-" << i << ".png";
+    const jxl::PaddedBytes compressed_frame = jxl::ReadTestData(filename.str());
+    jxl::CodecInOut frame;
+    ASSERT_TRUE(
+        SetFromBytes(jxl::Span<const uint8_t>(compressed_frame), &frame));
+    EXPECT_TRUE(jxl::SamePixels(*io.Main().color(), *frame.Main().color()));
+  }
+
+  // After all frames gotten, JxlDecoderProcessInput should return
```
veluca93, line 2285: gotten?

lib/jxl/dec_frame.cc
lvandeve: Thanks for the pointer! Used NeedsBlending's implementation to know what to write here, did not actually use NeedsBlending directly from here though because it requires implementing the same loop anyway to know the sources of the extra channels dependencies

lib/jxl/dec_frame.h
lvandeve: Fixed

lib/jxl/decode_test.cc
lvandeve: changed to "were decoded"

veluca93 (approved):

PR #178, opened 2021-06-15
Add a skeleton for a pipelined approach to FinalizeImageRect.

veluca93:


lib/jxl/dec_render_pipeline.h
```diff
@@ -0,0 +1,89 @@
...
+ public:
+  // Initial shifts for the channels (following the same convention as
+  // RenderPipelineStage for naming the channels).
+  void Init(const std::vector<std::pair<size_t, size_t>>& channel_shifts) {
+    JXL_ABORT("Not implemented");
+  }
+
+  // Adds a stage to the pipeline. The shifts for all the channels that are not
+  // kIgnored by the stage must be identical at this point.
+  void AddStage(RenderPipelineStage&& stage) { JXL_ABORT("Not implemented"); }
```
sboukortt, line 73: How do you plan to store these in a way that will maintain runtime dispatch?

veluca93: good point - either I'll store them in a lambda-converted-to-std::function or (more likely) I'll change this to take a unique ptr.

sboukortt (approved):

lib/jxl/dec_render_pipeline.h
```diff
@@ -0,0 +1,91 @@
...
+                          float* JXL_RESTRICT** output, size_t xsize,
+                          size_t xpos, size_t ypos) const = 0;
+  virtual ~RenderPipelineStage() {}
+
+  // Amount of padding required by each channel in the various directions.
+  // The value for c=0 indicates padding required for color channels, subsequent
+  // values refer to padding for non-color channels, in order.
+  virtual size_t GetPaddingX(size_t c) const = 0;
+  virtual size_t GetPaddingY(size_t c) const = 0;
+
```
deymo, line 49: I think this should be "border" not "padding". for me padding implies that the memory needs to be there but we don't use it (uninitialized is fine). I don't think we will ever need padding in Y.
Border in X and Y is often the same but it is ok to have two methods; padding in X would be a little bigger, but I think it could be BorderX rounded up to some vector size; anyway, what do you plan to use the padding (not border) information for? just that sum(all border + last padding) <= kRenderPipelineXOffset?

veluca93: You are correct, I meant border :)

PR #187, opened 2021-06-16
Make progress in the RenderPipeline implementation.

veluca93:
Applying the pipeline is still missing.


lib/jxl/dec_render_pipeline.h
```diff
@@ -24,28 +24,31 @@ enum class RenderPipelineChannelMode {
...
+                          size_t xsize, size_t xpos, size_t ypos) const = 0;
   virtual ~RenderPipelineStage() {}
 
   // Amount of padding required by each channel in the various directions.
   // The value for c=0 indicates padding required for color channels, subsequent
   // values refer to padding for non-color channels, in order.
-  virtual size_t GetPaddingX(size_t c) const = 0;
-  virtual size_t GetPaddingY(size_t c) const = 0;
+  virtual size_t GetBorderX(size_t c) const = 0;
+  virtual size_t GetBorderY(size_t c) const = 0;
```
deymo, line 51: Is different border per channel a use-case we need? I think that's simplify things a bit.

lib/jxl/dec_render_pipeline.h
```diff
@@ -24,28 +24,31 @@ enum class RenderPipelineChannelMode {
 
 class RenderPipelineStage {
  public:
-  // `input` points to `2*MaxPaddingY() + 1` pointers, each of which points to
-  // `3+num_non_color_channels` pointer-to-row. So, `input[MaxPaddingY()][0]` is
+  // `input` points to `2*MaxBorderY() + 1` pointers, each of which points to
+  // `3+num_non_color_channels` pointer-to-row. So, `input[MaxBorderY()][0]` is
   // the pointer to the center row of the first color channel.
-  //  `MaxPaddingY()` is the maximum value returned by `GetPaddingX()`;
+  //  `MaxBorderY()` is the maximum value returned by `GetBorderX()`;
```
deymo, line 30: MaxBorderY is the maximum of GetBorderX?

lib/jxl/dec_render_pipeline.h
veluca93: Unfortunately yes... Extra channels will have different borders in most cases.

lib/jxl/dec_render_pipeline.h
```diff
@@ -24,28 +24,31 @@ enum class RenderPipelineChannelMode {
 
 class RenderPipelineStage {
  public:
-  // `input` points to `2*MaxPaddingY() + 1` pointers, each of which points to
-  // `3+num_non_color_channels` pointer-to-row. So, `input[MaxPaddingY()][0]` is
+  // `input` points to `2*MaxBorderY() + 1` pointers, each of which points to
```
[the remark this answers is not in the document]
veluca93: Mhhhh... I'd prefer to avoid negative offsets if possible. Also in most (all?) cases MaxPaddingY() should be a class constant, so I wouldn't be too concerned about getting it wrong.

lib/jxl/dec_render_pipeline.h
```diff
@@ -20,70 +20,187 @@ enum class RenderPipelineChannelMode {
   kIgnored = 0,
   kInPlace = 1,
   kInOut = 2,
+  kInput = 3,
```
jonsneyers, line 23: Would be nice to have some comments to document what these different ChannelModes mean (what assumptions or invariants they imply, etc).

jonsneyers (approved):

lib/jxl/dec_render_pipeline.h
veluca93: True! We can do that once we actually know what the assumptions and invariants are :P

jonsneyers:
What is the status of this one? No point merging it until work on it continues?

PR #189, opened 2021-06-16
Fix: float-to-int-overflow

eustas:
Intermediate mul value for bpp=24..32 is bigger than expected due to limited float presision.
For bpp=32 it is bigger than MAX_INT. Use direct "cap" values for v >= 1.0

### DIFF
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -35,6 +35,8 @@ void FloatToU32(const float* in, uint32_t* out, size_t num, float mul,
   const HWY_FULL(float) d;
   const hwy::HWY_NAMESPACE::Rebind<uint32_t, decltype(d)> du;
   size_t vec_num = num;
+  const uint32_t cap = (1ull << bits_per_sample) - 1;
+  // TODO(eustas): investigate 24..31 bpp cases.
   if (bits_per_sample == 32) {
     // Conversion to real 32-bit *unsigned* integers requires more intermediate
     // precision that what is given by the usual f32 -> i32 conversion
@@ -57,9 +59,13 @@ void FloatToU32(const float* in, uint32_t* out, size_t num, float mul,
   }
   for (size_t x = vec_num; x < num; x++) {
     float v = in[x];
-    // Inverted condition grants that NaN is mapped to 0.0f.
-    v = (v >= 0.0f) ? (v > 1.0f ? mul : (v * mul)) : 0.0f;
-    out[x] = static_cast<uint32_t>(v + 0.5f);
+    if (v >= 1.0f) {
+      out[x] = cap;
+    } else if (v >= 0.0f) {  // Inverted condition => NaN -> 0.
+      out[x] = static_cast<uint32_t>(v * mul + 0.5f);
+    } else {
+      out[x] = 0;
+    }
   }
 }
 


jonsneyers:
Why do we even have a uint32 output format?

I think it makes more sense to have only uint8, uint16, and float32. If and when someone actually wants to work with > 16-bit integers, we can add it, but at the moment our encoder/decoder implementation is limited to 24-bit ints anyway (because of the intermediate float buffers that are used), and we don't have any input/output formats that support more than 16-bit ints.

Could we just get rid of the uint32 output completely?

veluca93:
@lvandeve

veluca93:
> @lvandeve

To answer Jon's question :)

lvandeve:
@jonsneyers there is uint32 output, but it's not actually implemented (the API gives error when trying to use `JXL_TYPE_UINT32`), it is there for future implementation because modular mode supports 32-bit integers (or 31-bit?), is that correct Jon?

@deymo, if we don't suport uint32, we could remove it from the API, WDYT?

jonsneyers:
Modular mode does support up to 31-bit uint (and 32-bit signed int, which is what is used when doing lossless float32), but with the current intermediate float representations, it's effectively limited to 24-bit uint (and lossless up to 32-bit float). I think it is sufficient to have uint8, uint16 and float32 output — we could later add other output formats like 3x10-bit-packed-in-4-bytes, uint24, fp24, int32, bitpacked 4-bit / 2-bit / 1-bit, etc if there's an actual need for it, but for now I think those things can just be done via float32 (except int32, but we don't support that anyway in the current implementation).

So let's remove uint32 output :)

eustas:
@lvandeve if 32-bit output is not implemented in decoder, how has djxl_fuzzer touched it then? =)

lvandeve:
@eustas it exists as a data type, and the API would return an error code, so to the fuzzer it would just look like the API returning an error code, this is not the same as a crash or similar

eustas:
I mean fuzzer really reaches the code changed in this PR, i.e. `bits_per_sample == 32`.

lvandeve:
@eustas Do you have a fuzzer reproduction file for this?

deymo:
Data precision in the image and the pixel format in the API are different things and don't need to match. There are some modes not implemented in the API but those are just a matter of implementing them in external image.

lvandeve:
I found the cause of why it's decodign with JXL_TYPE_UINT32 here: in this test case, `JxlDecoderSetImageOutCallback` is used instead of `JxlDecoderSetImageOutBuffer`, and the check for the data type is not done there, because the check for the data type happens where the buffer size is being computed (which is not done for JxlDecoderSetImageOutCallback), so that should be fixed.

I'll do that in a different merge request than this one.

By the way, I'm also all for submitting this merge request here myself: fixing this conversion can be the way towards enabling the uint32 option in the api eventually

deymo (approved):
I think the ternary operator could be kept (with type uint32_t) but this is fine and I don't think will make any difference for the compiler and it is just for the last bit.

eustas:
Ooops, manual rebase staled the review. @deymo

deymo (approved):

deymo:
Sorry, I got confused that this said "Merged" already, but that was for the linked PR, not this PR.

Issue #184, opened 2021-06-16
JXL_FAILURE: Invalid channel range


Jamaika1:
cjxl.exe image_21447_24bit.png image_21447_24bit.jxl -v -m -q 100 -s 5 -C 1 --num_threads=4
```
JPEG XL encoder v0.3.7 0.3.7-13649d2b [Scalar]
codec_png.cc:496: PNG: no color_space/icc_pathname given, assuming sRGB
Read 1563x1558 image, 20.7 MP/s
Encoding [Modular, lossless, hare], 4 threads.
transform.cc:70: JXL_FAILURE: Invalid channel range
```
https://github.com/libjxl/libjxl/commit/30dae3b5b0cb70134dd08964a6bb1f247f2ac412

jonsneyers:
Hi, thanks for the report. Using a random image, I cannot reproduce. Could you share the image on which this happens?

jonsneyers:
Nevermind, I found an image on which I can reproduce it.

Jamaika1:
https://www.sendspace.com/file/wx8wam

PR #190, opened 2021-06-16
remove another unnecessary modular image field

jonsneyers:
Get rid of `Image.nb_channels`, since it had become redundant, especially in the decoder.

Also fixes https://github.com/libjxl/libjxl/issues/184, removes/updates some stale comments, and removes a check that is already implied by an earlier check (so the condition will always fail).

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -255,7 +255,6 @@ Status ModularFrameDecoder::DecodeGroup(const Rect& rect, BitReader* reader,
     gc.vshift = fc.vshift;
     gi.channel.emplace_back(std::move(gc));
   }
-  gi.nb_channels = gi.channel.size();
   if (zerofill) {
     int gic = 0;
     for (c = beginc; c < full_image.channel.size(); c++) {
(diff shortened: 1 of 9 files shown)


veluca93 (commented):
Where's the bug exactly? I assume that it was encoder-only?

deymo:
do we want to backport the fix to 0.4? Could we split the commit to have the fix separated from the removal of the unnecessary field? I see a bit semantic changes here, sometimes is chanels.size() but sometimes is changed to channels.size() - nb_meta_channels.

Could we add a test for this fix as well?

jonsneyers:
One bug this fixes is a minor encoder regression where it would output an error in cases where palette is done and a manual global RCT is specified (it would still encode fine, just unnecessarily show an error).

Another bug it fixes is in the default squeeze params in the theoretical case where multiple squeeze transforms are done (something the encoder currently doesn't do).

jonsneyers:
I split off the change that is an actual (hypothetical) decoder bugfix, see https://github.com/libjxl/libjxl/pull/194.

The bug doesn't affect any existing bitstream and is unlikely to affect any future bitstream (the case in which it would make a difference is not something I think an encoder would want to do), but still, it would be good to backport that fix to 0.4 just to ensure it is forward-compatible.

veluca93 (approved):

PR #195, opened 2021-06-17
improve squeeze param check

jonsneyers:
Squeeze params were checked before the loop, which wasn't quite correct, since the number of channels changes during the loop that came after the check. This meant that custom squeeze params that would do squeezes on channels that don't exist yet before the loop would not be allowed (both in the encoder and in the decoder), while it could actually be a valid bitstream.

This PR puts the check inside the loop instead, so it checks things w.r.t. the actual channel vector at that point in the squeeze steps series.

The default squeeze params (which is the only thing the encoder currently uses) are not affected by this, because that always passes the too-strict check too. But valid custom squeeze params could lead to decode error.


veluca93 (approved):

jonsneyers:
@deymo you may want to cherrypick this one too for 0.4.

PR #196, opened 2021-06-17
Avoid %zu in cjxl_main.cc

deymo:
Printing size_t with printf requires a different format string depending
on the library. Other projects use macros like PRIuS similar to PRIu32
and PRIu64.

To workaround the issue we print a long long unsigned instead in this
case.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -105,9 +105,9 @@ int CompressJpegXlMain(int argc, const char* argv[]) {
       const size_t pixels = io.xsize() * io.ysize();
       const double bpp =
           static_cast<double>(compressed.size() * jxl::kBitsPerByte) / pixels;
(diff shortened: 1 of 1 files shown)


eustas (approved):

jonsneyers:
There are quite a lot of other cases of size_t being printed with `%zu`, e.g. https://github.com/libjxl/libjxl/blob/main/tools/cjxl.cc#L821

Might be worth checking all occurrences of `%zu` because I suspect most of them would refer to a size_t variable.

deymo:
Yeah, we would need a better solution; this is to fix the builder.
The library really shouldn't print anything unless you are debugging it.

jonsneyers:
It's a bit strange that the %zu size_t that cjxl prints just a little bit earlier doesn't cause trouble though...

deymo:
We might be missing some include or macro in this file. I don't think it is worth looking into it.

PR #198, opened 2021-06-17
push global RCT to groups if possible, parallelize otherwise

jonsneyers:
Makes global RCTs a bit faster (as e.g. used at fast modular speed settings like thunder and falcon).

Introduces a list of global transforms that are applied immediately after decoding each group instead of during finalization. For now only RCTs are moved from the global image to that list, could later add non-delta-palette transforms (channel palettes or simple color palettes) too. @veluca93 : this might be a start for a memory reduction by skipping the global image in case there are no global transforms left at that point.

In case there are still global RCTs left (e.g. when the global RCT is followed by a global squeeze or delta palette), use threads.

About 5% speedup of modular thunder decoding:
before: 4064 x 2704, geomean: 47.28 MP/s [38.06, 49.52], 30 reps, 4 threads.
after: 4064 x 2704, geomean: 49.72 MP/s [40.32, 52.68], 30 reps, 4 threads.

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -24,6 +24,8 @@
 #include "lib/jxl/epf.h"
 #include "lib/jxl/modular/encoding/encoding.h"
 #include "lib/jxl/modular/modular_image.h"
+#include "lib/jxl/modular/transform/transform.h"
+
 HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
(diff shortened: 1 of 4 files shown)


veluca93 (dismissed):

veluca93 (approved):

eustas:
@jonsneyers : it seems that this PR causes benchmark pipeline failures: `jxl:m:cheetah:P6 failed to decompress encoded image. Original source: /tmp/tmp.dvO3hD76wF/bench2.png` and other errors

jonsneyers:
Please revert then for now, I will look at it on Monday.

jonsneyers:
The issue was with the case where first a channel palette is done, then an RCT. It would then push that RCT to the groups, which could be done in principle but it's a bit more complicated since the channel indices in the groups are just the usual 0-2 while in the full_image the RCT is done on channels 2-4 (it did channel palette on two channels, which adds two metachannels with the palettes, but in the groups those are not visible).

In general the channel indices will always be messed up with any non-RCT transform that gets done before the RCT, so for now I"ll just only do it when there's only a global RCT and no other global transforms.

PR #199, opened 2021-06-17
Store frame dependency graph on rewind

lvandeve:
This stores the dependency graph between frames on rewind, so that
decoding of more frames can be avoided when skipping to a later frame
after rewind.

Improved the SkipFrameWithBlendingTest to have true dependencies between
frames that use a blending mode that depends on the pixels of both 
frames, to test this feature, and verified that making 
GetFrameDependencies return wrong output, breaks the test.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -285,6 +285,83 @@ struct Sections {
   std::vector<char> section_received;
 };
 
+/*
+Given list of frame references to storage slots, and storage slots in which this
+frame is saved, computes which frames are required to decode the frame at the
+given index and any frames after it. The frames on which this depends are
+returned as a vector of their indices, in no particular order. The given index
+must be smaller than saved_as.size(), and references.size() must equal
+saved_as.size(). Any frames beyond saved_as and references are considered
+unknown future frames and must be treated as if something depends on them.
+*/
+std::vector<size_t> GetFrameDependencies(size_t index,
+                                         const std::vector<int>& saved_as,
+                                         const std::vector<int>& references) {
+  JXL_ASSERT(references.size() == saved_as.size());
+  JXL_ASSERT(index < references.size());
+
+  std::vector<size_t> result;
+
+  constexpr size_t kNumStorage = 8;
+
+  // value which indicates nothing is stored in this storage slot
+  const size_t invalid = references.size();
+  // for each of the 8 storage slots, a vector that translates frame index to
+  // frame stored in this storage slot at this point, that is, the last
+  // frame that was stored in this slot before or at this index.
+  std::array<std::vector<size_t>, kNumStorage> storage;
+  for (size_t s = 0; s < kNumStorage; ++s) {
+    storage[s].resize(saved_as.size());
+    int mask = 1 << s;
+    size_t id = invalid;
+    for (size_t i = 0; i < saved_as.size(); ++i) {
+      if (saved_as[i] & mask) {
+        id = i;
+      }
+      storage[s][i] = id;
+    }
+  }
+
+  std::vector<char> seen(index + 1, 0);
+  std::vector<size_t> stack;
+  stack.push_back(index);
+  seen[index] = 1;
+
+  // For frames after index, assume they can depend on any of the 8 storage
+  // slots, so push the frame for each stored reference to the stack and result.
+  // All frames after index are treated as having unknown references and with
+  // the possibility that there are more frames after the last known.
+  // TODO(lode): take values of saved_as and references after index, and a
+  // input flag indicating if they are all frames of the image, to further
+  // optimize this.
+  for (size_t s = 0; s < kNumStorage; ++s) {
+    size_t frame_ref = storage[s][index];
+    if (frame_ref == invalid) continue;
+    if (seen[frame_ref]) continue;
+    stack.push_back(frame_ref);
+    seen[frame_ref] = 1;
+    result.push_back(frame_ref);
+  }
+
+  while (!stack.empty()) {
+    size_t frame_index = stack.back();
+    stack.pop_back();
+    if (frame_index == 0) continue;  // first frame cannot have references
+    for (size_t s = 0; s < kNumStorage; ++s) {
+      int mask = 1 << s;
+      if (!(references[frame_index] & mask)) continue;
+      size_t frame_ref = storage[s][frame_index - 1];
+      if (frame_ref == invalid) continue;
+      if (seen[frame_ref]) continue;
+      stack.push_back(frame_ref);
+      seen[frame_ref] = 1;
+      result.push_back(frame_ref);
+    }
+  }
+
+  return result;
+}
+
 }  // namespace
 
 // NOLINTNEXTLINE(clang-analyzer-optin.performance.Padding)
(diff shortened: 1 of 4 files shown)


lib/jxl/decode.cc
```diff
@@ -285,6 +285,72 @@ struct Sections {
...
 };
 
+/*
+Given list of frame references to storage slots, and storage slots in which this
+frame is saved, computes which frames are required to decode the frame at the
+given index and any frames after it. The frames on which this depends are
+returned as a vector of their indices, in no particular order. The given index
+must be smaller than saved_as.size(), and references.size() must equal
+saved_as.size(). Any frames beyond saved_as and references are considered
+unknonw future frames and must be treated as if something depends on them.
```
veluca93, line 295: typo

lib/jxl/decode.cc
```diff
@@ -285,6 +285,72 @@ struct Sections {
...
+*/
+std::vector<size_t> GetFrameDependencies(size_t index,
+                                         const std::vector<int>& saved_as,
+                                         const std::vector<int>& references) {
```
veluca93, line 299: Shouldn't this also take a parameter that specifies if the frames here are *all* the frames?

lib/jxl/decode.cc
```diff
@@ -285,6 +285,72 @@ struct Sections {
...
+  JXL_ASSERT(references.size() == saved_as.size());
+  JXL_ASSERT(index < references.size());
+
+  std::vector<size_t> result;
+
+  constexpr size_t num_storage = 8;
```
veluca93, line 305: kNumStorage

lib/jxl/decode.cc
```diff
@@ -285,6 +285,72 @@ struct Sections {
...
+
+  // value which indicates nothing is stored in this storage slot
+  const size_t invalid = references.size();
+  // for each of the 8 storage slots, a vector that translates frame index to
+  // frame stored in this storage slot at this point, that is, the last
+  // frame that was stored in this slot before or at this index.
+  std::vector<size_t> storage[num_storage];
```
veluca93, line 312: Could make this a std::vector<std::array<size_t, kNumStorage>>

lib/jxl/decode.cc
```diff
@@ -507,11 +602,24 @@ void JxlDecoderDestroy(JxlDecoder* dec) {
...
+  std::vector<int> frame_references;
+  std::vector<int> frame_saved_as;
+  std::vector<size_t> frame_external_to_internal;
+  std::vector<char> frame_required;
+  frame_references.swap(dec->frame_references);
+  frame_saved_as.swap(dec->frame_saved_as);
+  frame_external_to_internal.swap(dec->frame_external_to_internal);
+  frame_required.swap(dec->frame_required);
+
   // TODO(lode): for performace, rewind should keep state about frames to make
```
veluca93, line 614: Is this still true?

lib/jxl/decode.cc
```diff
@@ -975,15 +1113,35 @@ JxlDecoderStatus JxlDecoderProcessInternal(JxlDecoder* dec, const uint8_t* in,
...
+
+      if (internal_frame_index >= dec->frame_saved_as.size()) {
+        dec->frame_saved_as.push_back(saved_as);
+        JXL_ASSERT(dec->frame_saved_as.size() == internal_frame_index + 1);
+
+        // add the value 255 (which means all references) to new slots: we only
+        // know the references of the frame at FinalizeFrame, and fill in the
+        // correct values there. As long as this information is not known, the
+        // worst case where the frame depends on all storage slots is assumed.
+        dec->frame_references.push_back(255);
```
veluca93, line 1130: let's use hex for bitmasks

lib/jxl/decode.cc
```diff
@@ -285,6 +285,72 @@ struct Sections {
...
+      if (saved_as[i] & mask) {
+        id = i;
+      }
+      storage[s][i] = id;
+    }
+  }
+
+  std::vector<char> seen(index + 1, 0);
+  std::vector<size_t> stack;
+  stack.push_back(index);
```
veluca93, line 327: I think this should also add all the frames that are saved for the last time in their slot if index is not the last frame, or something like that, no?

lib/jxl/decode.cc
lvandeve: Fixed

lib/jxl/decode.cc

lib/jxl/decode.cc
lvandeve: Removed

lib/jxl/decode.cc

lib/jxl/decode.cc
lvandeve: See other comment about the same topic

lib/jxl/decode.cc
lvandeve: For now that's not implemented, as implemented now, it requires that all 8 referenceable storage slots are available after the frame with `index` and does not take the structure after that frame into account

This code could be upgraded with that, it would require adding everything from saved_as/references after index to the stack, and then depending on whether "seen_last" is true or not, require the 8 referenceable save slots or not as well. That is ok to do, but wouldn't do *that* much in the grand scheme of things given what the jxl encoder now does (which is to use slot 2 only, not taking dc into account).

So for now, I'd keep as is, and then taking the frames after index into account can be done as a future optimization.

veluca93: Fair enough, but shouldn't you push those slots in the stack too? I don't see that happening...

lvandeve: It happens in the code pasted below. It goes through the 8 storage slots, and does not do 'continue' for all 8 of them when frame_index == index. Does this look correct?

```
      // Include all storage slots this frame depends on, except for the
      // future frames, which may depend on anything, so for index itself
      // test all 8 storage slots. We know some information about future frames
      // (up to the length of saved_as/references), but there can come more
      // frames after it.
      if (!(references[frame_index] & mask) && frame_index != index) continue;
```

veluca93:
Ah, I did not notice this. It might perhaps be better to explicitly push
things in the stack, but it's OK like this.

lib/jxl/decode.cc

lib/jxl/decode.cc
lvandeve: I separated out that code into explicitely pushing these values on the stack and added a TODO for the potential optimzation with all frames seen flag, PTAL

lib/jxl/decode.cc
```diff
@@ -285,6 +285,83 @@ struct Sections {
...
+  std::vector<size_t> result;
+
+  constexpr size_t kNumStorage = 8;
+
+  // value which indicates nothing is stored in this storage slot
+  const size_t invalid = references.size();
+  // for each of the 8 storage slots, a vector that translates frame index to
+  // frame stored in this storage slot at this point, that is, the last
+  // frame that was stored in this slot before or at this index.
+  std::array<std::vector<size_t>, kNumStorage> storage;
```
veluca93, line 312: Ah, I actually meant vector-of-array instead of array-of-vectors here, but up to you

veluca93 (approved):

veluca93 (approved):

lib/jxl/decode.cc
lvandeve: Doesn't matter too much I think, it's accessed in both orders anyway, I won't change it anymore now :)

PR #200, opened 2021-06-17
Pass `decoded->extra_channels()` and the corresponding Rects separately to PrepareBlending

sboukortt:
This prevents the pointers that we keep in `extra_channels` from being
invalidated by reallocation.

### DIFF
--- a/lib/jxl/blending.cc
+++ b/lib/jxl/blending.cc
@@ -36,7 +36,8 @@ Status ImageBlender::PrepareBlending(
     const std::vector<ExtraChannelInfo>* extra_channel_info,
     const ColorEncoding& frame_color_encoding, const Rect& frame_rect,
     Image3F* output, const Rect& output_rect,
-    std::vector<std::pair<ImageF*, Rect>> output_extra_channels) {
+    std::vector<ImageF>* output_extra_channels,
+    std::vector<Rect> output_extra_channels_rects) {
   const PassesSharedState& state = *dec_state->shared;
   info_ = state.frame_header.blending_info;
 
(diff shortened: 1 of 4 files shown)


veluca93:
That looks perhaps a bit fragile - can we do resize instead?

sboukortt:
> That looks perhaps a bit fragile - can we do resize instead?

I looked it up when coming up with the fix, and (https://stackoverflow.com/a/46132842):

> No reallocation shall take place during insertions that happen after a call to `reserve()` until the time when an insertion would make the size of the vector greater than the value of `capacity()`

If we really want to avoid any such risk, how about populating the two vectors in separate passes? One for `decoded->extra_channels()`, and then one for `extra_channels` once we know that we are not going to touch the former any further.

veluca93:
> > That looks perhaps a bit fragile - can we do resize instead?
> 
> I looked it up when coming up with the fix, and (https://stackoverflow.com/a/46132842):
> 
> > No reallocation shall take place during insertions that happen after a call to `reserve()` until the time when an insertion would make the size of the vector greater than the value of `capacity()`
> 
> If we really want to avoid any such risk, how about populating the two vectors in separate passes? One for `decoded->extra_channels()`, and then one for `extra_channels` once we know that we are not going to touch the former any further.

The advantage of doing a resize first is that there is no push_back() in the loop, and it makes it more obvious (and likely easier to catch) that we are relying on the resize being called with the correct size.

deymo:
uugh, keeping raw pointers to vector elements sounds a recipe for bugs. Could we keep indices into the array instead?

sboukortt:
> uugh, keeping raw pointers to vector elements sounds a recipe for bugs. Could we keep indices into the array instead?

They are in the same order, so the way that things are at the moment, we can instead pass the additional data separately, as a parallel array. How is this?

lib/jxl/blending.cc
```diff
@@ -36,7 +36,8 @@ Status ImageBlender::PrepareBlending(
     const std::vector<ExtraChannelInfo>* extra_channel_info,
     const ColorEncoding& frame_color_encoding, const Rect& frame_rect,
     Image3F* output, const Rect& output_rect,
-    std::vector<std::pair<ImageF*, Rect>> output_extra_channels) {
+    std::vector<ImageF>* output_extra_channels,
+    std::vector<Rect> output_extra_channels_rects) {
```
deymo, line 40: wouldn't this be a `std::vector<Rect>&&`? or it doesn't matter because you call it with std::move() anyway?

deymo (approved):

lib/jxl/blending.cc
sboukortt: Yep, the latter. https://abseil.io/tips/117

PR #201, opened 2021-06-17
Verify JxlPixelFormat validity when setting out callback

lvandeve:
The format must be checked for having a supported amount of channels,
pixel type, etc... This was done in the size check for the image output
buffer setters, but was not done in the callback because buffer size
works differently for those.

Now the full check is done here as well to fix that.


lvandeve:
This is in response to the bug found through https://github.com/libjxl/libjxl/pull/189

veluca93 (approved):

eustas (approved):

PR #203, opened 2021-06-18
OssFuzz: Do not read/use undefined values in UndoXYB

eustas:
Refactor code to extract common part. This allows to deal with MSAN in one place (instead of 6).

### DIFF
--- a/lib/jxl/dec_reconstruct.cc
+++ b/lib/jxl/dec_reconstruct.cc
@@ -37,141 +37,114 @@ HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
 
-Status UndoXYBInPlace(Image3F* idct, const Rect& rect,
+template <typename Op>
+void DoUndoXYBInPlace(Image3F* idct, const Rect& rect, Op op,
                       const OutputEncodingInfo& output_encoding_info) {
-  PROFILER_ZONE("UndoXYB");
-
+  // TODO(eustas): should it still be capped?
+  const HWY_CAPPED(float, GroupBorderAssigner::kPaddingXRound) d;
+  const size_t xsize = rect.xsize();
+  const size_t xsize_v = RoundUpTo(xsize, Lanes(d));
   // The size of `rect` might not be a multiple of Lanes(d), but is guaranteed
   // to be a multiple of kBlockDim or at the margin of the image.
   for (size_t y = 0; y < rect.ysize(); y++) {
     float* JXL_RESTRICT row0 = rect.PlaneRow(idct, 0, y);
     float* JXL_RESTRICT row1 = rect.PlaneRow(idct, 1, y);
     float* JXL_RESTRICT row2 = rect.PlaneRow(idct, 2, y);
-
-    const HWY_CAPPED(float, GroupBorderAssigner::kPaddingXRound) d;
-
-    if (output_encoding_info.color_encoding.tf.IsLinear()) {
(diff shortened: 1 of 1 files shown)


lib/jxl/common.h
```diff
@@ -30,6 +30,15 @@
 namespace jxl {
 // Some enums and typedefs used by more than one header file.
 
+// Define MEMORY_SANITIZER if memory sanitizer is enabled.
```
deymo, line 33: MEMORY_SANITIZER enables a bunch of other behavior which we should re-evaluate, in particular the initialization of padding, which may not be an issue with recent versions. This should probably be in a different commit.

eustas: I could make a separate commit, but the fix won't work without this change.
IIUC MEMORY_SANITIZER macro is a legacy way, currently documentation says "In some cases one may need to execute different code depending on whether MemorySanitizer is enabled. __has_feature can be used for this purpose.". In brotli we do check both macro and has_feature to detect MSAN build.

eustas: Or we could move this to compiler options.

lib/jxl/dec_reconstruct.cc
```diff
@@ -68,9 +68,17 @@ Status UndoXYBInPlace(Image3F* idct, const Rect& rect,
...
+        const auto mask = Iota(d, x) < Set(d, rect.xsize());
+        const auto sentinel = Set(d, kSanitizerSentinel);
+        const auto in_opsin_x = IfThenElse(mask, Load(d, row0 + x), sentinel);
+        const auto in_opsin_y = IfThenElse(mask, Load(d, row1 + x), sentinel);
+        const auto in_opsin_b = IfThenElse(mask, Load(d, row2 + x), sentinel);
+#else
         const auto in_opsin_x = Load(d, row0 + x);
         const auto in_opsin_y = Load(d, row1 + x);
         const auto in_opsin_b = Load(d, row2 + x);
+#endif
```
deymo, line 81: Can we poison the portion of row{0,1,2} + x in line 91?

eustas: Likely, yes. Do you propose to unpoison-process-poison here?

deymo: yes. The unpoison-process-poison here will be founded on the fact that XybToRgb uses the undefined values but it is safe to do so; but we shouldn't assume that the output xybtorgb is initialized in other places because of this.
A comment about why next to the #ifdef MEMORY_SANITIZER explaining why would be appreciated.

eustas:
@deymo PTAL. In the next PR I'll add lib/sanitizers.h and after that apply unpoison-process-poison scheme to this code.

eustas:
@deymo PTALl nota-bene, now MSAN fails at later stage, but I think PR is big enough, to split more fixes in other PR

lib/jxl/dec_reconstruct.cc
```diff
@@ -37,141 +37,114 @@ HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
 
-Status UndoXYBInPlace(Image3F* idct, const Rect& rect,
+template <typename Op>
```
deymo, line 40: I think there's some potential here to reduce the binary size by removing this template and using a virtual function. It won't be inlined of course but given that we only do a Store afterwards I don't think it will make much difference in runtime performance. I'll do this in a follow up.

deymo (approved):

PR #205, opened 2021-06-18
avoid full modular image copy

jonsneyers:
In cases where there are no global modular transforms except RCT (as is the case at speed falcon and faster), avoid the unnecessary copy from decoded groups to the modular `full_image` (and then to the float `decoded`).

For `-m -e 2`:

Before:
4064 x 2704, geomean: 47.20 MP/s [38.94, 49.02], 30 reps, 4 threads.
Allocations: 21465 (max bytes in use: 4.360149E+08)

After:
4064 x 2704, geomean: 57.45 MP/s [46.70, 59.76], 30 reps, 4 threads.
Allocations: 21465 (max bytes in use: 3.020612E+08)


jonsneyers:
Includes a corrected version of https://github.com/libjxl/libjxl/pull/198 so RCT is only pushed to the groups if it is the only global transform (if there are other transforms, things get more complicated quickly and https://github.com/libjxl/libjxl/pull/198 made the false assumption that other transforms are fine if they are undone _after_ the RCT — they're not, since they still mess up the channel indices).

veluca93 (approved):

PR #209, opened 2021-06-20
Squeeze: skip setting DefaultSqueezeParameters when all channels are meta

lovell:
Hello, this is a possible solution to prevent a buffer/container overflow at line 145 when a (most likely invalid) image contains only meta channels.

https://github.com/libjxl/libjxl/blob/f6d3a89f48e7a66d235ebee73cd5db694d69b63b/lib/jxl/modular/transform/squeeze.cc#L140-L145

An example image that triggers this, found via fuzz testing, is [default-squeeze-params-overflow.txt](https://github.com/libjxl/libjxl/files/6682092/default-squeeze-params-overflow.txt)

```
READ of size 8 at 0x6140000101b0 thread T4 (pool)
--
SCARINESS: 23 (8-byte-read-container-overflow)
0x2df4915 in jxl::DefaultSqueezeParameters(std::__1::vector&lt;jxl::SqueezeParams, std::__1::allocator&lt;jxl::SqueezeParams&gt; &gt;*, jxl::Image const&amp;) libjxl/lib/jxl/modular/transform/squeeze.cc:145:52
0x2df4c27 in jxl::MetaSqueeze(jxl::Image&amp;, std::__1::vector&lt;jxl::SqueezeParams, std::__1::allocator&lt;jxl::SqueezeParams&gt; &gt;*) libjxl/lib/jxl/modular/transform/squeeze.cc:210:5
0x2622704 in jxl::Transform::MetaApply(jxl::Image&amp;) libjxl/lib/jxl/modular/transform/transform.cc:54:14
```

The approach in this PR returns early and allows the subsequent call to `CheckMetaSqueezeParams` to fail as the parameters are empty. I'm sure there are other approaches to solve this, but this is hopefully the smallest/simplest change to do so.

(Name added to AUTHORS as per contributing checklist.)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,3 +22,4 @@ Pieter Wuille
 Marcin Konicki <ahwayakchih@gmail.com>
 Ziemowit Zabawa <ziemek.zabawa@outlook.com>
 Petr Diblík
+Lovell Fuller
--- a/lib/jxl/modular/transform/squeeze.cc
+++ b/lib/jxl/modular/transform/squeeze.cc
@@ -140,6 +140,9 @@ void InvVSqueeze(Image &input, int c, int rc, ThreadPool *pool) {
 void DefaultSqueezeParameters(std::vector<SqueezeParams> *parameters,
                               const Image &image) {
   int nb_channels = image.channel.size() - image.nb_meta_channels;
+  if (nb_channels <= 0) {
+    return;
+  }
 
   parameters->clear();
   size_t w = image.channel[image.nb_meta_channels].w;


jonsneyers:
Thanks for the find and for the fix!

I'm a bit curious how you can end up in this situation - I thought that if there are no channels, then modular decode gets skipped before it tries to decode transforms. Only Palette makes metachannels, but it needs at least one channel to operate on. So maybe there's an underlying bug or earlier fail is possible.

lovell:
The sample image has 12 channels, all of which are meta channels / palettes. If such an image is invalid then I agree it would make sense to try to catch this earlier if possible.

jonsneyers:
There is an earlier problem indeed, quite an interesting one.

The spec only says that palette has to be applied to channels of the same dimensions. It doesn't say that these channels have to be non-meta-channels. That is probably an oversight / problem.

So what happens in this fuzzer-found example is the following. We start with a normal 1x1 pixel 3-channel image. Then a bunch of Palette transforms are applied (and RCTs, but those are irrelevant here) to turn it into a 14-channel image of which 11 channels are meta channels (single-channel palettes) and 3 are still non-meta (obviously this is a silly thing to do to a 1-pixel image, but that's the point of fuzzing - finding the silly bitstreams).

Now a funny thing happens: a 4-channel palette is done on 4 of those metachannels that happen to have the same dimensions. This transform adds one palette metachannel and removes 3 channels, turning the image into a 12-channel image with 12 metachannels. Yes, both spec and code just always increment `nb_meta_channels` when a Palette transform happens, even if it actually net removed meta channels.

This is a conceptual problem that is caused by not requiring Palette to operate on actual (non-meta) channels. I _think_ we may have thought at some point that it could be good to allow such "meta-palettes" (I can e.g. imagine a big multi-channel (delta) palette where there is a minor compression benefit in doing such a thing), but it's certainly not something the current encoder will ever do. The theoretical compression advantage of allowing it is quite minor.

If we do want to allow it, it would require updating spec and code to adjust `nb_meta_channels` correctly in case palette is removing some meta channels.
If we do not want to allow it, then it requires updating spec and code to check that the channels Palette operates on are not meta-channels.

@veluca93 @jyrkialakuijala @sboukortt opinions on how to handle this?
My preference is to just not allow Palette to operate on meta-channels (i.e. doing palette on data that represents a palette). If we do want to allow it, then I would only allow it on a series of channels that are either all non-meta or all meta, with a correct update of `nb_meta_channels` in the latter case. We _could_ also allow it on any series of channels (also mixed meta / nonmeta), but that's semantically a very weird thing to do.

veluca93:
I like the all-meta or all-non-meta approach the best.

lovell:
Thank you for the detail Jon, I'm (slowly) starting to understand the JPEG-XL bitstream and had been trying to figure out where the non-meta channels had gone.

I'll close this in favour of https://github.com/libjxl/libjxl/pull/210

jonsneyers:
In any case, even when the underlying issue gets fixed, this check in DefaultSqueezeParameters is probably a good idea to keep, just in case, but maybe it should be turned into a non-silent error (maybe a JXL_ASSERT? or make it return a failed Status so the decoder can actually fail when this happens).

As it currently is, DefaultSqueezeParameters() would return, and then there would be no params at all, and Squeeze would just end up being a no-op since the loops in MetaSqueeze and InvSqueeze would just be skipped. That's better than trying to read width and height of a channel that doesn't exist, but returning error would be even better since this is not supposed to happen. I'm just not sure what kind of error would be best, @deymo or @veluca93 do you have a suggestion? This would be a programming error, even malicious bitstreams should not be able to reach a state where DefaultSqueezeParameters() is called on an image that does not have any (nonmeta) channels, so maybe JXL_ABORT would make most sense? Or WDYT?

Meta-comment: these flexible signalled transform chains are quite nice and they are certainly useful, but it was to be expected that it would bite us in the rear in terms of malicious/fuzzed bitstreams doing weird/invalid stuff. Thanks @lovell  for helping us catch yet another such case! I hope we are converging now to having only well-behaved transform chains accepted by the decoder, but please keep fuzzing :)

Issue #129, opened 2021-06-09
Possible divide by zero in DrawGaussian function of splines.cc


lovell:
Hello, the following file, discovered via fuzz testing, causes a divide by zero when decoded using the latest code on the `main` branch.

[clusterfuzz-testcase-minimized-jpegsave_buffer_fuzzer-5146219264475136.txt](https://github.com/libjxl/libjxl/files/6622107/clusterfuzz-testcase-minimized-jpegsave_buffer_fuzzer-5146219264475136.txt)

```
/src/libjxl/lib/jxl/splines.cc:105:39: runtime error: division by zero
#0 0x14cb3e8 in jxl::N_AVX2::(anonymous namespace)::DrawGaussian(jxl::Image3<float>*, jxl::Rect const&, jxl::Rect const&, jxl::Spline::Point const&, float, float const*, float, std::__1::vector<int, std::__1::allocator<int> >&, std::__1::vector<int, std::__1::allocator<int> >&, std::__1::vector<float, std::__1::allocator<float> >&) libjxl/lib/jxl/splines.cc:105:39
#1 0x14c7cbf in jxl::N_AVX2::(anonymous namespace)::DrawFromPoints(jxl::Image3<float>*, jxl::Rect const&, jxl::Rect const&, jxl::Spline const&, bool, std::__1::vector<std::__1::pair<jxl::Spline::Point, float>, std::__1::allocator<std::__1::pair<jxl::Spline::Point, float> > > const&, float) libjxl/lib/jxl/splines.cc:162:5
#2 0x14c2de5 in jxl::Status jxl::Splines::Apply<true>(jxl::Image3<float>*, jxl::Rect const&, jxl::Rect const&, jxl::ColorCorrelationMap const&) const libjxl/lib/jxl/splines.cc:507:5
#3 0x1479729 in jxl::FinalizeImageRect(jxl::Image3<float>*, jxl::Rect const&, std::__1::vector<std::__1::pair<jxl::Plane<float>*, jxl::Rect>, std::__1::allocator<std::__1::pair<jxl::Plane<float>*, jxl::Rect> > > const&, jxl::PassesDecoderState*, unsigned long, jxl::ImageBundle*, jxl::Rect const&) libjxl/lib/jxl/dec_reconstruct.cc:900:5
#4 0x1481683 in operator() libjxl/lib/jxl/dec_reconstruct.cc:1160:12
```

The point of failure is:

[...]

PR #211, opened 2021-06-21
Splines: skip DrawGaussian when sigma is 0, prevents possible divide-by-zero

lovell:
This seems to be the preferred approach discussed in https://github.com/libjxl/libjxl/issues/129

(Name added to AUTHORS as per contributing checklist.)

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -6,6 +6,7 @@
 #include "lib/jxl/splines.h"
 
 #include <algorithm>
+#include <cmath>
 
 #include "lib/jxl/ans_params.h"
 #include "lib/jxl/base/status.h"
@@ -61,6 +62,11 @@ void DrawGaussian(Image3F* const opsin, const Rect& opsin_rect,
                   const float sigma, std::vector<int32_t>& xs,
                   std::vector<int32_t>& ys,
                   std::vector<float>& local_intensity_storage) {
+  // Sanity check sigma, inverse sigma and intensity
+  if (!(std::isfinite(sigma) && std::isfinite(1.0f / sigma) &&
+      std::isfinite(intensity))) {
+    return;
+  }
   constexpr float kDistanceMultiplier = 4.605170185988091f;  // -2 * log(0.1)
   // Distance beyond which exp(-d^2 / (2 * sigma^2)) drops below 0.1.
   const float maximum_distance = sigma * sigma * kDistanceMultiplier;
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,3 +23,4 @@ Pieter Wuille
 Marcin Konicki <ahwayakchih@gmail.com>
 Ziemowit Zabawa <ziemek.zabawa@outlook.com>
 Petr Diblík
+Lovell Fuller


lib/jxl/splines.cc
```diff
@@ -159,8 +159,10 @@ void DrawFromPoints(
     }
     const float sigma =
         ContinuousIDCT(spline.sigma_dct, (32 - 1) * progress_along_arc);
-    DrawGaussian(opsin, opsin_rect, image_rect, point, multiplier, color, sigma,
-                 xs, ys, local_intensity_storage);
+    if (sigma != 0.0f) {
```
jonsneyers, line 162: Likely the float calculations become numerically unstable also for values very close to 0, which may or may not be compensated by the fact that what is drawn has near-zero intensity.

I would feel more comfortable by something like `abs(sigma) > .0.0000001f` in the condition here (if that fails, it should be so faint that it's invisible anyway), to not just protect against the division-by-zero but also against potential numerical instability when things are getting close to the precision limits of floats.

@sboukortt WDYT?

eustas:
(note: Sami is OOO until Monday)

lib/jxl/splines.cc
sboukortt: Right, that would make sense, as we likely don’t want to splat NaNs all over. But then I wonder if perhaps we wouldn’t want such a protection to be more downstream where it’s more likely to be accurate? For example doing all the computations normally and then only replacing pixel values with the result of the addition if the latter is sensible.

jonsneyers: What is sensible though? Not inf/NaN? If sigma gets really close to 0, dividing by sigma might give huge values, I'm not sure if that would result in sensible outputs. It might help to make a test image with a long spline that goes smoothly from large sigma to negative large sigma and see what the different options produce.

@sboukortt  could you add splines to `jxl_from_tree`? Just manually entering all the dct coeffs for colors/sigma is fine (`Spline [128 numbers] x0 y0 x1 y1 ... EndSpline` or some other easy to parse syntax), we don't need something user friendly, just something we can make art and test bitstreams with :)

sboukortt: > What is sensible though? Not inf/NaN?

Yes, probably something like that.

> could you add splines to `jxl_from_tree`?

That sounds helpful, I’ll try to look into this. Thanks :)

sboukortt: Done in #247 if you would like to have a look.

lovell: Thank you all for the feedback. Would `isfinite` checks within `DrawGaussian` itself suffice? In addition to `inv_sigma`, these could also apply to `sigma` and `intensity`, as if either of these values are NaN or infinite then the rest of the Gaussian calculations will result in NaN or infinite values.

lovell: I've updated this PR to move the `sigma` and `intensity` sanity check into `DrawGaussian`.

lib/jxl/splines.cc
```diff
@@ -61,6 +62,9 @@ void DrawGaussian(Image3F* const opsin, const Rect& opsin_rect,
                   const float sigma, std::vector<int32_t>& xs,
                   std::vector<int32_t>& ys,
                   std::vector<float>& local_intensity_storage) {
+  // Sanity check sigma, inverse sigma and intensity
+  if (!(std::isfinite(sigma) && std::isfinite(1.0f / sigma) &&
+      std::isfinite(intensity))) return;
```
sboukortt, line 67: With the check taking up two lines, we would prefer to have braces around the `return`: https://google.github.io/styleguide/cppguide.html#Conditionals

```suggestion
  if (!(std::isfinite(sigma) && std::isfinite(1.0f / sigma) &&
      std::isfinite(intensity))) {
    return;
  }
```

sboukortt (approved):
Oops, my apologies for not responding sooner. Thanks for the change! Just one small comment.

sboukortt (approved):

PR #216, opened 2021-06-23
Make GetSomeTestImage more varied

lvandeve:
Make the randomly generated image from GetSomeTestImage more varied:
before, it always had a quarter circle of the same size in the top left
corner, and random noise elsewhere.

Now, it has a circle with random radius at random location, a rectangle
with random noise, and a random colored gradient as background.

I needed this to generate a jxl image with blending that has the
blending visible (multiple circles at random locations overlaid, which
was not possible to see if circle was always the same), but
this is worth committing because it improves the tests in general.

### DIFF
--- a/lib/jxl/test_utils.h
+++ b/lib/jxl/test_utils.h
@@ -273,8 +273,37 @@ std::vector<ColorEncodingDescriptor> AllEncodings() {
 std::vector<uint8_t> GetSomeTestImage(size_t xsize, size_t ysize,
                                       size_t num_channels, uint16_t seed) {
   // Cause more significant image difference for successive seeds.
-  std::mt19937 rng(seed);
-  std::uniform_int_distribution<uint16_t> dark(0, 32767);
+  std::mt19937 std_rng(seed);
+  std::uniform_int_distribution<uint16_t> std_distr(0, 65535);
+
+  // Returns random integer in interval (0, max_value - 1)
+  auto rng = [&std_rng, &std_distr](size_t max_value) -> size_t {
+    return static_cast<size_t>(std_distr(std_rng) / 65536.0f * max_value);
+  };
+
+  // Dark background gradient color
+  uint16_t r0 = rng(32768);
+  uint16_t g0 = rng(32768);
+  uint16_t b0 = rng(32768);
+  uint16_t a0 = rng(32768);
+  uint16_t r1 = rng(32768);
+  uint16_t g1 = rng(32768);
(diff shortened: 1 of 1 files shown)


lib/jxl/test_utils.h
```diff
@@ -273,25 +273,60 @@ std::vector<ColorEncodingDescriptor> AllEncodings() {
 std::vector<uint8_t> GetSomeTestImage(size_t xsize, size_t ysize,
                                       size_t num_channels, uint16_t seed) {
   // Cause more significant image difference for successive seeds.
-  std::mt19937 rng(seed);
-  std::uniform_int_distribution<uint16_t> dark(0, 32767);
+  std::mt19937 std_rng(seed);
+  std::uniform_int_distribution<uint16_t> std_distr(0, 65535);
+
+  // Returns random integer in interval (0, max_value - 1)
```
jan-wassenberg, line 279: Nit: uniform_int_distribution takes a closed interval, so this is in [0, max_value).

lib/jxl/test_utils.h
```diff
@@ -273,25 +273,60 @@ std::vector<ColorEncodingDescriptor> AllEncodings() {
...
       // put some shape in there for visual debugging
-      if (x * x + y * y < 1000) {
+      if ((x - circle_x) * (x - circle_x) + (y - circle_y) * (y - circle_y) <
+          circle_r * circle_r) {
         r = (65535 - x * y) ^ seed;
         g = (x << 8) + y + seed;
         b = (y << 8) + x * seed;
         a = 32768 + x * 256 - y;
+      } else if (x > rect_x0 && x < rect_x1 && y > rect_y0 && y < rect_y1) {
+        r = rng(65536);
```
jan-wassenberg, line 326: This could actually overflow u16 if the RNG returns the max value 65535.

lib/jxl/test_utils.h
lvandeve: Thanks for spotting that, dividing through 65536.0f instead should have fixed this

lib/jxl/test_utils.h
lvandeve: Thanks for spotting that! The fix from the other comment should have fixed this

lib/jxl/test_utils.h
jan-wassenberg: Sounds good, thanks for making the change!

jan-wassenberg (approved):

PR #233, opened 2021-06-28
Merge multiple DoUndoXYBInPlace() paths.

deymo:
Remove the template parameter from DoUndoXYBInPlace and replace the call
to Op::Transform with a virtual function call to a common base class
type. This stops inlining the call to Transform but allows the compiler
to not have multiple versions of DoUndoXYBInPlace().

In the x86 build with 4 SIMD targets, dec_reconstruct.o contribution to
libjxl_dec.so goes from 67 KB to 51 KB. In arm 32-bits with only one
target this reduces 2.7 KB.

### DIFF
--- a/lib/jxl/dec_reconstruct.cc
+++ b/lib/jxl/dec_reconstruct.cc
@@ -37,8 +37,23 @@ HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
 
-template <typename Op>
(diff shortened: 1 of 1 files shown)


deymo:
@veluca93 what do you think about the performance impact of this?

veluca93:
> @veluca93 what do you think about the performance impact of this?

Can you measure it?

How is the binary size reduction of having an op for a whole line?

deymo:
Added gbench in #830, this is quite a bit slower in the Linear case.

PR #237, opened 2021-06-28
correctly convert CMYK to RGB

jonsneyers:
Makes CMYK jxl images (images with a kBlack extrachannel) get converted correctly to RGB, i.e. using 4-channel input, 3-channel output.

This works with skcms, I haven't tested if it also works with lcms (what's the magic incarnation to compile things with lcms instead of skcms?)


jonsneyers:
My clang-format seems to have slightly different opinions about minor things (a space here or there) compared to the one in the CI.
I am using Debian clang-format version 11.0.1-2. @deymo do you have a suggestion on what to do with that?

jonsneyers:
Might want to have this in 0.5 too, to make sure djxl properly converts cmyk jxl images to png. @deymo @veluca93

deymo:
> Might want to have this in 0.5 too, to make sure djxl properly converts cmyk jxl images to png. @deymo @veluca93

Can this wait for 0.6?  I'm not sure what images/use cases would this break.

veluca93:
> > Might want to have this in 0.5 too, to make sure djxl properly converts cmyk jxl images to png. @deymo @veluca93
> 
> Can this wait for 0.6? I'm not sure what images/use cases would this break.

no strong opinion, it mostly matters for CMYK which is not *super* common...

jonsneyers:
Use cases are limited mostly to the printing industry, where cmyk is very common. Doing something correct here is nice if we want to make djxl convert all conformance bitstreams correctly in 0.5. In general, if we want to claim jxl supports cmyk, it is nice that the reference decoder tool handles it correctly (as opposed to just ignoring the black and the color profile, showing an incorrect image instead).

deymo:
> Use cases are limited mostly to the printing industry, where cmyk is very common. Doing something correct here is nice if we want to make djxl convert all conformance bitstreams correctly in 0.5. In general, if we want to claim jxl supports cmyk, it is nice that the reference decoder tool handles it correctly (as opposed to just ignoring the black and the color profile, showing an incorrect image instead).

I agree with what you say. I'm not opposed to this change at all, the question is whether we should cherry-pick this to v0.5.x instead of waiting for it to have more testing in main (fuzzing, using a decoder with this change for images *other* than cmyk and not detecting any problem). There's already a lot in v0.5.

tools/djxl.cc
```diff
@@ -242,8 +241,11 @@ jxl::Status WriteJxlOutput(const DecompressArgs& args, const char* file_out,
   // (Writing large PNGs is slow, so allow skipping it for benchmarks.)
   if (file_out == nullptr) return true;
 
-  // Override original color space with arg if specified.
+  // Stay in original color space unless something else is needed.
   jxl::ColorEncoding c_out = io.metadata.m.color_encoding;
+  // Override original color space with sRGB if input is CMYK.
+  if (io.Main().HasBlack()) c_out = jxl::ColorEncoding::SRGB(false);
```
deymo, line 247: Isn't there any change required for the C API to handle this? @lvandeve

deymo:
> I am using Debian clang-format version 11.0.1-2. @deymo do you have a suggestion on what to do with that?

CI checks if clang-format 7, 8 or 9 are happy with the patch; if any of them is happy, CI is happy. We can add more. clang-format evolved over time.

jonsneyers:
> > Use cases are limited mostly to the printing industry, where cmyk is very common. Doing something correct here is nice if we want to make djxl convert all conformance bitstreams correctly in 0.5. In general, if we want to claim jxl supports cmyk, it is nice that the reference decoder tool handles it correctly (as opposed to just ignoring the black and the color profile, showing an incorrect image instead).
> 
> I agree with what you say. I'm not opposed to this change at all, the question is whether we should cherry-pick this to v0.5.x instead of waiting for it to have more testing in main (fuzzing, using a decoder with this change for images _other_ than cmyk and not detecting any problem). There's already a lot in v0.5.

I'm fine with waiting a bit to test it better. I just would like at least djxl and preferably also applications to handle CMYK correctly as soon as possible, but it doesn't matter much if that happens in 0.5 or in 0.6.

tools/djxl.cc
lvandeve: One question to ask is: is sRGB the only option here, or can we choose others and may others be better in some cases? sRGB cannot represent all CMYK images

In the API, I don't think a change is needed to handle this (other than in the implementation, since djxl is only the decoder binary) since it already can choose and force the output color format now too

However other options to handle CMYK need to be added to the API, such as to get the 4 CMYK channels without conversion. That's different than this discussion though. The extra channels feature now can only get K.

jonsneyers: It's certainly not the only option - this is just what it does by default when nothing else is specified. It does have to be an RGB space if the goal is to output to any of the output formats djxl supports, and sRGB seemed like a reasonable choice, but indeed a wider gamut space that contains more of the typical CMYK gamuts might make more sense.

Getting the 4 channels without conversiom is possible in the api afaik: the color image contains CMY with 0=max ink (pretending to be RGB with 1=max light, but with an ICC profile that is CMYK), and you can get get K as an extra channel.

jonsneyers: It might be good to add an input/output format like TIFF that supports CMYK (currently we can use CMYK PSD input but there's no output format in djxl to get CMYK as CMYK).

jonsneyers:
This was postponed for v0.5, could we merge it now?

In a future high-level decode API that also does color management, we'll need to make sure to properly deal with CMYK. What this PR proposes for djxl is what would need to end up in the high-level decode API.

Applications that are using the current low-level API will most likely not handle CMYK correctly, since it requires passing 4 channels to the CMS and they're likely not doing that (they'll likely pass 3 channels, which will probably result in a CMS error since the ICC profile will be a CMYK one).

jonsneyers:
@sboukortt PTAL.

This PR makes things work in djxl, but for the API implementation, we have to think about how we want to do things. I would try to make it in such a way that if you plug in a CMS (or use the default jxl one), then CMYK gets automagically converted to RGB, so applications don't have to deal with the special case of 4 color channels. When there is no CMS, you'll a CMYK ICC profile and you can get the CMY in the RGB channels and have to fetch the kBlack separately - it would also be nice to have a way then to get a CMYK buffer instead.

lib/jxl/enc_image_bundle.cc
```diff
@@ -50,6 +50,33 @@ Status CopyToT(const ImageMetadata* metadata, const ImageBundle* ib,
...
+              rect.ConstPlaneRow(ib->color(), 2, y);
+          const float* JXL_RESTRICT row_in3 = rect.ConstRow(ib->black(), y);
+          for (size_t x = 0; x < rect.xsize(); x++) {
+#if JPEGXL_ENABLE_SKCMS
+            // SKCMS does CMYK like JXL: 0 = max ink, 1 = white
+            mutable_src_buf[4 * x + 0] = row_in0[x];
+            mutable_src_buf[4 * x + 1] = row_in1[x];
+            mutable_src_buf[4 * x + 2] = row_in2[x];
+            mutable_src_buf[4 * x + 3] = row_in3[x];
+#else
```
sboukortt, line 72: This has a similar problem to what #958 fixed. Is there a way that we could have this logic inside of `JxlCms`? Perhaps as a correction pass that only occurs in the LCMS case? It could look like the “Expand from 1 to 3 channels” bit in that other PR for example.

sboukortt (changes requested):

lib/jxl/enc_image_bundle.cc
jonsneyers: OK, I'll move it to the same place where gray gets expanded to 3 channels for SKCMS.

jonsneyers: Done, PTAL.

lib/jxl/enc_color_management.cc
```diff
@@ -244,10 +253,13 @@ Status DoColorSpaceTransform(void* cms_data, const size_t thread,
     }  // else: in-place, no need to copy
   } else {
 #if JPEGXL_ENABLE_SKCMS
-    JXL_CHECK(skcms_Transform(
-        xform_src, skcms_PixelFormat_RGB_fff, skcms_AlphaFormat_Opaque,
-        &t->profile_src, buf_dst, skcms_PixelFormat_RGB_fff,
-        skcms_AlphaFormat_Opaque, &t->profile_dst, xsize));
+    JXL_CHECK(
+        skcms_Transform(xform_src,
+                        (t->channels_src != 4 ? skcms_PixelFormat_RGB_fff
```
sboukortt, line 258: This is quite minor but I feel that perhaps this would be more legible with the condition inverted.

sboukortt (approved):
Thanks! Looks good.

lib/jxl/enc_color_management.cc
jonsneyers: updated.

sboukortt (approved):

PR #238, opened 2021-06-28
Allow arbitrary center for group order

mo271:
This adds command line parameters `center_x` and `center_y` allowing to
specify the first tile, i.e. the center for tile order.
When no value is given the geomemtric middle of the image is used as a
center, as it was before.

### DIFF
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -1339,9 +1339,27 @@ Status EncodeFrame(const CompressParams& cparams_orig,
     std::vector<coeff_order_t> ac_group_order(num_groups);
     std::iota(ac_group_order.begin(), ac_group_order.end(), 0);
     size_t group_dim = frame_dim.group_dim;
-    // The center of the image.
-    int64_t imag_cx = ib.xsize() / 2;
-    int64_t imag_cy = ib.ysize() / 2;
+
+    // The center of the image is either given by parameters or chosen
+    // to be the middle of the image by default if center_x, center_y resp.
+    // are not provided.
+
+    int64_t imag_cx;
+    if (cparams.center_x != static_cast<size_t>(-1)) {
+      JXL_RETURN_IF_ERROR(cparams.center_x < ib.xsize());
+      imag_cx = cparams.center_x;
+    } else {
+      imag_cx = ib.xsize() / 2;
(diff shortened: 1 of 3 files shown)


lib/jxl/enc_frame.cc
```diff
@@ -1339,9 +1339,20 @@ Status EncodeFrame(const CompressParams& cparams_orig,
     std::vector<coeff_order_t> ac_group_order(num_groups);
     std::iota(ac_group_order.begin(), ac_group_order.end(), 0);
     size_t group_dim = frame_dim.group_dim;
-    // The center of the image.
-    int64_t imag_cx = ib.xsize() / 2;
-    int64_t imag_cy = ib.ysize() / 2;
+    // The center of the image is either given by parameters or chosen
+    // to be the middle of the image by default if center_x, center_y resp.
+    // are not provided.
+    if (cparams.center_x != size_t(-1)) {
```
veluca93, line 1345: `static_cast<size_t>(-1)`

lib/jxl/enc_frame.cc
```diff
@@ -1339,9 +1339,20 @@ Status EncodeFrame(const CompressParams& cparams_orig,
...
+      JXL_RETURN_IF_ERROR(cparams.center_x < ib.xsize());
+    }
+    if (cparams.center_y != size_t(-1)) {
+      JXL_RETURN_IF_ERROR(cparams.center_y < ib.ysize());
+    }
+    int64_t imag_cx =
+        cparams.center_x == size_t(-1) ? ib.xsize() / 2 : cparams.center_x;
```
veluca93, line 1352: this should probably go in the if

lib/jxl/enc_frame.cc
mo271: Done. Should I also cast it when setting the defaults in enc_params.h?

lib/jxl/enc_frame.cc
mo271: Done

veluca93:
Yes, please static cast everywhere (or use {} instead of ()...)

mo271:
> Yes, please static cast everywhere (or use {} instead of ()...)

Done

veluca93 (approved):

PR #243, opened 2021-06-29
Always define the fuzzer build targets

deymo:
When JPEGXL_ENABLE_FUZZERS is off, we weren't defining any of the fuzzer
targets except in coverage builds (when JPEGXL_ENABLE_COVERAGE is on).
This patch instead defines the fuzzer targets as "djxl_fuzzer_runner"
when JPEGXL_ENABLE_FUZZERS is off but excluded from the "all" target.

There are no changes to the targets defined when JPEGXL_ENABLE_FUZZERS
is on, except that djxl_fuzzer depends on the decoder side only, helping
the fuzzer library with the coverage analysis by removing all the unused
encoder and extras library code.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -2183,8 +2183,6 @@ JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
   return JXL_DEC_SUCCESS;
 }
 
-#if JXL_IS_DEBUG_BUILD || defined(JXL_ENABLE_FUZZERS)
 void SetDecoderMemoryLimitBase_(size_t memory_limit_base) {
   memory_limit_base_ = memory_limit_base;
 }
-#endif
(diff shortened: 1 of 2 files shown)


lib/jxl/decode.cc
```diff
@@ -2183,8 +2183,6 @@ JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
   return JXL_DEC_SUCCESS;
 }
 
-#if JXL_IS_DEBUG_BUILD || defined(JXL_ENABLE_FUZZERS)
 void SetDecoderMemoryLimitBase_(size_t memory_limit_base) {
```
eustas, line 2186: So, this function will become part of almost public API? @lvandeve WDYT?

eustas (approved):

lib/jxl/decode.cc
deymo: No, this is not exported and won't be on the shared library. The fuzzer_runner is using the static library which has access to all the internal symbols.

A functionality to limit the memory usage could be added to the API, or gracefully handling of OOM conditions could be added to the library functions. I'm not sure this particular function is the best approach to handle the situation and it was a workaround that needs cleanup.

PR #250, opened 2021-06-30
Encoder: override color_transform for grayscale JPEG images with AdobeRGB ICC.

eustas:
Otherwise decoder puts all zeroes instead of gray channel.

### DIFF
--- a/lib/jxl/jpeg/enc_jpeg_data.cc
+++ b/lib/jxl/jpeg/enc_jpeg_data.cc
@@ -355,7 +355,7 @@ Status DecodeImageJPG(const Span<const uint8_t> bytes, CodecInOut* io) {
 
   io->Main().chroma_subsampling = cs;
   io->Main().color_transform =
-      !is_rgb ? ColorTransform::kYCbCr : ColorTransform::kNone;
+      (!is_rgb || nbcomp == 1) ? ColorTransform::kYCbCr : ColorTransform::kNone;
 
   io->metadata.m.SetIntensityTarget(
       io->target_nits != 0 ? io->target_nits : kDefaultIntensityTarget);


veluca93 (approved):

lib/jxl/dec_external_image.cc
```diff
@@ -350,7 +350,7 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
             const float* JXL_RESTRICT row_in[4];
             size_t c = 0;
             for (; c < color_channels; c++) {
-              row_in[c] = color->PlaneRow(c, y);
+              row_in[c] = color->PlaneRow(color_channels == 1 ? 1 : c, y);
```
lvandeve, line 353: Why 1 exactly? Wouldn't all 3 channels be the same when the image is grayscale?

jonsneyers: The channels are in Cb, Y, Cr order (it's that way to match XYB order so chroma from luma can be used) and in the case of grayscale the Cb and Cr channels are not used.

lvandeve: But this is in ConvertToExternal, nothing specifies what color format the image is in.

Now, if rgb and ycbcr are the only possibilities, where both happen to work (since for grayscale, rgb's g, and component 1 of Cb,Y,Cr it happens to match), then this is fine, but are we sure of this?

jonsneyers: Agreed this is probably not the right place to fix it.

eustas:
@lvandeve PTAL

jonsneyers:
The following tests FAILED:
	1623 - RoundtripTest.FloatFrameRoundtripTest (Failed)

eustas:
Oh, that is, actually, encoder bug...
PTAL

veluca93:
> Oh, that is, actually, encoder bug...
> PTAL

*facepalm*... Nice find!

veluca93 (approved):

PR #259, opened 2021-07-01
Allocate output in case of rasterisation of partially-decoded DC

### LINKED ISSUE #254: Tests fail on ASAN build
**Describe the bug**
```
        1576 - DecodeTest.PixelTestOpaqueSrgbLossyNoise (Failed)
        1637 - DecodeTest/DecodeTestParam.PixelTest/301x33RGBtoRGBu8#GetParam()=301x33RGBtoRGBu8 (Failed)
        1661 - DecodeTest/DecodeTestParam.PixelTest/301x33RGBtoRGBAu8#GetParam()=301x33RGBtoRGBAu8 (Failed)
        1957 - RoundtripTest.Uint8FrameRoundtripTest (Failed)
        1960 - RoundtripTest.TestICCProfile (Failed)
```

**To Reproduce**
/ci.sh asan

eustas:


veluca93 (approved):

lib/jxl/dec_frame.cc
```diff
@@ -909,6 +909,12 @@ Status FrameDecoder::FinalizeFrame() {
         "FinalizeFrame called before the frame was fully decoded");
   }
 
+  if (!finalized_dc_) {
+    JXL_DASSERT(allow_partial_frames_);
```
veluca93, line 913: Are we sure that this DASSERT cannot be triggered by erraneous usage?

jonsneyers: Probably best to use regular JXL_ASSERT in cases like this where it's not in a hot loop — makes it more likely to catch things.

eustas: Yup, will change to DASSERT.
decoded_ac_global_ checked in preceding `if` is set to true at the end of ProcessACGlobal that is invoked if finalized_dc_. This assertion is more for readability than for safety.

eustas: PTAL

veluca93 (approved):

PR #264, opened 2021-07-01
Dissable ci.sh ossfuzz_ninja for msan.

deymo:
Calling ninja directly doesn't work in the msan case because the
msan libraries need to be copied to the docker container every time.
Disable this option since there's otherwise no indication that
ninja built with the non-instrumented std libraries.


eustas (approved):
Yup, faced the same problem, but ossfuzz_msan rebuilds are rather quick for me, so that ninja was not useful anyways.

lvandeve (approved):

deymo:
I had this hooked up to my IDE so I would rebuild all the time. The difference between just ninja and calling cmake adds up.

Issue #266, opened 2021-07-02
Attempting to create a lossy JXL from a lossless JXL produces a dark image


cagelight:
cjxl v0.3.7 [AVX2,SSE4,Scalar]

A few times now, I've wanted to decode a lossless JXL and re-encode it into a lossy JXL. Since cjxl does not allow input of JXL files, I use an intermediary PNG. At some point in the chain however, color information seems to be lost as see here:
![2021-07-01_20-59](https://user-images.githubusercontent.com/3761279/124205579-875bbb80-daaf-11eb-85a3-ccd6a86b2502.png)
1.png: Source PNG
2.jxl: Lossless JXL (`cjxl 1.png 2.jxl -d 0`)
3.png: Intermediary PNG (`djxl 2.jxl 3.png`) (NOTE: Same thing happens with PPM)
4.jxl: Lossy JXL (`cjxl 3.png 4.jxl -d 1`)

If I do step 4 directly with 1.png, the issue does not occur, leading me to believe it is a bug with `djxl`

Images: [img.tar.gz](https://github.com/libjxl/libjxl/files/6751477/img.tar.gz)

TheHardew:
How do you display the files? With `djxl `4.jxl 5.png`, 5.png looks identical to 1.png.

cagelight:
It looks identical for me too with another layer of djxl, but I'm using the Qt plugin to view them. While there could be a bug with the Qt plugin here playing a part, ultimately I strongly feel 1.png and 3.png should have identical color data such that this bug with the Qt plugin shouldn't even be possible.

jonsneyers:
`jxlinfo 2.jxl` gives something that is weird:
```
  transfer_function: 65535
  transfer_function gamma: 0.454550
```

@lvandeve @sboukortt what does that transfer_function mean? that's not a valid value, is that a bug?

sboukortt:
It’s the value that signals a gamma TRC:

https://github.com/libjxl/libjxl/blob/3c54e982df76f46f1162e87a7235342ba592581a/lib/include/jxl/color_encoding.h#L82-L84

DavidBuchanan314:
`3.png` has an ICC colour profile, added by djxl

sboukortt:
With #270 now merged, would you give that another try starting from `djxl 2.jxl 3.png`?

cagelight:
All is well on my end, thank you for the quick fix!

PR #270, opened 2021-07-02
PNG: try to preserve the original color tags when writing

sboukortt:
Currently, we are quick to convert the color information to an ICC profile (https://github.com/libjxl/libjxl/issues/266#issuecomment-873052852), which means that when decoding to PNG (with iCCP and gAMA), reencoding to JXL (reading iCCP but ignoring gAMA), and decoding to PNG again (writing iCCP only), the final PNG does not have the original gAMA tag, although it does have a functionally equivalent ICC profile. With this, its set of tags is closer to the original’s.

Because we recognized the sRGB primaries, we also used not to write them back, but now we do. We also include the recommended gAMA chunk for sRGB images.

### DIFF
--- a/lib/extras/codec_png.cc
+++ b/lib/extras/codec_png.cc
@@ -510,7 +510,9 @@ class ColorEncodingWriterPNG {
     if (c.IsSRGB()) {
       JXL_RETURN_IF_ERROR(AddSRGB(c, info));
       // PNG recommends not including both sRGB and iCCP, so skip the latter.
-    } else {
+    } else if (!c.HaveFields() || !c.tf.IsGamma()) {
+      // Having a gamma value means that the source was a PNG with gAMA and
+      // without iCCP.
       JXL_ASSERT(!c.ICC().empty());
       JXL_RETURN_IF_ERROR(AddICC(c.ICC(), info));
     }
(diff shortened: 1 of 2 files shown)


lib/extras/codec_png.cc
```diff
@@ -569,8 +571,16 @@ class ColorEncodingWriterPNG {
...
                              LodePNGInfo* JXL_RESTRICT info) {
-    if (!c.tf.IsGamma()) return true;
-    const double gamma = c.tf.GetGamma();
+    double gamma;
+    if (c.tf.IsGamma()) {
+      gamma = c.tf.GetGamma();
+    } else if (c.tf.IsLinear()) {
+      gamma = 1;
+    } else if (c.tf.IsSRGB()) {
+      gamma = 0.45455;
```
lvandeve, line 580: This is only an approximation to sRGB, would you say it's useful to add this at all?

Fortunately, sRGB and ICC chunk override the gama chunk, so it doesn't hurt at least, so I'm ok to keep

jonsneyers (approved):

lvandeve (approved):

lib/extras/codec_png.cc
sboukortt: Yep :) and more generally:

> An application that writes the iCCP chunk should also write gAMA and cHRM chunks that approximate the ICC profile's transfer function, for compatibility with applications that do not use the iCCP chunk.

So I think this value should be fine even for colorspaces that are not sRGB but only share its TRC (e.g. Display P3).

PR #272, opened 2021-07-03
Force disable XYB when lossless mode detected

### LINKED ISSUE #257: 14-bit XYB encoding incorrectly used when JxlEncoderOptionsSetLossless is set to JXL_TRUE
I was asked to create this issue from the discord. libjxl produces substantially larger files than cjxl in lossless mode due to the usage of 14-bit XYB, which in addition to being larger is also not truly lossless.

Code used below:
```CPP
m_opt = JxlEncoderOptionsCreate(m_enc, nullptr);
if (quality < 0) quality = 0;
JXLEE(JxlEncoderOptionsSetDistance(m_opt, quality));
JXLEE(JxlEncoderOptionsSetLossless(m_opt, quality == 0 ? JXL_TRUE : JXL_FALSE));
JXLEE(JxlEncoderOptionsSetEffort(m_opt, 3));

static constexpr JxlPixelFormat PFMT {
	.num_channels = 3,
	.data_type = JXL_TYPE_UINT8,
	.endianness = JXL_NATIVE_ENDIAN,
	.align = 0
};

m_info.xsize = width;
m_info.ysize = height;
m_info.bits_per_sample = 8;
m_info.num_color_channels = 3;
m_info.intensity_target = 255;
m_info.orientation = JXL_ORIENT_IDENTITY;
JXLEE(JxlEncoderSetBasicInfo(m_enc, &m_info));

JxlColorEncoding color_profile;
JxlColorEncodingSetToSRGB(&color_profile, JXL_FALSE);
JXLEE(JxlEncoderSetColorEncoding(m_enc, &color_profile));
JXLEE(JxlEncoderAddImageFrame(m_opt, &PFMT, buf, width * height * 3));
JxlEncoderCloseInput(m_enc);
```
(`quality` is set to 0 in this circumstance)

cagelight:
I admit I'm not 100% if this is the best approach to fix it, but I thought I'd try my hand regardless. It solves the problem for me, but I am unsure if this solution may infringe upon other scenarios.
Fixes #257


cagelight:
@googlebot I signed it!

ziemek99:
AUTHORS has just been fixed, PTAL and enable workflows since it's @cagelight's first contribution and they can't be run otherwise.

lib/jxl/encode.cc
```diff
@@ -378,6 +378,11 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
     return JXL_ENC_ERROR;
   }
 
+  if (options->values.lossless) {
+    // XYB does not make sense for lossless encoding, ensure it is disabled.
+    options->enc->metadata.m.xyb_encoded = false;
```
lvandeve, line 383: Thank you for proposing the fix. Modifying the value here when adding image frame may not be the correct approach, maybe it could return an error instead here, and JxlEncoderOptionsSetLossless should set this instead. I wonder how this worked before actually!

In which scenario did you encounter this issue? That might help to find the best solution.

cagelight: In the linked issue, you can see the code that generates the issue. The encoder at the moment seems to always want to use XYB encoding, lossless or not.

cagelight: `JxlEncoderOptionsSetLossless` could set it, but in the current code it would be overwritten by `JxlEncoderSetBasicInfo`
This just seemed like the most simple and elegant solution to me at least.

lvandeve: I responded to the bug: the solution is to set m_info.uses_original_profile = JXL_TRUE before calling JxlEncoderSetBasicInfo. So lossless, and xyb color are two independent settings, and there's no need to force one in the encoder. Does this fix it?

jonsneyers:
What is the status of this?

AFAIU, if you want to do lossless, you have to disable XYB (set use_original_profile=true) _and_ use lossless options on all frames. Once you started writing a lossy frame in XYB, all frames will have to be in XYB, so true lossless is no longer possible — though you can still do 'lossless XYB', which could be useful e.g. when having a photo layer with a logo/text overlay.

While it might theoretically be useful to do 'lossless XYB', it also leads to applications accidentally doing that instead of actual lossless, which leads to worse compression and pixel values not being actually lossless. I think it is too easy to make that mistake right now. Perhaps it would make sense to return API usage error instead when you set use_original_profile=false and then try to encode the first and only frame with lossless settings, since it is almost certainly a mistake. There could be valid reasons to use combinations of XYB and lossless (in multi-frame / multi-layer scenarios), but I think it can only be a mistake when lossless XYB is done on a single frame.

lib/jxl/encode.cc
jonsneyers: The problem is that xyb_encoded is an image setting (part of SetBasicInfo) so it is too late when adding a frame to change it (since the image header is already written here).

Returning an error when trying to use lossless when xyb_encoded is true would make sense though, since actual lossless is no longer possible in that case.

jonsneyers:
Closing, https://github.com/libjxl/libjxl/pull/1270 addresses this.

PR #277, opened 2021-07-05
Fix md-links

eustas:
Paths are relative - files in doc/ directory should not use doc/ path prefix in links.


deymo (approved):

deymo:
That's unfortunate; oh well.

PR #278, opened 2021-07-05
Fixing the build in msan

jyrkialakuijala:
msan needs a higher butteraugli score than non-msan

This is of course not ideal, the result should be the same.

PR 273 that surfaced this however did not introduce this discrepancy,
and rolling it back would not solve the root cause -- so 'fixing'
the problem with upping the limit. I have filed a bug about the
msan and non-msan not matching and we will discuss that later.


jan-wassenberg (approved):

eustas:
Would you mind creating an issue? The behaviour should be fixed at some point, so it is better to leave the "TODO" for that (+ nice repro case, by adjusting just one value).

PR #280, opened 2021-07-05
Use Poison/Unpoison instead of changing code behavior in msan.

deymo:
This uses Poison and Unpoison on the last portion of the last vector to
handle the case where we do NearestInt or DemoteTo on uninitialized
memory in dec_external_image.cc.

### DIFF
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -24,6 +24,7 @@
 #include "lib/jxl/base/compiler_specific.h"
 #include "lib/jxl/color_management.h"
 #include "lib/jxl/common.h"
+#include "lib/jxl/sanitizers.h"
 #include "lib/jxl/transfer_functions-inl.h"
 
 HWY_BEFORE_NAMESPACE();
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_external_image.cc
```diff
@@ -67,30 +70,29 @@ void FloatToU32(const float* in, uint32_t* out, size_t num, float mul,
...
-  size_t vec_num = num;
-#if JXL_IS_DEBUG_BUILD
-  // Avoid accessing partially-uninitialized vectors with memory sanitizer.
-  vec_num &= ~(Lanes(d) - 1);
-#endif  // JXL_IS_DEBUG_BUILD
-  for (size_t x = 0; x < vec_num; x += Lanes(d)) {
+
+  // Unpoison accessing partially-uninitialized vectors with memory sanitizer.
+  // This is because we run DemoteTo() on the vector which triggers msan on
+  // scalar implementation.
```
veluca93, line 84: Uhm, this feels wrong - unpoison/poison should be a noop on scalar, no?

eustas: Hopefully they are. Those are no-op in non-MSAN build; and semantical no-op for scalar MSAN build (num_round_up == num)

eustas (dismissed):

lib/jxl/dec_external_image.cc
veluca93: Right, but then the comment is suspicious...

lib/jxl/dec_external_image.cc
```diff
@@ -43,10 +44,12 @@ void FloatToU32(const float* in, uint32_t* out, size_t num, float mul,
     // instructions, so we run the non-SIMD path for those.
     vec_num = 0;
```
eustas, line 45: I belive we should also remove vec_num; now it is either 0 or num, so it will be more readable, if we check `bits_per_sample == 32` around the places that depend on that.

lib/jxl/dec_external_image.cc
eustas: Yup, it will be more clear if we mention that transformation is lane-wise and DemoteTo() might use comparison results in computation (and MSAN hates that).

deymo: hmmm.. well the DemoteTo implementation in scalar definitively triggers msan because of its if(); it is possible that the simd implementation of DemoteTo doesn't; but it is true that in scalar this would be a no-op. We were already processing the last vector separately in msan so I would keep this code here but will update the comment.

veluca93 (dismissed):

lib/jxl/dec_external_image.cc
deymo: Sorry, didn't see this comment before; updated the patch with this change.

veluca93 (approved):

PR #283, opened 2021-07-06
Remove ancient encoder flag

ziemek99:
To say that this flag is deprecated would be an understatement since it's been marked as such from the very first public commit ff09371267315c39fe0220082943c5834db04ab9 from 2019.


jonsneyers (approved):

deymo:
Yeah, it doesn't look like any client is setting this anymore.

deymo (approved):

PR #284, opened 2021-07-06
Abort `jxl_from_tree` if output filename same as input

ziemek99:
@sboukortt reported on Discord that `jxl_from_tree` can overwrite its input if the provided output filename is the same as input; now it should throw an error and exit, just like in case of wrong number of args.

### DIFF
--- a/tools/jxl_from_tree.cc
+++ b/tools/jxl_from_tree.cc
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file.
 
 #include <stdio.h>
+#include <string.h>
 
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


sboukortt:
Oh, thanks 😁 if you feel like it, perhaps a more generic check could be to verify that the extension of the output is `.jxl`, or possibly just “not `.txt`”. The `Extension` function from `jxl/base/file_io.h` should be useful for that:

https://github.com/libjxl/libjxl/blob/df9a35578234447490a5c423e69729ef2cd7507f/lib/jxl/base/file_io.h#L22-L29

ziemek99:
It's more like a "sanity check" quickfix that just gets the job done, but big thanks for pointing that code out, I might use it in the future! 😃

sboukortt (approved):

sboukortt:
Sounds good, thanks again 😄

PR #286, opened 2021-07-06
Add PoisonImage and CheckImageInitialized to sanitizers.h

deymo:
This adds two new Image-related msan helper functions: PoisonImage will
mark all the Image memory (including padding) as uninitialized while
CheckImageInitialized will check that the pixels inside a given Rect are
is initialized in an image.

When running a multi-thread step with per-thread temporary images we can
use PoisonImage to avoid leaking initialized memory from a previous run
in the same thread to the next run. This helps identifying where was the
uninitialized memory from.

Functions accepting images, specially those that take chunks of images
with a given rect should use CheckImageInitialized over the passed Rect
to trigger msan early when uninitialized memory bugs arise, otherwise
msan would only trigger when the uninitialized values are used in a
comparison (like NearestInt or IfThenElse).

Since now there are more functions in the sanitizers.h header and all of
them are msan-only, this also moves the functions inside the jxl::msan
namespace.

### DIFF
--- a/lib/extras/codec_gif.cc
+++ b/lib/extras/codec_gif.cc
@@ -81,13 +81,15 @@ Status DecodeImageGIF(Span<const uint8_t> bytes, ThreadPool* pool,
     return JXL_FAILURE("Failed to read GIF: %s", GifErrorString(gif->Error));
   }
 
-  UnpoisonMemory(gif.get(), sizeof(*gif));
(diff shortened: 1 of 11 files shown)


veluca93 (approved):

eustas (commented):
Cool!

PR #296, opened 2021-07-07
Fix clang-tidy bugprone-argument-comment errors

deymo:
The only actual bug fixed here is in modular_test which was passing
maxval instead of bitdepth.


veluca93 (approved):
Should we add this to some clang-tidy settings too? (i.e. make it become an error)

deymo:
> Should we add this to some clang-tidy settings too? (i.e. make it become an error)

We are not blocking anything on clang-tidy, so error/no-error doesn't make a difference. There are some GitHub apps that can show clang-tidy errors on PRs, but I haven't set that up yet.

PR #298, opened 2021-07-08
Fuzz: introduce CPU limit alongside memory limit.

eustas:
With ANS coding it is easy to make CPU-bomb.

In example discovered by fuzzer there were hundreds of frames,
each encoded with just 39 bytes. Nevertheless, this fires the
full image decoding / processing machinery.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -29,6 +29,8 @@ namespace {
 // TODO(eustas): this is a poor-mans replacement for memory-manager approach;
 //               remove, once memory-manager actually works.
 size_t memory_limit_base_ = 0;
+size_t cpu_limit_base_ = 0;
+size_t used_cpu_base_ = 0;
 
 bool CheckSizeLimit(size_t xsize, size_t ysize) {
   if (!memory_limit_base_) return true;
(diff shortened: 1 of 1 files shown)


lib/jxl/decode.cc
```diff
@@ -2185,4 +2201,6 @@ JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
 
 void SetDecoderMemoryLimitBase_(size_t memory_limit_base) {
   memory_limit_base_ = memory_limit_base;
+  // Allow 5 x max_image_size processing units.
+  cpu_limit_base_ = 5 * memory_limit_base;
```
veluca93, line 2205: Doesn't this result in a hard limit on the # of frames in an image?

eustas: nope. it is sum(w * h) limit.

veluca93: Sure, but if you have a long animation you'll go over this number, no?

eustas: Of course. 1x1 animation in this case is allowed to have no more than 10M frames.
But it does not limit the number of frames - limitation is applied to frames being decoded. IIUC skipped frames won't count for that limit.

veluca93:
Is this applied only in msan mode?

eustas:
*san. The only user of `SetDecoderMemoryLimitBase_` is `djxl_fuzzer`

veluca93:
> *san. The only user of `SetDecoderMemoryLimitBase_` is `djxl_fuzzer`

Can we make a comment about this then? Thanks!

veluca93 (approved):

eustas:
Done. PTAL

veluca93 (approved):

PR #303, opened 2021-07-08
Relax x0 % kPaddingXRound requirement in FilterPipeline

deymo:
The FilterPipeline required that the x0 in all three input, output and
image rects be a multiple of kPaddingXRound. This has the issue that it
is 4 in ARM but 8 in x86, however Sigma is always an 8x8 downsampled
image.

This patch relaxes the requirement so that input, output and image rect
all have the same alignment % kPaddingXRound but it doesn't need to be
0, removing the logic that extends the filters ranges from
dec_reconstruct.

The EPF functions were receiving a image_x_mod_8 value as
image_rect.x0() % 8, which was always 0 in x86, but may be 4 in ARM.
This value now can be anything between 0 and 7 in both architectures
since it depends on the value of x0.

This patch overall reduces a tiny bit the total range of [x0, x1) where
we run intermediate filters since the FilterPipeline now handles the
values of image_rect.x0() that are not a multiple of kPaddingXRound.

This removes the rect parameter from ApplyFiltersRow() since it is
already passed on initialization and it must be the same every time.

### DIFF
--- a/lib/jxl/epf.cc
+++ b/lib/jxl/epf.cc
@@ -141,7 +141,7 @@ void GaborishVector(const D df, const float* JXL_RESTRICT row_t,
 
 void GaborishRow(const FilterRows& rows, const LoopFilter& /* lf */,
                  const FilterWeights& filter_weights, size_t x0, size_t x1,
-                 size_t /*image_x_mod_8*/, size_t /* image_y_mod_8 */) {
+                 size_t /*sigma_x_offset*/, size_t /* image_y_mod_8 */) {
   JXL_DASSERT(x0 % Lanes(df) == 0);
 
   const float* JXL_RESTRICT gab_weights = filter_weights.gab_weights;
@@ -192,7 +192,7 @@ void GaborishRow(const FilterRows& rows, const LoopFilter& /* lf */,
 // plus-shaped). So this makes this filter a 7x7 filter.
 void Epf0Row(const FilterRows& rows, const LoopFilter& lf,
              const FilterWeights& filter_weights, size_t x0, size_t x1,
-             size_t image_x_mod_8, size_t image_y_mod_8) {
+             size_t sigma_x_offset, size_t image_y_mod_8) {
   JXL_DASSERT(x0 % Lanes(df) == 0);
   const float* JXL_RESTRICT row_sigma = rows.GetSigmaRow();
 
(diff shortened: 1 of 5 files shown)


lib/jxl/filters.h
```diff
@@ -312,6 +334,10 @@ class FilterPipeline {
   // Whether we need to compute the sigma_row_ during application.
   bool compute_sigma = false;
 
+  // Rect to be processed in the image coordinated. This doesn't include any
```
veluca93, line 337: coordinates

lib/jxl/epf.cc
```diff
@@ -648,13 +657,24 @@ FilterPipeline* PrepareFilterPipeline(
     const Rect& input_rect, size_t image_ysize, size_t thread,
     Image3F* JXL_RESTRICT out, const Rect& output_rect) {
   const LoopFilter& lf = dec_state->shared->frame_header.loop_filter;
-  JXL_DASSERT(image_rect.x0() % GroupBorderAssigner::kPaddingXRound == 0);
-  JXL_DASSERT(input_rect.x0() % GroupBorderAssigner::kPaddingXRound == 0);
-  JXL_DASSERT(output_rect.x0() % GroupBorderAssigner::kPaddingXRound == 0);
-  JXL_DASSERT(input_rect.x0() >= lf.Padding());
+  // image_rect, input and output must all have the same kPaddingXRound
+  // alignment for SIMD, but it doesn't need to be 0.
+  JXL_DASSERT(image_rect.x0() % GroupBorderAssigner::kPaddingXRound ==
```
veluca93, line 662: I think we actually need it to be 0 because we do aligned loads

veluca93:
Did you check that you get the exact same results in benchmark_xl on x86?

lib/jxl/epf.cc
deymo: The filter function will get a different x0 because it needs to process border needed for future stages; instead of `filter_x0 = RoundDown(RoundDown(x0, 4) - border, 4)` this patch does `filter_x0 = RoundDown(x0 - border, 4)` so the image_rect.x0 here doesn't need to aligned.

veluca93: Ah, I didn't notice this was epf.cc :) nvm...

deymo:
Tested with benchmark_xl for codecs `jxl:fast:d4:gab{0,1}:epf{0,1,2,3}` in x86, same butteraugli numbers. I'll test with aarch64

lib/jxl/filters.h
deymo: Done

veluca93 (approved):

deymo:
This produced no changes on the jyrki31 in aarcdh64. The numbers are different between aarch64 and x86; but this commit didn't change them

PR #305, opened 2021-07-09
fix psd decoder bug

jonsneyers:
Fixes loading of PSD files that have multiple layers, spot colors and a merged image without alpha (while the layers do of course have alpha).


deymo (approved):
I'm not particularly familiar with the PSD codec but I don't see any problem with this PR.

PR #307, opened 2021-07-09
Add an option to generate noise of more photorealistic intensity

sboukortt:
For now, whenever the parameter is set, it is used as-is without
attenuation for lower distances.

The sparseness of the LUT at the bottom end makes this not quite as
fine-grained (no pun intended) as would be ideal but it’s a start.


veluca93 (commented):
Test please? :)

sboukortt:
> Test please? :)

What sort of test do you think would be appropriate? Maybe just a few LUTs for a few input values?

veluca93:
Yes, and/or roundtrip tests

sboukortt:
How is this?

veluca93 (approved):

PR #316, opened 2021-07-12
Define JPEGXL_VERSION in main CMakeLists

eddiezato:
Define `JPEGXL_VERSION` in the main CMakeLists to use it to provide additional version information when building. This flag is already used in [tools/CMakeLists](https://github.com/libjxl/libjxl/blob/main/tools/CMakeLists.txt#L58).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,8 @@ set(JPEGXL_DEP_LICENSE_DIR "" CACHE STRING
     "Directory where to search for system dependencies \"copyright\" files.")
 set(JPEGXL_FORCE_NEON false CACHE BOOL
     "Set flags to enable NEON in arm if not enabled by your toolchain.")
-
+set(JPEGXL_VERSION "" CACHE STRING
+    "Build JPEGXL with additional version information.")
 
 # Force system dependencies.
 set(JPEGXL_FORCE_SYSTEM_GTEST false CACHE BOOL
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,8 +17,9 @@ Google LLC <*@google.com>
 
 # Individuals:
 Dirk Lemstra <dirk@lemstra.org>
+Dmitry <eddie.zato@gmail.com>
 Jon Sneyers <jon@cloudinary.com>
-Pieter Wuille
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -109,7 +109,8 @@ set(JPEGXL_DEP_LICENSE_DIR "" CACHE STRING
     "Directory where to search for system dependencies \"copyright\" files.")
 set(JPEGXL_FORCE_NEON false CACHE BOOL
     "Set flags to enable NEON in arm if not enabled by your toolchain.")
-
+set(JPEGXL_VERSION "" CACHE STRING
```
deymo, line 112: What's the idea behind this? The PR description says to provide additional version information but this patch alone doesn't provide additional information. You can already pass `-DJPEGXL_VERSION=123456` to cmake, but I don't think it would be beneficial to cache this value in cmake.

eddiezato: Now the `-DJPEGXL_VERSION` flag is ignored when building.

deymo: ok, that's a different problem. The issue here is the automatic variable JPEGXL_VERSION that cmake creates based on the project(); any other name would work. We should fix that instead.

eddiezato: Simply defining `JPEGXL_VERSION` in CMakelists solves the problem for me. But if there is more proper solution, I'd be glad to have it. 😉

deymo: #321 explains what's the issue here. The more proper solution would be to set the major.minor.patch version at the project line, but I'd need to move around some macros since they are currently only defined in the lib.gni and lib/CMakeLists.txt and we have some tooling to keep them in sync. However, that would define the JPEGXL_VERSION as the version number (like 0.3.7) which is fine, but we also want to have a version string you can include in the tools defined from packages which was the purpose of this macro; so changing the project name to avoid the collision with JPEGXL_VERSION sounds a reasonable first step.

eddiezato: Thanks for the explanation. If this PR is not needed, please feel free to close it.

deymo: Thank you for reporting the issue!

PR #319, opened 2021-07-14
Properly unpoison/poison memory in SIMDfied noise decoding.

veluca93:
This silences msan.

### DIFF
--- a/lib/jxl/dec_noise.cc
+++ b/lib/jxl/dec_noise.cc
@@ -22,6 +22,7 @@
 #include "lib/jxl/chroma_from_luma.h"
 #include "lib/jxl/image_ops.h"
 #include "lib/jxl/opsin_params.h"
+#include "lib/jxl/sanitizers.h"
 #include "lib/jxl/xorshift128plus-inl.h"
 HWY_BEFORE_NAMESPACE();
 namespace jxl {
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_noise.cc
```diff
@@ -215,13 +216,18 @@ void AddNoise(const NoiseParams& noise_params, const Rect& noise_rect,
...
+  const size_t xsize_v = RoundUpTo(xsize, Lanes(d));
+
   for (size_t y = 0; y < ysize; ++y) {
     float* JXL_RESTRICT row_x = opsin_rect.PlaneRow(opsin, 0, y);
     float* JXL_RESTRICT row_y = opsin_rect.PlaneRow(opsin, 1, y);
     float* JXL_RESTRICT row_b = opsin_rect.PlaneRow(opsin, 2, y);
     const float* JXL_RESTRICT row_rnd_r = noise_rect.ConstPlaneRow(noise, 0, y);
     const float* JXL_RESTRICT row_rnd_g = noise_rect.ConstPlaneRow(noise, 1, y);
     const float* JXL_RESTRICT row_rnd_c = noise_rect.ConstPlaneRow(noise, 2, y);
+    msan::UnpoisonMemory(row_x + xsize, (xsize_v - xsize) * sizeof(float));
```
deymo, line 228: Please explain in a comment why this is needed and why it is safe to do. I think that actually only row_x and row_y are used, and that's because of the Floor() call in NoiseStrength(). AddNoiseToRGB seems to be ok.

deymo (approved):

PR #323, opened 2021-07-15
Fuzz: Initialize all modular channels in all cases

deymo:
When reading a channel larger than the max_chan_size we were leaving the
remaining channels uninitialized. This patch makes sure all the channels
are initialized (in msan builds) after calling ModularDecode() and
initializes the remaining ones in that case.

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -12,6 +12,7 @@
 
 #include "lib/jxl/modular/encoding/context_predict.h"
 #include "lib/jxl/modular/options.h"
+#include "lib/jxl/sanitizers.h"
 
 namespace jxl {
 
@@ -469,6 +470,8 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
     }
     if (i >= image.nb_meta_channels && (channel.w > options->max_chan_size ||
                                         channel.h > options->max_chan_size)) {
+      ZeroFillImage(&channel.plane);
+      while (++i < nb_channels) ZeroFillImage(&image.channel[i].plane);
       break;
     }
     JXL_RETURN_IF_ERROR(DecodeModularChannelMAANS(br, &reader, *context_map,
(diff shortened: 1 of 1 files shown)


jonsneyers:
This does ensure everything will always be initialized, but it does like come at a bit of a decode speed cost - could you compare before/after decode speed of a large image encoded with e.g. `cjxl -m -q 80 -s 2`?

The channels after the break _should_ always be initialized later — either when the Modular DC or AC groups are decoded and the decoded values are written there, or when that decode gets skipped because it's a truncated / partially loaded bitstream and `ModularFrameDecoder::DecodeGroup()` gets called with `zerofill = true`.

If there is no speed difference, I'm fine with this change (but then the zerofill codepath becomes unnecessary and should be removed). If there is a speed difference, then I'd prefer to find out how there is something uninitialized left at the point transforms are undone, and fix that.

jonsneyers:
One issue (besides speed) I have with always initializing , is that it will hide the problem for msan but doesn't solve the semantic problem of values that do not get written — from a security point of view, all that matters is not exposing uninitialized memory, but from a correctness point of view, it doesn't solve the problem to use zeroes instead of the values that should have been there.

In principle, there should only be uninitialized memory in the partial/truncated/progressively-decoding bitstream case, not in the complete bitstream case. Do you have an example bitstream that causes uninitialized memory to be read? I can take a look at what is happening:

- if it's a valid (complete) bitstream, it's a bug that does not get fixed by zero-initializing;
- if it's a valid but incomplete bitstream, it's a bug that does get fixed by this but that would better be fixed by zero-initializing later (and only do it for the missing groups);
- if it's an invalid bitstream, then likely an error should have been thrown instead of continuing to a point where uninitialized memory gets read.

jonsneyers:
The underlying bug is the following (it's quite similar to the one in https://github.com/libjxl/libjxl/pull/210):
- Palette introduces metachannels
- Custom squeeze script then squeezes metachannels (adding more metachannels) but doesn't update nb_meta_channels
- This results in trouble that eventually leads to uninitialized memory getting read.

Again, question is what to do with such cases of weird custom Squeeze scripts. I think it makes most sense to do the same thing as with Palette and RCT: only allow it to operate on either all meta-channels or all non-meta-channels, and in case of all meta-channels, to only allow in-place squeeze (not putting meta-channels at the end), and increment the nb_meta_channels counter accordingly.
@veluca93 WDYT?

jonsneyers:
Instead of explicitly zero-initializing channels all the time, which will come at a decode speed cost and will hide bugs, could we instead add a flag to channels to see if they contain uninitialized memory or not? Basically init the flag to true, set it to false after decoding a channel, check flags before undoing transforms. If there are channels with uninitialized memory, it's a bug that needs to be fixed.

PR #325, opened 2021-07-15
Support extra channel output

lvandeve:
Allows outputting extra channels to individual single-channel buffers,
similar to the main image output

Adds two new functions to the API: JxlDecoderExtraChannelBufferSize
and JxlDecoderSetExtraChannelBuffer

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -828,6 +828,58 @@ JxlDecoderSetImageOutCallback(JxlDecoder* dec, const JxlPixelFormat* format,
                               JxlImageOutCallback callback, void* opaque);
 
 /**
+ * Returns the minimum size in bytes of an extra channel pixel buffer for the
+ * given format. This is the buffer for JxlDecoderSetExtraChannelBuffer.
+ * Requires the basic image information is available in the decoder.
+ *
+ * @param dec decoder object
+ * @param format format of the pixels. The num_channels value is ignored and is
+ * always treated to be 1.
+ * @param size output value, buffer size in bytes
+ * @param index which extra channel to get, matching the index used in @see
+ * JxlDecoderGetExtraChannelInfo. Must be smaller than num_extra_channels in the
+ * associated JxlBasicInfo.
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as
+ *    information not available yet or invalid index.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderExtraChannelBufferSize(
+    const JxlDecoder* dec, const JxlPixelFormat* format, size_t* size,
+    uint32_t index);
+
+/**
+ * Sets the buffer to write an extra channel to. This can be set when
+ * the JXL_DEC_FRAME or JXL_DEC_NEED_IMAGE_OUT_BUFFER event occurs, and applies
+ * only for the current frame. The size of the buffer must be at least as large
+ * as given by JxlDecoderExtraChannelBufferSize. The buffer follows the format
+ * described by JxlPixelFormat, but where num_channels is 1. The buffer is owned
+ * by the caller. The amount of extra channels is given by the
+ * num_extra_channels field in the associated JxlBasicInfo, and the information
+ * of individual extra channels can be queried with @see
+ * JxlDecoderGetExtraChannelInfo. To get multiple extra channels, this function
+ * must be called multiple times, once for each wanted index. Not all images
+ * have extra channels. The alpha channel is an extra channel and can be gotten
+ * as part of the color channels when using an RGBA pixel buffer with
+ * JxlDecoderSetImageOutBuffer, but additionally also can be gotten separately
+ * as extra channel. The color channels themselves cannot be gotten this way.
+ *
+ *
+ * @param dec decoder object
+ * @param format format of the pixels. Object owned by user and its contents
+ * are copied internally. The num_channels value is ignored and is always
+ * treated to be 1.
+ * @param buffer buffer type to output the pixel data to
+ * @param size size of buffer in bytes
+ * @param index which extra channel to get, matching the index used in @see
+ * JxlDecoderGetExtraChannelInfo. Must be smaller than num_extra_channels in the
+ * associated JxlBasicInfo.
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as
+ * size too small or invalid index.
+ */
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetExtraChannelBuffer(JxlDecoder* dec, const JxlPixelFormat* format,
+                                void* buffer, size_t size, uint32_t index);
+
+/**
  * Sets output buffer for reconstructed JPEG codestream.
  *
  * The data is owned by the caller
@@ -860,8 +912,6 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
  * JxlDecoderReleaseJPEGBuffer was already called.
  */
 JXL_EXPORT size_t JxlDecoderReleaseJPEGBuffer(JxlDecoder* dec);
-
-/* TODO(lode): add way to output extra channels */
 
 /**
  * Outputs progressive step towards the decoded image so far when only partial
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -512,5 +512,17 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
       pool, out_image, out_size, out_callback, out_opaque, undo_orientation);
 }
 
+Status ConvertToExternal(const jxl::ImageF& channel, size_t bits_per_sample,
+                         bool float_out, JxlEndianness endianness,
+                         size_t stride, jxl::ThreadPool* pool, void* out_image,
+                         size_t out_size, JxlImageOutCallback out_callback,
+                         void* out_opaque, jxl::Orientation undo_orientation) {
+  const ImageF* channels[1];
+  channels[0] = &channel;
+  return ConvertChannelsToExternal(
+      channels, 1, bits_per_sample, float_out, endianness, stride, pool,
+      out_image, out_size, out_callback, out_opaque, undo_orientation);
+}
+
 }  // namespace jxl
 #endif  // HWY_ONCE
--- a/lib/jxl/dec_external_image.h
+++ b/lib/jxl/dec_external_image.h
@@ -40,6 +40,22 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
                          size_t out_size, JxlImageOutCallback out_callback,
                          void* out_opaque, jxl::Orientation undo_orientation);
 
+// Converts single-channel image to interleaved void* pixel buffer with the
+// given format, with a single channel.
+// bits_per_sample: must be 8, 16 or 32, and must be 32 if float_out
+// is true. 1 and 32 int are not yet implemented.
+// This supports the features needed for the C API to get extra channels.
+// stride_out is output scanline size in bytes, must be >=
+// output_xsize * output_bytes_per_pixel.
+// undo_orientation is an EXIF orientation to undo. Depending on the
+// orientation, the output xsize and ysize are swapped compared to input
+// xsize and ysize.
+Status ConvertToExternal(const jxl::ImageF& channel, size_t bits_per_sample,
+                         bool float_out, JxlEndianness endianness,
+                         size_t stride_out, jxl::ThreadPool* thread_pool,
+                         void* out_image, size_t out_size,
+                         JxlImageOutCallback out_callback, void* out_opaque,
+                         jxl::Orientation undo_orientation);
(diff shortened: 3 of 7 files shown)


jonsneyers:
For the CMYK case, it might make more sense to get output as 4-channel CMYK (like we do with RGBA now) than to get a "main image" that is CMY and then a separate extra channel for the K.

For the RGBG case (Bayer data), the same is probably true: you likely want to get all channels together.

For now, this is good enough, but perhaps it would make sense to have other 4-channel output paths than RGBA, in case there is some other extra channel that is not alpha but still interesting to interleave with the main RGB data (kBlack or kCFA spring to mind, but possibly kDepth too?). This should be relatively easy to implement (just selecting a different extra channel to act as the "alpha channel" in the RGBA code path).

The general case of interleaved N-channel output (for N>4) is probably not worth doing and just planar access to the extra channels is good enough in those cases (e.g. when CMYKA is used or multiple spot colors).

lib/jxl/decode.cc
```diff
@@ -438,10 +438,15 @@ struct JxlDecoderStruct {
   size_t preview_out_size;
   size_t image_out_size;
 
-  // TODO(lode): merge these?
   JxlPixelFormat preview_out_format;
   JxlPixelFormat image_out_format;
 
+  // For extra channels
+  std::vector<bool> extra_channel_wanted;  // Empty if none, reset each frame
```
veluca93, line 445: Can we avoid vector<bool> here?

lib/jxl/decode_test.cc
```diff
@@ -2415,6 +2416,87 @@ TEST(DecodeTest, AnimationTestStreaming) {
...
+  }
+
+  EXPECT_EQ(JXL_DEC_SUCCESS, JxlDecoderSetImageOutBuffer(
+                                 dec, &format, image.data(), image.size()));
+  EXPECT_EQ(JXL_DEC_SUCCESS, JxlDecoderSetExtraChannelBuffer(
+                                 dec, &format, extra.data(), extra.size(), 0));
+
+  EXPECT_EQ(JXL_DEC_FULL_IMAGE, JxlDecoderProcessInput(dec));
+
+  // After the full image is gotten, JxlDecoderProcessInput should return
```
veluca93, line 2475: "is gotten"?

veluca93 (dismissed):
Code LGTM

lib/jxl/decode.cc
lvandeve: You mean for all or only the bool one? What do you propose instead?

lib/jxl/decode_test.cc
lvandeve: Changed to "was output"

lib/jxl/decode.cc
veluca93: uh, github ate my answer :P `vector<bool>`

lvandeve: Done, removed

lib/jxl/decode.cc
```diff
@@ -882,29 +889,28 @@ static size_t GetStride(const JxlDecoder* dec, const JxlPixelFormat& format,
   return stride;
 }
 
-static JxlDecoderStatus ConvertImageInternal(const JxlDecoder* dec,
-                                             const jxl::ImageBundle& frame,
-                                             const JxlPixelFormat& format,
-                                             void* out_image, size_t out_size,
-                                             JxlImageOutCallback out_callback,
-                                             void* out_opaque) {
+static JxlDecoderStatus ConvertImageInternal(
```
deymo, line 892: Can we have documentation of what's this function doing? It does quite different things depending on `want_extra_channel`; if it is true we get the extra channel only; but if it is false we get the color image. Maybe worth it to  have a different function for the "want_extra_channel==true" case, since it uses a different set of parameters, but for example do we need the RGBBuffer if want_extra_channel is true? We should assert that num_channels==1 in that case too since that should have been handled before.

lib/jxl/dec_external_image.cc
```diff
@@ -281,6 +290,8 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
   Image3F temp_color, unpremul;
   const ImageF* alpha = ib.HasAlpha() ? &ib.alpha() : nullptr;
   ImageF temp_alpha;
+  const ImageF* extra = &ib.extra_channels()[extra_channel_index];
```
deymo, line 293: `extra = want_extra_channel ? ... : nullptr;`  extra_channel_index is not relevant if want_extra_channel is false and even if you specify that it must be 0 it could be possible that extra_channels is empty.

lib/jxl/dec_external_image.cc
```diff
@@ -408,6 +429,9 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
             if (want_alpha) {
               row_in[c++] = ib.HasAlpha() ? alpha->Row(y) : ones.Row(0);
             }
+            if (want_extra_channel) {
```
deymo, line 432: I know this is correct, but it would make it more obvious what's going on if you do `if (want_extra_channel) { row_in[c++] = extra->Row(y); } else { ... set color and alpha ... }`. I think the fact that this logic is repeated 3 times merits a refactor of this code, for example you can split setting up which Planes need to be converted and separately the code that actually converts them, transposes them, etc.

lib/jxl/dec_external_image.cc
```diff
@@ -303,20 +314,27 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
     color = &unpremul;
```
deymo, line 314: Do we need to do this unpremul processing in the want_extra_channel case? I think this ends up not being used so we don't need to spend time here. This shows the problems of reusing this function to do something different when a bool parameter is true.

lib/jxl/dec_external_image.h
```diff
@@ -35,6 +35,7 @@ namespace jxl {
 // xsize and ysize.
 Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
```
deymo, line 36: Please document what's the meaning of want_extra_channel and what are the restrictions about calling this function with want_extra_channel=true.

lib/jxl/decode.cc
```diff
@@ -1073,7 +1079,8 @@ JxlDecoderStatus JxlDecoderProcessInternal(JxlDecoder* dec, const uint8_t* in,
       if (want_preview) {
         if (dec->preview_out_buffer) {
           JxlDecoderStatus status = ConvertImageInternal(
```
deymo, line 1081: Are there extra channels in the preview?

deymo (commented):
Please add a call to these function to the djxl_fuzzer and run it for a bit, I'm pretty sure it would have detected indexing the empty vector.

lib/jxl/decode.cc
lvandeve: Yes, the same image metadata applies to it. Outputting those can be added in a later merge request with a function such as `JxlDecoderSetExtraChannelPreviewOutBuffer`

lib/include/jxl/decode.h
```diff
@@ -827,6 +827,57 @@ JXL_EXPORT JxlDecoderStatus
...
+ * channel is an extra channel and can be gotten as part of the color channels
+ * when using an RGBA pixel buffer with JxlDecoderSetImageOutBuffer, but
+ * additionally also can be gotten separately as extra channel. The color
+ * channels themselves cannot be gotten this way.
+ *
+ *
+ * @param dec decoder object
+ * @param format format of the pixels. Object owned by user and its contents
+ * are copied internally. The num_channels value is ignored and is always
+ * treated to be 1.
```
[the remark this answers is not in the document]
lvandeve: Done

lib/include/jxl/decode.h
```diff
@@ -827,6 +827,57 @@ JXL_EXPORT JxlDecoderStatus
...
+ * @param size size of buffer in bytes
+ * @param index which extra channel to get, matching the index used in @see
+ * JxlDecoderGetExtraChannelInfo. Must be smaller than num_extra_channels in the
+ * associated JxlBasicInfo.
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as
+ * size too small or invalid index.
+ */
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetExtraChannelBuffer(JxlDecoder* dec, const JxlPixelFormat* format,
+                                void* buffer, size_t size, size_t index);
```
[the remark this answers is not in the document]
lvandeve: Done

lib/jxl/decode.cc
```diff
@@ -438,10 +438,14 @@ struct JxlDecoderStruct {
   size_t preview_out_size;
   size_t image_out_size;
 
-  // TODO(lode): merge these?
   JxlPixelFormat preview_out_format;
   JxlPixelFormat image_out_format;
 
+  // For extra channels. Empty of none, and they are reset each frame
+  std::vector<JxlPixelFormat> extra_channel_format;
```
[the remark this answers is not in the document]
lvandeve: Done, using a struct now

lib/jxl/dec_external_image.cc
lvandeve: No longer an issue after https://github.com/libjxl/libjxl/pull/338

lib/jxl/dec_external_image.h
lvandeve: Done

lib/jxl/dec_external_image.cc
lvandeve: No longer an issue after https://github.com/libjxl/libjxl/pull/325

lib/jxl/dec_external_image.cc
lvandeve: No longer an issue after https://github.com/libjxl/libjxl/pull/325

lib/jxl/decode.cc
lvandeve: Fixed for external_image with https://github.com/libjxl/libjxl/pull/325, and documented for this internal function. I believe the internal function is simple enough to not split it into 3 functions (RGB(A) case, extra channel case, and the shared implementation), and documentation is added.

lvandeve:
> Please add a call to these function to the djxl_fuzzer and run it for a bit, I'm pretty sure it would have detected indexing the empty vector.

Added to fuzzer. Not sure which empty vector you mean though.

deymo:
> Added to fuzzer. Not sure which empty vector you mean though.

`  const ImageF* extra = &ib.extra_channels()[extra_channel_index];`

lvandeve:
> > Added to fuzzer. Not sure which empty vector you mean though.
> 
> ` const ImageF* extra = &ib.extra_channels()[extra_channel_index];`

I see, that code is gone. The API returns error if you try to get external image with smaller index. The fuzzer now 

> > Added to fuzzer. Not sure which empty vector you mean though.
> 
> ` const ImageF* extra = &ib.extra_channels()[extra_channel_index];`

I see, that code is gone. The API checks if you give a valid index, and it's now also fuzzed.

I also added back error checking for grayscale vs 3 channels in JxlDecoderImageOutBufferSize and JxlDecoderPreviewOutBufferSize, which the fuzzer expects. I had removed this due to reusing the code for extra channels where 1 channel does not mean grayscale, however the fuzzer expects it, and it makes sense that the "Size" functions do the same error checks as the "set buffer" functions.

PTAL

tools/djxl_fuzzer.cc
```diff
@@ -334,6 +337,22 @@ bool DecodeJpegXl(const uint8_t* jxl, size_t size, size_t max_pixels,
         seen_need_image_out = true;
       }
 
+      if (info.num_extra_channels > 0) {
+        size_t ec_index = info.num_extra_channels - 1;
```
deymo, line 341: Probably a good idea to make the requested index a random number % (info.num_extra_channels). Or better % (info.num_extra_channles+1) and use one value to not request extra channels even if there are extra channels.

deymo (approved):

tools/djxl_fuzzer.cc
lvandeve: Done

lib/jxl/decode.cc
```diff
@@ -438,10 +444,12 @@ struct JxlDecoderStruct {
   size_t preview_out_size;
   size_t image_out_size;
 
-  // TODO(lode): merge these?
   JxlPixelFormat preview_out_format;
   JxlPixelFormat image_out_format;
 
+  // For extra channels. Empty of none, and they are reset each frame
```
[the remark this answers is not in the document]
lvandeve: Was typo for "if", made the sentence longer

lib/jxl/dec_external_image.h
```diff
@@ -40,6 +40,12 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
                          size_t out_size, JxlImageOutCallback out_callback,
                          void* out_opaque, jxl::Orientation undo_orientation);
 
+Status ConvertToExternal(const jxl::ImageF& channel, size_t bits_per_sample,
```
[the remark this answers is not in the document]
lvandeve: Done

deymo (approved):
Thanks for the updates!

PR #326, opened 2021-07-15
Fix name mismatch "associated" and "premultiplied"

lvandeve:
Fix name mismatch in the API. These are corresponding fields and
it was a historical error that they appeared independently with
a different name.

Also add to the comments that they match that of the main alpha
extra channel.

This renames a field in the API, but I believe that because the API
did not yet support outputting extra channels, usage of this field
should be extremely rare.

### DIFF
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -203,15 +203,20 @@ typedef struct JxlBasicInfo {
   uint32_t num_extra_channels;
 
   /** Bit depth of the encoded alpha channel, or 0 if there is no alpha channel.
+   * If present, matches the alpha_bits value of the JxlExtraChannelInfo
+   * associated with this alpha channel.
    */
   uint32_t alpha_bits;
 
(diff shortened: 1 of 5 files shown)


jonsneyers (approved):

lib/include/jxl/codestream_header.h
```diff
@@ -257,7 +262,7 @@ typedef struct {
   /** Whether alpha channel uses premultiplied alpha. Only applicable if
    * type is JXL_CHANNEL_ALPHA.
    */
-  JXL_BOOL alpha_associated;
+  JXL_BOOL alpha_premultiplied;
```
deymo, line 265: Mention this change in the changelog in a new Unreleased.Changed section since this will go in 0.6

deymo (approved):

lib/include/jxl/codestream_header.h
lvandeve: I'll do that in a next merge request soon since this involves chapter changes in the changelog during the 0.5/0.6 transition

Issue #28, opened 2021-05-26
Add check that libjxl builds with clang-5.0


deymo:
Mozilla uses clang-5.0 for some builds; but it doesn't work when compiling in C++17 mode (it is fine in C++11 mode) for what appears to be a clang-5 compiler bug.

We should add a check that libjxl compiles in release mode with clang-5 to not regress here.

Details: https://gitlab.com/wg1/jpeg-xl/-/issues/227 @saschanaz FYI

eustas:
Currently blocked by:
```
[34/418] Building CXX object third_party/highway/CMakeFiles/hwy_contrib.dir/hwy/contrib/image/image.cc.o
FAILED: third_party/highway/CMakeFiles/hwy_contrib.dir/hwy/contrib/image/image.cc.o 
ccache /usr/bin/clang++-5.0 -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -I../third_party/highway -ffunction-sections -fdata-sections -funwind-tables -Xclang -mrelax-all -Xclang -mconstructor-aliases -fno-omit-frame-pointer -O3 -DNDEBUG -O2 -fPIC -Wno-builtin-macro-redefined -Wall -Werror -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -fmerge-all-constants -Wextra -Wformat-security -Wno-unused-function -Wnon-virtual-dtor -Woverloaded-virtual -Wvla -Wc++2a-extensions -Wfloat-overflow-conversion -Wfloat-zero-conversion -Wfor-loop-analysis -Wgnu-redeclared-enum -Winfinite-recursion -Wself-assign -Wstring-conversion -Wtautological-overlap-compare -Wthread-safety-analysis -Wundefined-func-template -fno-cxx-exceptions -fno-slp-vectorize -fno-vectorize -fdiagnostics-show-option -fcolor-diagnostics -fmath-errno -fno-exceptions -std=c++11 -MD -MT third_party/highway/CMakeFiles/hwy_contrib.dir/hwy/contrib/image/image.cc.o -MF third_party/highway/CMakeFiles/hwy_contrib.dir/hwy/contrib/image/image.cc.o.d -o third_party/highway/CMakeFiles/hwy_contrib.dir/hwy/contrib/image/image.cc.o -c ../third_party/highway/hwy/contrib/image/image.cc
[...]

PR #337, opened 2021-07-20
Fix clang-5 warnings

eustas:
See #28


veluca93 (approved):
I'm not sure how well this go but worth a try...

deymo:
It didn't go that far :-/ Maybe try running this workflow on `ubuntu-18.04` (bionic) instead of latest.

eustas:
Yup, trying on ubuntu-18

eustas:
Depends on HWY roll after https://github.com/google/highway/commit/8fa39c099f26265e76a961f5ceaa70427a4d4669

.github/workflows/build_test.yml
```diff
@@ -17,40 +17,56 @@ on:
...
+            os: ubuntu-latest
             test_in_pr: true
           - name: debug
+            os: ubuntu-latest
           # Build scalar-only hwy instructions.
           - name: scalar
+            os: ubuntu-latest
             mode: release
             cxxflags: -DHWY_DISABLED_TARGETS=~HWY_SCALAR
+          # TODO(eustas): why ASAN has no params?
```
deymo, line 37: `mode` is the parameter we pass to ./ci.sh, and if mode is empty we use `name`.

asan, msan and debug don't need any parameters and that's why they are listed above. The rest don't need to be listed afaik.

.github/workflows/build_test.yml
```diff
@@ -17,40 +17,56 @@ on:
 jobs:
   ubuntu_build:
     name: Ubuntu Build ${{ matrix.name }}
-    runs-on: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
```
deymo, line 20: `${{ matrix.os || 'ubuntu-latest' }}` and you can remove all the `os: ubuntu-latest` lines.

.github/workflows/build_test.yml
```diff
@@ -17,40 +17,56 @@ on:
...
+            os: ubuntu-latest
             mode: release
             cxxflags: -DJXL_DEBUG_ON_ABORT=0
             cmake_args: >-
               -DJPEGXL_ENABLE_TRANSCODE_JPEG=OFF
               -DJPEGXL_ENABLE_PLUGINS=OFF
               -DJPEGXL_ENABLE_VIEWERS=OFF
           # Builds with gcc in release mode
-          - name: release:gcc8
+          - name: release-gcc8
```
deymo, line 57: why the rename? in gitlab we had the compiler as `:compiler`

.github/workflows/build_test.yml
eustas: Got it. Should we remove then `release, msan, scalar` from list and empty `debug` expando?

.github/workflows/build_test.yml
eustas: Done. Thanks

.github/workflows/build_test.yml
eustas: Because ":" can not be part of name in flat list in yml. Rolled back.

eustas:
Rolled HWY dep, fixed clang-5 specific syntax problems (unsupported curly brackets elimination). Now it builds successfully. PTAL @deymo.

deymo:
I think it would be better to have the highway update on a separated PR.
I see this is using 0.14 tag now, but I think there are a few new bugs in 0.14 due to the new SSSE3 target (for example a bug that makes it fail to run on older CPUs with no AES -- this is the fix https://github.com/google/highway/commit/241162b235b788dbfec98a89ae6d08290b0af4c4 or the issue that kCLMUL is aliased with kAVX in that file). I would prefer to update to a version of hwy that doesn't have these regressions.

The other fixes (to .cc and .h) SGTM and we can land those now.

eustas:
Separated code changes from HWY rolling + actions workflow changes.
PTAL

deymo (approved):

PR #338, opened 2021-07-20
Refactor ConvertToExternal to use an array of ImageF*

deymo:
The new `ConvertChannelsToExternal` uses a list of channels (ImageF*)
instead of repeating the logic of deciding which channels to include
in every lambda. This should make it easier to include a function to
convert extra channels.

### DIFF
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -248,14 +248,29 @@ void StoreFloatRow(const float* JXL_RESTRICT* rows_in, size_t num_channels,
 
 void JXL_INLINE Store8(uint32_t value, uint8_t* dest) { *dest = value & 0xff; }
 
-}  // namespace
+// Maximum number of channels for the ConvertChannelsToExternal function.
+const size_t kConvertMaxChannels = 4;
 
-Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
-                         bool float_out, size_t num_channels,
-                         JxlEndianness endianness, size_t stride,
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_external_image.cc
```diff
@@ -349,17 +347,12 @@ Status ConvertToExternal(const jxl::ImageBundle& ib, size_t bits_per_sample,
...
-              row_in[c++] = ib.HasAlpha() ? alpha->Row(y) : ones.Row(0);
+            const float* JXL_RESTRICT row_in[kConvertMaxChannels];
+            for (size_t c = 0; c < num_channels; c++) {
+              row_in[c] = channels[c] ? channels[c]->Row(y) : ones.Row(0);
             }
-            JXL_ASSERT(c == num_channels);
-            hwy::float16_t* JXL_RESTRICT row_f16[4];
-            for (size_t r = 0; r < c; r++) {
+            hwy::float16_t* JXL_RESTRICT row_f16[kConvertMaxChannels];
+            for (size_t r = 0; r < num_channels; r++) {
```
lvandeve, line 355: small nit (and fine to keep to prevent touching more code), now one could as well use 'c' as the loop variable everywhere for consistency with the other loops

lvandeve (approved):

lvandeve:
LGTM, yes this should help for the extra channels

lib/jxl/dec_external_image.cc
deymo: Yeah, that's a good point, I'll change it.

lvandeve (approved):

Issue #347, opened 2021-07-22
multibyte read out of bounds reported by oss-fuzz in libjxl master


jcupitt:
Hello everyone,

oss-fuzz thinks it has a multibyte read of of bounds in libjxl master, or more specifically in highway. I'm reporting here since the error is presumably in libjxl, although I've not checked.

asan output:

```
  | ==312304==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x631000012e00 at pc 0x000002d4a870 bp 0x7fa53b3f74b0 sp 0x7fa53b3f74a8
  | READ of size 32 at 0x631000012e00 thread T2
  | SCARINESS: 26 (multi-byte-read-heap-buffer-overflow)
  | #0 0x2d4a86f in Load<int> libjxl/third_party/highway/hwy/ops/x86_256-inl.h:1327:7
  | #1 0x2d4a86f in jxl::N_AVX2::RgbFromSingle(unsigned long, int const*, float, jxl::Image3<float>*, unsigned long, unsigned long, jxl::Rect&) libjxl/lib/jxl/dec_modular.cc:63:21
  | #2 0x2569861 in jxl::ThreadPool::SequentialRunnerStatic(void*, void*, int (*)(void*, unsigned long), void (*)(void*, unsigned int, unsigned long), unsigned int, unsigned int) libjxl/lib/jxl/base/data_parallel.cc:18:5
```

Minimised testcase:

www.rollthepotato.net/~john/.libjxl-buffer-overflow.jxl

jonsneyers:
Should be fixed now. Closing...
(please reopen if needed)

PR #348, opened 2021-07-23
Fuzz: Fix ModularImageToDecodedRect crop and add checks

deymo:
See #347.


veluca93 (approved):

deymo:
FTR, I'm not entirely sure this is the right fix, but the situation is that `decoded` is an image of 773x369 and the requested rect is Rect(0,256, 128x128) which is OOB.

jonsneyers (approved):

jonsneyers (approved):

deymo:
The code this is fixing was introduced in #205  which is not in v0.5.x branch, removing label.

PR #352, opened 2021-07-23
Fuzz: Fix Rect rounding error decoding extra channels.

deymo:
When using an upsampling factor > 1 for extra channels with an odd size
(or not a multiple of the upsampling factor) we need to round up the
Rect copied in ModularImageToDecodedRect().

This fixes an msan bug where some pixels near the right or bottom edge
can be uninitialized.

Added more msan checks to the extra channels to fail faster on this type
of errors.


jonsneyers (approved):

eustas:
@deymo 52 tests fail with `../lib/jxl/dec_modular.cc:574: JXL_DASSERT: r.IsInside(ch_in.plane)`

deymo:
The code this is fixing was introduced in #205 which is not in v0.5.x branch, removing label.

Issue #354, opened 2021-07-24
`cjxl` - incorrect `ExitCode` on failure


eddiezato:
**Describe the bug**
If `cjxl` failed to encode the input file, it still returns a successful `ExitCode`.

**To Reproduce**
Just try to encode any invalid file:
```Powershell
PS > cjxl 0.jpg 1.jxl
JPEG XL encoder v0.5.0 [SSE4,Scalar]
Failed to read image 0.jpg.
PS > $LastExitCode
0
```
```Cmd
cjxl 0.jpg 1.jxl
JPEG XL encoder v0.5.0 [SSE4,Scalar]
Failed to read image 0.jpg.

echo %errorlevel%
0
```

**Expected behavior**
The `ExitCode` must be different from 0.

**Environment**
 - OS: Windows 10 x64
 - Compiler version: mingw/clang 12.0.1
 - CPU type: x86_64
 - cjxl/djxl version string: v0.5.0 [SSE4,Scalar]

PR #356, opened 2021-07-26
cjxl exit codes

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/354

Also makes cjxl return different exit codes for different error conditions.


veluca93 (commented):
Constants please!

veluca93 (approved):

Issue #346, opened 2021-07-22
cjxl silently fails when losslessly transcoding a jpg


TheHardew:
**Describe the bug**
Encoding `1.jpg` appears to work, but no file is created.
`cjxl` output omits the `Including container: [size] bytes ([bpp] bpp).` message.
If the flags `-j` or `--strip` are used however, it does work.

**To Reproduce**
```
~/P/test $ ls
1.jpg
~/P/test $ cjxl 1.jpg 1.jxl
JPEG XL encoder v0.5.0 [AVX2]
Read 1024x768 image, 116.5 MP/s
Encoding [Container | JPEG, lossless transcode, squirrel | JPEG reconstruction data | 38973-byte Exif], 8 threads.
Compressed to 129617 bytes (1.319 bpp).
1024 x 768, 14.23 MP/s [14.23, 14.23], 1 reps, 8 threads.
~/P/test $ ls
1.jpg
```

**Expected behavior**
It should be possible to transcode this image or `cjxl` should output an error message and a non-zero exit status if the transcoding is not possible.

**Environment**
 - OS: Arch Linux
 - Compiler version: clang 12.0.1
 - CPU type: ryzen 2700x x86_64
 - cjxl/djxl version string: 31d3ba750c9a705689490cc57159ebf898b6d64c

**Additional context**
Compiled with https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libjxl-git and
```
CFLAGS="-march=native -O3 -pipe -fstack-protector-strong --param=ssp-buffer-size=4 -fno-plt -pthread -Wno-error -w"
CXXFLAGS="$CFLAGS"
```
The image:
![1](https://user-images.githubusercontent.com/7374061/126671020-0d292ffe-dbb1-44cf-b47c-c431ad987d70.jpg)

TheHardew:
It seems running `jpegtran -copy all -outfile 2.jpg 1.jpg` makes it, so that the file can be transcoded.

jonsneyers:
I get the following errors on this image, indicating that there's something strange about the original jpeg (it has 3 quantization tables but only uses 2 of them):
```
./lib/jxl/jpeg/jpeg_data.cc:200: JXL_FAILURE: Not all quant tables are used (3 tables, 3 used table mask)
../lib/jxl/fields.cc:711: JXL_RETURN_IF_ERROR code=1: CanEncode(fields, &extension_bits, &total_bits)
[...]

PR #358, opened 2021-07-26
handle case when jbrd fails at encoding

jonsneyers:
See https://github.com/libjxl/libjxl/issues/346

When for whatever reason, the `jbrd` could not be created, instead of silently failing, it will now still encode a jxl file (without `jbrd`) and produce a warning. The exit code returned is different from 0 in this case, so you can tell that something went wrong.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -25,6 +25,7 @@ enum CjxlRetCode : int {
   ERR_ENCODING,
   ERR_CONTAINER,
   ERR_WRITE,
+  DROPPED_JBRD,
 };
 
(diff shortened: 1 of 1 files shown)


jonsneyers:
This (to some extent) addresses https://github.com/libjxl/libjxl/issues/803

jonsneyers:
Can we merge this one? It will have to be rewritten once cjxl migrates to the api, but I think it's good to have this fix in asap. Now it just crashes without output in the release build whenever it has to transcode a jpeg where writing `jbrd` fails, which is Not So Good behavior.

(ping @lvandeve and @veluca93)

jyrkialakuijala (approved):

PR #359, opened 2021-07-26
allow redundant quant table in jbrd

jonsneyers:
See https://github.com/libjxl/libjxl/issues/346

This changes the behavior of `jbrd` so that if there are quantization tables that are not actually used (so the jxl codestream will not actually contain the values), it does not throw an error but it assumes that the unused table is exactly equal to the previous table. At encode time it checks if this is true, at decode time it is assumed that it is true.

The spec (18181-2) is a bit vague on this — it only says

> Qk are corresponding quantization factors from the codestream (see ISO/IEC 18181-1 / C.6.2)

but it's not really clarified/specified what exactly "corresponding" means here. We could clarify/extend the spec later, and specify what a decoder is supposed to do when there are no "corresponding" quantization factors.

Since there do seem to be JPEGs in the wild that do have a redundant quantization table, and it always seems to be just a duplication of the chroma quant table, I think it's worthwhile to interpret the spec a bit freely here and consider "corresponding" to mean "corresponding to what's in the jxl codestream if it is used, identical to the previous QuantTable otherwise".

### DIFF
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -542,16 +542,26 @@ Status FrameDecoder::ProcessACGlobal(BitReader* br) {
     size_t num_components = jpeg_data->components.size();
     bool is_gray = (num_components == 1);
     auto jpeg_c_map = JpegOrder(frame_header_.color_transform, is_gray);
+    size_t qt_set = 0;
     for (size_t c = 0; c < num_components; c++) {
       // TODO(eustas): why 1-st quant table for gray?
       size_t quant_c = is_gray ? 1 : c;
       size_t qpos = jpeg_data->components[jpeg_c_map[c]].quant_idx;
       JXL_CHECK(qpos != jpeg_data->quant.size());
+      qt_set |= 1 << qpos;
       for (size_t x = 0; x < 8; x++) {
         for (size_t y = 0; y < 8; y++) {
           jpeg_data->quant[qpos].values[x * 8 + y] =
               (*qe[0].qraw.qtable)[quant_c * 64 + y * 8 + x];
         }
+      }
+    }
+    for (size_t i = 0; i < jpeg_data->quant.size(); i++) {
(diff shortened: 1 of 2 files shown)


jonsneyers:
@deymo : I'm using clang-format-11, which sometimes does things differently than clang-format-7, 8 or 9, causing the format check to fail. Would it be possible to add versions 10 and 11 of clang-format?

deymo:
> @deymo : I'm using clang-format-11, which sometimes does things differently than clang-format-7, 8 or 9, causing the format check to fail. Would it be possible to add versions 10 and 11 of clang-format?

https://github.com/libjxl/libjxl/pull/401

ziemek99:
You should [re-run a workflow](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow) in order to allow merging this PR. You have to hurry up, though, because according to that page you've got 30 days after the initial run to do that!

ziemek99:
I've noticed that re-running workflows didn't help. Maybe it's because those workflows are from before clang-format commit? Try creating an empty commit (`git commit --allow-empty -m "trigger workflows"; git push`) and let's see if it works.

deymo:
or just rebase the change and reupload.

ziemek99:
> or just rebase the change and reupload.

Ok, I didn't think of it. That actually might be better solution.

jonsneyers:
This makes lossless jpeg recompression work on jpegs produced by various digital cameras where it previously didn't work. Worth adding into v0.6 imo.

veluca93 (dismissed):

jonsneyers:
Rebased, please re-approve.

veluca93 (approved):

ziemek99:
I think that this PR deserves the newly created "spec" label. @jonsneyers WDYT?

jonsneyers:
Indeed should be clarified in part 2 of the spec how to interpret "corresponding" in this case.

PR #362, opened 2021-07-26
Fix DASSERT() introduced in #352

deymo:
The code is copying from 0,0 in the ch_in channel to x0,y0 in the `r`
Rect. This updates the assert to match that.


eustas (approved):

jonsneyers (approved):

deymo:
The code this is fixing was introduced in #352 which is not in v0.5.x branch and we are not cherry-picking it, removing label.

PR #365, opened 2021-07-26
fix use_full_image==false case

jonsneyers:
Some fixes to the case where the full modular image is skipped:

- don't assume that everything happens at the modular AC group level (`minShift==0`), there can also be upsampling causing channels to have nonzero shift even when there's no Squeeze
- for partial decodes (when `zerofill` is true), don't try to fill the full image when it's not used. Instead initialize the decoded image with zeroes and skip the decoding.

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -274,44 +274,36 @@ Status ModularFrameDecoder::DecodeGroup(const Rect& rect, BitReader* reader,
     Rect r(rect.x0() >> fc.hshift, rect.y0() >> fc.vshift,
            rect.xsize() >> fc.hshift, rect.ysize() >> fc.vshift, fc.w, fc.h);
     if (r.xsize() == 0 || r.ysize() == 0) continue;
-    Channel gc(r.xsize(), r.ysize());
-    gc.hshift = fc.hshift;
-    gc.vshift = fc.vshift;
-    gi.channel.emplace_back(std::move(gc));
-  }
-  if (zerofill) {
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_modular.cc
```diff
@@ -274,44 +274,35 @@ Status ModularFrameDecoder::DecodeGroup(const Rect& rect, BitReader* reader,
...
   }
+  if (zerofill && use_full_image) return true;
   ModularOptions options;
-  if (!ModularGenericDecompress(
-          reader, gi, /*header=*/nullptr, stream.ID(frame_dim), &options,
-          /*undo_transforms=*/-1, &tree, &code, &context_map))
+  if (!zerofill && !ModularGenericDecompress(reader, gi, /*header=*/nullptr,
+                                             stream.ID(frame_dim), &options,
+                                             /*undo_transforms=*/-1, &tree,
+                                             &code, &context_map))
```
veluca93, line 295: Braces pls

jonsneyers: Updated. Also changed the `&&` to two ifs, to make it a bit clearer that short-circuit evaluation is important.

veluca93 (approved):

deymo:
@jonsneyers  please put the PR description in the commit message, you can do so manually when doing "Squash and merge" (which also tags the commit with the PR number).

deymo:
The code this is fixing was introduced in #205 which is not in v0.5.x branch, removing label. (`use_full_image` is not present in v0.5.x)

PR #369, opened 2021-07-27
Fuzz: Check for missing channels in ModularImageToDecodedRect

deymo:
gi.channels may not contain all the channels we expect so we need to
check for them before accessing the array.

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -462,7 +462,7 @@ Status ModularFrameDecoder::ModularImageToDecodedRect(
   }
   JXL_DASSERT(rect.IsInside(decoded));
 
-  int c = 0;
+  size_t c = 0;
   if (do_color) {
     const bool rgb_from_gray =
         metadata->m.color_encoding.IsGray() &&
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_modular.cc
```diff
@@ -473,14 +473,19 @@ Status ModularFrameDecoder::ModularImageToDecodedRect(
...
+      size_t c_in = c;
       if (frame_header.color_transform == ColorTransform::kXYB) {
         factor = dec_state->shared->matrices.DCQuants()[c];
         // XYB is encoded as YX(B-Y)
         if (c < 2) c_in = 1 - c;
       } else if (rgb_from_gray) {
         c_in = 0;
       }
+      if (c_in >= gi.channel.size()) {
+        return JXL_FAILURE(
```
jonsneyers, line 485: It's of course good to make sure that it's in bounds, but are there bitstreams where this can happen? It would be good to understand how it can happen, so we can be sure it cannot happen for valid bitstreams, and also so maybe we can fail earlier on invalid ones.

deymo: I think this was happening because of the bug fixed in #370

jonsneyers (approved):

lib/jxl/dec_modular.cc
jonsneyers: OK.
Maybe turn it into an assert instead? If this happens, I think it's an implementation bug, and just failing will not allow it to be caught by fuzzing...

deymo: Sorry, #370 is unrelated. This still fails even with #370. Do you think that there is some other bug that would cause this?

jonsneyers: I wouldn't expect things to fail here; maybe there's still a problem with the "try to skip the full modular image" codepath. Do you have a repro case?

deymo:
Replaced by #379

PR #370, opened 2021-07-27
Fuzz: Update the nb_meta_channels in squeeze when applying.

deymo:
MetaSqueeze() started updating the nb_meta_channels value in #331, but
we also need to update the value back when applying the transform. This
was leading to the transforms getting out of sync with when
meta-applying versus inverse applying them later.


jonsneyers (approved):

jonsneyers:
Good catch!

PR #371, opened 2021-07-27
Fuzz: Return error if Gaborish weights lead to invalid kernel.

deymo:
Spec says that the coefficients of the kernel used in Gaborish are
normalized to sum 1, but if the sum is already near 0 this is undefined.

This patch catches that case and returns a JXL_FAILURE instead of
creating NaNs or infs in the output.

### DIFF
--- a/lib/jxl/adaptive_reconstruction_test.cc
+++ b/lib/jxl/adaptive_reconstruction_test.cc
@@ -134,10 +134,10 @@ void EnsureUnchanged(const float background, const float foreground,
   jxl::PassesDecoderState state;
   JXL_CHECK(
       jxl::InitializePassesSharedState(frame_header, &state.shared_storage));
-  state.Init();
+  JXL_CHECK(state.Init());
   state.InitForAC(/*pool=*/nullptr);
 
-  state.filter_weights.Init(lf, frame_dim);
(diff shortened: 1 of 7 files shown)


deymo:
Should we restrict gab weights to be >= 0 when parsing them instead?

veluca93 (approved):

veluca93 (approved):

jonsneyers:
> Should we restrict gab weights to be >= 0 when parsing them instead?

No, imo we should allow negative gab weights. It could be useful if we want to make a `jxltran` tool that lets you sharpen or blur images in a cheap and reversible way by just adjusting the gab weights: in principle you could do an unsharp mask using negative gab weights.

jyrkialakuijala:
Ouch. Let's make the divisor 1e-6 away from zero if it that close to zero. Let's preserve its sign and let's keep allowing negative values -- they may turn out useful for ML approaches, online sharpness 'enhancement', and generative arts.

deymo:
Added `cmath` to make msan builder happy.

@jyrkialakuijala current PR code returns failure if `abs(divisior) < 1e-6`, is that you mean?

veluca93 (approved):

PR #373, opened 2021-07-27
fix modular nb_meta_channels counter

jonsneyers:
Fuzzing hasn't found a problem case yet afaik, but in the palette transform there was a similar problem as in https://github.com/libjxl/libjxl/pull/370 : the nb_meta_channels counter was not properly decremented in the inverse transform.

Also now checking for doing transforms on an invalid mix of meta/nonmeta channels in CheckEqualChannels(), so it will also cause RCT to fail when tried to be applied to a mix of meta and nonmeta channels.

### DIFF
--- a/lib/jxl/modular/transform/palette.h
+++ b/lib/jxl/modular/transform/palette.h
@@ -271,7 +271,13 @@ static Status InvPalette(Image &input, uint32_t begin_c, uint32_t nb_colors,
           "UndoDeltaPaletteNoWP");
     }
   }
-  input.nb_meta_channels--;
+  if (c0 >= input.nb_meta_channels) {
+    input.nb_meta_channels--;
+  } else {
+    JXL_ASSERT(c0 + nb - 1 < input.nb_meta_channels);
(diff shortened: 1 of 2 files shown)


lib/jxl/modular/transform/palette.h
```diff
@@ -271,7 +271,13 @@ static Status InvPalette(Image &input, uint32_t begin_c, uint32_t nb_colors,
           "UndoDeltaPaletteNoWP");
     }
   }
-  input.nb_meta_channels--;
+  if (c0 >= input.nb_meta_channels) {
+    input.nb_meta_channels--;
+  } else {
+    JXL_ASSERT(c0 + nb - 1 < input.nb_meta_channels);
+    JXL_ASSERT((int)input.nb_meta_channels >= 2 - nb);
```
deymo, line 278: nit: we normally use C++ casting: `static_cast<int>(input.nb_meta_channels)`

deymo (dismissed):
Please include the PR description in the commit message (you can do so when squashing the change).

deymo (approved):

PR #374, opened 2021-07-27
Remove usage of %n.

deymo:
%n is error prone and some environments start to complain about its
usage. Remove %n usage and add a check to avoid re-introducing it.

Tested that `cjxl third_party/sjpeg/tests/testdata/test_exif_xmp.png`
still parses the metadata correctly.

### DIFF
--- a/bash_test.sh
+++ b/bash_test.sh
@@ -208,6 +208,21 @@ EOF
   return $ret
 }
 
+# Test that we don't use %n in C++ code to avoid using it in printf and scanf.
+# This test is not very precise but in cases where "module n" is needed we would
+# normally have "% n" instead of "%n". Using %n is not allowed in Android 10+.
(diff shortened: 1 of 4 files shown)


veluca93 (approved):

bash_test.sh
```diff
@@ -208,6 +208,21 @@ EOF
   return $ret
 }
 
+# Test that we don't use %n in C++ code to avoid using it in printf and scanf.
+# This test is not very precise but in cases where "module n" is needed we would
```
lvandeve, line 212: Perhaps write why %n should be avoided (security? platform compatibility?) for future reference

deymo: Added a mention that it is not allowed in Android anymore.

lvandeve (approved):

PR #375, opened 2021-07-27
Fix division through zero issue in AdaptToXYZD50

lvandeve:
The division through zero was not the division through wy, but a value
that happened to be 0 after the bradford matrix multiplication.
Requiring the values to be in range 0-1, similar to skcms, should fix
this.
Also did this in the similar PrimariesToXYZD50.
Added TODO for also checking the r,g,b primaries like this: the ACES
P0 color space supports out of range values here.

### DIFF
--- a/lib/jxl/color_encoding_internal.cc
+++ b/lib/jxl/color_encoding_internal.cc
@@ -718,7 +718,13 @@ static const float kBradfordInv[9] = {
 
 // Adapts whitepoint x, y to D50
 Status AdaptToXYZD50(float wx, float wy, float matrix[9]) {
+  if (wx < 0 || wx > 1 || wy <= 0 || wy > 1) {
+    // Out of range values can cause division through zero
+    // further down with the bradford adaptation too.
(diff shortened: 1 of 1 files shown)


lib/jxl/color_encoding_internal.cc
```diff
@@ -718,8 +718,12 @@ static const float kBradfordInv[9] = {
 
 // Adapts whitepoint x, y to D50
 Status AdaptToXYZD50(float wx, float wy, float matrix[9]) {
+  if (wx < 0 || wx > 1 || wy <= 0 || wy > 1) {
```
deymo, line 721: what about float numbers that are veeery close to 0?

lvandeve: I tested 1 / 5e-324 (smallest float) and indeed that gives infinity, added back the std::finite checks

deymo (approved):

deymo:
tests were stuck for some reason. I cancelled them and retry

PR #377, opened 2021-07-28
ix unbalanced nb_metachannel changes.

eustas:
Otherwise DefaultTransforms on inverse pass generates different list than on meta pass.

### DIFF
--- a/lib/jxl/modular/transform/palette.h
+++ b/lib/jxl/modular/transform/palette.h
@@ -122,6 +122,7 @@ static Status InvPalette(Image &input, uint32_t begin_c, uint32_t nb_colors,
   size_t w = input.channel[c0].w;
   size_t h = input.channel[c0].h;
   if (nb < 1) return JXL_FAILURE("Corrupted transforms");
+  input.nb_meta_channels += nb - 1;
   for (int i = 1; i < nb; i++) {
     input.channel.insert(
         input.channel.begin() + c0 + 1,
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/transform/palette.h
```diff
@@ -281,18 +282,20 @@ static Status MetaPalette(Image &input, uint32_t begin_c, uint32_t end_c,
   JXL_RETURN_IF_ERROR(CheckEqualChannels(input, begin_c, end_c));
 
   size_t nb = end_c - begin_c + 1;
+  if (nb < 1) return JXL_FAILURE("Corrupted transforms");
   if (begin_c >= input.nb_meta_channels) {
-    input.nb_meta_channels++;
+    JXL_DASSERT(nb == 1);
```
jonsneyers, line 287: This assert will fail for anything that is not single-channel palette.

lib/jxl/modular/transform/palette.h
```diff
@@ -281,18 +282,20 @@ static Status MetaPalette(Image &input, uint32_t begin_c, uint32_t end_c,
   JXL_RETURN_IF_ERROR(CheckEqualChannels(input, begin_c, end_c));
 
   size_t nb = end_c - begin_c + 1;
+  if (nb < 1) return JXL_FAILURE("Corrupted transforms");
   if (begin_c >= input.nb_meta_channels) {
-    input.nb_meta_channels++;
```
jonsneyers, line 285: Why remove this line? Palette does add a metachannel.

jonsneyers: Nevermind, you moved it below.

jonsneyers:
I think this is not needed: https://github.com/libjxl/libjxl/pull/373 does the same thing.

PR #379, opened 2021-07-28
fuzz: fix assert in the exotic case of doing palette on palettes

jonsneyers:
Made a braino in the original assert here: the assert talks about the situation _after_ the inverse transform (checking that it is indeed a case where all affected channels were metachannels), so it has to use the updated `nb_meta_channels` and `begin_c` instead of `c0`.

### DIFF
--- a/lib/jxl/modular/transform/palette.h
+++ b/lib/jxl/modular/transform/palette.h
@@ -272,11 +272,13 @@ static Status InvPalette(Image &input, uint32_t begin_c, uint32_t nb_colors,
     }
   }
   if (c0 >= input.nb_meta_channels) {
+    // Palette was done on normal channels
     input.nb_meta_channels--;
   } else {
-    JXL_ASSERT(c0 + nb - 1 < input.nb_meta_channels);
+    // Palette was done on metachannels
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/transform/palette.h
```diff
@@ -274,9 +274,9 @@ static Status InvPalette(Image &input, uint32_t begin_c, uint32_t nb_colors,
   if (c0 >= input.nb_meta_channels) {
     input.nb_meta_channels--;
   } else {
-    JXL_ASSERT(c0 + nb - 1 < input.nb_meta_channels);
     JXL_ASSERT(static_cast<int>(input.nb_meta_channels) >= 2 - nb);
```
deymo, line 277: I'd appreciate if you can add a comment here of what's going on in this case. Like, why the `if` at all and what's the intention of the assertion. A reviewer should be able to understand the code and reason about whether it is correct.

jonsneyers: Added comments here and in the similar MetaPalette code to explain the two cases.

deymo (approved):
thanks!

PR #380, opened 2021-07-28
fuzz: return early when there is no modular data to decode

jonsneyers:
When a modular dc or ac group is empty, a 0-channel image gets decoded, and in the old flow that just meant that all of those 0 channels would get copied to the modular full_image.

In the new flow however (where full_image is avoided), it created trouble because a 0-channel image would still cause a call to ModularImageToDecodedRect(), which would then try to access non-existing channels.

Returning early when a 0-channel group image is about to get decoded solves that.
Still adding an assert to check for access to non-existing channels.

~~There could still be a remaining bug that can arise when a modular group e.g. contains data for some extrachannels but not for others (because they have a different dim_shift / upsampling factor). I haven't seen a fuzzcase for that yet, but it is something that can happen so it needs to be fixed (not in this pull request though).~~

Also don't do the no-full-image thing when not all modular channels have the same shift. That avoids the potential problem where some channels are encoded in the DC groups and others in the AC groups, which would cause trouble since ModularImageToDecodedRect() assumes all channels are available. This is a temporary fix; a better but more complicated fix would be to update ModularImageToDecodedRect() to work on the subset of channels that was just decoded.

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -185,6 +185,7 @@ Status ModularFrameDecoder::DecodeGlobalInfo(BitReader* reader,
   Image gi(frame_dim.xsize, frame_dim.ysize, metadata.bit_depth.bits_per_sample,
            nb_chans + nb_extra);
 
+  all_same_shift = true;
   if (frame_header.color_transform == ColorTransform::kYCbCr) {
     for (size_t c = 0; c < nb_chans; c++) {
       gi.channel[c].hshift = frame_header.chroma_subsampling.HShift(c);
@@ -194,6 +195,9 @@ Status ModularFrameDecoder::DecodeGlobalInfo(BitReader* reader,
       size_t ysize_shifted =
           DivCeil(frame_dim.ysize, 1 << gi.channel[c].vshift);
       gi.channel[c].shrink(xsize_shifted, ysize_shifted);
+      if (gi.channel[c].hshift != gi.channel[0].hshift ||
+          gi.channel[c].vshift != gi.channel[0].vshift)
+        all_same_shift = false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/jxl/dec_modular.cc
```diff
@@ -474,6 +475,7 @@ Status ModularFrameDecoder::ModularImageToDecodedRect(
       } else if (rgb_from_gray) {
         c_in = 0;
       }
+      JXL_ASSERT(c_in < gi.channel.size());
```
deymo, line 478: please add a similar check for `c < gi.channel.size()` near line 554

lib/jxl/dec_modular.cc
```diff
@@ -288,6 +288,7 @@ Status ModularFrameDecoder::DecodeGroup(const Rect& rect, BitReader* reader,
     }
   }
   if (zerofill && use_full_image) return true;
+  if (gi.channel.empty()) return true;
```
deymo, line 291: please add a comment explaining why is this check here. The idea is to provide more context to the next person who needs to debug an issue in this code.

jonsneyers: Added a comment

lib/jxl/dec_modular.cc
jonsneyers: Done

jonsneyers:
Changed things a bit, see updated description and code.

deymo (approved):
I would have preferred to split out the all_same_shift change to another PR, even if its related, it is another fix for another related problem.

deymo:
The code this is fixing was introduced in #205  which is not in v0.5.x branch, removing label.

@jonsneyers Do we need the `all_same_shift` fix in v0.5.x?

jonsneyers:
> The code this is fixing was introduced in #205 which is not in v0.5.x branch, removing label.
> 
> @jonsneyers Do we need the `all_same_shift` fix in v0.5.x?

No, that fix is only needed for the new !use_full_image case.

PR #387, opened 2021-07-31
Remove unnecessary submodules.

veluca93:
The projects are now cloned when running prepare_metrics.sh, as they
were only used for running the metrics.

### DIFF
--- a/tools/benchmark/metrics/prepare_metrics.sh
+++ b/tools/benchmark/metrics/prepare_metrics.sh
@@ -12,6 +12,18 @@ MYDIR=$(dirname $(realpath "$0"))
 main() {
   cd "${MYDIR}/../../../third_party"
   local zipurl
+  local repourl
+  for repourl in \
+    'https://github.com/veluca93/IQA-optimization.git'
+    'https://github.com/Netflix/vmaf.git'
+    'https://github.com/thorfdbg/difftest_ng.git'
+  do
+    local reponame=$(basename "${repourl}")
+    local dirname=$(basename "${reponame}")
+    if [[ ! -e "${dirname}" ]]; then
+      git clone "${repourl}"
+    fi
+  done
   for zipurl in \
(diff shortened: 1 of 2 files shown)


tools/benchmark/metrics/prepare_metrics.sh
```diff
@@ -12,6 +12,18 @@ MYDIR=$(dirname $(realpath "$0"))
 main() {
   cd "${MYDIR}/../../../third_party"
   local zipurl
+  local repourl
+  for repourl in \
+    'https://github.com/veluca93/IQA-optimization.git'
+    'https://github.com/Netflix/vmaf.git'
+	  'https://github.com/thorfdbg/difftest_ng.git'
```
sboukortt, line 19: The indentation for this line seems off.

tools/benchmark/metrics/prepare_metrics.sh
```diff
@@ -12,6 +12,18 @@ MYDIR=$(dirname $(realpath "$0"))
...
+  do
+    local reponame=$(basename "${repourl}")
+    local dirname=$(basename "${reponame}")
+    if [[ ! -e "${dirname}" ]]; then
+      git clone "${repourl}"
```
sboukortt, line 24: Should we perhaps check out a specific revision, and clone it shallowly if applicable?

sboukortt (approved):

tools/benchmark/metrics/prepare_metrics.sh
veluca93: Something put a tab in there... oh well.

tools/benchmark/metrics/prepare_metrics.sh
veluca93: Nah, too much effort...

deymo: so lazy!

deymo (approved):

PR #393, opened 2021-08-02
Allow blending to work when the output is the foreground

sboukortt:
When copying the background to the output in PrepareBlending, we now leave a hole where the foreground will be, in case it is in fact already there.

### DIFF
--- a/lib/jxl/blending.cc
+++ b/lib/jxl/blending.cc
@@ -10,6 +10,66 @@
 
 namespace jxl {
 
+namespace {
+
+// Given two rects A and B, returns a set of rects whose union is A \ B. This
+// may require from 0 to 4 rects, one for each non-empty side of B. `storage`
+// must have room to accommodate that many rects. The order is consistent when
+// called successively with "parallel" rects.
+//
+// +----------------------+
+// |         top          |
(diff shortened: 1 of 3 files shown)


lib/jxl/blending.cc
```diff
@@ -10,6 +10,60 @@
...
+// consistent when called successively with "parallel" rects.
+//
+// +----------------------+
+// |         top          |
+// +------+-------+-------+
+// | left | inner | right |
+// +------+-------+-------+
+// |        bottom        |
+// +----------------------+
+std::vector<Rect> SubtractRect(const Rect& outer, const Rect& inner) {
```
veluca93, line 26: Maybe I'm just over-optimizing, but how about filling an output array and returning the number?

sboukortt: I was thinking of that too but wasn’t sure whether it would be worth it. Let me give it a try.

sboukortt: There we go.

sboukortt: (And just removed the leftover `std::vector<Rect> result;`.)

veluca93 (approved):

PR #406, opened 2021-08-03
Build release binaries on post-submit

deymo:
Release binaries include the Debian packages for Ubuntu 18.04 and 20.04
and static linux binaries for x86_64.

The binaries are built from main and the release branches whenever a
new commit is pushed there, and the artifacts are automatically attached
to releases in GitHub when a release is created.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,169 @@
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Workflow for building the release binaries.
+#
+# This workflow runs as a post-submit step, when pushing to main or the release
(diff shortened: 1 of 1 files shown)


deymo:
Tested this with a release in my own fork: https://github.com/deymo/libjxl/releases/tag/test3

veluca93 (approved):

veluca93:
Is it possible to also upload nightly builds? (possibly in future versions)

jo620kix:
Great!

In addition to binaries on GitHub, I believe an official PPA repository will be useful, so that Ubuntu users will not have to visit GitHub and download manually on new releases, and instead they can use `apt` to install new libjxl releases automatically.

deymo:
> Is it possible to also upload nightly builds? (possibly in future versions)

It should be doing that right now. The "Upload artifacts" step will make those available in the action on push. There's a 90-day retention policy on those.

PR #415, opened 2021-08-05
Set JPEGXL_VERSION macro from git hash at build time.

deymo:
This sets the JPEGXL_VERSION macro used in the tools from the short git
hash at build time, updating it only when needed to avoid rebuilding
every time. Unfortunately, ninja checks for the header timestamp before
running the target that generates it so when only updating the hash, the
tool_version.cc.o file is not regenerated. This doesn't affect make, or
when running `ci.sh release` or similar commands.

GitHub/GitLab builders use the hash of the built reference, so this code
doesn't run in the workflow/pipelines anyway.

Tested with:
  - `./ci.sh release`
  - `./ci.sh release -DJPEGXL_VERSION="foo bar"`
  - `mkdir build; cd build; cmake ..`

Test rebuilt after amending a commit.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -56,11 +56,6 @@ else()
 endif()
 endif()  # JPEGXL_ENABLE_VIEWERS
 
-# The JPEGXL_VERSION is set from the builders.
-if(NOT DEFINED JPEGXL_VERSION OR JPEGXL_VERSION STREQUAL "")
-  set(JPEGXL_VERSION "(unknown)")
-endif()
-
 set(TOOL_BINARIES
   cjxl
   djxl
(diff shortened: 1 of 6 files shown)


tools/CMakeLists.txt
```diff
@@ -69,36 +64,80 @@ set(TOOL_BINARIES
...
+  add_dependencies(jxl_tool tool_version_git)
+
+  set_source_files_properties(tool_version.cc PROPERTIES
+    COMPILE_DEFINITIONS JPEGXL_VERSION_FROM_GIT=1
+    INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}"
+  )
+  # Note: Ninja looks for dependencies on the jxl_tool target before running
+  # the the tool_version_git targets, so when updating the tool_version_git.h
+  # the jxl_tool target is not rebuilt. This forces to generate it at configure
+  # time if needed.
```
sboukortt, line 115:
```suggestion
  # the tool_version_git targets, so when updating the tool_version_git.h the
  # jxl_tool target is not rebuilt. This forces to generate it at configure time
  # if needed.
```

sboukortt (dismissed):

sboukortt (approved):

PR #431, opened 2021-08-11
Remove ConvertImageInternal HasRGBBuffer assert

lvandeve:
ConvertImageInternal Used to have a much larger implementation and
be specific to the RGB output case, the assert then made sense.

Now ConvertImageInternal is a small wrapper around ConvertImageExternal
and is used for much more than just RGB: also preview image and extra channels.

The assert was checking that when calling this function, the slow code
path of the decoder (without internal RGB buffer) was being taken. However
now that this function is also used for preview image and extra channels,
such check does not belong here.

The cases where this function is used in the RGB context already have
a clear `HasRGBBuffer` if condition, so the assert does not need to
be moved there, it can just be removed.

This fixes a fuzzer-found DASSERT issue.


veluca93 (approved):

deymo (approved):
thanks for the explanation!

PR #438, opened 2021-08-12
Stamp .deb packages with the commit date.

deymo:
To provide .deb packages with increasing versions from main, this patch
updates the version on the package being built to include a suffix
with the date.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -149,6 +149,27 @@ jobs:
         submodules: true
         fetch-depth: 1
 
+    - name: Stamp non-release versions
+      # Stamps the built package with the commit date as part of the version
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

andrius4669:
maybe consider using `git rev-list --count HEAD` too? I guess two commits at the same second wouldn't be likely, but who knows..

For reference, that's what archlinux recommends for `-git` AUR pkgbuilds: https://wiki.archlinux.org/title/VCS_package_guidelines#Git

deymo:
I think it is very unlikely to be an issue. The date has a more visible meaning than the number of commits, and the string is already quite long. I was actually thinking to move this to a nightly build with a schedule instead of building on every commit.

PR #440, opened 2021-08-12
Add tooling to update the version on main and bump it to 0.6.

deymo:
./ci.sh bump_version will automatically mark the version as released in
the Debian package and create a new entry with the UNRELEASED label.
This now marks the .deb packages as released when published in a GitHub
release, although we should probably mark them as such in the source
code before tagging a release.


debian/changelog
```diff
@@ -1,8 +1,14 @@
...
+ -- JPEG XL Maintainers <jpegxl@google.com>  Thu, 12 Aug 2021 23:49:40 +0200
+
+jpeg-xl (0.5.0) unstable; urgency=medium
 
   * Bump JPEG XL version to 0.5.0.
 
- -- Alex Deymo <deymo@google.com>  Thu, 15 Jul 2021 14:27:29 +0200
+ -- JPEG XL Maintainers <jpegxl@google.com>  Thu, 12 Aug 2021 23:49:40 +0200
 
 jpeg-xl (0.3.7) UNRELEASED; urgency=medium
```
veluca93, line 13: Is this line correct?

veluca93 (approved):

debian/changelog
deymo: when you mark at as released it updates the date. I think it is fine or it doesn't matter. I feel like having the debian/ files here doesn't work together with our release (branching) process, but meanwhile having it here is convenient.

PR #441, opened 2021-08-13
CMake setting to bundle libskcms.a into libjxl.a

### LINKED ISSUE #428: libjxl.a depends on missing skcms static library
The static libjxl.a we install depends on skcms.a (when using skcms which is the default). In the libjxl.so case we bundle the skcms.a into the shared library, but in the static case the skcms.a is missing. Unlike with hwy, brotli, etc, the skcms dependency is not listed in the pkg-config .pc file since I'm not aware of any package that would provide it. This prevents external projects from using the static jxl library.

We need to either bundle skcms.a (and somehow avoid collisions with other projects using it), switch to lcms (which has its own set of problems) or provide a dependency injection hook for having arbitrary color conversions in the encoder.

deymo:
skcms is only a single .cc file with a C API, but not widely available
as a standalone package or library. We currently require to use a CMS in
the encoder side of the library to convert from arbitrary color spaces
when encoding, among probably other things. The user can select from
either skcms (the default) or lcms2 for the CMS library. In the skcms
case libjxl.so was bundling the libskcms.a symbols (since skcms is built
as a static library) but libjxl.a wasn't, resulting in an unusable
static library being packaged and installed.

This patch adds a default-on setting to bundle skcms.cc into libjxl.a
when using skcms, however, we rename all the function from `skcms_` to
`jxl_skcms_` to avoid issues with potential projects that would
include skcms from elsewhere. Until there's a packaged version of skcms
or we provide an external way to provide a CMS on the encoder side this
at least makes libjxl.a link a sample program.

Fixes #428

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,8 @@ set(JPEGXL_ENABLE_OPENEXR true CACHE BOOL
     "Build JPEGXL with support for OpenEXR if available.")
 set(JPEGXL_ENABLE_SKCMS true CACHE BOOL
     "Build with skcms instead of lcms2.")
+set(JPEGXL_BUNDLE_SKCMS true CACHE BOOL
+    "When building with skcms, bundle it into libjxl.a.")
 set(JPEGXL_ENABLE_VIEWERS false CACHE BOOL
     "Build JPEGXL viewer tools for evaluation.")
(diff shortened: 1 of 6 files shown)


deymo:
What do you think about this horrible hack?

sboukortt:
> What do you think about this horrible hack?

It’s pragmatic 😄 thanks!

sboukortt (approved):

PR #444, opened 2021-08-13
Make skcms-obj an OBJECT library

deymo:
When bundling skcms into libjxl.a we want to include the object files
with `$<TARGET_OBJECTS:lib>` which is only supported for OBJECT
libraries in older versions of cmake. This patch makes a new skcms-obj
library similar to what we do with jxl_dec-obj and jxl_enc-obj.
This fixes the build for debian:buster.

In MSVC, the flag to include a header is /FI, this patch splits that
case which fixes the MSVC windows build.


veluca93 (approved):

deymo:
added the fix for MSVC build. I'm testing that now

sboukortt (approved):

deymo:
windows build passed.

PR #446, opened 2021-08-13
update debian packaging

xiota:
* add build-depends: asciidoc, xmlto
   - needed to build man pages
 * add man pages to jxl package
 * add libhwy-dev to depends for libjxl-dev
 * update homepage to https://github.com/libjxl/libjxl

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,20 +3,23 @@ Maintainer: JPEG XL Maintainers <jpegxl@google.com>
 Section: misc
 Priority: optional
 Standards-Version: 3.9.8
-Build-Depends: cmake,
-               debhelper (>= 9),
-               libbrotli-dev,
-               libgif-dev,
-               libgmock-dev,
-               libgoogle-perftools-dev,
-               libgtest-dev,
-               libhwy-dev,
-               libjpeg-dev,
-               libopenexr-dev,
-               libpng-dev,
-               libwebp-dev,
-               pkg-config,
-Homepage: https://gitlab.com/wg1/jpeg-xl
+Build-Depends:
+ asciidoc,
+ cmake,
+ debhelper (>= 9),
+ libbrotli-dev,
+ libgif-dev,
+ libgmock-dev,
+ libgoogle-perftools-dev,
+ libgtest-dev,
+ libhwy-dev,
+ libjpeg-dev,
+ libopenexr-dev,
+ libpng-dev,
+ libwebp-dev,
+ pkg-config,
+ xmlto,
+Homepage: https://github.com/libjxl/libjxl
 Rules-Requires-Root: no
 
 Package: jxl
@@ -36,6 +39,7 @@ Package: libjxl-dev
 Architecture: any
 Section: libdevel
 Depends: libjxl (= ${binary:Version}), ${misc:Depends}
+ libhwy-dev,
 Description: JPEG XL Image Coding System - "JXL" (development files)
  The JPEG XL Image Coding System (ISO/IEC 18181) is a lossy and
  lossless image compression format. It has a rich feature set and is
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,8 +19,9 @@ Google LLC <*@google.com>
 Alexander Sago <cagelight@gmail.com>
 Dirk Lemstra <dirk@lemstra.org>
 Jon Sneyers <jon@cloudinary.com>
-Pieter Wuille
-Marcin Konicki <ahwayakchih@gmail.com>
-Ziemowit Zabawa <ziemek.zabawa@outlook.com>
-Petr Diblík
 Lovell Fuller
+Marcin Konicki <ahwayakchih@gmail.com>
+Petr Diblík
+Pieter Wuille
+xiota
+Ziemowit Zabawa <ziemek.zabawa@outlook.com>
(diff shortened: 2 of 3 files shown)


xiota:
@googlebot I signed it!

debian/control
```diff
@@ -36,6 +40,7 @@ Package: libjxl-dev
 Architecture: any
 Section: libdevel
 Depends: libjxl (= ${binary:Version}), ${misc:Depends}
+ libhwy-dev,
```
deymo, line 43: I don't think this is correct. The public API in libjxl-dev doesn't require any hwy header. Is there something in libjxl-dev that's currently depending on libhwy-dev?

debian/control
```diff
@@ -4,19 +4,23 @@ Section: misc
...
+ libgmock-dev,
+ libgoogle-perftools-dev,
+ libgtest-dev,
+ libhwy-dev,
+ libjpeg-dev,
+ libopenexr-dev,
+ libpng-dev,
+ libwebp-dev,
+ pkg-config,
+ doxygen,
```
deymo, line 19: I don't think doxygen is needed to build the debian package. If I remember correctly we don't install the doxygen docs.

debian/control
```diff
@@ -4,19 +4,23 @@ Section: misc
...
+ asciidoc,
+ xmlto,
```
deymo, line 21: What are `xmlto` and `xdg-utils` for?

debian/control
```diff
@@ -4,19 +4,23 @@ Section: misc
...
-               libgoogle-perftools-dev,
-               libgtest-dev,
-               libhwy-dev,
-               libjpeg-dev,
-               libopenexr-dev,
-               libpng-dev,
-               libwebp-dev,
-               pkg-config,
-Homepage: https://gitlab.com/wg1/jpeg-xl
+ debhelper (>= 9),
```
deymo, line 7: lets keep the list sorted. Was there a problem with the indentation?

debian/control
deymo: hmmm.. I guess libjxl.a (installed by libjxl-dev) needs libhwy.a (installed by libhwy-dev)

debian/control
xiota: @deymo

* libhwy-dev – Everything I try to build against libjxl-dev fails when libhwy-dev is not also installed.  Nothing else currently  uses libhwy, except libjxl.  Looks like you found the specific files involved.
* doxygen, xmlto  – When trying to build the man pages on launchpad, I got various errors because of missing files that are provided these packages.  I think they were related to `lintian` checking `xml` output.  I would have to look for the logs if you want more details.
* xdg-utils – This is needed to provide `xdg-mime` referenced in `plugins/gdk-pixbuf/CMakeLists.txt`
* spacing – It was just annoying to have to type so many spaces every time I added a package to the depends list.  Most other packages seem to use just one space.
* alphabetization – I just added the packages to the end of the list in the order that I encountered and fixed build errors.  I will revise the commit to alphabetize.

debian/control
xiota: xmlto provides some files required by lintian

xdg-utils provides xdg-mime, which is referenced in plugins/gdk-pixbuf/CMakeLists.txt

debian/control
xiota: When trying to build the man pages on launchpad, I got the following error: `-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)`

debian/control
xiota: Everything I try to build against libjxl-dev fails when libhwy-dev is not also installed.  Nothing else currently  uses libhwy, except libjxl.  Looks like you found the specific files involved.

deymo: Could you post the error? Unless you are linking statically it shouldn't fail.

xiota: autoconf just didn't detect libjxl as being installed unless libhwy was also installed.

xiota: When I try to build the libjxl/examples folder without libhwy-dev installed, this is what cmake reports:
```
-- Checking for module 'libjxl'
--   Package 'libhwy', required by 'libjxl', not found
CMake Error at /usr/share/cmake-3.18/Modules/FindPkgConfig.cmake:545 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPkgConfig.cmake:733 (_pkg_check_modules_internal)
  CMakeLists.txt:14 (pkg_check_modules)
```

debian/control
deymo: It is not a fatal error, it just won't build the documentation pages which are not installed in debian anyway. The code does:
```
find_package(Doxygen)
if(DOXYGEN_FOUND)
...
```

We could make this QUIET or put a flag to disabled it to silence the warning in the debian build.

debian/control
deymo: `JPEGXL_ENABLE_PLUGINS` defaults to false and we don't enable it in the debian/rules, so the plugins/gdk-pixbuf shouldn't even be parsed. Eventually it would be good to build the pluings to a new .deb package but that's a bit out of scope here and would likely require more work.

xiota: I'll update the commit to remove xdg-utils.

debian/control
xiota: I'll rebuild after uninstalling doxygen to see if the man pages build.  If they do, I'll update the commit to remove doxygen.

xiota:
@deymo I've updated the commit to remove doxygen and xdg-utils

deymo (approved):
Thanks for your patch! :-D

xiota:
@deymo Thanks for the code review and help.

PR #451, opened 2021-08-16
Fix manpage generation on debian:10 / ubuntu:18.04

deymo:
asciidoc in debian:10 and ubuntu:18.04 uses python2 as an interpreter.
This patch detects the python version requested from the shebang in a2x
and looks for that Python interpreter version.


veluca93 (approved):

deymo:
@xiota FYI

PR #452, opened 2021-08-17
fix pkg-config files

landfillbaby:
apparently some programs expect `libdir` to be a fully qualified path


landfillbaby:
@googlebot I signed it!

landfillbaby:
wait this isn't why it's failing in imagemagick static. it's the `-R` stuff in brotli!

Issue #100, opened 2021-06-04
GIMP plugin: Add encode options dialog


petrdiblik:
It would be great if the plugin showed a dialog with encode options before actual saving.
The options could be: Distance, Quality, Effort, Progressive (which corresponds with `cjxl -h`).
A brief explanation of the options could be there as well.

(The plugin can't easily decide itself whether it should use lossy/lossless in the same way as cjxl because the input are pixels, not image formats.)

Issue #417, opened 2021-08-06
GIMP file-jxl plugin: image imported incorrectly


novomesk:
Hello,

when I open [jpegxl-logo.jxl](https://github.com/novomesk/qt-jpegxl-image-plugin/blob/main/testfiles/jpegxl-logo.jxl) image in GIMP 2.10.24, the image looks too transparent:
![gimp1](https://user-images.githubusercontent.com/52529860/128533673-8a397be2-b7b3-49e8-839d-2d79556d7136.png)

When converted to png via djxl, it looks this way:
![gimp2](https://user-images.githubusercontent.com/52529860/128533788-3ed8ed8d-045f-4c78-aae6-8e949e36664f.png)

novomesk:
Possibly related issue: https://gitlab.com/wg1/jpeg-xl/-/issues/234

jonsneyers:
Supposedly fixed, please reopen if needed.

rubyFeedback:
Is there an educated guess when gimp will allow us to use jxl? Right now it all seems to be "it is beta", almost since 2 years by now at this point.

novomesk:
You can use GIMP 2.99.10 today where JXL works out of box in Windows build or in Linux flatpak.
It is called Beta but it works.

Or you can use the old GIMP 2.10.x and manually install file-jxl plug-in afterwards.

Issue #229, opened 2021-06-26
GIMP plugin fails to open JXLs which were transcoded from JPEG


cagelight:
Arch Linux
GCC 11.1.0
JPEG XL encoder v0.3.7 [AVX2,SSE4,Scalar]

I tried to load an assortment of JXL images and there seems to be a bug specifically with JXLs that were transcoded from JPEG, it only seems to happen to the GIMP plugin specifically.

The issue does not occur with lossless JXL, or JXLs encoded with a -q < 100, only JPEG -> JXL

I get a GIMP error: "Opening '...' failed: JPEG XL image plug-in could not open image"
and then the following output in the console:
```
/home/.cache/yay/libjxl-git/src/libjxl/lib/jxl/image_metadata.cc:354: JXL_FAILURE: invalid min 0.001234 vs max 0.001001
/home/.cache/yay/libjxl-git/src/libjxl/lib/jxl/fields.cc:697: JXL_RETURN_IF_ERROR code=1: visitor.Visit(fields, PrintVisitors() ? "-- Read\n" : "")
/home/.cache/yay/libjxl-git/src/libjxl/lib/jxl/dec_file.cc:33: JXL_RETURN_IF_ERROR code=1: ReadImageMetadata(reader, &io->metadata.m)
/home/.cache/yay/libjxl-git/src/libjxl/lib/jxl/dec_file.cc:108: JXL_RETURN_IF_ERROR code=1: DecodeHeaders(&reader, io)
/home/.cache/yay/libjxl-git/src/libjxl/plugins/gimp/file-jxl-load.cc:95: JXL_RETURN_IF_ERROR code=1: DecodeFile(dparams, compressed, &io, &pool)
```

ToBeFree:
Confirmed. I thought I had compiled this incorrectly.

ToBeFree:
Example image (public domain) that can't be opened by the GIMP plugin:
[publicdomain-green-butterfly-2-avatar-scale100px-lossless.zip](https://github.com/libjxl/libjxl/files/6719948/publicdomain-green-butterfly-2-avatar-scale100px-lossless.zip)

Source image:
![publicdomain-green-butterfly-2-avatar-scale100px](https://user-images.githubusercontent.com/66693882/123510229-99d29100-d67a-11eb-9e0e-5b1417be32b4.jpg)

Converted using cjxl [v0.3.7 | SIMD supported: AVX2,SSE4,Scalar] -q 100 -s 8

[...]

PR #453, opened 2021-08-17
Update GIMP plugin to use public API and add save dialog

xiota:
* Remove libjxl internals
* Use multi-threaded parallel runner
* Allow saving without requiring compiler flag
* Load multiple frames as layers
* Save non-linear uint8/16 and linear float16/32
* Save alpha channel in uint8/16 formats
* Add save options dialog

The following issues are resolved: #100, #229, #417, [wg1#234](https://gitlab.com/wg1/jpeg-xl/-/issues/234), [wg1#184](https://gitlab.com/wg1/jpeg-xl/-/issues/184).

save_dialog

### DIFF
--- a/plugins/gimp/common.cc
+++ b/plugins/gimp/common.cc
@@ -1,0 +1,27 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "plugins/gimp/common.h"
+
+namespace jxl {
+
+JpegXlGimpProgress::JpegXlGimpProgress(const char *message) {
+  cur_progress = 0;
+  max_progress = 100;
+
+  gimp_progress_init_printf("%s\n", message);
+}
+
+void JpegXlGimpProgress::update() {
+  gimp_progress_update((float)++cur_progress / (float)max_progress);
+  return;
+}
+
+void JpegXlGimpProgress::finished() {
+  gimp_progress_update(1.0);
+  return;
+}
+
+}  // namespace jxl
--- a/plugins/gimp/common.h
+++ b/plugins/gimp/common.h
@@ -7,49 +7,38 @@
 #define PLUGINS_GIMP_COMMON_H_
 
 #include <libgimp/gimp.h>
+#include <libgimp/gimpui.h>
+#include <math.h>
+
+#include <fstream>
+#include <iterator>
+#include <string>
+#include <vector>
+
+#define PLUG_IN_BINARY "file-jxl"
+#define SAVE_PROC "file-jxl-save"
+
+// Defined by both FUIF and glib.
+#undef MAX
+#undef MIN
+#undef CLAMP
+
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/resizable_parallel_runner_cxx.h"
 
 namespace jxl {
 
-// FromFloat expects a value between 0 and 1, and ToFloat returns such values
-// from GIMP values.
-template <GimpPrecision>
-struct BufferFormat;
-template <>
-struct BufferFormat<GIMP_PRECISION_U8_GAMMA> {
-  using Sample = uint8_t;
-  static Sample FromFloat(const float x) {
-    return static_cast<Sample>(std::round(x * 255.f));
-  }
-  static float ToFloat(const Sample s) { return s; }
-};
-template <>
-struct BufferFormat<GIMP_PRECISION_U16_GAMMA> {
-  using Sample = uint16_t;
-  static Sample FromFloat(const float x) {
-    return static_cast<Sample>(std::round(x * 65535.f));
-  }
-  static float ToFloat(const Sample s) { return s * (1.f / 65535.f); }
-};
-template <>
-struct BufferFormat<GIMP_PRECISION_U32_GAMMA> {
-  using Sample = uint32_t;
-  static Sample FromFloat(const float x) {
-    return static_cast<Sample>(std::round(x * 4294967295.f));
-  }
-  static float ToFloat(const Sample s) { return s * (1.f / 4294967295.f); }
-};
-template <>
-struct BufferFormat<GIMP_PRECISION_HALF_GAMMA> {
-  using Sample = float;
-  static Sample FromFloat(const float x) { return x; }
-  static float ToFloat(const Sample s) { return s; }
-};
-template <>
-struct BufferFormat<GIMP_PRECISION_FLOAT_GAMMA> {
-  using Sample = float;
-  static Sample FromFloat(const float x) { return x; }
-  static float ToFloat(const Sample s) { return s; }
-};
+class JpegXlGimpProgress {
+ public:
+  explicit JpegXlGimpProgress(const char *message);
+  void update();
+  void finished();
+
+ private:
+  int cur_progress;
+  int max_progress;
+
+};  // class JpegXlGimpProgress
 
 }  // namespace jxl
 
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -5,168 +5,335 @@
 
 #include "plugins/gimp/file-jxl-load.h"
 
-// Defined by both FUIF and glib.
-#undef MAX
-#undef MIN
-#undef CLAMP
-
-#include "lib/jxl/alpha.h"
-#include "lib/jxl/base/file_io.h"
-#include "lib/jxl/base/thread_pool_internal.h"
-#include "lib/jxl/dec_file.h"
-#include "plugins/gimp/common.h"
+#define _PROFILE_TARGET_ JXL_COLOR_PROFILE_TARGET_DATA
 
 namespace jxl {
 
-namespace {
+bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
+  std::vector<uint8_t> icc_profile;
+  GimpColorProfile *profile = nullptr;
+  bool is_linear = false;
 
-template <GimpPrecision precision, bool has_alpha, size_t num_channels>
-void FillBuffer(
-    const CodecInOut& io,
-    std::vector<typename BufferFormat<precision>::Sample>* const pixel_data) {
-  pixel_data->reserve(io.xsize() * io.ysize() * (num_channels + has_alpha));
-  for (size_t y = 0; y < io.ysize(); ++y) {
-    const float* rows[num_channels];
-    for (size_t c = 0; c < num_channels; ++c) {
-      rows[c] = io.Main().color().ConstPlaneRow(c, y);
-    }
-    const float* const alpha_row =
-        has_alpha ? io.Main().alpha().ConstRow(y) : nullptr;
-    for (size_t x = 0; x < io.xsize(); ++x) {
-      const float alpha = has_alpha ? alpha_row[x] : 1.f;
-      const float alpha_multiplier =
-          has_alpha && io.Main().AlphaIsPremultiplied()
-              ? 1.f / std::max(kSmallAlpha, alpha)
-              : 1.f;
-      for (const float* const row : rows) {
-        pixel_data->push_back(BufferFormat<precision>::FromFloat(
-            std::max(0.f, std::min(1.f, alpha_multiplier * row[x]))));
-      }
-      if (has_alpha) {
-        pixel_data->push_back(
-            BufferFormat<precision>::FromFloat(255.f * alpha));
-      }
-    }
-  }
-}
+  gint32 layer;
 
-template <GimpPrecision precision>
-Status FillGimpLayer(const gint32 layer, const CodecInOut& io,
-                     GimpImageType layer_type) {
-  std::vector<typename BufferFormat<precision>::Sample> pixel_data;
-  switch (layer_type) {
-    case GIMP_GRAY_IMAGE:
-      FillBuffer<precision, /*has_alpha=*/false, /*num_channels=*/1>(
-          io, &pixel_data);
-      break;
-    case GIMP_GRAYA_IMAGE:
-      FillBuffer<precision, /*has_alpha=*/true, /*num_channels=*/1>(
-          io, &pixel_data);
-      break;
-    case GIMP_RGB_IMAGE:
-      FillBuffer<precision, /*has_alpha=*/false, /*num_channels=*/3>(
-          io, &pixel_data);
-      break;
-    case GIMP_RGBA_IMAGE:
-      FillBuffer<precision, /*has_alpha=*/true, /*num_channels=*/3>(
-          io, &pixel_data);
-      break;
-    default:
-      return false;
-  }
-
-  GeglBuffer* buffer = gimp_drawable_get_buffer(layer);
-  gegl_buffer_set(buffer, GEGL_RECTANGLE(0, 0, io.xsize(), io.ysize()), 0,
-                  nullptr, pixel_data.data(), GEGL_AUTO_ROWSTRIDE);
-  g_clear_object(&buffer);
-  return true;
-}
-
-}  // namespace
-
-Status LoadJpegXlImage(const gchar* const filename, gint32* const image_id) {
-  PaddedBytes compressed;
-  JXL_RETURN_IF_ERROR(ReadFile(filename, &compressed));
-
-  // TODO(deymo): Use C API instead of the ThreadPoolInternal.
-  ThreadPoolInternal pool;
-  DecompressParams dparams;
-  CodecInOut io;
-  JXL_RETURN_IF_ERROR(DecodeFile(dparams, compressed, &io, &pool));
-
-  const ColorEncoding& color_encoding = io.metadata.m.color_encoding;
-  JXL_RETURN_IF_ERROR(io.TransformTo(color_encoding, &pool));
-
-  GimpColorProfile* profile = nullptr;
-  if (color_encoding.IsSRGB()) {
-    profile = gimp_color_profile_new_rgb_srgb();
-  } else if (color_encoding.IsLinearSRGB()) {
-    profile = gimp_color_profile_new_rgb_srgb_linear();
-  } else {
-    profile = gimp_color_profile_new_from_icc_profile(
-        color_encoding.ICC().data(), color_encoding.ICC().size(),
-        /*error=*/nullptr);
-  }
-  if (profile == nullptr) {
-    return JXL_FAILURE(
-        "Failed to create GIMP color profile from %zu bytes of ICC data",
-        color_encoding.ICC().size());
-  }
+  gpointer pixels_buffer;
+  size_t buffer_size;
 
   GimpImageBaseType image_type;
-  GimpImageType layer_type;
+  GimpImageType layer_type = GIMP_RGB_IMAGE;
+  GimpPrecision precision = GIMP_PRECISION_U16_GAMMA;
+  JxlBasicInfo info = {};
+  JxlPixelFormat format = {};
 
-  if (io.Main().IsGray()) {
-    image_type = GIMP_GRAY;
-    if (io.Main().HasAlpha()) {
-      layer_type = GIMP_GRAYA_IMAGE;
-    } else {
-      layer_type = GIMP_GRAY_IMAGE;
-    }
-  } else {
-    image_type = GIMP_RGB;
-    if (io.Main().HasAlpha()) {
-      layer_type = GIMP_RGBA_IMAGE;
-    } else {
-      layer_type = GIMP_RGB_IMAGE;
-    }
+  format.num_channels = 4;
+  format.data_type = JXL_TYPE_UINT8;
+  format.endianness = JXL_NATIVE_ENDIAN;
+  format.align = 0;
+
+  JpegXlGimpProgress gimp_load_progress(
+      ("Opening:" + (std::string)filename).c_str());
+  gimp_load_progress.update();
+
+  // read file
+  std::ifstream instream(filename, std::ios::in | std::ios::binary);
+  std::vector<uint8_t> compressed((std::istreambuf_iterator<char>(instream)),
+                                  std::istreambuf_iterator<char>());
+  instream.close();
+
+  gimp_load_progress.update();
+
+  // multi-threaded parallel runner.
+  auto runner = JxlResizableParallelRunnerMake(nullptr);
+
+  auto dec = JxlDecoderMake(nullptr);
+  if (JXL_DEC_SUCCESS !=
+      JxlDecoderSubscribeEvents(dec.get(), JXL_DEC_BASIC_INFO |
+                                               JXL_DEC_COLOR_ENCODING |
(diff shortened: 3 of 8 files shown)


plugins/gimp/file-jxl-load.cc
```diff
@@ -10,163 +10,252 @@
...
-      if (has_alpha) {
-        pixel_data->push_back(
-            BufferFormat<precision>::FromFloat(255.f * alpha));
-      }
-    }
-  }
-}
+bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
+  std::vector<uint8_t> icc_profile;
+  GimpColorProfile *profile = NULL;
```
sboukortt, line 25: This is rather minor but could we use `nullptr`, and rename `loadlinear` to `load_linear`?

plugins/gimp/file-jxl-load.cc
```diff
@@ -10,163 +10,252 @@
...
+        if (JXL_DEC_SUCCESS != JxlDecoderGetColorAsICCProfile(
+                                   dec.get(), &format,
+                                   JXL_COLOR_PROFILE_TARGET_DATA,
+                                   icc_profile.data(), icc_profile.size())) {
+          g_printerr("Error: JxlDecoderGetColorAsICCProfile failed\n");
+          return false;
+        }
+
+        GError **error = NULL;
+        profile = gimp_color_profile_new_from_icc_profile(
```
sboukortt, line 180: I think this will reintroduce https://gitlab.com/wg1/jpeg-xl/-/issues/182. Ideally, we would also try to get the color profile as a `JxlColorEncoding` and, if successful, compare it against sRGB and use `gimp_color_profile_new_rgb_srgb` if it matches. Would you like to give it a try in this pull request or should we do it later?

plugins/gimp/file-jxl-save.cc
```diff
@@ -10,122 +10,368 @@
 #undef MIN
 #undef CLAMP
 
-#include "lib/jxl/alpha.h"
+#include <glib/gi18n.h>
+#include <libgimp/gimp.h>
+#include <libgimp/gimpui.h>
+//#include <glib/gstdio.h>
```
sboukortt, line 16: This seems to be a leftover.

plugins/gimp/file-jxl-save.cc
```diff
@@ -10,122 +10,368 @@
...
-      if (has_alpha) {
-        alpha_row[x] = alpha_multiplier *
-                       BufferFormat<GIMP_PRECISION_FLOAT_GAMMA>::ToFloat(
-                           *current_sample++);
-      }
-    }
-  }
+typedef struct {
+  gboolean tmp;
+} JpegXlSaveVals;
```
sboukortt, line 34: In C++, you can write this simply as:

```suggestion
struct JpegXlSaveVals {
  gboolean tmp;
};
```

plugins/gimp/file-jxl-save.cc
```diff
@@ -10,122 +10,368 @@
...
-  if (has_alpha) {
-    io->metadata.m.SetAlphaBits(alpha_bits);
-    io->Main().SetAlpha(std::move(alpha), /*alpha_is_premultiplied=*/false);
-  }
-  return true;
+static JpegXlSaveVals jxl_vals = {
+    TRUE  // tmp
+};
+
+bool save_dialog(void) {
```
sboukortt, line 40: Likewise, `(void)` is a remnant from C:

```suggestion
bool save_dialog() {
```

plugins/gimp/file-jxl-save.cc
```diff
@@ -10,122 +10,368 @@
...
-  CompressParams params;
-  PassesEncoderState encoder_state;
-  PaddedBytes compressed;
-  ThreadPoolInternal pool;
-  params.butteraugli_distance = 1.f;
-  JXL_RETURN_IF_ERROR(EncodeFile(params, &io, &encoder_state, &compressed,
-                                 /*aux_out=*/nullptr, &pool));
-  JXL_RETURN_IF_ERROR(WriteFile(compressed, filename));
+  JxlColorEncoding color_encoding = {};
+  JxlColorEncodingSetToSRGB(&color_encoding,
```
sboukortt, line 306: I feel that I must be missing something but shouldn’t we get the profile from GIMP?

sboukortt (commented):
Thank you, it’s great to have this!

plugins/gimp/file-jxl-load.cc
xiota: No problem.  I'll make the change and update the commit.

plugins/gimp/file-jxl-load.cc
xiota: I didn't notice the regression because I had GIMP set to keep embedded profiles.  The problem is Thread Pool Internal has access to CodecInOut, but Parallel Runner doesn't, as far as I can tell.  I'll take a look at it.

plugins/gimp/file-jxl-save.cc
xiota: Thanks.  I'll change this.  It's been a long time since I've done anything with a C-like language, and all of the plugins in the GIMP repository are written in C...

plugins/gimp/file-jxl-save.cc
xiota: `glib/gi18n.h` is used for the `_(string)` wrappers.  Not sure whether to keep them or not.

I also don't know whether to put `libgimp/gimp*` in `file-jxl-save.h`.

`glib/gstdio.h` is commented out because I didn't know whether I'd need it later.  So far, it looks like I won't, so I'll probably just remove it.

plugins/gimp/file-jxl-save.cc
xiota: Yes, getting the profile from GIMP would make sense.  I either overlooked the old code that did that or didn't know how to do it.  I'll take a look at it and update the commit if I can figure it out.

xiota: So my left hand doesn't know what my right hand is doing... there's some icc processing at the beginning of the file, but it doesn't go anywhere... Also looks like another CodecInOut access difference between Thread Pool Internal and Parallel Runner.  I'll poke around some more to see if there's something I overlooked earlier..

plugins/gimp/file-jxl-load.cc
xiota: I think the regression is fixed now...  it's a bit complicated because I'm avoiding using internal headers.  (I'd like to remove "lib/jxl/base/file_io.h" someday.)

plugins/gimp/file-jxl-save.cc
xiota: ICC profiles should be working more or less as expected now.  If internal GIMP profiles are used in the image, then internal JPEG XL profiles will be embedded in the image.  Otherwise the user-specified ICC profile is embedded.

There are also some limits on when linear or perceptual gamma can be used.  Linear only works with float and perceptual only with int because images with linear int or perceptual float don't display properly.  Conversion by GIMP without any user intervention.

xiota:
@sboukortt I made the changes you suggested.  ICC profiles should work more or less as expected.  Linear/perceptual are converted without user intervention depending on the bit format because certain combinations don't display properly.  The limit can be removed later if the cause is discovered.

I'm also curious about your opinion about whether to set `jxl_info.uses_original_profile` to true or false.  When set to true, only one color profile is saved in the jxl file.  When false, there is an original profile and a data profile.  Files generated by `cjxl` usually have the same profile set for original and data.  However, when saving with the modified GIMP plugin, it's possible to have an original AdobeRGB profile (for instance), and a data sRGB profile.  Then I don't know whether the image should be loaded with the original or data profile.  (I don't know what color transformations the JXL encoder is doing, if any.)

Right now, the modified plugin is saving both profiles and loading using the data profile.  It's a simple change if only one profile should be saved or if images should be loaded using the original profile.

novomesk:
It is a good idea to rewrite the plug-in to use public API.

In order to increase chances for this plug-in to be included by official GIMP, I suggest to try following GIMP coding style: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/CODING_STYLE.md

From GIMP's perspective it would be preferred to use only C instead of C++, that's just FYI.

I have alternative implementation for GIMP 2.99 (it has different API than 2.10.x) in C only. It is not finished but if you want to take a look: https://github.com/novomesk/gimp-jpegxl-plugin/blob/main/src/file-jpegxl.c

deymo:
@novomesk do you plan to upstream your plugin to GIMP repo?

xiota:
@novomesk Converting to C and following GIMP plugin conventions would be a major undertaking that I'll leave for someone else.  I expect GIMP won't have an official plugin until after the JPEG XL standard is published.

Even if I wanted to follow GIMP coding style, I couldn't.  To push to this repository requires passing an automated format check.  It's nice not having to type spaces everywhere, then just letting `clang-format` fix it.

novomesk:
> @novomesk do you plan to upstream your plugin to GIMP repo?

@deymo Yes. GIMP developer knows about the plug-ins and waiting for MR when everything is OK according the expectations.

When contributing to GIMP, it has to be for 2.99.x first and later the new features can be backported to old 2.10.x

In my plug-in I always export in 16bit depth with non-linear sRGB. I chose this for compatibility reasons because lot of viewers do not have color management yet.

veluca93:
> > @novomesk do you plan to upstream your plugin to GIMP repo?
> 
> @deymo Yes. GIMP developer knows about the plug-ins and waiting for MR when everything is OK according the expectations.
> 
> When contributing to GIMP, it has to be for 2.99.x first and later the new features can be backported to old 2.10.x
> 
> In my plug-in I always export in 16bit depth with non-linear sRGB. I chose this for compatibility reasons because lot of viewers do not have color management yet.

I don't think that will work for non-SDR content, among other things... I think it would be a better idea to export things in their current colorspace, and leave it to users to do compatibility decisions :)

Anyway, if you plan to upstream your plugin, I don't think upstreaming this plugin is necessary, no?

deymo:
> Anyway, if you plan to upstream your plugin, I don't think upstreaming this plugin is necessary, no?

Yeah, it looks like @novomesk plugin is already ported to the external API. We could delete this one and add a reference to that one in the README. WDYT?

It would be good to update it to support HDR properly in gimp.

deymo:
Actually, let's keep this plugin around (in C++, with libjxl code style) for using it in 2.10.

xiota:
@deymo  Nooooo.... don't delete this plugin...  Aside from the work I just put into it, considering the historical pace of GIMP development, the new API may not be considered stable for the next decade.  Also, this plugin is able to export to grayscale, 8- and 16-big non-linear, and 32-bit linear formats.

novomesk:
Keep this plug-in for GIMP 2.10, don't delete.

And I will add option for 2.99 plug-in to use original profile during export (otherwise compatible sRGB will be used).

xiota:
@novomesk 
In GIMP 2.10, trying to load 16-bit linear jxl files with the public API doesn't seem to work correctly.  Maybe 2.99 handles it better.

This plugin tries to figure out what profile GIMP is using to set appropriate flags for libjxl (and vice versa).  Otherwise, you can just embed the ICC profile GIMP provides, but that seems to result in disaster since the API doesn't seem to extract the settings from ICC profiles.  When `uses_original_profile=false`, it looks like the public API decodes XYB to sRGB no matter what.  However, if `uses_original_profile=true`, I noticed file sizes double.

deymo:
> When `uses_original_profile=false`, it looks like the public API decodes XYB to sRGB no matter what.

@xiota that's correct. Unlike many other formats, in JXL it is possible to have images encoded in absolute color space like XYB. In other formats like png you provide an array of integers to the encoder, and on the other side you when you decode you get the same integers (or ~similar integers in lossy ones). There's a blob somewhere that tells you what those integers mean in terms of colors (like the ICC blob), but the codec doesn't encode "colors", it encodes numbers.

In the `uses_original_profile=false` case JXL decode API tells you the color encoding of the pixels it returns (normally sRGB) and it tells your the color encoding of the original image as an FYI. The right thing to do depends what you intend to do with the decoded pixels. For example, let's say the original image was 8-bit and had some specific ICC attached (which is not sRGB) but in JXL the image is encoded in XYB, then:
 * if you want to obtain the original 8-bit pixel values you need to decode to float in the jxl api, convert those floats from the jxl-provided (`JXL_COLOR_PROFILE_TARGET_DATA`) color encoding (likely sRGB in this case) to the ICC attached in the file and then reduce the precision to 8-bit integers (you'd likely do this on the output callback to save memory). If you instead get 8-bit integers in sRGB from the API you would incur into clamping and rounding issues.
 * OTOH, if you want to *display* the pixels on a monitor then the original color encoding is basically irrelevant because you'd be converting that image to the color encoding that your output buffer for the display needs, so doing a color conversion directly to the display is likely faster.

I think that for GIMP you always want the first case. If `uses_original_profile=false` then you want to request FLOAT and do the cms conversion in the plugin (the decoder api doesn't use a CMS), in the =true case the pixel data is already in that profile so no conversion is needed in the example above you could directly request 8-bit integers from the API and skip the conversion altogether.

@sboukortt @veluca93 WDYT?

xiota:
> I think that for GIMP you always want the first case. If uses_original_profile=false then you want to request FLOAT and do the cms conversion in the plugin (the decoder api doesn't use a CMS), in the =true case the pixel data is already in that profile so no conversion is needed in the example above you could directly request 8-bit integers from the API and skip the conversion altogether.

@deymo That makes sense, and it seems to be what people would expect from existing image formats and software.  I think I'd like to leave it for a future pull request though.

plugins/gimp/file-jxl-save.h
```diff
@@ -6,15 +6,61 @@
...
-#include "lib/jxl/base/status.h"
+#include "jxl/encode.h"
+#include "jxl/encode_cxx.h"
+#include "plugins/gimp/common.h"
 
 namespace jxl {
 
-Status SaveJpegXlImage(gint32 image_id, gint32 drawable_id,
-                       gint32 orig_image_id, const gchar* filename);
+bool save_dialog();
```
sboukortt, line 15: Given that this and `JpegXlSaveOpts` don’t seem to be used outside of the `.cc` file, it would perhaps be preferable not to declare them in the header file, and to put them in an anonymous namespace to give them internal linkage. (The local naming convention would also be to name the function `SaveDialog`.)

plugins/gimp/file-jxl-save.cc
```diff
@@ -3,130 +3,813 @@
...
+
+  basic_info.alpha_bits = 0;
+  basic_info.alpha_exponent_bits = 0;
+  basic_info.uses_original_profile = false;
+  basic_info.intensity_target = 255.0f;
+  basic_info.orientation = JXL_ORIENT_IDENTITY;
+
+  babl_format_str = "RGB u16";
+  babl_type_str = "u16";
+  babl_model_str = "RGB";
```
sboukortt, line 578: FYI, in modern C++, it is also possible to initialize members in the class declaration directly. (But up to you.)

https://en.cppreference.com/w/cpp/language/data_members#Member_initialization (2)

plugins/gimp/file-jxl-save.cc
```diff
@@ -3,130 +3,813 @@
...
+JpegXlSaveOpts jxl_save_opts;
+
+bool SaveJpegXlImage(const gint32 image_id, const gint32 drawable_id,
+                     const gint32 orig_image_id, const gchar* const filename) {
+  if (!save_dialog()) {
+    return true;
+  }
+
+  JpegXlGimpProgress gimp_save_progress(
+      ("Saving JPEG XL file:" + (std::string)filename).c_str());
```
sboukortt, line 26: C++-style cast would be preferred:

```suggestion
      ("Saving JPEG XL file:" + std::string(filename)).c_str());
```

sboukortt (approved):
Thanks again!

plugins/gimp/file-jxl-save.h
```diff
@@ -6,15 +6,61 @@
...
+
+bool SaveJpegXlImage(gint32 image_id, gint32 drawable_id, gint32 orig_image_id,
+                     const gchar* filename);
+
+class JpegXlSaveOpts {
+ public:
+  float distance;
+  float quality;
```
veluca93, line 23: Which one ends up getting used? `distance` or `quality`? I see that in the dialog you have both...

plugins/gimp/file-jxl-save.cc
```diff
@@ -3,130 +3,813 @@
...
+  gtk_box_pack_start(GTK_BOX(vbox), toggle, false, false, 0);
+  gtk_widget_set_sensitive(toggle, false);
+  gtk_widget_show(toggle);
+
+  advanced_opts.push_back(toggle);
+
+  // Encoding Effort/Speed
+  static gchar effort_help[] =
+      "Encoding Effort: Higher number is more effort (slower).\n\tDefault = 7.";
+  frame = gtk_frame_new("Encoding Effort/Speed");
```
veluca93, line 460: Just from the option name, I'd consider "effort/speed" somewhat confusing and I'd just leave Effort there. WDYT?

plugins/gimp/file-jxl-save.h
veluca93: Oh, nvm me, apparently they are tied together :)

plugins/gimp/file-jxl-save.cc
xiota: So this shows how long it's been since I've done anything with C++...  I'd personally rather initialize them directly.  Not sure if I should do that for this commit or at a later time, since copy/paste can lead to errors.

plugins/gimp/file-jxl-save.cc
xiota: I was planning to change it to just "Encoding Effort", as you suggest.  But I thought all the force pushes may have been delaying review, so I was holding off on it.

plugins/gimp/file-jxl-save.cc
xiota: Thanks.  Will fix this.

plugins/gimp/file-jxl-save.cc
```diff
@@ -3,130 +3,813 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// #include "plugins/gimp/file-jxl-save.h"
+
```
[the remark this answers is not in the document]
xiota: Yes.  Will remove.

plugins/gimp/common.cc
```diff
@@ -0,0 +1,27 @@
...
+  gimp_progress_update((gdouble)++cur_progress / (gdouble)max_progress);
+  return;
+}  // JpegXlGimpProgress::update
+
+void JpegXlGimpProgress::finished() {
+  gimp_progress_update(1.0);
+  return;
+}  // JpegXlGimpProgress::finished
+
+};  // namespace jxl
```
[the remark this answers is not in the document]
xiota: Will remove.  Thanks.

plugins/gimp/file-jxl-save.h
xiota: What about the naming of the other functions, like the gnome callback functions?  Would those become GuiOnChangeSuchAndSuch()?

plugins/gimp/file-jxl-save.cc
veluca93: I think what Sami meant is that you can do 

```
JpegXlSaveOpts::JpegXlSaveOpts(): advanced_mode(false),  lossless(false), ... {
  set_distance(1.0f);
}
```

which I personally would find more readable - but I'm not sure and I don't really have that strong of an opinion :P

plugins/gimp/common.h
```diff
@@ -7,49 +7,38 @@
...
-};
-template <>
-struct BufferFormat<GIMP_PRECISION_FLOAT_GAMMA> {
-  using Sample = float;
-  static Sample FromFloat(const float x) { return x; }
-  static float ToFloat(const Sample s) { return s; }
-};
+class JpegXlGimpProgress {
+ public:
+  JpegXlGimpProgress(const char *message);
```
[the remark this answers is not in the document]
xiota: Will change.  Thanks.

plugins/gimp/file-jxl-save.cc
sboukortt: I was thinking of even putting them outside of the constructor, directly in the class declaration:

```c++
class JpegXlSaveOpts {
 public:
  bool advanced_mode = false;
  bool lossless = false;
  // …
};
```

xiota (commented):
@mo271 @sboukortt @veluca93  I've pushed what I hope will be the final changes for this PR.  I made the changes suggested earlier.  Fixed some typos.  Added float16 image saving, and added loading multiple frames as layers.  Thank you for your help.

plugins/gimp/common.cc
```diff
@@ -0,0 +1,27 @@
...
+JpegXlGimpProgress::JpegXlGimpProgress(const char *message) {
+  cur_progress = 0;
+  max_progress = 10;
+
+  gimp_progress_init_printf("%s\n", message);
+}  // JpegXlGimpProgress::JpegXlGimpProgress
+
+void JpegXlGimpProgress::update() {
+  gimp_progress_update((gdouble)++cur_progress / (gdouble)max_progress);
+  return;
```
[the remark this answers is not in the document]
sboukortt: I mean the return statement, since the end of the function is reached anyway.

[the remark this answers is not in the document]
xiota: O, haha.  I try to explicitly return functions.  My editor highlights them so it helps me identify where the end is.

plugins/gimp/file-jxl-save.h
xiota: My favorite part of the plugin is those sliders.

sboukortt (approved):

deymo (approved):
I think there some cleanup to do here but we can do it later.

xiota:
@deymo I know the function names don't follow the local style and should probably be changed.  Is there any other general cleanup you had in mind?

jonsneyers:
Merging now, code style cleanup is not urgent imo for 'external' code like this plugin.

plugins/gimp/file-jxl-save.cc
```diff
@@ -5,128 +5,884 @@
...
+  }
+
+  if (JXL_ENC_SUCCESS !=
+      JxlEncoderSetBasicInfo(enc.get(), &jxl_save_opts.basic_info)) {
+    g_printerr("JXL Error: JxlEncoderSetBasicInfo failed\n");
+    return false;
+  }
+
+  // try to use ICC profile
+  if (icc.size() > 0 && !gimp_color_profile_is_gray) {
```
deymo, line 765: `gimp_color_profile_is_gray` is using the address of the function as a boolean, instead of calling it. I'll send a fix for this.

xiota: Thanks.  I saw that after it had already been merged.  It was supposed to be `!jxl_save_opts.is_gray`.

PR #469, opened 2021-08-19
Add information about jbrd presence to jxlinfo

EugeneVert:
I came across a situation where I need to losslessly decode an image collection, which may include images with reconstruction data. I wanted to use `jxlinfo` in a shell script to separate what to decode in png and what to reconstruct to jpg, but there was no information about reconstruction data presence.

Unfortunately, `JxlBasicInfo` does not contain information about the presence of the `jbrd` box.


deymo (approved):
Yeah, the jbrd box may come a lot later than the basic info. @lvandeve WDYT?

EugeneVert:
Moved printf to branch with basic_info to avoid printing reconstruction info on error

deymo (approved):

examples/jxlinfo.c
```diff
@@ -78,6 +81,11 @@ int PrintBasicInfo(FILE* file) {
       printf("xsize: %u\n", info.xsize);
       printf("ysize:  %u\n", info.ysize);
       printf("have_container: %d\n", info.have_container);
+      if (info.have_container) {
```
deymo, line 84: I'd need to check if JXL_DEC_JPEG_RECONSTRUCTION is required to occur before JXL_DEC_BASIC_INFO. It may depend on the order of the boxes, but I'm not sure if it is allowed to be in the other order. @lvandeve ?

jonsneyers: The spec allows the boxes to be in either order (`jbrd` before or after `jxlc`). The current encoder always puts `jbrd` before `jxlc`. Both orders make sense: putting `jbrd` first makes it easy to see that it can be reconstructed to jpeg, and maybe (in some cases) makes it possible to do a streaming reconstruction; putting it at the end makes sense for better progressive rendering, since you don't need the `jbrd` to show the image.

Do we want to ensure that `jxlinfo` only reads the beginning of a file? Or is it OK if it does some seeking (just no actual decode)? If we want to display information like the presence of exif/xml or the number of frames / duration of an animation, then seeking is needed in any case, and then it would make sense to iterate over all boxes (and to do skip-frame parsing of the codestream).

deymo: I wouldn't expect jxlinfo to be limited to just reading the beginning of the file. If the issue here is that we read the file chunk by chunk that can be addressed with providing other modes of input in the API (planned) or by using mmap (at least in unix); but anyway reading the file is not the most expensive part of decoding.

jonsneyers (approved):

PR #470, opened 2021-08-19
splines: prevent divide by zero in DrawGaussian

lovell:
As discussed in https://github.com/libjxl/libjxl/issues/129


sboukortt (approved):
Thanks!

PR #475, opened 2021-08-20
Allow greater range of butteraugli distances in cjxl

xiota:
The encoder is hard coded to allow a minimum distance of 0.01.
It's reasonable for cjxl match the encoder's capability.

### DIFF
--- a/tools/cjxl.cc
+++ b/tools/cjxl.cc
@@ -616,8 +616,8 @@ jxl::Status CompressArgs::ValidateArgs(const CommandLineParser& cmdline) {
   }
 
   if (got_distance) {
-    constexpr float butteraugli_min_dist = 0.1f;
-    constexpr float butteraugli_max_dist = 15.0f;
+    constexpr float butteraugli_min_dist = 0.01f;
+    constexpr float butteraugli_max_dist = 45.0f;
     if (!(0 <= params.butteraugli_distance &&
           params.butteraugli_distance <= butteraugli_max_dist)) {
       fprintf(stderr, "Invalid/out of range distance, try 0 to %g.\n",


tools/cmdline.cc
```diff
@@ -53,9 +53,30 @@ bool CommandLineParser::Parse(int argc, const char* argv[]) {
       i++;
       continue;
     }
+    if (!strcmp("-hv", argv[i]) || !strcmp("-vh", argv[i])) {
```
deymo, line 56: I'd probably leave this out of this PR. I also think it is possible to implement it such you can bundle any once char flags rather than checking for these specific combinations.

tools/cjxl.cc
```diff
@@ -616,8 +616,8 @@ jxl::Status CompressArgs::ValidateArgs(const CommandLineParser& cmdline) {
   }
 
   if (got_distance) {
-    constexpr float butteraugli_min_dist = 0.1f;
-    constexpr float butteraugli_max_dist = 15.0f;
+    constexpr float butteraugli_min_dist = 0.01f;
```
deymo, line 619: The idea of having this 0.1 to 15 range in butteraugli distances is that is a reasonable range you should use. While it is possible to generate images with much larger distance like 45 in this case, it is probably not the intended use. I'd defer to @jyrkialakuijala for this.

deymo (commented):
Thanks for this proposal

tools/cmdline.cc
xiota: It seems like `v` and `h` are handled separately from the other short flags in this section of code.  So bundling short options probably won't affect `v` and `h`.  Should this be in a separate pull request?

deymo: yeah, let's do this in another PR and discuss the distance range here instead.

xiota: Okay.  I'll push an update to remove this.

TheHardew: I would also expect it to handle options like `-me9`, which would be modular encoding with effort 9, similar to how most GNU programs work. Maybe even `-e9d2` for effort 9 and distance 2, although I don't know if any programs support this. `du -d1t20` doesn't work, but `du -d1 -t20` does.

xiota: Maybe I should open it as a feature request rather than a pull request.

tools/cjxl.cc
xiota: For a reasonable range, there is the `q` interface to `d`.  If the control logic to turn on lossless mode is ignored, setting q=100 would result in d=0.1.  People using `d` directly may want to go outside of what `q` allows.  For instance, I wanted to know what really bad JXL artifacts look like, but d=15 didn't really show much.  d=0.01 would be useful for someone who wants/needs the highest quality level possible, but doesn't quite need lossless.  For photographic images, file sizes at d=0.01 appear to be about half of what lossless produces.

jyrkialakuijala: The internet quality range is about d0.5 to d3. Above d8 cjxl no longer works as designed.

Already at d15 the artefacts look nothing like the image quality degradation in the normal range (d0.8 to d3).

Increasing the quantization and looking at errors is not going to reveal much from behaviour at normal qualities. A better alternative is to zoom and improve viewing conditions (brightness of the monitor, no glare in the room, wipe the monitor before viewing to remove dust, clean glasses, use objective metrics to find where to look in the image, flip between original and compressed).

We have traditionally maintained a d0.1 level on distance because some parts craches or didn't operate well at lower distances. If those have been fixed, I have no issues making the lower limit smaller. I don't understand well how much risk that is going to give us.

xiota: I've been intermittently using d=0.01.  I chose d=0.01 because d=0.009 raised some error messages.  I think they were !fp assertion messages.  Is there a collection of test images?

xiota: @jyrkialakuijala I found that `kMinButteraugliDistance = 0.01f` in `enc_params.h` is the reason I couldn't reduce the min in `cjxl` lower than 0.01 without error.  When I change both, to 0.001, I was able to compress and decompress photographic images without raising any errors with d=0.001.  But speed was very slow, and there wasn't any noticeable difference from d=0.01.  I think it's reasonable to set the min in `cjxl` to match `enc_params.h` (currently 0.01).

jonsneyers: I don't see a reason to have a different range in cjxl than in the api or in benchmark_xl. Obviously for normal use cases, such extreme settings will not be useful, but people are used to that, e.g. you can make a -q 1 jpeg that is just useless, or a -q 100 jpeg which is probably useless too. We could maybe display a warning message in cjxl when values are used outside the 'reasonable' range (say outside d0.3 to d8), but just refusing to encode seems a bit paternalistic to me.

xiota: @jonsneyers  I'll look into how to add a reasonable range warning.

jonsneyers: I'd say allow the entire range that the api / actual encoder allows, but print a warning to stderr (unless --quiet is used) in case values outside the "sane" range are used.

Obviously exact boundaries are a matter of taste, but I'd propose 0.3 to 6 as the "sane range" where we know everything should work well and with little surprises. Above 6, I'd produce a warning that says something like
```
WARNING: this encode setting is likely to produce severe compression artifacts
```

and below 0.3 (but above 0) I'd produce a warning saying something like

```
WARNING: this encode setting could produce unnecessarily large files. Mathematically lossless compression, or higher distance lossy compression might be more suitable to your use case.
```

mo271:
We are considering to allow this in cjxl_ng, but not for cjxl right now.

PR #478, opened 2021-08-23
Fix running the hdr-vdp metric.

veluca93:


tools/benchmark/metrics/hdrvdp.sh
```diff
@@ -4,6 +4,6 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-./compute_octave_metric.sh "$@" \
+"$(dirname "$0")"/compute_octave_metric.sh "$@" \
```
sboukortt, line 7: Uh. So we thought of this for the next two lines but not this one?

sboukortt (approved):

tools/benchmark/metrics/hdrvdp.sh
veluca93: Beats me...

PR #479, opened 2021-08-23
Fix patch decoding fuzzer-found slow behaviour.

### LINKED ISSUE #476: Possible reduction in the maximum number of patches
The following 73 byte image, found via fuzz testing, claims to contain a patch dictionary with 491521 entries.
 
https://github.com/libjxl/libjxl/files/7023786/patch-491521.txt

This is less than the 1<<24 `kMaxPatches` limit enforced in `PatchDictionary::Decode`

https://github.com/libjxl/libjxl/blob/e00ac5788b769b14af8a543d6572aee615513649/lib/jxl/dec_patch_dictionary.cc#L55-L56

Attempting to process this image takes many minutes and the `PatchPosition` vector it builds gradually consumes all available memory.

Is 16777216 (1<<24) an appropriate maximum? Perhaps significantly reducing this to something like 4096 (1<<12) might be sufficient?

veluca93:
Fixes #476. Two issues are fixed here:
- reserve() on a vector does not apply the usual doubling trick, making
  the loop quadratic
- checking separately for kMaxPatches for the total number of distinct
  patches and the number of occurrences of a patch effectively squares
  the actual bound on # of patches, bringing it to 1<<48, which is not
  OK.

### DIFF
--- a/lib/jxl/dec_patch_dictionary.cc
+++ b/lib/jxl/dec_patch_dictionary.cc
@@ -35,7 +35,7 @@
 
 namespace jxl {
 
-constexpr int kMaxPatches = 1 << 24;
+constexpr size_t kMaxPatches = 1 << 24;
 
 Status PatchDictionary::Decode(BitReader* br, size_t xsize, size_t ysize,
                                bool* uses_extra_channels) {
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_patch_dictionary.cc
```diff
@@ -35,7 +35,7 @@
 
 namespace jxl {
 
-constexpr int kMaxPatches = 1 << 24;
+constexpr size_t kMaxPatches = 1 << 24;
```
jonsneyers, line 38: In the profile amendment, the actual limit is `min(1<<24, width*height/16))`, but I guess we are free to not do the size-dependent thing and decode some bitstreams that are technically using too many patches.

jonsneyers (approved):

Issue #468, opened 2021-08-18
Failed to read certain JPGs


Fraetor:
**Describe the bug**
cjxl is not able to read certain JPGs. The files open successfully in my image viewer (Eye of GNOME), and Firefox

**To Reproduce**
Steps to reproduce the behaviour: Try and encode the JPG files in the attached archive: [[Non readable JPEGs.zip](https://github.com/libjxl/libjxl/files/7010747/Non.readable.JPEGs.zip)]

**Expected behaviour**
The files to be read successfully, then encoded.

**Screenshots**
```
$ cjxl 04.jpg 04.jxl
JPEG XL encoder v0.5.0 c98f133 [AVX2,SSE4,Scalar]
Failed to read image 04.jpg.
```

**Environment**
 - OS: Debian 11 Bullseye
 - Compiler version: gcc (Debian 10.2.1-6) 10.2.1 20210110
 - CPU type: x86_64
 - cjxl/djxl version string: cjxl v0.5.0 c98f133 [AVX2,SSE4,Scalar]

**Additional context**
Here is the output of the `file` command, with newlines added for clarity:
```
$ file *
04.jpg:   JPEG image data, JFIF standard 1.01, resolution (DPI), density 72x72, segment length 16, Exif Standard: [TIFF image data, little-endian, direntries=7, orientation=upper-left, xresolution=98, yresolution=106, resolutionunit=2, software=Adobe Photoshop CS5.1 Windows, datetime=2012:12:20 17:27:02], baseline, precision 8, 1200x1680, components 4

19-2.jpg: JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=7, orientation=upper-left, xresolution=98, yresolution=106, resolutionunit=2, software=Adobe Photoshop CS6 (Windows), datetime=2014:05:13 12:01:10], baseline, precision 8, 1100x1600, components 4

19.jpg:   JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=7, orientation=upper-left, xresolution=98, yresolution=106, resolutionunit=2, software=Adobe Photoshop CS6 (Windows), datetime=2014:08:31 15:09:20], progressive, precision 8, 1100x1600, components 4

23.jpg:   JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=12, height=1518, bps=0, PhotometricIntepretation=CMYK, orientation=upper-left, width=1076], baseline, precision 8, 1076x1518, components 4

[...]

PR #480, opened 2021-08-23
print error when encountering non-recompressable (cmyk or corrupt) JPEG

jonsneyers:
See https://github.com/libjxl/libjxl/issues/468

Corrupt JPEGs and CMYK JPEGs cannot be losslessly recompressed to jxl.
Instead of just saying "failed to read", write a more useful error message in this case.

### DIFF
--- a/lib/extras/codec_jpg.cc
+++ b/lib/extras/codec_jpg.cc
@@ -244,7 +244,9 @@ Status DecodeImageJPG(const Span<const uint8_t> bytes, ThreadPool* pool,
 
   // Use brunsli JPEG decoder to read quantized coefficients.
   if (target == DecodeTarget::kQuantizedCoeffs) {
-    return jxl::jpeg::DecodeImageJPG(bytes, io);
+    Status could_decode = jxl::jpeg::DecodeImageJPG(bytes, io);
+    if (!could_decode) fprintf(stderr, "Corrupt or CMYK JPEG.\n");
+    return could_decode;
   }
 
 #if JPEGXL_ENABLE_JPEG


veluca93 (approved):

veluca93 (changes requested):
Actually - shouldn't this return JXL_FAILURE instead, or fill an error message for the caller? I am not sure a print should happen here.

jonsneyers:
It will return failure, but in the caller failure is handled silently and the next codec is tried (before it gets there, it has already tried decoding it as a PNG, GIF, PPM etc and failed silently each time). It would maybe be nicer to use a different mechanism where decoders can fail silently ("this is not the right decoder") and non-silently ("right decoder, but bad bitstream"), but for that the all the codecs in `extras` need to be refactored to no longer use `Status` but some new struct with 3 return values.

veluca93:
> It will return failure, but in the caller failure is handled silently and the next codec is tried (before it gets there, it has already tried decoding it as a PNG, GIF, PPM etc and failed silently each time). It would maybe be nicer to use a different mechanism where decoders can fail silently ("this is not the right decoder") and non-silently ("right decoder, but bad bitstream"), but for that the all the codecs in `extras` need to be refactored to no longer use `Status` but some new struct with 3 return values.

Status already has multiple values:

```
enum class StatusCode : int32_t {
  // Non-fatal errors (negative values).
  kNotEnoughBytes = -1,

  // The only non-error status code.
  kOk = 0,

  // Fatal-errors (positive values)
  kGenericError = 1,
};
```
We could add stuff to that ;)

jonsneyers:
Ah cool, OK that makes things easier. Could add `kWrongDecoder = -2` as a non-fatal error (which then causes codec.cc to try others) and reserve the fatal errors for cases where it's the right decoder but a bad input (e.g. `kCmykJpeg`), and codec.cc shouldn't bother to try other codecs (that will also get rid of trying openexr when the input is a bad jpeg, for example).

But where should the printing of an error message happen? Propagate specific fatal errors all the way to cjxl and print different user-facing errors there?

veluca93:
> Ah cool, OK that makes things easier. Could add `kWrongDecoder = -2` as a non-fatal error (which then causes codec.cc to try others) and reserve the fatal errors for cases where it's the right decoder but a bad input (e.g. `kCmykJpeg`), and codec.cc shouldn't bother to try other codecs (that will also get rid of trying openexr when the input is a bad jpeg, for example).
> 
> But where should the printing of an error message happen? Propagate specific fatal errors all the way to cjxl and print different user-facing errors there?

IIRC that's how it's usually done; one could also add a std::string / pointer to static char* to Status... @deymo any opinions?

deymo:
This being in the `extras` library, I think the fprintf is ok. Adding a status enum value for errors generated outside the lib doesn't seem correct. I'm working on refactoring this code to use less of the internal API so they can be used with the external one.

deymo:
For library-originated errors we have `JxlDecoderStatus` but there's no way to know more details like an error code. The issue with having more granular error codes for the failure case is that they become part of the API so you might accidentally start relying on them for specific behavior (if error_code == JPEG_IS_CMYK then ...).

veluca93 (approved):

jonsneyers:
OK, so let's keep it like this then.

PR #481, opened 2021-08-23
flag to stop parsing options, fixes #433

### LINKED ISSUE #433: Add flag to cjxl/djxl to stop parsing options.
Filenames with leading `-` characters are hard to feed in and out of the CLI encoder/decoder as it tries to parse them as options.

Add a special option (`--` seems to be pretty much ubiquitous for this) to tell the program to stop parsing further arguments as options.

**Describe alternatives you've considered**
There are workarounds, such as by specifying a full path (e.g `./-filename.jpg`), but this in inconvenient for use within scripts.

jonsneyers:
See #433. If for whatever reason you need things to work on filenames that start with a dash, you can now do `cjxl -q 80 -- -myweirdfile.png -myweirdoutputfile.jxl`.

### DIFF
--- a/tools/cmdline.h
+++ b/tools/cmdline.h
@@ -41,7 +41,7 @@ class CommandLineParser {
     virtual bool matched() const = 0;
 
     // Returns whether this option matches the passed command line argument.
-    virtual bool Match(const char* arg) const = 0;
+    virtual bool Match(const char* arg, bool parse_options) const = 0;
 
     // Parses the option. The passed i points to the argument with the flag
     // that matches either the short or the long name.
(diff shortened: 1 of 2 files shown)


tools/cmdline.h
```diff
@@ -136,8 +136,8 @@ class CommandLineParser {
     // Only match non-flag values. This means that you can't pass '-foo' as a
```
deymo, line 136: This comment needs to be updated.

tools/cmdline.h
```diff
@@ -41,7 +41,7 @@ class CommandLineParser {
     virtual bool matched() const = 0;
 
     // Returns whether this option matches the passed command line argument.
```
deymo, line 43: Please explain what's the meaning of `parse_options` in the comment.

deymo (approved):

tools/cmdline.h
jonsneyers: Done.

tools/cmdline.h
jonsneyers: Done.

deymo (approved):

PR #483, opened 2021-08-23
Add float16 support to public API

xiota:
A parameter to flag whether the input is float or not was added to  ConvertFromExternal() so it can differentiate between uint16 and float16 images.  Existing calls to ConvertFromExternal were modified.  Checks to reject float16 in the external API were removed.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -30,9 +30,7 @@
    JXL_ENC_ERROR)
 #endif  // JXL_CRASH_ON_ERROR
 
-namespace jxl {
-
-}  // namespace jxl
+namespace jxl {}  // namespace jxl
 
 uint32_t JxlEncoderVersion(void) {
   return JPEGXL_MAJOR_VERSION * 1000000 + JPEGXL_MINOR_VERSION * 1000 +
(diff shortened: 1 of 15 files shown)


lib/jxl/encode.cc
```diff
@@ -157,9 +155,10 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
     return JXL_ENC_ERROR;
   }
   if (info->exponent_bits_per_sample) {
-    if (info->exponent_bits_per_sample != 8) return JXL_ENC_NOT_SUPPORTED;
```
deymo, line 160: we should keep this check, this line would move inside the bits_per_sample==32 case; and for the 16 case we should have this but checking == 5. Those are the only float types supported by this implementation.

deymo (dismissed):
Looks good! I checked all the float_in usages and the LGTM.
Just one small comment to keep the exponent check.

lib/jxl/encode.cc
xiota: Ah... yes, moving it would make sense.  I'll make the change.

xiota: I'm updated the code to make sure 8 is paired with 32 and 5 is paired with 16.

deymo (approved):

xiota:
@deymo Thanks for the help.

PR #489, opened 2021-08-24
Add convenience flags to access verbose help

xiota:
This just adds `-vv`, `-vh`, and `-vvh` to make accessing verbose help more convenient.  Since help and verbosity flags are already handled independently of the options parser, it's reasonable to also add these flags outside of the options parser.

### DIFF
--- a/tools/cmdline.cc
+++ b/tools/cmdline.cc
@@ -53,9 +53,30 @@ bool CommandLineParser::Parse(int argc, const char* argv[]) {
       i++;
       continue;
     }
+    if (!strcmp("-hv", argv[i]) || !strcmp("-vh", argv[i])) {
+      help_ = true;
+      verbosity++;
+      i++;
+      continue;
+    }
+    if (!strcmp("-hvv", argv[i]) || !strcmp("-vvh", argv[i]) ||
+        !strcmp("-vhv", argv[i])) {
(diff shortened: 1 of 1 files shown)


jonsneyers:
It's nice to have that, but now I expect something like `cjxl -vmq 80` to also work.

Also the combinatorial explosion of handling all combinations of -v and -h is not so nice, it's OK with two -v's, but now I really don't want to add more obscure settings that require three -v's because then I need to change that code too to deal with more combinations.

Can this perhaps be rewritten in such a way that `-` followed by something that is not `-` (so it's not a long option) puts the command line parser in short option mode, and then single letters are matched repeatedly until the end of the arg?

xiota:
@jonsneyers   I don't know how to put the parser into short option mode.  Is there documentation for the parser I can look at?  But it looks to me like `v` and `h` are handled outside of the parser, so even if it were switched to short option mode, it wouldn't work with `vh`/`vvh`.

The patch could be simplified to just accepting `vh` and `vvh`.  The other combinations were just added for "completeness".  I doubt anyone would naturally type it as `vhv` for instance.

mo271:
The help for cjxl_ng will be handled in a different way, thanks for the input on this, but we don't plan on fixing this for the old cjxl now.

PR #490, opened 2021-08-24
Update Highway to 0.14.2

jan-wassenberg:


ziemek99:
I guess it should be closed in favor of #844...?

jan-wassenberg:
Yes, thank you :)

Issue #432, opened 2021-08-12
Lossy / lossless indication in JXL file


eddiezato:
**Is your feature request related to a problem? Please describe.**

A JXL file can represent several different things in terms of image compression. But there is no easy way to know how an image was encoded in JXL. Is it lossy or lossless?

**Describe the solution you'd like**

The JXL image should indicate the type of encoding process. This can be a flag in the metadata.

**Additional context**

If the user needs a bunch of lossless images for further editing, I think he would prefer to use another lossless format, such as PNG, instead of JXL, because PNG gives a clear understanding that images are compressed losslessly, which JXL lacks.

Or maybe I'm wrong, and such indicator already exists.

xiota:
It would be useful to have the library version and encoding parameters included in the metadata.  Also a tool for showing encoding internals.

To determine lossless vs lossy, I noticed that `jxlinfo` output for lossy files have two color profile sections "original" and "data", while lossless files don't seem to make that distinction. (`jxlinfo` is in the examples folder.  I don't know why it isn't made into a bona fide tool, along with `cjxl` and `djxl`.)

DamonHD:
Adding metadata for clients that don't need it simply detracts from the space-saving virtues of JXL.

eddiezato:
@DamonHD
1. Lossy/lossless flag can only take: 0 [lossless], 1 [lossy], 2 [jpeg]. Not so much, I guess.

2. Do "clients" need all these values?
```
PS > .\jxlinfo 1.jxl
xsize: 1920
ysize:  1080
have_container: 1
uses_original_profile: 0
bits_per_sample: 8
exponent_bits_per_sample: 0
intensity_target: 255.000000
min_nits: 0.000000
relative_to_max_display: 0
linear_below: 0.000000
have_preview: 0
have_animation: 0
orientation: 1
num_extra_channels: 0
alpha_bits: 0
alpha_exponent_bits: 0
alpha_premultiplied: 0
original color profile:
  format: JPEG XL encoded color profile
  color_space: 0
  white_point: 1
  white_point XY: 0.312700 0.329000
  primaries: 1
  red primaries XY: 0.639999 0.330010
  green primaries XY: 0.300004 0.600003
  blue primaries XY: 0.150002 0.059997
  transfer_function: 13
  rendering_intent: 1
data color profile:
  format: JPEG XL encoded color profile
  color_space: 0
  white_point: 1
  white_point XY: 0.312700 0.329000
  primaries: 1
[...]

PR #491, opened 2021-08-25
add color transform to frame header api struct

jonsneyers:
Split from https://github.com/libjxl/libjxl/pull/487

Adds a field to the frame header api struct so applications can see if frames are using XYB, RGB or YCbCr.

See also the discussion in https://github.com/libjxl/libjxl/issues/432

### DIFF
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -282,6 +282,18 @@ typedef struct {
   uint64_t extensions;
 } JxlHeaderExtensions;
 
+/** Color transform of a frame.
+    XYB: likely indicates that the frame is lossy.
+    RGB: no transform was used; possibly the frame was encoded in a lossless
+   way. YCBCR: JPEG's YCbCr transform was used, e.g. because it is a
+   recompressed JPEG.
+ */
+typedef enum {
+  JXL_XYB = 0,
+  JXL_RGB = 1,
+  JXL_YCBCR = 2,
+} JxlColorTransform;
+
 /** The header of one displayed frame. */
 typedef struct {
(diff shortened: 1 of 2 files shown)


lvandeve:
In theory applications don't need to see this, as the decoder converts the color to RGB.

On the other hand, this is helpful for the jxlinfo tool, but other fields could then also be helpful, such as whether it's modular or lossless, etc...

Perhaps there could be a different API to get this type of low level information though, with its own structs. This also in combination with API to get colors in XYB or other encoded values to allow to do the transform on GPU for example. WDYT?

deymo:
From the discussion in the other PR, I wanted to comment here that the main focus I proposed for the decoder API was to be an "image decoder" library; not a "JXL parser" library. We can make the argument form many things in the JXL file format that could be exposed in the API for the user, but my proposal is that if they are not needed for decoding an image or animation then we should avoid including them in the API. For example, should we expose an API to tell the user in which order are the groups stored on the file? we have that information on the header, it would be nice to have in a jxlinfo tool (to tell whether it is sequential or spiral, or custom... etc), it could maybe be useful for some use case; but regardless I don't think we should include it here. We have other priorities too.

A separate set of functions targeted more towards "parsing JXL files" where you can ask about all the internals could an option here, but I'd rather just don't have that now.

jonsneyers:
I think we can consider this information to be "internal info" that does not need to be exposed.

For the purpose of 'guessing' whether the image was encoded with lossy or lossless settings, the `uses_original_profile` BasicInfo field is just as good.

If we at some point want to allow e.g. yuv420 output in cases where that is possible, it should probably be done in a different way.

Closing this PR.

PR #493, opened 2021-08-25
Fix build for GIMP plugin with new compilers

deymo:
`!gimp_color_profile_is_gray` was converting the address of the function
to bool instead of calling it and then using the result. New compilers
actually have a warning for this.


sboukortt (approved):

xiota:
@deymo That was a copy/paste mistake.  It should have been referring to `!jxl_save_opts.is_gray`.

PR #496, opened 2021-08-25
Fix multi-frame container output from API

xiota:
Calling JxlEncoderAddImageFrame multiple times produces output that cannot be decoded with `djxl`.  The cause appears to be the omission of container structures between frames.


xiota:
agh... i got my git branches mixed up... will fix in a bit

Issue #484, opened 2021-08-23
problem to encode with grayscale ICC profile


novomesk:
I am trying to encode JXL file using C API.

I set basic info `uses_original_profile = JXL_TRUE;` and to set grayscale ICC profile via `JxlEncoderSetICCProfile` but I get following error during the process:

```
lib/jxl/enc_external_image.cc:113: JXL_FAILURE: Buffer size is too small
lib/jxl/enc_external_image.cc:328: JXL_RETURN_IF_ERROR code=1: ConvertFromExternal( jxl::Span<const uint8_t>(static_cast<const uint8_t*>(buffer), size), xsize, ysize, c_current, pixel_format.num_channels == 2 || pixel_format.num_channels == 4, false, bitdepth, pixel_format.endianness, false, pool, ib, float_in)
```

I used `num_color_channels = 1;` and pixel format `num_channels = 1;` and `bits_per_sample = 16;` and `data_type = JXL_TYPE_UINT16;`

Similar code with `uses_original_profile = JXL_FALSE;` and `JxlEncoderSetColorEncoding` works with grayscale images so I believe I calculate buffer size correctly.

When I want to save RGB ICC profile, everything works but my attempts to save grayscale ICC lead to failure.

The grayscale ICC profile is generated by GIMP.

xiota:
What ICC profile are you using?  Is it one of the default GIMP profiles or one you actually have a file for?

novomesk:
The default non-linear gray profile. But also with the gray linear.

I understand that only one of JxlEncoderSetColorEncoding and JxlEncoderSetICCProfile can be used.

xiota:
If it's the default gray profile from GIMP, you can call JxlColorEncodingSetToLinearSRGB(&color_encoding, gray) or JxlColorEncodingSetToSRGB(&color_encoding, gray) instead of attaching the ICC profile.

xiota:
Would you show the code you use to calculate the buffer size?

It looks like libjxl doesn't translate translate ICC profiles into the internal color encoding, so it doesn't know that the ICC profile is grayscale.  As a result, [checks in `image_bundle.cc` fail](https://github.com/libjxl/libjxl/blob/2681dc2f4d0a028a38d8cb1d48c38a24cce9e14a/lib/jxl/image_bundle.cc#L41).  Unfortunately, the workaround is to not use the ICC profile on grayscale images.

novomesk:
My code to calculate buffer size with 16bit depth is:
[...]

PR #498, opened 2021-08-26
Parse the ICC profile in JxlEncoderSetICCProfile

xiota:
JxlEncoderSetICCProfile attaches ICC profiles without parsing them.  This prevents a JxlColorEncoding from being set, which causes a default RGB profile to be used.  When a grayscale image is passed to the encoder, a JXL_CHECK error is normally raised.  However, the check is bypassed when `uses_original_profile=true`.  This causes unusual outputs to be produced (as with #484).

This change allows the encoder to extract information required for encoding from the ICC profile.  If the profile cannot be parsed, it is discarded, and a different profile may be tried, or JxlEncoderSetColorEncoding may be used instead.

This causes the roundtrip test to fail because it uses junk data instead of a valid ICC profile.  The test has been modified to use kEncodedTestProfile from `icc_codec_test.cc`.


xiota:
@sboukortt Any chance you can review this change?  I understand it would be a temporary fix until a better implementation of the underlying mechanics comes along.  Thanks.

sboukortt:
> @sboukortt Any chance you can review this change? I understand it would be a temporary fix until a better implementation of the underlying mechanics comes along. Thanks.

Oops, sorry, yes, I was meaning to wait for Alex’s comment to be resolved but then somehow didn’t notice that it was.

sboukortt (approved):

xiota:
@sboukortt Thank you for your help.

PR #499, opened 2021-08-26
implement brob box

jonsneyers:
Implements `brob` box support in the cjxl/djxl tooling. This still has to be migrated to the api (but for that we first need to decide on an api for it, and deal with streaming).

Obviously, brotli-compressing metadata can result in nice savings if this metadata is large.
For example, for this image:
https://www.iptc.org/std/photometadata/examples/IPTC-PhotometadataRef-Std2019.1.jpg

IPTC-PhotometadataRef-Std2019.1.jpg: 129915 bytes
No `brob` (before): cjxl recompresses to 103807 bytes
With `brob` (after): cjxl recompresses to 76713 bytes

### DIFF
--- a/tools/box/box.h
+++ b/tools/box/box.h
@@ -8,6 +8,9 @@
 #ifndef TOOLS_BOX_BOX_H_
 #define TOOLS_BOX_BOX_H_
 
+#include <brotli/decode.h>
+#include <brotli/encode.h>
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


tools/box/box.h
```diff
@@ -53,38 +56,122 @@ jxl::Status ParseBoxHeader(const uint8_t** next_in, size_t* available_in,
...
+          b.udata_size = tot_dec;
+        }
+        return &b;
+      }
+    }
+    return nullptr;
+  }
+
+  void addBlob(const char* type, const uint8_t* data, size_t data_size,
+               bool compress = true) {
```
lvandeve, line 158: I'm not sure if compress should be the default, maybe just don't have a default. While compressing is the best idea, there may also be cases where you want it uncompressed to be compatible with exif, jumb, ... tools that don't support this. Since addBlob looks generic for adding any type of box, it might also be used to add small application-specific boxes where compression wouldn't help. WDYT?

lvandeve (approved):

tools/box/box.h
jonsneyers: Makes sense, I'll remove the default. For now, I suppose there are no tools that can interpret compressed exif/xmp so it would probably be useful to have a cjxl option to produce uncompressed boxes, in case compatibility with existing tools is important.

tools/box/box.h
```diff
@@ -53,38 +56,122 @@ jxl::Status ParseBoxHeader(const uint8_t** next_in, size_t* available_in,
...
+    - in the compressed case, udata* points to the uncompressed input, and
+      when adding the box with addBlob(), pbytes will store the compressed
+      data, and cdata* will point to it.
+    - when writing, a "brob" box will be written if cdata* is nonzero,
+      otherwise an uncompressed box will be written.
+*/
+struct BrobBlob {
+  uint8_t type[4];
+  // uncompressed data (mutable: decompressed data can be added or removed)
+  mutable const uint8_t* udata = nullptr;  // Not owned
```
deymo, line 96: why is this mutable?

tools/box/box.h
```diff
@@ -53,38 +56,122 @@ jxl::Status ParseBoxHeader(const uint8_t** next_in, size_t* available_in,
...
-  // JUMBF superbox data, or null if not present in the container.
-  // The parsing of the nested boxes inside is not handled here.
-  const uint8_t* jumb = nullptr;  // Not owned
-  size_t jumb_size = 0;
-
+  std::vector<BrobBlob> blobs;
+
+  // Returns a pointer to a BrobBlob with uncompressed data (or nullptr if not
+  // found). In case there are multiple of that type, return the index-th one.
+  const BrobBlob* getBlob(const char* type, size_t index = 0) const {
```
deymo, line 112: Is there a reason to not have this in the .cc file instead? These should be CamelCase method names.

deymo (changes requested):
We need to migrate this implementation to the library. If you add this here it is just more work to re-implement this while migrating it to libjxl "without regressing". Before you expand on this implementation here I think it would be better to resolve the box handling in the library/api instead of increasing the workload for migrating it.

tools/box/box.h
jonsneyers: Because I want to be able to take a const BrobBlob and add (or I guess remove) its alternative, redundant representation (decompress it if it was compressed, compress it if it was uncompressed), which does change the struct but it doesn't conceptually change its semantical value, so it should be something you can do to const BrobBlobs.

Probably there's a cleaner way to do this, and if not, then probably it's a better idea to make those mutable fields private.

jonsneyers:
> We need to migrate this implementation to the library. If you add this here it is just more work to re-implement this while migrating it to libjxl "without regressing". Before you expand on this implementation here I think it would be better to resolve the box handling in the library/api instead of increasing the workload for migrating it.

I agree this needs to be migrated to the library, but do you want to postpone cjxl/djxl dealing with brob boxes until they are using the library? Because then we might have a situation where djxl cannot e.g. reconstruct a recompressed jpeg that has its XMP metadata brob'ed...

deymo:
We can have it in the C++ library in a way that *can* be exposed in the C API, even if djxl is using the internal C++ functions for this and everything else, taking into account that we should be able to build a decode-to-pixel version of the library without the need of brotli support. The C API already supports decoding to JPEG and doesn't include this library when doing that, so I don't think there's any blocker for replacing the jxl to jpeg workflow with something that uses the C API to the the processing; but even calling some internal decoder functions is fine. Implementing it here would bring the functionality to djxl in a way that's meant to deleted; and would not bring the functionality to the C API decoder which is the way meant to stay. What I'm proposing is that you implement it in the library.

jonsneyers:
OK, let's put this pull request on hold then until the libjxl api is decided.

jonsneyers:
Closing this pull request since it will become obsolete when cjxl/djxl start using the API.

PR #501, opened 2021-08-27
metadata api proposal

jonsneyers:
Proposal for an  encode/decode api for metadata blobs (Exif, XMP, JUMBF). This is just the api, no implementation yet.

Aim is to make compression of metadata completely transparent on the decode side: you always get the uncompressed metadata, no matter if it's e.g. an `xml ` box or a `brob` box containing compressed `xml ` data. On the encode side, you can choose whether to use compression or not.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -931,6 +931,29 @@ JXL_EXPORT size_t JxlDecoderReleaseJPEGBuffer(JxlDecoder* dec);
  */
 JXL_EXPORT JxlDecoderStatus JxlDecoderFlushImage(JxlDecoder* dec);
 
+/**
+ * Retrieves a metadata blob that is stored in the JPEG XL container.
+ * Returns the uncompressed blob in case it was Brotli-compressed.
+ * The data is owned by the decoder object; the returned pointer becomes invalid
+ * when the decoder is destroyed or reset.
+ *
+ * @param dec decoder object
+ * @param type 4-byte code representing the type of metadata. Examples are
+ * "Exif", "xml " (for XMP), and "jumb".
+ * @param index the index of the metadata blob of this type (there could be
+ * multiple blobs of the same type). Typically 0.
+ * @param data output pointer to the blob data
+ * @param size size in bytes of the blob data
+ * @return JXL_DEC_SUCCESS if the requested blob was found and successfully
+ * decompressed, JXL_DEC_ERROR if the requested blob was not found or an error
+ * happened during decompression, JXL_DEC_NEED_MORE_INPUT if not enough data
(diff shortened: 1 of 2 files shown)


jonsneyers:
To be clear on terminology: the container format is an instance of ISOBMFF (not JUMBF), which is a generic box format that is also used by mp4, jpeg 2000, etc.

JUMBF is a JPEG standard for extensible metadata (JPEG Universal Metadata Box Format), which is used e.g. for 360 degree images, and it can be embedded in various file formats, including jxl. As far as we are concerned, it's just an optional blob that is stored in a box called `jumb`.

There are already quite some tools out there that can parse ISOBMFF, and for things like uncompressed Exif metadata they might just work even if they don't know about jxl.

None of them at the moment know what a `brob` box is though.

To display an image, the jxl decoder only needs to find the `jxlc` box (or concatenate the partial `jxlp` boxes) and it can ignore anything else. It does not need Brotli in this case.

To reconstruct a JPEG bitstream from a jxl containing a recompressed jpg, the jxl decoder does need to find the `jbrd` box and also any boxes for Exif/XMP/JUMBF, and it needs to decompress them in case they are compressed in a `brob` box. It does need Brotli in any case, since the `jbrd` box uses Brotli compression internally.

To let applications access Exif/XMP/JUMBF metadata, it would be useful if our API also had methods for that. As to where to draw the line in terms of functionality: I think the line should be that we handle metadata as black boxes, just blobs. It's up to the application to know what the semantics of the blob data is. So I don't think we should have methods to access/modify specific Exif/XMP fields or things like that. I think the functions we need are

- get a blob (when decoding)
- add a blob (when encoding)
- have a way to write a jxl container without doing an encode (copying the codestream from an existing jxl file)

jonsneyers:
So, API-wise, I think the two functions of this proposal will be needed in any case.
It could also be useful to add two more functions:

- A decoder function to 'subscribe' to a subset of metadata blob types (e.g. `Exif` only), so a streaming decoder does not have to keep (all) metadata in memory. Another option is to use a callback that gets called whenever metadata is encountered.
- An encoder function to just take the codestream from an existing jxl file, which could be used to add/strip/modify metadata, or e.g. to join or split partial codestream boxes.
 
The last one needs some further thought: do we want to treat jxl codestreams as a black box here, or do we also want to do things like joining multiple jxl files into an animation / image sequence / layered image, or changing some image/frame header fields (e.g. doing things like reversible crop by changing the image dimensions / frame offsets, renaming layers or channels, adding or removing extra channels, removing or changing the ICC profile of an XYB image, sharpening/blurring by changing the gaborish weights, etc).

jonsneyers:
Closing this one, https://github.com/libjxl/libjxl/pull/552 is where the discussion continues.

PR #504, opened 2021-08-27
plugins/gimp: adjust gimp_register_magic_load_handler magic

### LINKED ISSUE #503: gimp plugin calls gimp_register_magic_load_handler with outdated signatures
https://github.com/libjxl/libjxl/blob/7333919b7d1851624e97348b9f10ad8af781b7ca/plugins/gimp/file-jxl.cc#L50-L51
should be adopted to newer stuff, like in https://github.com/libjxl/libjxl/blob/main/plugins/mime/image-jxl.xml

andrius4669:
Fixes #503.


andrius4669:
oh duh, fmt again. i though that it'd be fine because it formatted kinda similar way before..

andrius4669:
fixed format.

deymo (approved):

PR #509, opened 2021-08-28
Add basic animation support to encoder API

xiota:
To use, Set JxlBasicInfo `have_animation` and send frames with `JxlEncoderAddImageFrame`.  If have_animation is not set first, only the first frame will be encoded.  This is to prevent generating codestreams that cannot be decoded by either the API or `djxl`.

-----

In `encode.cc`, frames are queued with an empty FrameInfo.  When multiple frames are encoded, `djxl` cannot process the output file:

    ../lib/jxl/dec_file.cc:172: JXL_FAILURE: DecodeFile reader position not at EOF.

Setting `is_last` (depending on whether the queue is empty and `JxlEncoderCloseInput` has been called) prevents the error from occurring, but then only one frame is decoded.  Setting animation metadata allows both the API and `djxl` to decode all frames.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -42,8 +42,24 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
       std::move(input_frame_queue[0]);
   input_frame_queue.erase(input_frame_queue.begin());
 
-  // TODO(zond): If the frame queue is empty and the input_closed is true,
-  // then mark this frame as the last.
+  // If the frame queue is empty and input_closed is true,
+  // mark this frame as the last.
+  jxl::FrameInfo frame_info;
+  if (input_closed && input_frame_queue.empty()) {
+    frame_info.is_last = true;
+  } else {
+    frame_info.is_last = false;
+  }
+
+  // don't encode more than one non-animation frame,
+  // until the decode API can return them
+  // or the encode API can set processing parameters
+  if (!metadata.m.have_animation && !frame_info.is_last) {
+    input_closed = true;
+    input_frame_queue.clear();
+    frame_info.is_last = true;
+    return JXL_ENC_NOT_SUPPORTED;
+  }
 
   jxl::BitWriter writer;
 
@@ -91,7 +107,7 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
   }
 
   jxl::PassesEncoderState enc_state;
-  if (!jxl::EncodeFrame(input_frame->option_values.cparams, jxl::FrameInfo{},
+  if (!jxl::EncodeFrame(input_frame->option_values.cparams, frame_info,
                         &metadata, input_frame->frame, &enc_state,
                         thread_pool.get(), &writer,
                         /*aux_out=*/nullptr)) {
@@ -217,6 +233,17 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
     default:
       return JXL_ENC_ERROR;
       break;
+  }
+  enc->metadata.m.have_animation = info->have_animation;
+  if (info->have_animation) {
+    if (info->animation.tps_numerator < 1 ||
+        info->animation.tps_denominator < 1) {
+      return JXL_ENC_ERROR;
+    }
+    enc->metadata.m.animation.have_timecodes = info->animation.have_timecodes;
+    enc->metadata.m.animation.tps_numerator = info->animation.tps_numerator;
+    enc->metadata.m.animation.tps_denominator = info->animation.tps_denominator;
+    enc->metadata.m.animation.num_loops = info->animation.num_loops;
   }
   enc->metadata.m.xyb_encoded = !info->uses_original_profile;
   if (info->orientation > 0 && info->orientation <= 8) {
@@ -428,6 +455,12 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
     }
   } else {
     c_current = options->enc->metadata.m.color_encoding;
+  }
+
+  // needed until API can adjust per-frame animation headers
+  if (options->enc->metadata.m.have_animation) {
+    queued_frame->frame.duration = 10;
+    options->enc->metadata.m.animation.have_timecodes = false;
   }
 
   if (!jxl::BufferToImageBundle(*pixel_format, options->enc->metadata.xsize(),
--- a/examples/encode_oneshot.cc
+++ b/examples/encode_oneshot.cc
@@ -191,6 +191,8 @@ bool EncodeJxlOneshot(const std::vector<float>& pixels, const uint32_t xsize,
     return false;
   }
 
+  JxlEncoderCloseInput(enc.get());
+
   compressed->resize(64);
   uint8_t* next_out = compressed->data();
   size_t avail_out = compressed->size() - (next_out - compressed->data());


jonsneyers:
LGTM as a temporary solution.

Longer term we need some api to properly deal with multiple frames, that lets you do these things:

- choose if it's an animation (with some way to adjust tick duration) or not (multi-layer still)
- set frame durations
- set frame blend mode
- use frames that have different dimensions than the image dimensions and have an offset

(also we probably want to add some bounding box detection so you can pass image-sized frames to the encoder but it still encodes it as smaller frames if there are parts that don't change between frames, and possibly we should add patch heuristics to find usable patches from previous frames - none of that is currently done in the encoder)

xiota:
@jonsneyers  I'm fine with the change being "temporary" until better Multi-Frame/Animation API is implemented.  There are some "Get" functions on the decode side that could have "Set" analogs on the encode side.

* JXLDecoderGetExtraChannelInfo, JXLDecoderGetExtraChannelName, JXLDecoderGetFrameHeader, JXLDecoderGetFrameName
* ⇒ JXLEncoderSetFrameHeader, JXLEncoderSetFrameName, etc.

xiota:
@sboukortt Any chance you can have a look at this too?  Basically it implements a couple of TODOs from comments to prevent the API from generating bad multi-frame JXL output that cannot be decoded.  I understand it may be superseded in the future when a better multi-frame API is established.

lib/jxl/encode.cc
```diff
@@ -401,6 +409,12 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
     c_current = options->enc->metadata.m.color_encoding;
   }
 
+  // treat multiple frames as animation
+  // until animation API is established
+  options->enc->metadata.m.have_animation = true;
```
deymo, line 414: Shouldn't this be set to true only if this is not the first frame? Or is there something else that ignores the have_animation if there's only one frame? I think the same would apply to duration of previous frames.

deymo (dismissed):

lib/jxl/encode.cc
```diff
@@ -401,6 +409,12 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
...
+  options->enc->metadata.m.animation.tps_numerator = 100;
```
lvandeve, line 415: The encoder API has the function JxlEncoderSetBasicInfo, with which the user can specify the fields have_animation, tps_numerator, tps_denominator, etc... Rather than set it here, a better place would be to set these metadata fields in JxlEncoderSetBasicInfo (also the tps_denominator, num_loops and have_timecodes fields from JxlAnimationHeader)

It looks like now it would always set have_animation, even for a still image. With JxlEncoderSetBasicInfo, the user can already choose whether have_animation is enabled or not, so it should already be possible to choose between a still made up of multiple frames, or an actual animation, that way.

The frame header for frame.duration is not implemented yet in the encoder, so the duration = 10 as a temporary solution is ok, much better would of course be if the user can set a JxlFrameHeader per frame.

xiota: @lvandeve  I missed that JxlBasicInfo includes animation.  So here should set duration in response to whether animation is already set or not.  If animation is not set, what should happen when a second frame is sent to the encoder?  Should it be rejected (set first frame as the last frame)?

jonsneyers: When animation is not set, the duration of frames is not signaled, and it is implicitly zero. This is what you would do in case of a multi-layer still image.

xiota: @jonsneyers The decoders don't work when animation is not set.  The API decodes only one frame, and `djxl` gives an error.

jonsneyers: The API doesn't provide a way (at least not atm) to extract non-coalesced frames, like the separate layers of a composite still image, it will only return the blended/merged image (which is good for viewing apps that wouldn't know how to properly blend the layers themselves).
But if djxl gives an error then I suppose the encode api somehow produces an invalid bitstream in this case.

deymo: I think what's missing is to set duration != 0 in the encoder. That could easily be a setting when adding the frame.

lib/jxl/encode.cc
deymo: @xiota not just saying that if there's only one frame it shouldn't have_animation = true. Lode's point that this should actually be set by the basic info is relevant here.

xiota: @deymo I missed that basic info already has an animation property.  I'll change it to set the duration based on the setting from basic info.  What I don't understand now is how the API should behave when it receives multiple frames to encode and animation is set to false.  Currently, non-animation, multiframe codestreams are undecodable by `djxl`.  Should the API refuse to encode multiple frames when animation is not set (until `djxl` and the decode API are able to handle such codestreams)?

deymo: Either we return an error the second time you add a frame if have_animation is false; or we take that as a way to set multiple frames in a still image (you can define multiple internal JXL "frames" that are all meant to be blended in the decoder, like layers in photoshop). We can return an error for now and work on the "layer" support (both encoder and decoder) later. There are many blending modes available to internal frames so this is not as simple as adding multiple frames in the encoder api anyway.

deymo: Note that the issue is not that `djxl` doesn't understand those streams; we are most likely encoding something wrong in the encoder side in that case related to boxes. `djxl` is meant to blend all the frames of a still image (have_animation=false).

xiota:
@deymo @jonsneyers @lvandeve  I've made some changes.  It now behaves as follows:

* `JxlEncoderCloseInput` works as apparently intended.  `encode_oneshot.cc` had to be modified.
* `JxlEncoderSetBasicInfo` copies over animation settings from JxlBasicInfo.
* When `have_animation` is *not* set, only the first frame is encoded.  This is to avoid generating codestreams that currently cannot be decoded by the API or `djxl`.
* When `have_animation` *is* set, the encoder uses a duration of 10.  This is to avoid generating codestreams that the API cannot decode.

Also, I'm getting PR/format failures even though `./ci.sh lint` says it's okay.  Is there a `clang-format` setting I'm missing somewhere?

lib/jxl/encode.cc
```diff
@@ -40,10 +40,25 @@ uint32_t JxlEncoderVersion(void) {
...
+  // mark this frame as the last.
+  jxl::FrameInfo frame_info;
+  if (input_closed && input_frame_queue.empty()) {
+    frame_info.is_last = true;
+  } else {
+    frame_info.is_last = false;
+  }
+
+  // don't encode more than one non-animation frame,
+  // until API is able to decode them
```
lvandeve, line 56: What do you mean here by API unable to decode them? For non-animation frames, the API is able to decode them, they can for example be frames that are blended together to form a single still image. The decoder API only returns the final blended frame as intended by default.

The decode API doesn't currently have an option to return the individual (pre-blended) non-animation frames. Such option could be added (e.g. for more low-level decoding with access to this), but it's not the default option: the intended goal of multiple non-animation frames is to display a single still image result, so it behaves correctly.

lib/jxl/encode.cc
```diff
@@ -191,6 +207,21 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
       return JXL_ENC_ERROR;
       break;
   }
+  if (info->have_animation) {
+    enc->metadata.m.have_animation = true;
```
lvandeve, line 211: This would better be outside of the if and be:

enc->metadata.m.have_animation = info->have_animation

because if enc->metadata.m.have_animation is true, but user now uses SetBasicInfo with have_animation = false, enc->metadata.m.have_animation should be set to false again too.

lib/jxl/encode.cc
```diff
@@ -191,6 +207,21 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
...
+    enc->metadata.m.animation.have_timecodes = false;
```
lvandeve, line 212: This would better take the value from the user: enc->metadata.m.animation.have_timecodes = info->have_timecodes

lib/jxl/encode.cc
```diff
@@ -191,6 +207,21 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
...
+    if (info->animation.tps_numerator > 0) {
+      enc->metadata.m.animation.tps_numerator = info->animation.tps_numerator;
+    } else {
+      enc->metadata.m.animation.tps_numerator = 100;
```
lvandeve, line 216: this could return an error status, instead of choosing a default: require the user to set the tps_numerator correctly.

lib/jxl/encode.cc
lvandeve, line 212: the num_loops field should also be copied:

enc->metadata.m.animation.num_loops = info->animation.num_loops

lvandeve:
> Also, I'm getting PR/format failures even though `./ci.sh lint` says it's okay. Is there a `clang-format` setting I'm missing somewhere?

I don't currently see the format failures (but possibly it's just not showing them for some reason), but it lists failing tests currently. Possibly some break due to no longer supporting multiple frames without animation, but it should be possible to allow that support again (see the comment about blended frames)

```
The following tests FAILED:
	1406 - EncodeTest.FrameEncodingTest (ILLEGAL)
	1407 - EncodeTest.EncoderResetTest (ILLEGAL)
	1408 - EncodeTest.OptionsTest (ILLEGAL)
	1629 - RoundtripTest.Uint16FrameRoundtripTest (ILLEGAL)
	1632 - RoundtripTest.ExtraBoxesTest (ILLEGAL)
```

lib/jxl/encode.cc
xiota: I have it set to false because the API doesn't currently have a way to set timecodes in the frame header.

lib/jxl/encode.cc
xiota: The API and `djxl` behave differently.  `djxl` saves each frame to a separate file.

xiota:
@lvandeve I'll look at the test failures and make changes according to your suggestions.

lib/jxl/encode.cc
xiota: I changed the comment: until the decode API can return them or the encode API can set processing parameters

xiota:
@lvandeve  It looks like the cause of the failed tests is uninitialized JxlBasicInfo structures.  The values vary on subsequent runs:
```
have_animation = 32568
animation.have_timecodes = 32568
animation.num_loops = 1016503500
animation.tps_numerator = 38
animation.tps_denominator = 0

have_animation = 21913
animation.have_timecodes = 21913
animation.num_loops = 4285465532
animation.tps_numerator = 2237483616
animation.tps_denominator = 32765

have_animation = 4294967295
animation.have_timecodes = 4294967295
animation.num_loops = 4294967295
animation.tps_numerator = 4294967295
animation.tps_denominator = 4294967295
```

When I set have_animation = false before calls to `JxlEncoderSetBasicInfo`, the tests pass (when run locally, the PR build test isn't finished yet).  I think it would be good to initialize the values to some reasonable defaults, but I don't know where that is supposed to occur.

jonsneyers:
See https://github.com/libjxl/libjxl/pull/535

Uninitialized basic info was indeed a problem. You're now supposed to call `JxlEncoderInitBasicInfo` on the struct before setting its values, which should solve this.

xiota:
@jonsneyers  Thanks for the info.  I'll take a look at JxlEncoderInitBasicInfo.  Should the tests be modified to use it also?

jonsneyers:
> @jonsneyers Thanks for the info. I'll take a look at JxlEncoderInitBasicInfo. Should the tests be modified to use it also?

That has already been done in that pull request. Best to rebase this one (`git fetch upstream; git rebase upstream/main`).

xiota:
@jonsneyers I've rebased and run the tests locally. Looks good.  Thanks.

lib/jxl/encode.cc
```diff
@@ -42,8 +42,23 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
...
+  // If the frame queue is empty and input_closed is true,
+  // mark this frame as the last.
+  jxl::FrameInfo frame_info;
+  if (input_closed && input_frame_queue.empty()) {
+    frame_info.is_last = true;
+  } else {
+    frame_info.is_last = false;
+  }
+
+  // don't encode more than one non-animation frame,
```
jonsneyers, line 54: Is this part still needed? It's of course not really useful yet to do non-animation multi-frame when we can't set blend modes / cropped frames yet, but it should work I think (it will just be full frames using whatever the default blend mode is, which is probably not useful).

xiota: I don't know if it's needed.  But it prevents users from creating files with frames they cannot recover yet.

jonsneyers: Fair enough, but maybe then return `JXL_ENC_NOT_SUPPORTED` in cases where someone is trying to produce a multi-layer file, instead of silently forcing the input to be closed after the first layer?

We don't want to have applications rely on behavior (implicit closing of input in case of !have_animation) that we know will change at some point.

xiota: Okay.  Will make that change.

lib/jxl/encode.cc
```diff
@@ -430,6 +457,12 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
     c_current = options->enc->metadata.m.color_encoding;
   }
 
+  // needed until API can adjust per-frame animation headers
+  if (options->enc->metadata.m.have_animation) {
+    queued_frame->frame.duration = 10;
+    options->enc->metadata.m.animation.have_timecodes = false;
+  }
+
```
jonsneyers, line 465: Instead of hardcoding all frames to have duration 10, could you propose a new function to set the per-frame info, e.g. a `JxlEncoderSetFrameHeader`?

It's OK if for now it only looks at the duration and is_last (there can be more frame header fields later, like the things I'm proposing to add in https://github.com/libjxl/libjxl/pull/831), but I think that would be a cleaner way to do it than this somewhat ad-hoc solution of a hardcoded duration 10 and deriving is_last from the frame queue being empty or not. If there's an explicit SetFrameHeader function (which would be optional and you would get a default frame header if you don't call it, i.e. is_last=true, duration=0, etc), then you can also encode one frame or a few frames with is_last=false without having to queue _all_ the frames — for long animations, you probably don't want to have to queue all frames simultaneously...

jonsneyers:
This one will be superseded by https://github.com/libjxl/libjxl/pull/909 so maybe it's better to move the discussion there and close this one?

xiota:
superseded by #909

PR #511, opened 2021-08-29
Add instructions to build on Windows with MSYS2

xiota:
An important point about MSYS2 that I did not see in other guides is that all package management tasks have to be done within the MSYS environment.  Otherwise, the other build environments will become unusable, and MSYS2 may have to be reinstalled.  If care is taken while setting up the build environment, `libjxl` can be built for 32- and 64-bit Windows, using the MSVCRT and UCRT runtimes, with the gcc/g++ or clang compilers.


doc/developing_in_windows_msys.md
```diff
@@ -0,0 +1,140 @@
...
+
+To remove packages, use:
+
+    pacman -Rsc [package-name]
+
+## Other Environments
+
+MSYS2 provides multiple development [environments](https://www.msys2.org/docs/environments/), which are layers on top of the MSYS environment.  They target slightly different platforms, runtime libraries, and compiler toolchains.  Remember to always make package changes from within the MSYS environment.  Then `exit` all MSYS2 environments and restart the desired build environment.
+
+* **MINGW32:**  To develop for 32-bit Windows, replace `x86_64` with `i686` in the package names.
```
sboukortt, line 64: Do you mean Mingw-w64?

doc/developing_in_windows_msys.md
```diff
@@ -0,0 +1,140 @@
...
+
+Start the MINGW64 environment, create a build directory within the source directory, and configure with `cmake`.
+```
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+   -DBUILD_TESTING=OFF -DJPEGXL_ENABLE_BENCHMARK=OFF \
+   -DJPEGXL_ENABLE_PLUGINS=ON -DJPEGXL_ENABLE_MANPAGES=OFF \
+   -DJPEGXL_FORCE_SYSTEM_BROTLI=ON \
+   -DJPEGXL_FORCE_SYSTEM_GTEST=ON ..
```
sboukortt, line 82: Also `-G Ninja`, right?

doc/developing_in_windows_msys.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Other Environments
+
+MSYS2 provides multiple development [environments](https://www.msys2.org/docs/environments/), which are layers on top of the MSYS environment.  They target slightly different platforms, runtime libraries, and compiler toolchains.  Remember to always make package changes from within the MSYS environment.  Then `exit` all MSYS2 environments and restart the desired build environment.
+
+* **MINGW32:**  To develop for 32-bit Windows, replace `x86_64` with `i686` in the package names. 
+
+* **UCRT64:**  The Universal C Runtime (UCRT) is used by recent versions of Microsoft Visual Studio.  It ships by default with Windows 10.  For older versions of Windows, it must be provided with the application or installed by the user.  To use the UCRT64 environment, follow the above instructions, but use the `mingw-w64-ucrt-x86_64` prefix in the package names.
+
+* **CLANG64:** Unfortunately, the `gimp` packages are not available for the CLANG64 environment.  However, `libjxl` will otherwise build in this environment if the appropriate packages are installed.  Packages are prefixed with `mingw-w64-ucrt64-x86_64`.
```
sboukortt, line 68:
```suggestion
* **CLANG64:** Unfortunately, the `gimp` packages are not available for the CLANG64 environment.  However, `libjxl` will otherwise build in this environment if the appropriate packages are installed.  Packages are prefixed with `mingw-w64-clang64-x86_64`.
```

sboukortt (dismissed):
This is great, thanks a lot!

doc/developing_in_windows_msys.md
xiota: I was using the 64-bit version of MSYS2.  I don't know if the package names are different on the 32-bit version.  I'll rewrite to clarify.

doc/developing_in_windows_msys.md
xiota: I didn't use `-G Ninja`.  When I run `cmake -G`, it has Ninja marked as the default.  I'll add a note that it may be useful to try if it doesn't work.

doc/developing_in_windows_msys.md
xiota: Thanks.  Will fix.

doc/developing_in_windows_msys.md
sboukortt: Sorry, you are right, the package group is indeed called mingw32/mingw64. I think I may have been thrown off by the upper case.

sboukortt (approved):

doc/developing_in_windows_msys.md
xiota: I refer to the build environment in uppercase (which link to click in the Start Menu).  The group names look like the build environment names, but in lower case.  Then there's the prefix and rest of the package name.  I'll see if I can add an explanation to clarify, since confusing the differences contributed to difficulties building libjxl on MSYS2.

sboukortt (approved):

PR #516, opened 2021-08-30
Remove the current GdkPixbuf plugin

sboukortt:


veluca93 (approved):

jonsneyers:
Can we first have an alternative implementation and then remove?

deymo (approved):

deymo:
I'm ok to remove now

xiota:
Could the headers be restored (from the FLIF version) and the plugin be sent upstream to the GTK folks?  It's only really useful when combined with their code anyway.

deymo:
It is not a lot of work to write a new one from scratch anyway.

veluca93:
I believe given #519 we don't need this anymore.

PR #519, opened 2021-08-30
Rewrite the GDK-Pixbuf plugin.

### LINKED ISSUE #512: Licensing Complications for the GDK Pixbuf Loader
The GDK Pixbuf Loader located at `plugins/gdk-pixbuf/pixbufloader-jxl.c` looks *remarkably* similar to the one in the FLIF repository, which I wrote. You can find it here: https://github.com/FLIF-hub/FLIF/blob/master/src/flif-pixbuf-loader.c

I do not mind this being used in the reference software, but there are some potential caveats that need to be considered.
- First, my code is not written from scratch. It, itself, is a derivative of the WebP version, which can be found here. https://github.com/aruiz/webp-pixbuf-loader
- That code, and also my derivative, is licensed under the GNU Lesser General Public License, version 2 or later.
- I signed a CLA which preserves the license and copyright of my work, but grants permission for you to use it. https://github.com/FLIF-hub/FLIF-CLAs/blob/master/signed/thebombzen-cla.txt

The GDK Pixbuf loader itself is a dynamic library loaded by `dlopen(3)` so licensing only this component under LGPLv2 or later should not break any existing code, but I'm not certain because I Am Not A Lawyer. I would also be happy to re-license the code to match the JXL project license, or even transfer or relinquish copyright, but IANAL and I am not sure if the previous authors' permission would need to be requested to do that.

This is what open source is about - sharing code, and I'm pretty happy to see this here, but I'm just not sure how the licensing works with regard to this. Thoughts?

veluca93:
The new version supports ICC profiles in the input JXL file, real
incremental loading and multithreading.

Pair programmed with Moritz.

Still TODO are animation and saving.

Fixes #512.

### DIFF
--- a/plugins/gdk-pixbuf/CMakeLists.txt
+++ b/plugins/gdk-pixbuf/CMakeLists.txt
@@ -15,7 +15,7 @@ endif ()
 add_library(pixbufloader-jxl SHARED pixbufloader-jxl.c)
 # Note: This only needs the decoder library, but we don't install the decoder
 # shared library.
-target_link_libraries(pixbufloader-jxl jxl jxl_threads PkgConfig::Gdk-Pixbuf)
+target_link_libraries(pixbufloader-jxl jxl jxl_threads skcms-interface PkgConfig::Gdk-Pixbuf)
 
 pkg_get_variable(GDK_PIXBUF_MODULEDIR gdk-pixbuf-2.0 gdk_pixbuf_moduledir)
 install(TARGETS pixbufloader-jxl LIBRARY DESTINATION "${GDK_PIXBUF_MODULEDIR}")
--- a/plugins/gdk-pixbuf/pixbufloader-jxl.c
+++ b/plugins/gdk-pixbuf/pixbufloader-jxl.c
@@ -7,254 +7,262 @@
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #undef GDK_PIXBUF_ENABLE_BACKEND
 
+#include "jxl/codestream_header.h"
 #include "jxl/decode.h"
-
-static void DestroyPixels(guchar *pixels, gpointer data) { free(pixels); }
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/types.h"
+#include "skcms.h"
 
 typedef struct {
-  GdkPixbufModuleSizeFunc size_func;
-  GdkPixbufModuleUpdatedFunc update_func;
-  GdkPixbufModulePreparedFunc prepare_func;
+  GdkPixbufModuleSizeFunc image_size_callback;
+  GdkPixbufModulePreparedFunc pixbuf_prepared_callback;
+  GdkPixbufModuleUpdatedFunc area_updated_callback;
   gpointer user_data;
-  GdkPixbuf *pixbuf;
-  GError **error;
 
-  FILE *increment_buffer;
-  char *increment_buffer_ptr;
-  size_t increment_buffer_size;
+  GdkPixbuf *output;
 
-} JxlContext;
+  JxlParallelRunner *parallel_runner;
+  JxlDecoder *decoder;
+  JxlPixelFormat pixel_format;
 
-uint8_t *JxlMemoryToPixels(const uint8_t *next_in, size_t size, size_t *stride,
-                           size_t *xsize, size_t *ysize, int *has_alpha) {
-  JxlDecoder *dec = JxlDecoderCreate(NULL);
-  *has_alpha = 1;
-  uint8_t *pixels = NULL;
-  if (!dec) {
-    fprintf(stderr, "JxlDecoderCreate failed\n");
-    return 0;
+  gboolean done;
+  gboolean alpha_premultiplied;
+
+  gpointer icc_buff;
+  skcms_ICCProfile icc;
+
+} GdkPixbufJXLDecoderState;
+
+static gpointer begin_load(GdkPixbufModuleSizeFunc size_func,
+                           GdkPixbufModulePreparedFunc prepare_func,
+                           GdkPixbufModuleUpdatedFunc update_func,
+                           gpointer user_data, GError **error) {
+  GdkPixbufJXLDecoderState *decoder_state =
+      g_malloc0(sizeof(GdkPixbufJXLDecoderState));
+  decoder_state->image_size_callback = size_func;
+  decoder_state->pixbuf_prepared_callback = prepare_func;
+  decoder_state->area_updated_callback = update_func;
+  decoder_state->user_data = user_data;
+
+  if (!(decoder_state->parallel_runner =
+            JxlResizableParallelRunnerCreate(NULL))) {
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_FAILED,
+                "Creation of the JXL parallel runner failed");
+    goto cleanup;
   }
-  if (JXL_DEC_SUCCESS !=
-      JxlDecoderSubscribeEvents(dec, JXL_DEC_BASIC_INFO | JXL_DEC_FULL_IMAGE)) {
-    fprintf(stderr, "JxlDecoderSubscribeEvents failed\n");
-    JxlDecoderDestroy(dec);
-    return 0;
+
+  if (!(decoder_state->decoder = JxlDecoderCreate(NULL))) {
+    g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_FAILED,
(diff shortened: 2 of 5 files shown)


plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,257 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+#include "jxl/codestream_header.h"
 #define GDK_PIXBUF_ENABLE_BACKEND
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #undef GDK_PIXBUF_ENABLE_BACKEND
 
+#include "../../third_party/skcms/skcms.h"
```
deymo, line 11: I don't think this will work with the shared lib.

plugins/gdk-pixbuf/CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ if (NOT Gdk-Pixbuf_FOUND)
   return ()
 endif ()
 
-add_library(pixbufloader-jxl SHARED pixbufloader-jxl.c)
+# TODO: nope!
```
lvandeve, line 15: this TODO could have a better description

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,257 @@
...
-  size_t data_size;
-  int status;
-  gpointer data;
+static gpointer begin_load(GdkPixbufModuleSizeFunc size_func,
+                           GdkPixbufModulePreparedFunc prepare_func,
+                           GdkPixbufModuleUpdatedFunc update_func,
+                           gpointer user_data, GError **error) {
+  GdkPixbufJXLDecoderState *decoder_state =
+      g_malloc(sizeof(GdkPixbufJXLDecoderState));
+  memset(decoder_state, 0, sizeof(*decoder_state));
```
thebombzen, line 42: Is it necessary to zero this memory first if you immediately populate it?

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,257 @@
...
+  JxlDecoderDestroy(decoder_state->decoder);
+  return NULL;
+}
 
-  if (context->update_func) {
-    (*context->update_func)(
-        context->pixbuf, 0, 0, gdk_pixbuf_get_width(context->pixbuf),
-        gdk_pixbuf_get_height(context->pixbuf), context->user_data);
+static gboolean stop_load(gpointer context, GError **error) {
+  GdkPixbufJXLDecoderState *decoder_state = context;
```
thebombzen, line 90: You should explicitly cast this here. Implicit casting of `void *` to anything else will fail if you compile this with a C++ compiler.

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: Well, it's not a lot of memory - but it's there for (a) initializing some struct members that are not initialized otherwise (no member initializers in C...) and (b) because in case of early failures, the code in the cleanup goto will free up NULLs rather than uninitialized memory (which would be bad)

thebombzen: If it is, then `g_malloc0()` does exactly this. Calls `calloc()` internally.

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: Compiling this with a C++ compiler will do Very Bad Things, such as mangle the fill_info and fill_vtable functions - this will result in a non-usable plugin, so I prefer failing at compile time :)

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: Ah, very useful to know, thanks! Done.

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,256 @@
...
 
-  context->pixbuf =
-      gdk_pixbuf__jxl_image_load(context->increment_buffer, error);
-
-  gint width = gdk_pixbuf_get_width(context->pixbuf);
-  gint height = gdk_pixbuf_get_height(context->pixbuf);
-  if (context->size_func) {
-    context->size_func(&width, &height, context->user_data);
-  }
+  decoder_state->pixel_format.data_type = JXL_TYPE_FLOAT;
```
lvandeve, line 78: For performance, FLOAT16 could be an option if the skcms usage below supports it. Another option is e.g. uint8 if you know that it's srgb or can be converted to srgb (setting it as preferred color space), uint16 or float otherwise.

If using full float buffers doesn't affect performance much, then, disregard this comment

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,256 @@
...
+        decoder_state->area_updated_callback(
+            decoder_state->output, 0, 0,
+            gdk_pixbuf_get_width(decoder_state->output),
+            gdk_pixbuf_get_height(decoder_state->output),
+            decoder_state->user_data);
+        // TODO(veluca): for animations, this will eventually only leave the
+        // last frame.
+        break;
+      }
+      default:;
```
lvandeve, line 245: the default case could instead return an error for the unexpected result

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: This uses the callback, so performance shouldn't be very different between f16 and f32 - probably slightly faster for f32 as f16 would need an extra intermediate conversion.

plugins/gdk-pixbuf/pixbufloader-jxl.c
lvandeve: Also, JXL_DEC_ERROR should be handled explicitely, and return FALSE.

veluca93: Done.

.github/workflows/build_test.yml
```diff
@@ -161,6 +161,7 @@ jobs:
         ./ci.sh test
 
 
+
```
deymo, line 164: :)

third_party/skcms.cmake
```diff
@@ -15,10 +15,15 @@
 add_library(skcms-obj OBJECT EXCLUDE_FROM_ALL skcms/skcms.cc)
 target_include_directories(skcms-obj PUBLIC "${CMAKE_CURRENT_LIST_DIR}/skcms/")
 
+add_library(skcms-interface INTERFACE)
```
deymo, line 18: A comment here about what's this lib would be appreciated by future you. Mention that the skcms-interface is meant to be compiled/used by external project (like the plugins) and not the bundled version in libjxl which uses the wrapper.

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,265 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+#include "jxl/codestream_header.h"
```
deymo, line 6: move this down with the rest?

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -3,258 +3,265 @@
...
+  decoder_state->pixel_format.endianness = JXL_NATIVE_ENDIAN;
 
-  if (context->prepare_func) {
-    (*context->prepare_func)(context->pixbuf, NULL, context->user_data);
-  }
+  return decoder_state;
+cleanup:
+  JxlResizableParallelRunnerDestroy(decoder_state->parallel_runner);
+  JxlDecoderDestroy(decoder_state->decoder);
+  return NULL;
```
deymo, line 85: you need to free `decoder_state` in this case, right?

deymo (approved):
sgtm

.github/workflows/build_test.yml
veluca93: Ups. Done.

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: Indeed...

deymo (approved):

PR #521, opened 2021-08-31
tweak size header encoding (512x256 can be 'small' too)

jonsneyers:
Silly little tweak in the encoding of SizeHeader: in some cases, even if the width is > 256, we can still use the shorter `small` encoding because the width can be derived from the height using the aspect ratio encoding.

This can save 1 byte for a 512x256 image.

### DIFF
--- a/lib/jxl/headers.cc
+++ b/lib/jxl/headers.cc
@@ -63,15 +63,15 @@ Status SizeHeader::Set(size_t xsize64, size_t ysize64) {
   const uint32_t xsize32 = static_cast<uint32_t>(xsize64);
   const uint32_t ysize32 = static_cast<uint32_t>(ysize64);
   if (xsize64 == 0 || ysize64 == 0) return JXL_FAILURE("Empty image");
-  small_ = xsize64 <= 256 && ysize64 <= 256 && (xsize64 % kBlockDim) == 0 &&
-           (ysize64 % kBlockDim) == 0;
+  ratio_ = FindAspectRatio(xsize32, ysize32);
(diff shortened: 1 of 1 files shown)


lib/jxl/headers.cc
```diff
@@ -63,15 +63,15 @@ Status SizeHeader::Set(size_t xsize64, size_t ysize64) {
   const uint32_t xsize32 = static_cast<uint32_t>(xsize64);
   const uint32_t ysize32 = static_cast<uint32_t>(ysize64);
   if (xsize64 == 0 || ysize64 == 0) return JXL_FAILURE("Empty image");
-  small_ = xsize64 <= 256 && ysize64 <= 256 && (xsize64 % kBlockDim) == 0 &&
-           (ysize64 % kBlockDim) == 0;
+  ratio_ = FindAspectRatio(xsize32, ysize32);
+  small_ = ysize64 <= 256 && (ysize64 % kBlockDim) == 0 &&
+           (ratio_ > 0 || (xsize64 <= 256 && (xsize64 % kBlockDim) == 0));
```
veluca93, line 68: should we use != 0 instead? I think it's clearer.

veluca93 (dismissed):

lib/jxl/headers.cc
jonsneyers: I am always tempted to write just `ratio_` in cases like this, but I guess that's bad style :)

In my head there's no difference between > 0 and != 0 for a uint, but if != 0 looks clearer to you, let's do that.

veluca93 (approved):

Issue #522, opened 2021-08-31
To support kThunder and kLightning in the JxlEncoderOptionsSetEffort function


sim9108:
Coding modifications are required to support kThunder and kLightning in the JxlEncoderOptionsSetEffort function.

			kThunder = 2,
			kLightning = 1

JxlEncoderStatus JxlEncoderOptionsSetEffort(JxlEncoderOptions* options,
                                            const int effort) {
  if (effort < 3 || effort > 9) {  //==> if (effort <1 || effort > 9)
    return JXL_ENC_ERROR;
  }
  options->values.cparams.speed_tier = static_cast<jxl::SpeedTier>(10 - effort);
  return JXL_ENC_SUCCESS;
}

**Screenshots**
If applicable, add screenshots or example input/output images to help explain your problem.

**Environment**
 - OS: [e.g. Windows]
 - Compiler version: [e.g. clang 11.0.1]
 - CPU type: [e.g. x86_64]
 - cjxl/djxl version string: [e.g. cjxl [v0.3.7 | SIMD supported: SSE4,Scalar]]

**Additional context**

jonsneyers:
Thanks for the catch, should be fixed as soon as the CI bots are done processing stuff.

PR #523, opened 2021-08-31
allow effort 1,2 in encode api

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/522


veluca93 (dismissed):
That was fast!

jonsneyers:
Ugh, we actually tested that effort 2 has to fail. Updated the test.

veluca93 (approved):

PR #525, opened 2021-08-31
GIMP Plugin Updates and Fixes

xiota:
* Reorganize save dialog
 * Rename functions, modify messages
 * Use JxlEncoderSetBasicInfo
 * Use float as intermediary format
 * Fix float image alpha channels
 * Fix grayscale ICC profile loading
 * Fix distance/quality conversion

screenshot

### DIFF
--- a/plugins/gimp/file-jxl-save.cc
+++ b/plugins/gimp/file-jxl-save.cc
@@ -23,6 +23,7 @@ class JpegXlSaveOpts {
   bool is_linear = false;
   bool has_alpha = false;
   bool is_gray = false;
+  bool icc_attached = false;
 
   bool advanced_mode = false;
   bool use_container = true;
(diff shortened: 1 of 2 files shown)


jonsneyers:
Thanks for working on this!

Some nitpicking remarks/suggestions, just looking at the screenshot:

- It might be good to annotate the sliders to indicate what they mean, e.g. write something like "higher fidelity" on the left of the "Distance" slider and "lower fidelity" on the right, or "faster" - "slower" for the "Effort" slider.
- It could also be a good idea, if possible, to draw some annotated 'ticks' on the slider to give further guidance, for example putting words like "excellent", "very good", "good", "fair", "poor" at points corresponding to, say, d0.3, d1, d2, d3, d6, or something like that. Photoshop does that for its jpeg export dialog, and I think it is useful since many users will not know what to expect from specific numeric values.
- Perhaps "Effort" could be considered an advanced setting, and just having the default 7 for lossy and 3 for lossless is good enough in most cases? Keeping it simple by default is maybe better.
- What does "Use Original Color Profile" mean? This is not clear to me. The color profile of the image should always be "used" in some way - but I suppose this refers to the corresponding API field, which I think is used to signal whether XYB can be used or not. I think this option should always be disabled when doing lossy, and always enabled when doing lossless (so I don't see a need to put it as an explicit option in the UI).

xiota:
@jonsneyers Thank you for the feedback.

* I would appreciate suggestions to improve the explanatory text in the tool tips.  They don't show up in screenshots, but can significantly affect users' ability to learn and use the interface.

* The GIMP-scale-entry widget doesn't seem to have the annotation features you describe.  I'll look into what else is available for a future revision.

* In the previous version, Lossless Mode and Effort were both in the Advanced section.  I moved Lossless Mode to the standard section because normal users can be expected to use it (when they might otherwise use PNG).  Since lossless affects the Distance, Quality, and Effort sliders, users could be confused if the settings were separated in the interface.  Users may also want to change Effort to match their computers' capabilities.

  Maybe a different label would help, like "Encoding Speed" with values subtracted from 10?  Then "Faster Decoding" could be renamed to "Decoding Speed".  They wouldn't match the `cjxl` options anymore though.

* "Use Original Color Profile" does refer to the similarly named API option.  This is an option I expect some photographers might require.  Perhaps it could be renamed something like "Do not use XYB".  I understand that XYB is supposed to be a major feature of JXL, but right now, it seems to hinder the use of color spaces other than sRGB (at least with the API).  It is in the Advanced section to discourage use unless users really feels it's needed.

veluca93:
> @jonsneyers Thank you for the feedback.
> 
> * I would appreciate suggestions to improve the explanatory text in the tool tips.  They don't show up in screenshots, but can significantly affect users' ability to learn and use the interface.
> * The GIMP-scale-entry widget doesn't seem to have the annotation features you describe.  I'll look into what else is available for a future revision.
> * In the previous version, Lossless Mode and Effort were both in the Advanced section.  I moved Lossless Mode to the standard section because normal users can be expected to use it (when they might otherwise use PNG).  Since lossless affects the Distance, Quality, and Effort sliders, users could be confused if the settings were separated in the interface.  Users may also want to change Effort to match their computers' capabilities.
>   Maybe a different label would help, like "Encoding Speed" with values subtracted from 10?  Then "Faster Decoding" could be renamed to "Decoding Speed".  They wouldn't match the `cjxl` options anymore though.

Just my 2c here: I don't think matching `cjxl` options is particularly important :)

> * "Use Original Color Profile" does refer to the similarly named API option.  This is an option I expect some photographers might require.  Perhaps it could be renamed something like "Do not use XYB".  I understand that XYB is supposed to be a major feature of JXL, but right now, it seems to hinder the use of color spaces other than sRGB (at least with the API).  It is in the Advanced section to discourage use unless users really feels it's needed.

jonsneyers:
> * "Use Original Color Profile" does refer to the similarly named API option.  This is an option I expect some photographers might require.  Perhaps it could be renamed something like "Do not use XYB".  I understand that XYB is supposed to be a major feature of JXL, but right now, it seems to hinder the use of color spaces other than sRGB (at least with the API).  It is in the Advanced section to discourage use unless users really feels it's needed.

Maybe something is going wrong then - XYB should work well with any color space, not just sRGB.

Ah but passing the input to the encoder has to be done in a color space that the encoder understands, I suppose.

@lvandeve @sboukortt do we (plan to) have an encode api for input with an arbitrary ICC profile, where the encoder does the necessary conversion? Like what cjxl does...

plugins/gimp/file-jxl-save.cc
```diff
@@ -194,69 +186,99 @@ bool SaveDialog() {
...
-  table = gtk_table_new(1, 3, false);
+  // Distance / Quality Table
+  table = gtk_table_new(20, 3, false);
   gtk_table_set_col_spacings(GTK_TABLE(table), 6);
-  gtk_container_add(GTK_CONTAINER(frame), table);
+  gtk_box_pack_start(GTK_BOX(vbox), table, false, false, 0);
   gtk_widget_show(table);
 
+  static gchar distance_help[] =
+      "Butteraugli distance.  Use lower values for higher quality.";
```
jonsneyers, line 206: Technically, it's a distance _target_, and the encoder doesn't necessarily hit the target (especially at low effort).

xiota:
@jonsneyers  The ICC profile issues I'm having may just be how it's currently implemented.  As far as I can tell, the ICC profile isn't really used.  For instance, the roundtrip test had used junk data instead of a real ICC profile.  If the encoder used the ICC profile, I'd expect junk data would have failed the test.

plugins/gimp/file-jxl-save.cc
xiota: How about:
```
Butteraugli distance target.  Lower values indicate higher quality.
   d = 0.3   Excellent
   d = 1     Very Good
   d = 2     Good
   d = 3     Fair
   d = 6     Poor
```

lvandeve:
> @lvandeve @sboukortt do we (plan to) have an encode api for input with an arbitrary ICC profile, where the encoder does the necessary conversion? Like what cjxl does...

I don't think so, at least nothing more than the reverse of what the decoder can do without a CMS

jonsneyers:
> @jonsneyers The ICC profile issues I'm having may just be how it's currently implemented. As far as I can tell, the ICC profile isn't really used. For instance, the roundtrip test had used junk data instead of a real ICC profile. If the encoder used the ICC profile, I'd expect junk data would have failed the test.

If I understand correctly, if you have `uses_original_profile` set to false, then the ICC profile will just be ignored by the encoder and the pixels will be assumed to be sRGB (if ints are used) or linear sRGB (if floats are used). It will store the bytes of the ICC profile in the file, but they're just a blob that is only informative and doesn't influence rendering. If you have `uses_original_profile` set to true, then the ICC profile will not be ignored by the encoder (to do the conversion to XYB).
So it should basically always be set to true, I think.

@lvandeve / @sboukortt can you comment? I don't really fully understand how the api works here...

jonsneyers:
> > @lvandeve @sboukortt do we (plan to) have an encode api for input with an arbitrary ICC profile, where the encoder does the necessary conversion? Like what cjxl does...
> 
> I don't think so, at least nothing more than the reverse of what the decoder can do without a CMS

So we really want (the full) `libjxl` not to depend on a CMS library?

For the decode library (`libjxl_dec`) I understand that we want to keep it minimal in terms of dependencies and binary size, but for the full library I don't think that's a big concern.


Anyway, does that mean that as it currently is, when doing lossy, the gimp encoder plugin would need to convert (whatever space the pixels are in) to linear sRGB floats, and call the encoder with uses_original_profile=false? And when doing lossless, it can pass the pixels directly and set uses_original_profile=true?

plugins/gimp/file-jxl-save.cc
jonsneyers: Yes, something like that makes sense to me. Of course such mappings from numbers to words are always a bit subjective and use case dependent.

xiota: "Butteraugli distance target.  Suggested starting values..."

plugins/gimp/file-jxl-save.cc
```diff
@@ -85,82 +85,77 @@ static bool gui_on_change_distance(GtkAdjustment* adj_dist,
...
-    jxl_save_opts.use_container = true;
-    gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(advanced_opts[2]), false);
-
     jxl_save_opts.basic_info.uses_original_profile = false;
-    gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(advanced_opts[3]), false);
+    gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(advanced_opts[1]), false);
 
     // save metadata
-    // gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(advanced_opts[4]), false);
+    // gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(advanced_opts[2]), false);
```
veluca93, line 152: Is this line meant to be uncommented, or is it not supposed to be here?

plugins/gimp/file-jxl-save.cc
```diff
@@ -194,69 +190,108 @@ bool SaveDialog() {
...
+                   G_CALLBACK(GuiOnChangeDistance), entry_quality);
   g_signal_connect(entry_quality, "value-changed",
-                   G_CALLBACK(gui_on_change_quality), entry_distance);
+                   G_CALLBACK(GuiOnChangeQuality), entry_distance);
+
+  // ----------
+  separator = gtk_vseparator_new();
+  gtk_table_attach(GTK_TABLE(table), separator, 0, 2, 2, 3, GTK_EXPAND,
+                   GTK_EXPAND, 9, 9);
+  // gtk_box_pack_start(GTK_BOX(vbox), separator, true, false, 6);
```
veluca93, line 241: remove?

plugins/gimp/file-jxl-save.cc
```diff
@@ -194,69 +190,108 @@ bool SaveDialog() {
...
+
+  // effort signal
+  g_signal_connect(entry_effort, "value-changed", G_CALLBACK(GuiOnChangeEffort),
+                   nullptr);
+
+  // ----------
+  separator = gtk_vseparator_new();
+  gtk_table_attach(GTK_TABLE(table), separator, 0, 2, 4, 5, GTK_EXPAND,
+                   GTK_EXPAND, 9, 9);
+  // gtk_box_pack_start(GTK_BOX(vbox), separator, true, false, 6);
```
veluca93, line 263: same

plugins/gimp/file-jxl-save.cc
```diff
@@ -194,69 +190,108 @@ bool SaveDialog() {
...
+
+  // Lossless Mode Convenience Checkbox
+  static gchar lossless_help[] =
+      "Compress using modular lossless mode.  "
+      "Speed\u00A0is adjusted to improve performance.";
+  toggle_lossless = gtk_check_button_new_with_label("Lossless Mode");
+  gimp_help_set_help_data(toggle_lossless, lossless_help, nullptr);
+  gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(toggle_lossless),
+                               jxl_save_opts.lossless);
+  // gtk_box_pack_start(GTK_BOX(vbox), toggle_lossless, false, false, 0);
```
veluca93, line 274: also same

plugins/gimp/file-jxl-save.cc
```diff
@@ -814,24 +812,21 @@ bool SaveJpegXlImage(const gint32 image_id, const gint32 drawable_id,
...
-  for (gint32 i = 0; i < nlayers; i++) {
+  for (int i = nlayers - 1; i >= 0; i--) {
     gimp_save_progress.update();
 
     // copy image into buffer...
     gpointer pixels_buffer;
     pixels_buffer = g_malloc(buffer_size);
 
+    gimp_layer_resize_to_image_size(layers[i]);
+    // layers[i] = gimp_image_flatten(layers[i]);
```
veluca93, line 828: should this be here?

veluca93 (dismissed):

veluca93:
> > > @lvandeve @sboukortt do we (plan to) have an encode api for input with an arbitrary ICC profile, where the encoder does the necessary conversion? Like what cjxl does...
> > 
> > 
> > I don't think so, at least nothing more than the reverse of what the decoder can do without a CMS
> 
> So we really want (the full) `libjxl` not to depend on a CMS library?
> 
> For the decode library (`libjxl_dec`) I understand that we want to keep it minimal in terms of dependencies and binary size, but for the full library I don't think that's a big concern.
> 
> Anyway, does that mean that as it currently is, when doing lossy, the gimp encoder plugin would need to convert (whatever space the pixels are in) to linear sRGB floats, and call the encoder with uses_original_profile=false? And when doing lossless, it can pass the pixels directly and set uses_original_profile=true?

If I understand correctly, the encoder API today has some limitations wrt colorspaces.

@xiota is it possible, today, to always pass pixels in linear floating point RGB(A) (asking GIMP to do a colorspace conversion if needed) if `d != 0` and pass pixels in their original colorspace and pixel format if `d == 0`, also passing the ICC? In this case, I believe "Use original profile" wouldn't be needed, right?

plugins/gimp/file-jxl-save.cc
xiota: It's meant to be commented, but I'll remove it entirely along with the "save metadata" placeholder option.

xiota:
@veluca93 I removed the commented lines you indicated. They were mostly placeholders or reminders to myself.  I recompiled and opened and saved a few files with the different settings.  So it should be ready to merge.

> is it possible, today, to always pass pixels in linear floating point sRGB(A) (asking GIMP to do a colorspace conversion if needed) if d != 0 and pass pixels in their original colorspace and pixel format if d == 0, also passing the ICC? In this case, I believe "Use original profile" wouldn't be needed, right?

When you say "today", do you mean make the changes today, or whether it's possible with the current API?  I think it can most likely be done, but I need to look at what functions are available to me through GIMP.  I'd also like to test the behavior to see if it behaves the way I'd expect.  I still think photographers might still want to have the option to skip XYB because other programs might not do the colorspace conversions.  It also occurs to me that it might be useful to use lossless/d=0 with XYB.  There are minimal (predictable?) losses from the conversion to XYB and file sizes are cut in half.

veluca93:
> @veluca93 I removed the commented lines you indicated. They were mostly placeholders or reminders to myself. I recompiled and opened and saved a few files with the different settings. So it should be ready to merge.

Sounds good - usually we go with "// TODO(xiota): some description here" for such reminders :)

> 
> > is it possible, today, to always pass pixels in linear floating point sRGB(A) (asking GIMP to do a colorspace conversion if needed) if d != 0 and pass pixels in their original colorspace and pixel format if d == 0, also passing the ICC? In this case, I believe "Use original profile" wouldn't be needed, right?
> 
> When you say "today", do you mean make the changes today, or whether it's possible with the current API? I think it can most likely be done, but I need to look at what functions are available to me through GIMP. I'd also like to test the behavior to see if it behaves the way I'd expect. I still think photographers might still want to have the option to skip XYB because other programs might not do the colorspace conversions. It also occurs to me that it might be useful to use lossless/d=0 with XYB. There are minimal (predictable?) losses from the conversion to XYB and file sizes are cut in half.

It seems to me that with that modification, that flag would be used in rather niche situations (I understand the issue of other programs not doing the color conversion - IIRC we want to change things so that you'd get sRGB by default in that case, which would likely be "more correct" than whatever the original colorspace is, but now that's not the case AFAIU). Anyway, I'm OK with leaving the option there at least for now, and removing it later if it becomes redundant/extremely niche. @lvandeve @sboukortt any opinions here?

petrdiblik:
Why is "Enable Advanced Settings" _below_ "Advanced Settings"? It would make more sense to have it _above_, WDYT?

plugins/gimp/file-jxl-save.cc
```diff
@@ -570,41 +538,41 @@ bool JpegXlSaveOpts::set_num_channels(int channels) {
...
 
-bool JpegXlSaveOpts::set_pixel_type(int type) {
+bool JpegXlSaveOpts::SetPixelType(int type) {
   switch (type) {
     case JXL_TYPE_FLOAT16:
-      set_babl_type("half");
+      SetBablType("half");
       pixel_format.data_type = JXL_TYPE_FLOAT16;
       basic_info.bits_per_sample = 16;
       basic_info.exponent_bits_per_sample = 5;
```
jonsneyers, line 571: The encode API is still a bit evolving here (see https://github.com/libjxl/libjxl/pull/562), and we need to discuss what exactly we want to support.

I think it's important to distinguish the internal pixel format GIMP is using, the pixel format used to pass images to the encoder, and the bit depth signaled in the jxl bitstream.

I assume `set_babl_type` is used to ask GIMP to produce the image buffer in a certain format.
If so, then it should indeed match the `pixel_format` which is used to tell the encoder how to interpret the buffer.

But the `basic_info.bits_per_sample` is kind of independent of that. You can pass the image as 32-bit float, but still set `basic_info.bits_per_sample = 8`, for example. This field will be interpreted in two different ways, depending on whether XYB is used (i.e. lossy encoding) or not (i.e. lossless encoding):

- in lossless (or more generally, non-XYB) encoding, this is how many bits are actually stored losslessly. E.g. if the image is passed as float, but bits_per_sample = 8, then what actually gets stored are 8-bit ints. If `exponent_bits_per_sample > 0`, then what is effectively done is lossless encoding of float values. The bitstream stores these values internally as integers (the floats are reinterpreted as integers before encoding) and the decoder uses the `basic_info` header to interpret them correctly.
- in lossy (XYB) coding, these values are only informative; they are a suggestion of how to represent the data (e.g. to decide whether to store the result as 8-bit PNG or as 16-bit PNG. Lossless floats are not supported in this case, and `exponent_bits_per_sample` should be zero.

xiota: I've been trying to pass data to the encoder in the same format that GIMP is using.  Should I always be sending float to the encoder?

xiota: I just tried a couple combinations of basic_info and pixel_format.  Mismatched basic_info and pixel_format doesn't seem to work.

jonsneyers: The pixel_format has to match the format of the buffer. Basic_info can have any bits_per_sample, and shouldn't use exponent_bits_per_sample>0 unless lossless/keep_original_profile is done.

xiota: The pixel format did match the buffer.  I'll play with it some more, but I would prefer to leave changes for a future pull request.

xiota: Using float to transfer the image doesn't seem to simplify much, if anything.  Suppose the plugin were to transfer the image to the encoder using float.
* It looks like the transfer image has to be linear when encoding to XYB. But when encoding to non-XYB, the transfer image gamma has to match the original.
* The switch statements are still needed to set up basic_info.

If basic_info is just informative for XYB encoding, I don't understand why the encoder should reject uint32.

jonsneyers: It rejects it because iirc the spec requires it to be < 32 in the integer case. We want decoders to be able to use int32_t to represent the (suggested) pixel format. Otherwise implementations would need to have a special case for uint32 where they use an int64_t type internally for pixel values, which would be an annoying complication.

xiota:
@veluca93 @jonsneyers  When using XYB, the plugin now converts to sRGB float before encoding, then back to the original profile and pixel format after decoding.

jonsneyers:
> @veluca93 @jonsneyers When using XYB, the plugin now converts to sRGB float before encoding, then back to the original profile and pixel format after decoding.

Please make sure it's converting to unclamped values, so wide gamut doesn't get clamped to the sRGB gamut. (E.g. take a ProPhoto or P3 image and check that you get negative values and values above 1.0f)

xiota:
I don't know how to check for that.

jonsneyers:
> I don't know how to check for that.

Easiest way is to take an image like this:

https://webkit.org/blog-files/color-gamut/Webkit-logo-P3.png

and see if you can still see the logo after a roundtrip through saving as jxl and loading.

xiota:
@jonsneyers I removed the sRGB conversion because it ruins the image you sent.  It still sends data to the encoder as float.  Now ICC profiles appear to behave the way I've been trying to get them to.  I had reordered some things to accommodate the format and colorspace conversion, so it may be that when the color profile is set makes a difference.

jonsneyers:
> I wonder if there's any benefit to converting an 8-bit image to jxl, but keeping it as float on import.

Shouldn't make any difference. As it is now, jxl uses float as an internal representation in any case (which is why it can only do lossless ints up to 24-bit: that's the maximum precision that can be converted between float and int losslessly). So if you give it a uint8 buffer, it will internally convert it to a float buffer and then start encoding, and if you give it a float buffer, it will just copy it tot an internal float buffer and then start encoding.

jonsneyers:
> @jonsneyers I removed the sRGB conversion because it ruins the image you sent. It still sends data to the encoder as float. Now ICC profiles appear to behave the way I've been trying to get them to. I had reordered some things to accommodate the format and colorspace conversion, so it may be that when the color profile is set makes a difference.

Likely the sRGB conversion also clamped the result to the nominal range of [0,1], causing gamut clipping.

deymo:
@jonsneyers looks like you did the review here. Is there any blocker here?

xiota:
> > I wonder if there's any benefit to converting an 8-bit image to jxl, but keeping it as float on import.
> 
> Shouldn't make any difference. As it is now, jxl uses float as an internal representation in any case (which is why it can only do lossless ints up to 24-bit: that's the maximum precision that can be converted between float and int losslessly). So if you give it a uint8 buffer, it will internally convert it to a float buffer and then start encoding, and if you give it a float buffer, it will just copy it tot an internal float buffer and then start encoding.

I was referring to the decode side.  Instead of converting the jxl back into an 8-bit image, keep it as float.  Maybe there's some latitude in the encode/decode process that could be beneficial for editing.  For instance, if there's a gradient that's prone to banding in 8-bit, maybe it somehow gets smoothed out a bit after enc/dec if kept as float.

jonsneyers:
> > > I wonder if there's any benefit to converting an 8-bit image to jxl, but keeping it as float on import.
> > 
> > 
> > Shouldn't make any difference. As it is now, jxl uses float as an internal representation in any case (which is why it can only do lossless ints up to 24-bit: that's the maximum precision that can be converted between float and int losslessly). So if you give it a uint8 buffer, it will internally convert it to a float buffer and then start encoding, and if you give it a float buffer, it will just copy it tot an internal float buffer and then start encoding.
> 
> I was referring to the decode side. Instead of converting the jxl back into an 8-bit image, keep it as float. Maybe there's some latitude in the encode/decode process that could be beneficial for editing. For instance, if there's a gradient that's prone to banding in 8-bit, maybe it somehow gets smoothed out a bit after enc/dec if kept as float.

Ah yes, when doing lossy compression of an 8-bit image, decoding it to higher bit depth or to float is beneficial, especially since we don't do any dithering yet in the decoder when it's quantizing things down to 8-bit.

jonsneyers:
> @jonsneyers looks like you did the review here. Is there any blocker here?

I don't think so. @xiota can you do some final sanity checks to see that things work properly?

xiota:
@jonsneyers  Sorry.  It's not quite ready yet.  Switching to sending data to the encoder as float was a big change that introduced some glitches I need to fix.

xiota:
@jonsneyers  I'm getting a `../lib/jxl/enc_modular.cc:506: JXL_ASSERT: !fp` with one particular image, but others are working fine.  It occurs with the float conversion turned off, so it's either not a new problem or it's a problem outside of the plugin.

jonsneyers:
> @jonsneyers I'm getting a `../lib/jxl/enc_modular.cc:506: JXL_ASSERT: !fp` with one particular image, but others are working fine. It occurs with the float conversion turned off, so it's either not a new problem or it's a problem outside of the plugin.

That's not a new problem and it is a problem outside of the plugin.  https://github.com/libjxl/libjxl/pull/562 should solve it.

xiota:
> That's not a new problem and it is a problem outside of the plugin.

Thanks.  The plugin is good for now then.

veluca93 (approved):

PR #531, opened 2021-09-01
Modular decode: reduce tree size limit by a factor of 16

### LINKED ISSUE #528: Decoding 1x1 image consumes 5.2GB RAM
Hello, the following 1x1 JPEG-XL image, found via fuzz testing, contains a 8163x2954 3-channel frame.

[fuzz-37679.txt](https://github.com/libjxl/libjxl/files/7085929/fuzz-37679.txt)

When decoding the tree in `DecodeGlobalInfo`, the `tree_size_limit` for this frame is calculated as 72341530 here:

https://github.com/libjxl/libjxl/blob/9d03bfe8111f49210e2d096688dd4db303654270/lib/jxl/dec_modular.cc#L163-L164

then clipped to 67108864 (1 << 26) here:

https://github.com/libjxl/libjxl/blob/9d03bfe8111f49210e2d096688dd4db303654270/lib/jxl/modular/encoding/dec_ma.cc#L100

The size of the `DecodeTree` eventually trips the  `tree_size_limit` check, but consumes 5.2GB of RAM at that point, and takes about 4 seconds to get there.

https://github.com/libjxl/libjxl/blob/9d03bfe8111f49210e2d096688dd4db303654270/lib/jxl/modular/encoding/dec_ma.cc#L43-L45

I suspect this rapid memory growth from a "small" image isn't desired.

Perhaps there could be an earlier sanity check on frame dimensions, or maybe the tree limit should be dependent on target dimensions rather than frame dimensions? I'm sure there are lots of other options too.

lovell:
As discussed in and fixes #528

The reduction applies to both the per-pixel limit and the overall limit, which means the maximum memory consumption of a frame's decode tree will be ~0.3MB (previously ~5.2GB).

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -161,7 +161,7 @@ Status ModularFrameDecoder::DecodeGlobalInfo(BitReader* reader,
   bool has_tree = reader->ReadBits(1);
   if (has_tree) {
     size_t tree_size_limit =
-        1024 + frame_dim.xsize * frame_dim.ysize * nb_chans;
+        1024 + frame_dim.xsize * frame_dim.ysize * nb_chans / 16;
     JXL_RETURN_IF_ERROR(DecodeTree(reader, &tree, tree_size_limit));
     JXL_RETURN_IF_ERROR(
         DecodeHistograms(reader, (tree.size() + 1) / 2, &code, &context_map));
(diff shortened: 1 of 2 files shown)


jonsneyers:
I was literally just making the exact same pull request :)

Note: currently in the amendment to 18181-1 that defines the (single) profile and levels, we have `1024 + width × height × nb_channels` as the limit on tree size, so we'll need to adjust the limit there too.

jonsneyers (approved):

PR #532, opened 2021-09-01
Add pipeline executions that allow ssh access.

veluca93:
Those pipelines will build a release, debug, asan, msan or opt build
when pushed to a branch ci-{release, debug, ...}-debug name, and then
start a tmate session in the resulting environment that can be used to
access the CI environment for debugging purposes.

### DIFF
--- a/.github/workflows/debug_ci.yml
+++ b/.github/workflows/debug_ci.yml
@@ -1,0 +1,59 @@
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Workflow for building and then debugging on a specific commit.
+
+name: Build and Test debugging
+on:
+  push:
+    branches:
+      - ci-*-debug
+
+jobs:
+  ubuntu_build:
+    name: Ubuntu Build and SSH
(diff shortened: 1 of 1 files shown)


.github/workflows/debug_ci.yml
```diff
@@ -0,0 +1,64 @@
...
+          ${{ matrix.apt_pkgs }} \
+        #
+        echo "CC=clang-7" >> $GITHUB_ENV
+        echo "CXX=clang++-7" >> $GITHUB_ENV
+    - name: Checkout the source
+      uses: actions/checkout@v2
+      with:
+        submodules: true
+        fetch-depth: 2
+    - name: Git environment
```
deymo, line 49: Don't need this here. This creates a variable (git-env.output.parent or something like that) that can be using within this file.

.github/workflows/debug_ci.yml
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Workflow for building and then debugging on a specific 
+
+name: Build/Test
```
deymo, line 8: Rename to not confuse with the other one

deymo (approved):

PR #534, opened 2021-09-02
Auto resample for high distance

lvandeve:
By default, let cjxl choose resampling=2 if requested distance >= 20

This also increases the allowed --distance given to cjxl to go up to 20,
previously it only was allowed up to 15.

For these bit rates, using the 2x2 resampling looks better on most
images.

The distance itself is lowered to 6 (and slightly more for higher
original distance), which roughly ends up at similar bit rate. This
depends on the image but is relatively close overall.

While there are lower distances where rescaling works better as well, it
depends on the image, 20 is a conservative choice.

This only implements it for the cjxl binary tool for now, not for the encoder
API (which doesn't yet support the resampling option; to be added soon)
nor for the benchmark.

To force using no resampling anyway, you can still use resampling=1.

### DIFF
--- a/tools/cjxl.cc
+++ b/tools/cjxl.cc
@@ -307,7 +307,7 @@ void CompressArgs::AddCommandLineOptions(CommandLineParser* cmdline) {
   // Target distance/size/bpp
   opt_distance_id = cmdline->AddOptionValue(
       'd', "distance", "maxError",
-      ("Max. butteraugli distance, lower = higher quality. Range: 0 .. 15.\n"
+      ("Max. butteraugli distance, lower = higher quality. Range: 0 .. 25.\n"
        "    0.0 = mathematically lossless. Default for already-lossy input "
        "(JPEG/GIF).\n"
        "    1.0 = visually lossless. Default for other input.\n"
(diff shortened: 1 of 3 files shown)


tools/cjxl.cc
```diff
@@ -617,7 +619,7 @@ jxl::Status CompressArgs::ValidateArgs(const CommandLineParser& cmdline) {
 
   if (got_distance) {
     constexpr float butteraugli_min_dist = 0.1f;
-    constexpr float butteraugli_max_dist = 15.0f;
+    constexpr float butteraugli_max_dist = 20.0f;
```
jonsneyers, line 622: Maybe bump it up a bit further? Otherwise a new encode technique is introduced for only one very specific distance target (20).

Also: maybe reconsider also updating the -q setting at which we switch to modular?

lvandeve: I don't mind extending it to 25, done

about the -q setting: that's an independent discussion, perhaps better done e.g. in a bug report than in this PR

jonsneyers (dismissed):

tools/cjxl.cc
```diff
@@ -420,9 +420,11 @@ void CompressArgs::AddCommandLineOptions(CommandLineParser* cmdline) {
...
                           "force disable/enable patches generation.",
                           &params.patches, &ParseOverride, 1);
-  cmdline->AddOptionValue('\0', "resampling", "1|2|4|8",
-                          "Subsample all color channels by this factor.",
-                          &params.resampling, &ParseUnsigned, 1);
+  cmdline->AddOptionValue(
+      '\0', "resampling", "0|1|2|4|8",
+      "Subsample all color channels by this factor, or use 0 to choose the "
+      "resampling factor based on distance.",
+      &params.resampling, &ParseUnsigned, 1);
```
jyrkialakuijala, line 427: 0 as default?

lvandeve: Indeed, fixed

jyrkialakuijala (approved):

PR #538, opened 2021-09-03
Properly clear the splines at the beginning of each frame

sboukortt:


lib/jxl/dec_frame.cc
```diff
@@ -358,6 +358,7 @@ Status FrameDecoder::ProcessDCGlobal(BitReader* br) {
   } else {
     shared.image_features.patches.Clear();
   }
+  shared.image_features.splines = Splines();
```
veluca93, line 361: Maybe we want to call a .clear() method that erases all splines but doesn't reallocate the vector? Likely doesn't really matter though

veluca93 (approved):

lib/jxl/dec_frame.cc
sboukortt: Done, thanks.

veluca93 (approved):

PR #539, opened 2021-09-03
Remove most other symbols from the shared library.

deymo:
hwy is always linked in statically (even when installed) and skcms is
sometimes bundled with libjxl.a, however functions from those libraries
shouldn't be exposed in the libjxl library. For hwy, we use
`-Wl,--exclude-libs=ALL` at link time, which would also work for skcms
but in the latter case we build it locally so there's no reason to also
mark it as hidden at build time.

The std:: namespace includes a lot of header-only functions that are not
always inlined (like template ones) and thus create weak symbols with
the functions. The std namespace is marked explicitly as default
visibility because those symbols should be exposed in shared libraries
when using std types in the library interface (functions or exceptions)
but our API is C only so we don't need them there. Removing the symbols
makes the library a little bit smaller in the case of templated
functions that were actually inlined and it removes about 100 symbols
from the dynamic symbol table.

This doesn't remove the typeinfo for all the other types when compiled
with rtti (like in asan/ubsan).

### DIFF
--- a/lib/jxl/jxl.version
+++ b/lib/jxl/jxl.version
@@ -1,4 +1,17 @@
 JXL_0 {
   global:
     Jxl*;
+
+  local:
+    # Hide all the std namespace symbols. std namespace is explicitly marked
+    # as visibility(default) and header-only functions or methods (such as those
+    # from templates) should be exposed in shared libraries as weak symbols but
+    # this is only needed when we expose those types in the shared library API
+    # in any way. We don't use C++ std types in the API and we also don't
+    # support exceptions in the library.
+    # See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36022 for a discussion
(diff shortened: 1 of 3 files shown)


lib/jxl/jxl.version
```diff
@@ -1,4 +1,17 @@
 JXL_0 {
   global:
     Jxl*;
+
+  local:
+    # Hide all the std namespace symbols. std namespace is explicitly marked
+    # as visibility(default) and header-only functions or methods (such as those
+    # from templates) should be expose in shared libraries as weak symbols but
```
veluca93, line 8: exposed

veluca93 (approved):

moubctez:
This breaks building on macOS:

```
ld: unknown options: --exclude-libs=ALL
```

Issue #540, opened 2021-09-04
Request of EXIF support for convert JXL to JPG (exif needs to be added after convertion)


MartinEesmaa:
Hello! I'm looking forward to builders needs to add DJXL for exif support after convertion from JXL to JPG, so JPG recompressed need to have same original metadata came from my camera picture. I tried using exiftool to add .exif metadata into JXL file, but updated file failed.

3 Major Developers, that I need request help to add EXIF support from JXL to JPG
@deymo 
@jonsneyers 
@jxl-bot 

Please add support to build.

Regards
- Martin Eesmaa

nathanielcwm:
what?

jloqfjgk:
@jxl-bot is a bot, so there is no need to ping it.

-----

If you would like to manually edit EXIF metadata, libjxl is not the right tool. Use exiftool, exiv2, etc. instead.

Since JXL is a new format, many tools do not support it well. For example [exiftool](https://github.com/exiftool/exiftool) can only read but not write metadata for JXL, and [exiv2](https://github.com/Exiv2/exiv2) only supports JXL/bmff ([JXL/codestream is a TODO](https://github.com/Exiv2/exiv2/issues/1506)).

-----

On the other hand, if you take photos with your camera and it contains EXIF metadata, when you `cjxl image.jpg image.jxl` (lossless JPEG transcode) or `djxl image.jxl image.jpg` (get back the original JPG), all the metadata is kept.

MartinEesmaa:
Impossible, I tried djxl image.jxl image.jpg of 8000x6000, but it successed convertion, but metadata failed.
Result: JPEG XL decoder v0.6.0 6e1e309 [Scalar]
Read 3871181 compressed bytes.
Warning: could not decode losslessly to JPEG. Retrying with --pixels_to_jpeg...
Decoded to pixels.
8000 x 6000, 79.68 MP/s [79.68, 79.68], 1 reps, 8 threads.
Allocations: 533 (max bytes in use: 6.993952E+08)

Exiftool image.jpg converted:
ExifTool Version Number         : 12.30
File Name                       : image.jpg
Directory                       : .
File Size                       : 11 MiB
File Modification Date/Time     : 2021:09:04 22:54:56+10:00
File Access Date/Time           : 2021:09:04 22:55:16+10:00
File Creation Date/Time         : 2021:09:04 22:54:56+10:00
File Permissions                : -rw-rw-rw-
File Type                       : JPEG
File Type Extension             : jpg
[...]

PR #544, opened 2021-09-06
better handling of exif (in particular orientation)

jonsneyers:
There were some issues with Exif metadata handling in cjxl/djxl (see also https://github.com/libjxl/libjxl/issues/540):

- `djxl foo.jxl bar.jpg -j --use_sjpeg` wrote no exif metadata
- `djxl foo.jxl bar.jpg -j` (using libjpeg) wrote malformed exif metadata
- `djxl foo.jxl bar.jpg -j` did not orient the image, which is not good (it would maybe be OK if we always wrote exif metadata that does the appropriate orientation, but there could also just not be any exif, e.g. when you do `cjxl oriented.jpg foo.jxl --strip`.
- `djxl foo.jxl bar.png` orients the image (which is good), but also writes exif metadata which might lead to repeated orientation (most viewers ignore exif orientation in png files, but some don't, and also cjxl doesn't)

This PR attempts to fix those issues in the following way:
- `djxl` now always writes (correct) exif metadata when outputting jpg
- `djxl` now orients the image before outputting a lossy jpg
- if the exif does orientation, its orientation field gets overwritten to the identity when writing png/lossy jpg output, so there is no repeated orientation
- the original exif is still preserved in the jxl bitstream, overwriting with the identity is only done just before writing png / lossy jpg output. When writing a reconstructed lossless jpg, we still want the original exif to be preserved.

This PR is not trying to get things right in the encode api, where I think currently no Exif inspection is done yet to set the codestream orientation correctly.

### DIFF
--- a/lib/jxl/exif.h
+++ b/lib/jxl/exif.h
@@ -1,0 +1,27 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#ifndef LIB_JXL_EXIF_H_
+#define LIB_JXL_EXIF_H_
+
+// Basic parsing of Exif (just enough for the render-impacting things
+// like orientation)
+
+#include "lib/jxl/base/padded_bytes.h"
+#include "lib/jxl/image_metadata.h"
+
+namespace jxl {
+
+// Parses the Exif data just enough to extract any render-impacting info.
+// If the Exif data is invalid or could not be parsed, then it is treated
+// as a no-op.
+void InterpretExif(const PaddedBytes& exif, CodecMetadata* metadata);
+
+// Sets the Exif orientation to the identity, to avoid repeated orientation
+void ResetExifOrientation(PaddedBytes& exif);
+
+}  // namespace jxl
+
+#endif  // LIB_JXL_EXIF_H_
--- a/lib/extras/enc/apng.cc
+++ b/lib/extras/enc/apng.cc
@@ -49,6 +49,7 @@
 #include "lib/jxl/dec_external_image.h"
 #include "lib/jxl/enc_color_management.h"
 #include "lib/jxl/enc_image_bundle.h"
+#include "lib/jxl/exif.h"
 #include "lib/jxl/frame_header.h"
 #include "lib/jxl/headers.h"
 #include "lib/jxl/image.h"
@@ -70,7 +71,12 @@ class BlobsWriterPNG {
  public:
   static Status Encode(const Blobs& blobs, std::vector<std::string>* strings) {
     if (!blobs.exif.empty()) {
-      JXL_RETURN_IF_ERROR(EncodeBase16("exif", blobs.exif, strings));
+      // PNG viewers typically ignore Exif orientation but not all of them do
+      // (and e.g. cjxl doesn't), so we overwrite the Exif orientation to the
+      // identity to avoid repeated orientation.
+      PaddedBytes exif(blobs.exif);
+      ResetExifOrientation(exif);
+      JXL_RETURN_IF_ERROR(EncodeBase16("exif", exif, strings));
     }
     if (!blobs.iptc.empty()) {
       JXL_RETURN_IF_ERROR(EncodeBase16("iptc", blobs.iptc, strings));
--- a/lib/extras/enc/jpg.cc
+++ b/lib/extras/enc/jpg.cc
@@ -19,8 +19,10 @@
 #include "lib/jxl/base/status.h"
 #include "lib/jxl/color_encoding_internal.h"
 #include "lib/jxl/common.h"
+#include "lib/jxl/dec_external_image.h"
 #include "lib/jxl/enc_color_management.h"
 #include "lib/jxl/enc_image_bundle.h"
+#include "lib/jxl/exif.h"
 #include "lib/jxl/image.h"
 #include "lib/jxl/image_bundle.h"
 #include "lib/jxl/sanitizers.h"
@@ -33,7 +35,6 @@ namespace extras {
 
 namespace {
 
-constexpr float kJPEGSampleMultiplier = MAXJSAMPLE;
 constexpr unsigned char kICCSignature[12] = {
     0x49, 0x43, 0x43, 0x5F, 0x50, 0x52, 0x4F, 0x46, 0x49, 0x4C, 0x45, 0x00};
 constexpr int kICCMarker = JPEG_APP0 + 2;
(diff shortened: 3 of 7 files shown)


lib/jxl/exif.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
```
deymo, line 1: This is encoder only, right? Could we name this enc_exif?

ResetExifOrientation() is only used in extras, are we gonna modify the exif if you request the pixels in some other orientation in the API? Looks like maybe this file should go to extras/ instead.

lib/extras/codec_jpg.cc
```diff
@@ -503,7 +498,7 @@ Status EncodeImageJPG(const CodecInOut* io, JpegEncoder encoder, size_t quality,
       break;
     case JpegEncoder::kSJpeg:
       JXL_RETURN_IF_ERROR(
-          EncodeWithSJpeg(ib, quality, chroma_subsampling, bytes));
+          EncodeWithSJpeg(ib, io, quality, chroma_subsampling, bytes));
```
deymo, line 501: could we pass just the blobs here?

lib/extras/codec_jpg.cc
```diff
@@ -195,11 +194,11 @@ void WriteExif(jpeg_compress_struct* const cinfo, const PaddedBytes& exif) {
   if (exif.size() < 4) return;
   jpeg_write_m_header(
       cinfo, kExifMarker,
-      static_cast<unsigned int>(exif.size() - 4 + sizeof kExifSignature));
+      static_cast<unsigned int>(exif.size() + sizeof kExifSignature));
   for (const unsigned char c : kExifSignature) {
     jpeg_write_m_byte(cinfo, c);
   }
-  for (size_t i = 4; i < exif.size(); ++i) {
+  for (size_t i = 0; i < exif.size(); ++i) {
```
deymo, line 201: why this change from skipping 4 bytes to not skipping them?

lib/jxl/exif.h
jonsneyers: Encoder-only atm, but maybe it could be useful to do ResetExifOrientation by default when we return Exif blobs in the decode api. If we just pass 'uncensored' Exif blobs, it's almost certain that many applications will interpret the Exif and incorrectly apply orientation twice.

lib/extras/codec_jpg.cc
jonsneyers: This is symmetric with the libjpeg function. I have no strong opinion on this though.

lib/extras/codec_jpg.cc
jonsneyers: I think it's a bug caused by confusion about whether the blob starts immediately with the tiff header, or it starts with a 4-byte offset (that is always zero) and then you get the tiff header.

In the file format, the `Exif` box does do the offset thing (because that's how it's defined in ISOBMFF and done by others like jpeg 2000). In the blobs we are passing here, we don't do the offset thing (we have an implicit offset 0). In jpeg and png, there's also no offset.

lib/jxl/exif.h
deymo: My request is that you don't add new calls from the tool/extras to internal API because I'm actually trying to do the opposite and it means just more tech debt that I have to clean up. Either we move this functionality to cjxl/djxl or the extras lib; or we implement it in the C API (which is in progress anyway).

jonsneyers: I think these functions should be integrated in the C api (called from the functions that get/set Exif metadata, conditioned on an option). Once cjxl/djxl use that api, they will no longer need to call these functions directly.

jonsneyers: For now, we need to fix a broken situation though: djxl was not writing correct exif metadata when outputting lossy jpeg, and problematic repeated orientation was happening when using cjxl/djxl to convert oriented jpg -> jxl -> png -> jxl.

Handling exif metadata properly in the api is something that also needs to be done, but atm it's not handled at all there so that's something else. Once it is handled there, we can integrate the exif handling stuff in the api and discuss if and what we want to expose there, but let's fix cjxl/djxl now already.

jyrkialakuijala (dismissed):

mo271:
This should close #540 once it is submitted.

lib/extras/codec_jpg.cc
```diff
@@ -408,27 +406,25 @@ Status EncodeWithLibJpeg(const ImageBundle* ib, const CodecInOut* io,
...
-    }
-    JSAMPROW rows[] = {row.get()};
-    jpeg_write_scanlines(&cinfo, rows, 1);
+  size_t stride = ib->oriented_xsize() * cinfo.input_components;
+  PaddedBytes raw_bytes(stride * ib->oriented_ysize());
+  JXL_RETURN_IF_ERROR(ConvertToExternal(
+      *ib, 8, /*float_out=*/false, cinfo.input_components, JXL_BIG_ENDIAN,
+      stride, nullptr, raw_bytes.data(), raw_bytes.size(),
+      /*out_callback=*/nullptr,
+      /*out_opaque=*/nullptr, ib->metadata()->GetOrientation()));
```
sboukortt, line 423: I believe this drops support for 12-bit libjpeg, is that correct? Are we fine with that? (I assume we are, just making sure 😁)

sboukortt (approved):

lib/extras/codec_jpg.cc
jonsneyers: Ah, I didn't think about that possibility. Does such a libjpeg have JSAMPLE/JSAMPROW defined as a 16-bit type then? This code could be adapted to do both 8 and 16 bit, I just didn't realize that JSAMPLE can be something else than uint8_t. I was just trying to leverage the orientation being done by ConvertToExternal, simplifying the code a bit in the process.

In my opinion, if you care enough about precision of your output to make 12-bit libjpeg worthwhile, then most likely you'll prefer your output as a 16-bit png and not as a jpeg. But I do not object to supporting 12-bit libjpeg, I suppose it's just a matter of multiplying `stride` and bitdepth `8` by `sizeof(JSAMPLE)`, right?

jonsneyers:
I rebased this one, it was a quite old pull request and the extras/* stuff has changed quite a bit in the mean time. PTAL.

Regarding encode api: I'm not 100% sure what is the best way to proceed there. There are two cases:

- AddJPEGFrame: there I think the only reasonable thing to do is to parse the exif that may be in the input jpeg file to set the jxl orientation correctly
- adding an Exif box that contains orientation: here we could do several things: 1) just do nothing (current situation), which means the exif orientation might be different from the jxl orientation, 2) parse the exif and set jxl orientation if that can still be done, 3) if exif is added after the image header is already written, parse the exif to check if the orientation matches; either return error or modify exif to match the image if needed, 4) whenever exif is added, first clear its orientation field to the identity since the jxl orientation takes precedence anyway and the best way to avoid double orientation is to set the exif orientation to identity.

lib/extras/codec_jpg.cc
jonsneyers: Haven't tested it with an actual 12-bit libjpeg, but I think I modified it so it should work when compiled against a 12-bit libjpeg.

sboukortt (approved):

PR #546, opened 2021-09-06
Fix signed-unsigned comparison warnings (errors)

eustas:
Suddenly discovered that libjxl does not build on my macbook.


deymo (dismissed):
It looks like all of these are in tests. Maybe it is related to the gtest version or flags? AFAIK we compile with -Wsign-comparison just fine (we used to have -Wno-sign-comparison)

eustas:
Yes, it is likely something about gtest.
Fixed lint / commit author. PTAL.

deymo (approved):

Issue #96, opened 2021-06-03
Add list of applications/projects using libjxl


jonsneyers:
It would be good to add a document that keeps track of software that has jxl support. Such a list serves several purposes:

- thank/acknowledge other projects for integrating jxl support
- point end-users to software that can read/write jxl
- keep track of the adoption status of jxl
- in case of a (security) bug, it's easier to see who might be affected and check if they are updated (in case they use static linking)

Here is a first attempt at making such a list. Please add missing software in the comments! When we have a more or less complete list, I'll make a pull request to add the list as a markdown document.

## Browsers

- Chromium: behind a flag since version 91, tracking bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1178058
- Firefox: behind a flag since version 90, tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1539075
- Safari: not supported, tracking bug: https://bugs.webkit.org/show_bug.cgi?id=208235
- Edge: behind a flag since version 91, start with `.\msedge.exe --enable-features=JXL`

## Image libraries

- ImageMagick: supported since 7.0.10-54 (https://imagemagick.org/)
- libvips: supported since 8.11 (https://libvips.github.io/libvips/)
- Imlib2: https://github.com/alistair7/imlib2-jxl

## OS-level support / UI frameworks / file browser plugins

- Qt / KDE: plugin available: https://github.com/novomesk/qt-jpegxl-image-plugin
- GDK-pixbuf: plugin available in libjxl repo
- gThumb: https://ubuntuhandbook.org/index.php/2021/04/gthumb-3-11-3-adds-jpeg-xl-support/
- MacOS viewer/QuickLook plugin: https://github.com/yllan/JXLook
- Windows Imaging Component: https://github.com/mirillis/jpegxl-wic
- Windows thumbnail handler: https://github.com/saschanaz/jxl-winthumb
- OpenMandriva Lx (since 4.3 RC)

## Image editors

- GIMP: plugin available in libjxl repo, no official support, tracking bug: https://gitlab.gnome.org/GNOME/gimp/-/issues/4681
- Photoshop: no plugin available yet, no official support yet

## Image viewers

- XnView: https://www.xnview.com/en/
- ImageGlass: https://imageglass.org/
- Any viewer based on Qt, KDE, GDK-pixbuf, ImageMagick, libvips or imlib2 (see above)
[...]

PR #547, opened 2021-09-07
add list of applications that use libjxl

jonsneyers:
See #96.


deymo:
I think it would be cleaner to fold the URLs into markdown links instead of having the URL text there.

deymo (approved):

jonsneyers:
> I think it would be cleaner to fold the URLs into markdown links instead of having the URL text there.

Agreed.

jonsneyers:
> I think it would be cleaner to fold the URLs into markdown links instead of having the URL text there.

Updated.

deymo (approved):

PR #551, opened 2021-09-07
Remove JPEG's DecodeTarget parameter.

deymo:
DecodeTarget was telling DecodeImageJPG whether we should load the pixel
or the JPEG coefficients. To load the JPEG coefficients we were actually
using the library internal API. Loading the coefficients is used for
pixel-lossless (but not byte-lossless) encoding of JPEG files. This is
not currently supported in the public API, but it would certainly be
a different workflow than loading the pixels (either we pass an image
with coefficient values or we pass the JPEG file like in lossless
recompression).

This patch splits out this workflow from DecodeImageJPG and
EncodeImageJPG so we can start migrating those functions to use the
external types.

### DIFF
--- a/lib/extras/codec_jpg.h
+++ b/lib/extras/codec_jpg.h
@@ -42,8 +42,17 @@ Status DecodeImageJPG(Span<const uint8_t> bytes, const ColorHints& color_hints,
 // Encodes into `bytes`.
 Status EncodeImageJPG(const CodecInOut* io, JpegEncoder encoder, size_t quality,
                       YCbCrChromaSubsampling chroma_subsampling,
-                      ThreadPool* pool, PaddedBytes* bytes,
-                      DecodeTarget target = DecodeTarget::kPixels);
+                      ThreadPool* pool, PaddedBytes* bytes);
+
+// Temporary wrappers to load the JPEG coefficients to a CodecInOut. This should
+// be replaced by calling the corresponding JPEG input and output functions on
+// the API.
+
+// Decodes the JPEG image coefficients to a CodecIO for lossless recompression.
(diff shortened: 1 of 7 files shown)


lib/extras/codec_jpg.h
```diff
@@ -42,8 +42,17 @@ Status DecodeImageJPG(Span<const uint8_t> bytes, const ColorHints& color_hints,
...
+                      ThreadPool* pool, PaddedBytes* bytes);
+
+// Temporary wrappers to load the JPEG coefficients to a CodecInOut. This should
+// be replaced by calling the corresponding JPEG input and output functions on
+// the API.
+
+// Decodes the JPEG image coefficients to a CodecIO for lossless recompression.
+Status DecodeImageJPGCoefficients(Span<const uint8_t> bytes, CodecInOut* io);
+
+// Reconstructs the JPEG from the coefficients and medatada in CodecInOut.
```
veluca93, line 54: metadata :)

jonsneyers:
Do cjxl/djxl still work as before with this change?

deymo:
@jonsneyers yes, the `if` checking for dec_target moved outside the function now in both cases.

there's an issue with the way cjxl handles the input because the flags by default basically say "if the input was a JPEG then parse the coefficients; otherwise then ignore that setting"; to translate this to the C API we need to know before decoding the image (with extras/) whether we want to use the AddJPEGFrame or AddImageFrame. But this is not handled here for now.

lib/extras/codec_jpg.h
deymo: tadaaa!

veluca93 (approved):

PR #552, opened 2021-09-07
Header only: box API design

lvandeve:
This MR proposes API for getting metadata boxes in the decoder.

It is integrated with the event based system the decoder uses,
using the JXL_DEC_BOX event to indicate a new box.

Boxes are output to a byte buffer owned by the user, similar to the
recompressed JPEG output.

Boxes can be gotten in an uncompressed mode or in a raw mode. For
non-brob boxes, these modes are the same. For brob boxes, the standard
mode automaticlaly decompresses them, the raw mode returns them as-is.


lib/include/jxl/decode.h
```diff
@@ -238,6 +243,39 @@ typedef enum {
    * image and always before JXL_DEC_FULL_IMAGE.
    */
   JXL_DEC_JPEG_RECONSTRUCTION = 0x2000,
+
+  /** Informative event by JxlDecoderProcessInput: The header of a box of the
+   * BMFF container format is decoded. The following API functions related to
```
veluca93, line 248: nit: "Box Metadata File Format container format" is a bit redundant :)

lib/include/jxl/decode.h
```diff
@@ -238,6 +243,39 @@ typedef enum {
...
+   * box typename.
+   * @see JxlDecoderBoxTypeRawEquals to compare with the raw box typename.
+   * @see JxlDecoderGetBoxSizeRaw to get the raw compressed box size.
+   * @see JxlDecoderSetBoxModeRaw to configure whether to get the box
+   * data uncompressed, or possibly compressed. Getting the raw box contents
+   * returns the same data and type unless JxlDecoderIsBoxCompressed is true. In
+   * the compressed box case, the raw typename is "brob" but the underlying
+   * typename can still be requested, and the data is brotli compressed. To get
+   * metadata, use the non-raw box mode. To perform more primitive operations
+   * with boxes, use the raw box mode and the non-raw type if needed. The buffer
```
veluca93, line 267: I think that the part starting with "The buffer" should be on a separate paragraph, no?

lib/include/jxl/decode.h
```diff
@@ -238,6 +243,39 @@ typedef enum {
...
+   * set with JxlDecoderSetBoxBuffer must be set again for each next box, or can
+   * be left unset to skip outputting this box. The output buffer contains the
+   * full box data when the next JXL_DEC_BOX event occurs or when
```
veluca93, line 270: What about unrelated events, say JXL_DEC_FRAME?

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ *
+ * Calling JxlDecoderReleaseBoxBuffer when no buffer is set is
+ * not an error and returns 0.
+ *
+ * @param dec decoder object
+ * @return the amount of bytes the decoder has not yet written to of the data
+ * set by JxlDecoderSetBoxBuffer, or 0 if no buffer is set or
+ * JxlDecoderReleaseBoxBuffer was already called.
+ */
+JXL_EXPORT size_t JxlDecoderReleaseBoxBuffer(JxlDecoder* dec);
```
veluca93, line 1006: nit: I'd move this one up.

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * @param type box type to compare with
+ * @return JXL_TRUE if the box type is available and equal to the given type,
+ * JXL_FALSE otherwise.
+ */
+JXL_EXPORT JXL_BOOL JxlDecoderBoxTypeEquals(JxlDecoder* dec, const char* type);
+
+/**
+ * Outputs the type of the current box, after a JXL_DEC_BOX event occured, as 4
+ * characters plus a null termination character. If the box is of type "brob",
+ * the underlying uncompressed box type is returned instead, not "brob".
```
veluca93, line 1034: I'm pretty sure this is not what is intended.

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+/**
+ * Outputs whether the given type equals the type returned by
+ * JxlDecoderGetBoxType.
+ *
+ * @param dec decoder object
+ * @param type box type to compare with
+ * @return JXL_TRUE if the box type is available and equal to the given type,
+ * JXL_FALSE otherwise.
+ */
+JXL_EXPORT JXL_BOOL JxlDecoderBoxTypeEquals(JxlDecoder* dec, const char* type);
```
veluca93, line 1029: What should you pass for type? 4 bytes? 4 bytes + null?

It might make sense to write this as `const char type[5]` to make the format clearer.

veluca93: (also for the Raw version)

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+JXL_EXPORT JXL_BOOL JxlDecoderBoxTypeRawEquals(JxlDecoder* dec,
+                                               const char* type);
+
+/**
+ * Outputs whether the raw box type equals "brob", indicating it is a brotli
+ * compressed box. Its underlying type is still returned with
+ * JxlDecoderGetBoxType. The data returned to the box buffer is transparently
+ * decompressed, unless JxlDecoderSetBoxModeRaw is enabled, so this function is
+ * not needed to just get the data, as long as the library is compiled with
+ * brotli.
```
veluca93, line 1062: I'd consider moving all the "brotli-only" functions in a separate part of the header, and writing comments of the form "The following functions are only available if JXL_SMALL_DECODER_LIBRARY is not set" (or whatever), as well as putting them behind an `#if`.

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ *
+ * @param dec decoder object
+ * @param type box type to compare with
+ * @return JXL_TRUE if the box type is available and is "brob",
+ * JXL_FALSE otherwise.
+ */
+JXL_EXPORT JXL_BOOL JxlDecoderIsBoxCompressed(JxlDecoder* dec,
+                                              const char* type);
```
veluca93, line 1070: I don't think we need the type here.

Also, isn't this function redundant wrt getting the Raw type?

lib/include/jxl/decode.h
```diff
@@ -913,6 +950,142 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * getting boxes in raw mode, in the default mode where brob boxes are
+ * uncompressed, this does not return the size of the uncompressed contents,
+ * and there is no way to get the uncompressed size before having already
+ * uncompressed and output this data.
+ *
+ * @param dec decoder object
+ * @param size raw size of the box in bytes
+ * @return JXL_DEC_ERROR if no box size is available, JXL_DEC_SUCCESS otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxSizeRaw(const JxlDecoder* dec,
```
veluca93, line 1086: Do we need this function?

lib/include/jxl/decode.h
jonsneyers: I would allow 4 bytes (not zero terminated) too, not requiring them to be zero terminated.

lib/include/jxl/decode.h
lvandeve: Those work independently, and it's not specified if you get box data before, after, or during the decoding of pixels. You could get a box in the middle of a frame (because a frame could be split over multiple partial boxes), or you could get a frame in the middle of a box (if it contains multiple), but I don't think the other events need to be mentioned, specifying this happens at the next box or the end should be complete.

lib/include/jxl/decode.h
lvandeve: We could make them unavailable at all at compile time, returning a compilation error, or instead we could make them exist in the API, but return "error" or "false" if brotli is not compiled in.

Not available behind #if has the advantage that you detect the problem sooner, but the disadvantage that it may break compilation of external libraries, even if they're only using the brotli decompression for non-core reasons.

Any thought on this?

lib/include/jxl/decode.h
lvandeve: When doing primitive box operations, such as concatenating boxes and leaving some out, then this helps: you know exactly what size to set on the output buffer, which makes it easier. The size field and type field are the two things you can get from the box header, not allowing to see the size field would disable possible functionality of the box format.

lib/include/jxl/decode.h
veluca93: Disabling brotli is not meant for compiled libraries that are to be distributed but rather for squeezing as many bytes as we can out of a static build - I wouldn't be concerned about external libraries.

lib/include/jxl/decode.h
lvandeve: Fixed, but leaving in the term BMFF for reference

lib/include/jxl/decode.h
lvandeve: Not sure what was the non intended part, but reworded the sentence

lib/include/jxl/decode.h
veluca93: I'm not sure what would be an use-case that would be made significantly simpler by having this function. I'd propose we leave this out until somebody asks for it, but I can be convinced otherwise ;)

lib/include/jxl/decode.h
veluca93: This is the documentation on the "BoxTypeRaw()" function, which was identical to the non-raw version - now the non-raw version was modified, but as far as I can see the documentation of the raw one is still incorrect.

lib/include/jxl/decode.h
lvandeve: Changed to type[5]. This function is intended as a convenience over `JxlDecoderGetBoxType` because I expect the 99% usecase exists out of checking if it's "Exif", "xml ", etc..., not getting its type as 4 characters.

With JxlDecoderGetBoxType, that would need the code: "char type[5]; JxlDecoderGetBoxType(dec, type); if (!strcmp(type, "Exif") {....}". In this use case, the string you give is null terminated, so I think type[5] instead of type[4] is ok.

lib/include/jxl/decode.h
lvandeve: I'd say it isn't redundant, and I removed the wording of the comments here that make it say that it's the same as box type being "brob". Checking against "brob" is a bit too much like manually parsing, having `JxlDecoderIsBoxCompressed` lets the API parse the file format specifications details such as the string "brob" for you.

veluca93: Well, do we need the GetBoxTypeRaw function then? Why is that one useful if you have the IsCompressed one?

lib/include/jxl/decode.h
lvandeve: It allows allocating a buffer of the exact correct size, so JXL_DEC_BOX_NEED_MORE_OUTPUT won't occur and you won't allocate more than needed. For quick concatenation operatrions, this can be helpful.

lib/include/jxl/decode.h
lvandeve: Ah, fixed

lib/include/jxl/decode.h
jonsneyers: If you give an explicit string, then it's null terminated, and in the output of JxlDecoderGetBoxType I would return a null-terminated string, but here it might be useful to also be able to pass a pointer that only has 4 bytes, e.g. in the case the type you're comparing against comes from another ISOBMFF stream.

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * must be kept alive until then.
+ *
+ * @param dec decoder object
+ * @param data pointer to next bytes to write to
+ * @param size amount of bytes available starting from data
+ * @param get_raw_box_data whether to get raw, uncompressed box data in the case
+ * of a "brob" box. If false, automatically decompresses the box with brotli if
+ * it was compressed. This requires the decoder has been compiled with brotli
+ * support.
+ * @return JXL_DEC_ERROR if input was already set without releasing,
```
jonsneyers, line 970: What does "input was already set without releasing" mean?

jonsneyers:
It might be useful to add a (currently hypothetical) example application that uses this new API. For example an application that takes a jxl and dumps the XMP metadata to a file. Or an application that takes a jxl and produces a new jxl where all the compressed boxes are uncompressed (say so they can be processed by 'dumb' metadata tools).

I think the API as it is described here is OK, but it would be instructive to have an example application to see if we're not missing something.

Also (not for this pull request, obviously), I think (as we have discussed before) we should add a simpler high-level API that doesn't do things in the streaming+events way but that you just give a filename or memory pointer and gives you an image buffer + uncompressed metadata blobs without you having to worry about memory allocation and all that. For simple applications, it would be good to have a convenient high-level one-shot API that you can use very easily, something like
```
char *filename;
uint8_t *display_profile; // points to an ICC profile describing the display space, or nullptr for sRGB
uint8_t *rgba; // will contain 8-bit interleaved RGBA data, possibly multiple frames 
JxlBlobs *blobs;  // struct that contains a list of tuples of (blobtype, data, data_size)
size_t w, h;
JxlAnimInfo *animinfo; // struct that contains nb of frames and frame duration

JxlRetCode = JxlSimpleDecodeRGBA8(filename, display_profile, rgba, blobs, &w, &h, animinfo);
if (JxlRetCode != JxlOK) { ... } // oops, something went wrong
if (animinfo->nb_frames > 1) { ... } // render as an animation
// yay we have an image we can put on the screen
JxlBlob exif = JxlGetBlob(blobs, "Exif");
if (exif.size > 0) {
  MyViewerAppShowExifMetadata(exif.data, exif.size);
}
```

jonsneyers (approved):

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ */
+JXL_EXPORT size_t JxlDecoderReleaseBoxBuffer(JxlDecoder* dec);
+
+/**
+ * Configures whether to get boxes in raw mode or in uncompressed
+ * mode.
+ *
+ * The default mode is non-raw (JXL_FALSE), where "brob" boxes are transparently
+ * decompressed. This setting can only be changed
+ * before decoding, or directly after a JXL_DEC_BOX event, and is preserved
```
deymo, line 996: preserved until? until reset?

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * JxlDecoderGetBoxType.
+ *
+ * @param dec decoder object
+ * @param type box type to compare with, e.g. "Exif". Typenames are 4 non-null
+ * characters, passing a const char* string with null terminator is possible but
+ * not required.
+ * @return JXL_TRUE if the box type is available and equal to the given type,
+ * JXL_FALSE otherwise.
+ */
+JXL_EXPORT JXL_BOOL JxlDecoderBoxTypeEquals(JxlDecoder* dec,
```
deymo, line 1030: Why do we need this function if we have a getter for the box type? isn't this just memcmp() of the GgetBoxType value?

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * characters plus a null termination character. In case of a compressed box,
+ * this returns the type of the underlying uncompressed box, not "brob". Use
+ * @see JxlDecoderGetBoxTypeRaw to get the raw type.
+ *
+ * @param dec decoder object
+ * @param type buffer to copy the type into
+ * @return JXL_DEC_SUCCESS if the value is available, JXL_DEC_ERROR if not, for
+ *    example the JXL file does not use the container format.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxType(JxlDecoder* dec, char type[5]);
```
deymo, line 1017: I would propose to use a `typedef struct { char box_type[4]; } JxlBoxType` instead of the raw char type[] that's sometimes 5 bytes and sometimes 4 bytes.

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * decompressed, unless JxlDecoderSetBoxModeRaw is enabled, so this function is
+ * not needed to just get the data, as long as the library is compiled with
+ * brotli.
+ *
+ * @param dec decoder object
+ * @return JXL_TRUE if the current box is a compressed box, false if it's an
+ * uncompressed box, or there's no box such as in case the container format is
+ * not used.
+ */
+JXL_EXPORT JXL_BOOL JxlDecoderIsBoxCompressed(JxlDecoder* dec);
```
deymo, line 1075: Is this equivalent to BoxTypeRaw == "brob", right? Not sure this deserves its own function.

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * set by JxlDecoderSetBoxBuffer, or 0 if no buffer is set or
+ * JxlDecoderReleaseBoxBuffer was already called.
+ */
+JXL_EXPORT size_t JxlDecoderReleaseBoxBuffer(JxlDecoder* dec);
+
+/**
+ * Configures whether to get boxes in raw mode or in uncompressed
+ * mode.
+ *
+ * The default mode is non-raw (JXL_FALSE), where "brob" boxes are transparently
```
deymo, line 994: Would the default still be non-raw if brotli-decoder support is not enabled/compiled? I would propose that the default is true so you have the same behavior whether you have support enabled or not. Applications that don't support decoding "brob" boxes would declare so by calling this function; applications that do support calling brotli (or just iterate over all the boxes without looking into their contents) would not call it.

I understand that the common case of a decoder getting an exif box would need to call this function to enable brotli-decoding; but that would give the application a point of failure where to detect that it is not supported.

lib/include/jxl/decode.h
```diff
@@ -913,6 +953,144 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+ * decompressed. This setting can only be changed
+ * before decoding, or directly after a JXL_DEC_BOX event, and is preserved
+ * otherwise.
+ *
+ * @param dec decoder object
+ * @param enable_raw whether to enable or disable raw mode
+ * @return JXL_DEC_ERROR if input was already set without releasing,
```
deymo, line 1001: return error if brotli decoding is not supported too?

lib/include/jxl/decode.h
```diff
@@ -238,6 +243,42 @@ typedef enum {
...
+   * @see JxlDecoderIsBoxCompressed to query whether the box was stored as a
+   * compressed box of type "brob", or directly uncompressed.
+   * @see JxlDecoderGetBoxTypeRaw get the 4-character raw, "brob" if compressed,
+   * box typename.
+   * @see JxlDecoderBoxTypeRawEquals to compare with the raw box typename.
+   * @see JxlDecoderGetBoxSizeRaw to get the raw compressed box size.
+   * @see JxlDecoderSetBoxModeRaw to configure whether to get the box
+   * data uncompressed, or possibly compressed.
+   *
+    Getting the raw box contents returns the same data and type unless
```
deymo, line 264: nit: missing `*` here and next line.

lib/include/jxl/decode.h
```diff
@@ -238,6 +243,42 @@ typedef enum {
    * image and always before JXL_DEC_FULL_IMAGE.
    */
   JXL_DEC_JPEG_RECONSTRUCTION = 0x2000,
+
+  /** Informative event by JxlDecoderProcessInput: The header of a box of the
```
deymo, line 247: The data we get in the BoxBuffer would be meant to be decompressed by the application or not based on   the value that can be set with `JxlDecoderSetBoxModeRaw`; so its default value is very important. This is like having a default value for the JxlPixelFormat of RGB 8bit. I think this API would be more robust (harder to use wrong) if we don't have a reliance on this default value behavior, for example if the "raw mode" is set at the time you set the buffer like in the the pixel format case. WDYT?

lib/include/jxl/decode.h
lvandeve: This is for convenience because in 99% of the usecases, you only want to compare to a type, like `JxlDecoderBoxTypeEquals("Exif", dec)`, which saves a temporary variable and C style string comparison

However, both you and Jon pointed out the redundancy, so I can remove this function after all and make the API more barebones

lib/include/jxl/decode.h
lvandeve: This redundancy too was pointed out by two people, so I'll remove it too.

Note that handling type == "brob" requires specification knowledge, so in a sense the user is writing the parser, while JxlDecoderIsBoxCompressed would wrap this

lib/include/jxl/decode.h
jonsneyers: I think if you subscribe to the event, in most applications you'll want to get the uncompressed blob. The exception would be bitstream muxing tools that want to copy boxes as-is.

lvandeve: I considered that, the difficulty is that you may have to call `JxlDecoderSetBoxBuffer` multiple times, because you don't know the size beforehand. So that means you need to pass this parameter multiple times and keep it in sync. It's better if you make the decision once, when you just get the JXL_DEC_BOX event. In fact, having to set it at each `JxlDecoderSetBoxBuffer` requires the user to keep extra state for this, which is too much.

A solution is to require calling `JxlDecoderSetBoxModeRaw` after each JXL_DEC_BOX event, right before the first `JxlDecoderSetBoxBuffer` that belongs to this box

WDYT?

lib/include/jxl/decode.h
lvandeve: I've done this but with `typedef char JxlBoxType[4]`

This solves the ambiguity of 4 vs 5. But without struct, it also prevents double naming like "box_type.box_type"

lib/include/jxl/decode.h
lvandeve: Done

lib/include/jxl/decode.h
lvandeve: > It might be useful to add a (currently hypothetical) example application that uses this new API. For example an application that takes a jxl and dumps the XMP metadata to a file. Or an application that takes a jxl and produces a new jxl where all the compressed boxes are uncompressed (say so they can be processed by 'dumb' metadata tools).
> 
> I think the API as it is described here is OK, but it would be instructive to have an example application to see if we're not missing something.
> 
> Also (not for this pull request, obviously), I think (as we have discussed before) we should add a simpler high-level API that doesn't do things in the streaming+events way but that you just give a filename or memory pointer and gives you an image buffer + uncompressed metadata blobs without you having to worry about memory allocation and all that. For simple applications, it would be good to have a convenient high-level one-shot API that you can use very easily, something like
> 
> ```
> char *filename;
> uint8_t *display_profile; // points to an ICC profile describing the display space, or nullptr for sRGB
> uint8_t *rgba; // will contain 8-bit interleaved RGBA data, possibly multiple frames 
> JxlBlobs *blobs;  // struct that contains a list of tuples of (blobtype, data, data_size)
> size_t w, h;
> JxlAnimInfo *animinfo; // struct that contains nb of frames and frame duration
> 
> JxlRetCode = JxlSimpleDecodeRGBA8(filename, display_profile, rgba, blobs, &w, &h, animinfo);
> if (JxlRetCode != JxlOK) { ... } // oops, something went wrong
> if (animinfo->nb_frames > 1) { ... } // render as an animation
> // yay we have an image we can put on the screen
> JxlBlob exif = JxlGetBlob(blobs, "Exif");
> if (exif.size > 0) {
>   MyViewerAppShowExifMetadata(exif.data, exif.size);
> }
> ```

Adding an example is a good idea, done!

I have not yet integrated the example in the build script because it of course can't yet link

lvandeve:
> It might be useful to add a (currently hypothetical) example application that uses this new API. For example an application that takes a jxl and dumps the XMP metadata to a file. Or an application that takes a jxl and produces a new jxl where all the compressed boxes are uncompressed (say so they can be processed by 'dumb' metadata tools).
> 
> I think the API as it is described here is OK, but it would be instructive to have an example application to see if we're not missing something.
> 
> Also (not for this pull request, obviously), I think (as we have discussed before) we should add a simpler high-level API that doesn't do things in the streaming+events way but that you just give a filename or memory pointer and gives you an image buffer + uncompressed metadata blobs without you having to worry about memory allocation and all that. For simple applications, it would be good to have a convenient high-level one-shot API that you can use very easily, something like
> 
> ```
> char *filename;
> uint8_t *display_profile; // points to an ICC profile describing the display space, or nullptr for sRGB
> uint8_t *rgba; // will contain 8-bit interleaved RGBA data, possibly multiple frames 
> JxlBlobs *blobs;  // struct that contains a list of tuples of (blobtype, data, data_size)
> size_t w, h;
> JxlAnimInfo *animinfo; // struct that contains nb of frames and frame duration
> 
> JxlRetCode = JxlSimpleDecodeRGBA8(filename, display_profile, rgba, blobs, &w, &h, animinfo);
> if (JxlRetCode != JxlOK) { ... } // oops, something went wrong
> if (animinfo->nb_frames > 1) { ... } // render as an animation
> // yay we have an image we can put on the screen
> JxlBlob exif = JxlGetBlob(blobs, "Exif");
> if (exif.size > 0) {
>   MyViewerAppShowExifMetadata(exif.data, exif.size);
> }
> ```

Adding an example is a good idea, done!

I have not yet integrated the example in the build script because it of course can't yet link

lib/include/jxl/decode.h
lvandeve: I don't mind changing the default mode to raw, and don't do brotli compression in the API implementation by default.

@veluca93 @jonsneyers any objections to this from your side?

jonsneyers: Ok for a low-level api. In a future high-level api (that is only available in non-size-constrained full builds), we can have convenience functions for applications that only need simple functionality (i.e. getting a decoded image and uncompressed metadata from a file on disk).

veluca93: I think it's OK.

lib/include/jxl/decode.h
deymo: If you make the default raw, basically forcing the user to call `JxlDecoderSetBoxModeRaw()`, I think it would be ok to remember the last state. I would perhaps use a different name than "box mode raw", something like JxlDecoderDecompressBoxes(true/false).

lib/include/jxl/decode.h
lvandeve: This was a copypasting error from "JxlDecoderSetInput", fixed

lib/include/jxl/decode.h
lvandeve: Resolved, since the equals function is removed

lib/include/jxl/decode.h
```diff
@@ -162,14 +162,19 @@ typedef enum {
    */
   JXL_DEC_NEED_IMAGE_OUT_BUFFER = 5,
 
-  /** Informative event by JxlDecoderProcessInput: JPEG reconstruction buffer is
-   * too small for reconstructed JPEG codestream to fit.
-   * JxlDecoderSetJPEGBuffer must be called again to make room for remaining
-   * bytes. This event may occur multiple times after
-   * JXL_DEC_JPEG_RECONSTRUCTION
+  /** The JPEG reconstruction buffer is too small for reconstructed JPEG
+   * codestream to fit. JxlDecoderSetJPEGBuffer must be called again to make
```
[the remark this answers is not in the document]
lvandeve: The implementation of the API would ideally output at least some data before calling this function again, but it shouldn't be required to by documentation and user should be prepared to handle the case where it doesn't, to be more generic.

That doesn't mean we shouldn't make the implementation do it because it's more effective when it outputs something, than when it continuously asks the user for more bytes without outputting anything, but no promises in the header should be made.

Speaking about this, I added comment to JxlDecoderSetBoxBuffer and JxlDecoderSetJPEGBuffer that specifies if you released a buffer, the next time  you set it, you should not include the already-output bytes in the buffer.

lib/include/jxl/decode.h
lvandeve: Done and renamed to JxlDecoderDecompressBoxes

lib/include/jxl/decode.h
lvandeve: Done, and remembers the setting until decoder reset now

lib/include/jxl/decode.h
lvandeve: It now only requires it for decompressed mode. Having the small vs non-small library with brotli dependency is for a different pull request.

lib/include/jxl/types.h
```diff
@@ -109,6 +109,10 @@ typedef struct {
   size_t align;
 } JxlPixelFormat;
 
+/** Data type holding the 4-character type name of an ISOBMFF box.
+ */
+typedef char JxlBoxType[4];
```
deymo, line 114: I think it would be useful to add some constants/defines for some of the boxes, maybe for BROB, just `{'b', 'r', 'o', 'b'}`, but we can do this later.

deymo (approved):

lib/include/jxl/decode.h
```diff
@@ -913,6 +952,102 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxType(JxlDecoder* dec,
+                                                 JxlBoxType* type,
+                                                 JXL_BOOL decompressed);
+
+/**
+ * Returns the size of a box as it appears in the container file, after the
+ * JXL_DEC_BOX event. For a non-compressed box, this is the size of the
+ * contents, excluding the 4 bytes indicating the box type. For a compressed
+ * "brob" box, this is the size of the compressed box contents plus the
+ * additional 4 byte indicating the underlying box type, but excluding the 4
```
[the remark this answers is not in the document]
lvandeve: Yes it includes this, updated the comment to say this gives the size of the data written in the output buffer when in raw mode

lib/include/jxl/decode.h
```diff
@@ -410,8 +446,8 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec);
  * @param dec decoder object
  * @param data pointer to next bytes to read from
  * @param size amount of bytes available starting from data
- * @return JXL_DEC_ERROR if input was already set without releasing,
- * JXL_DEC_SUCCESS otherwise
+ * @return JXL_DEC_ERROR if input was already set without releasing, or brotli
+ * is not available and non-raw output is requested, JXL_DEC_SUCCESS otherwise.
```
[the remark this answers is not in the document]
lvandeve: Removed this stale comment

lib/include/jxl/decode.h
```diff
@@ -913,6 +952,102 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+
+/**
+ * Outputs the type of the current box, after a JXL_DEC_BOX event occured, as 4
+ * characters plus a null termination character. In case of a compressed "brob"
+ * box, this will return "brob" if the decompressed argument is JXL_FALSE, or
+ * the underlying box type if the decompressed argument is JXL_TRUE.
+ *
+ * @param dec decoder object
+ * @param type buffer to copy the type into
+ * @param decompressed whether to get the raw box type, or the underlying
```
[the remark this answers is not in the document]
lvandeve: Reworded to mention the true and the false behavior

lib/include/jxl/decode.h
```diff
@@ -913,6 +952,102 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetJPEGBuffer(JxlDecoder* dec,
...
+                                                 JXL_BOOL decompressed);
+
+/**
+ * Returns the size of a box as it appears in the container file, after the
+ * JXL_DEC_BOX event. For a non-compressed box, this is the size of the
+ * contents, excluding the 4 bytes indicating the box type. For a compressed
+ * "brob" box, this is the size of the compressed box contents plus the
+ * additional 4 byte indicating the underlying box type, but excluding the 4
+ * bytes indicating "brob". This function is only useful when getting boxes in
+ * raw mode, in the default mode where brob boxes are decompressed, this does
```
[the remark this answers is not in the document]
lvandeve: Fixed

lib/include/jxl/types.h
lvandeve: I agree that we can do it later

deymo (approved):

PR #553, opened 2021-09-08
JNI: add an option to perform color transform for output pixels.

eustas:


jonsneyers:
Could you add a description that is slightly more elaborate than the commit message? What's the name of this option, what exactly does it do, etc.

deymo:
This is departing the JNI wrapper more from the C API. What's the overall plan for supporting JNI here? Last time we discussed this topic I got the impression that the JNI wrapper was more of an experiment.

lvandeve:
We may also add a simpler library that does color conversions to the C API, so might be better to only support this then. This could become a utility, or an example, in Java instead?

KoromaruKoruko:
> This is departing the JNI wrapper more from the C API. What's the overall plan for supporting JNI here? Last time we discussed this topic I got the impression that the JNI wrapper was more of an experiment.

Currently my plan is to see about tying in the jni wrapper with the wasm wrapper. by using the same form of preprocessed code generation. but before I work on other interfaces, I have to get wasm fully implemented.

check out #1136 that's where I'll be posting the preprocessor, and converted public facing api part of the code, along with were the discussion on it should be.

PR #557, opened 2021-09-08
Add GIMP Plugin to debian packaging

xiota:
Added changes to the `debian` folder that I've been using to build the GIMP plugin on launchpad.
Also added `clang` to the build dependencies, though it's not needed and can be removed.
Removed `debian/tmp/` from the install files, since it doesn't seem to be necessary.


debian/control
```diff
@@ -63,3 +66,12 @@ Description: JPEG XL Image Coding System - "JXL" (shared libraries)
  several features that help transition from the legacy JPEG format.
  .
  This package installs shared libraries.
+
+Package: libjxl-gimp
+Architecture: any
+Multi-Arch: same
+Section: universe/graphics
```
veluca93, line 73: I guess "universe" here is ubuntu-specific, no?

xiota: o, I don't know... I'll go grab a debian package and see.  Think I just copied from an existing gimp plugin at the time.

debian/control
```diff
@@ -5,10 +5,12 @@ Priority: optional
 Standards-Version: 3.9.8
 Build-Depends:
  asciidoc,
+ clang,
```
deymo, line 8: we weren't compiling the debian package with clang. We could, but that's a totally separated discussion from GIMP, unless GIMP plugin doesn't compile with gcc which would be a problem we need to fix.

debian/control
```diff
@@ -63,3 +66,12 @@ Description: JPEG XL Image Coding System - "JXL" (shared libraries)
  several features that help transition from the legacy JPEG format.
  .
  This package installs shared libraries.
+
+Package: libjxl-gimp
```
deymo, line 70: I was just discussing this with @veluca93 . What do you think about `libjxl-gimp-plugin` for the binary package?

debian/libjxl-gimp.install
```diff
@@ -0,0 +1 @@
+usr/lib/gimp/2.0/plug-ins/file-jxl/file-jxl
```
deymo, line 1: This path comes from `gimp-2.0` pkg-config. I guess it is fine to hard-code it here since we hard-code the gimp-2.0? I don't know what's the common thing to do.

debian/rules
```diff
@@ -13,4 +13,5 @@ override_dh_auto_configure:
 	  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
 	  -DJPEGXL_FORCE_SYSTEM_GTEST=ON \
 	  -DJPEGXL_FORCE_SYSTEM_BROTLI=ON \
-	  -DJPEGXL_FORCE_SYSTEM_HWY=ON
+	  -DJPEGXL_FORCE_SYSTEM_HWY=ON \
+	  -DJPEGXL_ENABLE_PLUGINS=ON
```
deymo, line 17: I think it is about time to enable JPEGXL_ENABLE_GIMP_SAVING by default, but probably in the cmakefile, right? (not related to this PR).

debian/control
```diff
@@ -63,3 +66,12 @@ Description: JPEG XL Image Coding System - "JXL" (shared libraries)
...
+Architecture: any
+Multi-Arch: same
+Section: universe/graphics
+Depends: ${shlibs:Depends}, ${misc:Depends}
```
veluca93, line 74: Probably should depend on the libjxl package and on gimp.

debian/control
```diff
@@ -18,6 +20,7 @@ Build-Depends:
  libpng-dev,
  libwebp-dev,
  pkg-config,
+ xdg-utils,
```
deymo, line 23: what do we use this package for? I don't see it directly being used but maybe I'm missing something.

deymo (commented):
sgtm, thanks

debian/rules
veluca93: That flag was actually removed, so not a concern :)

debian/control
veluca93: (no strong opinions for me here)

debian/control
deymo: I think debian does some magic and looks at the shared library dependencies of your binaries, that's what the shlibs:Depends does. For other cases (like in the .a to .a dependency in libjxl-dev to libhwy-dev) you have to spell it out; but otherwise... magic 🤯️

debian/control
xiota: It is ubuntu-specific, I'll remove it.

debian/control
xiota: The plugin builds fine without clang.  Was just wondering, since the docs say clang is recommended.   Will remove.

debian/control
xiota: I'm fine with that... another option was gimp-libjxl, since I noticed most gimp plugins are named gimp-*  I'll change it to libjxl-gimp-plugin

debian/libjxl-gimp.install
xiota: I just looked in another plugin, and they just have `/usr/lib/gimp`.  I'll change it.

debian/control
xiota: I'll remove it and see if the test builds work.  I don't remember what it was needed for, but something didn't work in the past without it.

debian/control
xiota: I think the plugin may be linked statically.  On Windows, it works without any of the other libjxl dlls.

veluca93: no no, I checked - the dependency is added in the debian package automagically :)

veluca93 (approved):

debian/control
xiota: xdg-utils used to be required by the old gdk plugin (at least, it was in cmakelists), so the entire plugins folder wouldn't build.  Looks like it isn't needed anymore.

xiota:
@googlebot I consent

PR #558, opened 2021-09-08
Add a debian package for the GDK plugin

veluca93:


debian/control
```diff
@@ -19,6 +20,7 @@ Build-Depends:
  libpng-dev,
  libwebp-dev,
  pkg-config,
+ xdg-utils,
```
deymo, line 23: This is for the test only, right? I think we should install xvfb for this test to pass in the builders. Did you test this in github?

deymo: I'm also ok not running this test in the Debian package build.

veluca93: No no, this is for xdg-mime, which apparently gets executed I think in cmake install... The pipeline fails without this either way (and works with this line ;))

debian/libjxl-gdk-pixbuf.install
```diff
@@ -0,0 +1,3 @@
+usr/lib/x86_64-linux-gnu/*/*/loaders/*
```
deymo, line 1: This wouldn't work for other archs, right?

veluca93: Very true. Fixed.

deymo (approved):

PR #560, opened 2021-09-08
Remove GitLab asan/msan fast_benchmarks

deymo:
asan/msan build were already removed.


deymo:
Validated it with ci/lint in GitLab.

veluca93 (approved):

Issue #447, opened 2021-08-14
enc_modular.cc:506: JXL_ASSERT: !fp


dlemstra:
**Describe the bug**
When encoding a lossless image with the `C API` the following assert is being raised: `enc_modular.cc:506: JXL_ASSERT: !fp`.

**To Reproduce**
We are getting this exception inside @ImageMagick and it suspect that it is happening because we are doing this:

```c
basic_info.bits_per_sample=32;
basic_info.exponent_bits_per_sample=8;
JxlEncoderSetBasicInfo(encoder,&basic_info)
```
This will set `enc->metadata.m.xyb_encoded` to `true` because we did not set `info->uses_original_profile` to `JXL_TRUE` because we did not include a color profile.

Then we call the method `JxlEncoderOptionsSetLossless` that sets `options->values.lossless` to `JXL_TRUE`. The next method we call is `JxlEncoderAddImageFrame` and this is happening inside here:

```c
if (options->values.lossless) {
  queued_frame->option_values.cparams.SetLossless();
}
```

And that does this:

```c
void SetLossless() {
  modular_mode = true;
  quality_pair.first = 100;
  quality_pair.second = 100;
  color_transform = jxl::ColorTransform::kNone;
}
```

At this point the `color_transform` is set to `kNone` as it should be.

Then we start to call `JxlEncoderProcessOutput` that at some point calls the method `RefillOutputByteQueue` that does the following:

```C
  if (metadata.m.xyb_encoded) {
    input_frame->option_values.cparams.color_transform =
        jxl::ColorTransform::kXYB;
  } else {
    // TODO(zond): Figure out when to use kYCbCr instead.
    input_frame->option_values.cparams.color_transform =
        jxl::ColorTransform::kNone;
  }
```

Because the value `metadata.m.xyb_encoded` is set to true the `color_transform` is incorrectly changed to `kXYB`.

**Environment**
 - OS: Windows
 - Compiler version: Visual Studio 2019
 - CPU type: x86_64
 - cjxl/djxl version string: v0.5

jonsneyers:
I suggest we make `xyb_encoded` an Override instead of a bool, and initialize it to a default value that will be interpreted as false if lossless frames are added and true if lossy frames are added. WDYT, @lvandeve ?

jonsneyers:
https://github.com/libjxl/libjxl/pull/562 makes modular float + XYB work, but I think we still need to better document the encode API and make it more clear that

[...]

PR #562, opened 2021-09-09
allow combination of float + XYB in modular mode

jonsneyers:
See also https://github.com/libjxl/libjxl/issues/447

Original pull request title: ~~don't allow invalid combination of lossless-fp + XYB/YCbCr ~~

~~The image header can signal that the image data is floating point (so not unsigned integers but a sign-exp-mantissa representation). This is meant to be used for lossless compression of PFM or EXR data. It is not meant to be used in combination with XYB or YCbCr color transforms. In VarDCT mode, it did more or less work to signal fp data while doing XYB, since the original bitdepth is generally just informative when doing XYB. The problem was that Patches would still be modular-encoded, in XYB, leading to a bad combination of XYB with fp.~~

~~This pull request makes it obligatory to signal a uint bitdepth in the image header when using XYB, and just refuse (in the encode api) to set the basicinfo to do fp when XYB will be done. This is the easiest solution that makes most sense to me. There are some alternative options:~~

- ~~disable patches when the image header says it's fp~~
- ~~make fp + XYB work, somehow (likely not possible with int32 modular buffers)~~
- ~~silently override the float bitdepth from basicinfo to become uint when doing lossy~~

~~Note that pixel data can still be provided to the encoder in float form; that has nothing to do with what is signaled in the image header.~~

~~Also don't allow uint32 (which we don't support) and do allow signalling any integer bitdepth up to 24 in the encode api — for VarDCT this is just informative, but for lossless it is the precision that will actually be encoded, so we need to allow other bitdepths than 8 and 16 if we want to do for example lossless 10-bit.~~

After discussion with @lvandeve and checking what the spec actually says about this, it makes more sense to simply allow any bitdepth (int or float) in the case of XYB. We already ignore the bits_per_sample, so it makes sense to also ignore the exp_bits_per_sample. The spec is not super clear on how to deal with it, but we'll make it clearer in the second edition of the spec, and already anticipate it here. In any case it was wrong to consider float bitdepth + XYB an invalid combination, since there's nothing in the spec that forbids doing that.


CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Improved the 2x2 downscaling method in the encoder for the optional color
    channel resampling for low bit rates.
 
+### Changed
+ - API: When encoding in lossy mode, the exponent_bits_per_sample in JxlBasicInfo
```
lvandeve, line 25: this field was originally designed for lossy and represents only metadata about the original image. Is it maybe possible to only disable it in the situation where it causes issues?

lvandeve:
> This is meant to be used for lossless compression of PFM or EXR data. It is not meant to be used in combination with XYB or YCbCr color transforms.

I wouldn't say that's true, it can also signal that your original was PFM, when you encode with lossy vardct

jonsneyers:
Changed this PR and its description quite heavily :)

This is a better way to solve things than what I proposed originally.

CHANGELOG.md
jonsneyers: Since this PR changed completely, this question is no longer relevant :)

veluca93 (approved):

veluca93 (approved):

PR #564, opened 2021-09-09
Update win64:clang:release to use clang-8

eustas:


jonsneyers:
It would be good to add a PR description explaining why this is done, e.g. a pointer to an issue that this resolves.

deymo:
did you test it in the gitlab ci?

deymo:
I sent this to the CI and didn't actually work but for a different reason (gtest).

.gitlab-ci.yml
```diff
@@ -224,6 +224,9 @@ test:i686:clang:release:
 build:win64:clang:release:
   <<: *linux_host_build_template
   stage: build
+  variables:
```
deymo, line 227: ok, this is very annoying, but if you create a variables: dict here it replaces the one in the template. I think you are missing the recursive checkout. Just pass CC=clang-8 CXX=clang++-8 in the build line.

eustas: Updated PR.

How do I push it to gitlab CI?

eustas: Tested on local docker; builds and passes tests. PTAL

docker/scripts/mingw_w64_float_fix.patch
```diff
@@ -0,0 +1,25 @@
+commit 82b169c5734a6198d3b4c51a48f82e7b7104f143
```
deymo, line 1: Can you copy the contents of https://github.com/msys2-contrib/mingw-w64/blob/master/COPYING at the top of this file?

deymo: @eustas  ping?

PR #569, opened 2021-09-10
Fix upsampling in combination with line-by-line output.

veluca93:


lvandeve (approved):

lvandeve:
It would be good if we have tests for the conformance test images that decode with multiple ways (line output, not line output, ...) to catch cases like this

PR #574, opened 2021-09-10
Add a test PQ image

sboukortt:
It goes from 10% PQ signal level to 90%, so from ~0.33 cd/m² to ~4000 cd/m², uniformly in PQ space.

Created using:

```shell
convert -size 17x1 -depth 16 -define gradient:direction=East gradient:gray10-gray90 -sample 6400% pq_gradient.pgm
tools/cjxl -m -x color_space=Gra_D65_Rel_PeQ pq_gradient.pgm pq_gradient.jxl
```


veluca93 (approved):

jonsneyers:
What is the goal of this test image?

In non-HDR mode Chrome, this gets rendered in a probably very similar way as in HDR mode Chrome: as 17 shades of gray, getting brighter from left to right. So it's won't really help to know based on that image if your HDR is working or not...

veluca93:
> What is the goal of this test image?
> 
> In non-HDR mode Chrome, this gets rendered in a probably very similar way as in HDR mode Chrome: as 17 shades of gray, getting brighter from left to right. So it's won't really help to know based on that image if your HDR is working or not...

At least on Linux, in non-HDR mode the first few steps are indistinguishable - and in HDR mode on Windows, the last few aren't distinguishable unless you have a *very* bright monitor.

PR #575, opened 2021-09-10
Implement cmake's BUILD_SHARED_LIBS setting.

deymo:
The current behavior was to build both shared and static libraries since
tests and some tools use internal symbols not available in the shared
libraries. We support passing `-DJPEGXL_STATIC=ON` to build all the
binaries as static binaries, which causes them to link against the
static version of the libraries but also link as a static binaries with
all the problems that such thing brings.

The new `-DBUILD_SHARED_LIBS=OFF` setting, which is the standard name
in cmake for this behavior, would make jxl, jxl_dec and jxl_threads all
be aliased to the -static targets so tools and plugins link against the
static version of those.


deymo:
@xiota this should do what you want of building all tools against the jxl static libs. libpng, libjpeg, etc depends on how you get those dependencies, but since we don't bundle them here if they are already shared libs they would be already "installed" in your system. Brotli, if built from source here, will use the static lib when this setting is off.

xiota:
Thanks.  I'll try it.

veluca93 (approved):

xiota:
On Windows, building with `-DBUILD_SHARED_LIBS=OFF` complains about dlls it didn't complain about before.   I'll try `-DJPEGXL_STATIC=ON` again to see if that works now.

xiota:
Binaries built on msys2 with clang and `BUILD_SHARED_LIBS=OFF` work.

PR #582, opened 2021-09-13
Fix building with LLVM LLD: check "-Wl,--exclude-libs=ALL"

misakikasumi:
In 63a481a, linker flag `-Wl,--exclude-libs=ALL` is added. However, LLVM LLD does not support it. Building with LLVM toolchain would result in failure.

This PR check whether the linker support this flag. If it does not, the flag will not be added. A not-so-pretty method with `CheckCSourceCompiles` is used, because CMake 3.10 does not support `CheckLinkerFlag` or `CMAKE_REQUIRED_LINK_OPTIONS`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,4 @@ Pieter Wuille
 xiota
 Ziemowit Zabawa <ziemek.zabawa@outlook.com>
 Andrius Lukas Narbutas <andrius4669@gmail.com>
+Misaki Kasumi <misakikasumi@outlook.com>
--- a/lib/jxl.cmake
+++ b/lib/jxl.cmake
@@ -523,6 +523,13 @@ set_target_properties(jxl_dec PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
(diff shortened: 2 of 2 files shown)


misakikasumi:
Further investigation shows that only ELF lld supports ``exclude-libs`` but not MinGW, COFF, or MachO.

misakikasumi:
I'm wondering what's the equivalent of `exclude-libs` on those platforms. Is it `exclude-all-symbols`?

SchrodingerZhu:
@misakikasumi 

> ``Specifies a list of archive libraries from which symbols should not be automatically exported. The library names may be delimited by commas or colons. Specifying --exclude-libs ALL excludes symbols in all archive libraries from automatic export. This option is available only for the i386 PE targeted port of the linker and for ELF targeted ports. For i386 PE, symbols explicitly listed in a .def file are still exported, regardless of this option. For ELF targeted ports, symbols affected by this option will be treated as hidden.

misakikasumi:
@SchrodingerZhu Got it. Thank you.

For MinGW backend, a symbol will not be exported unless it is exported explicitly. So no special flag needed.

misakikasumi:
@googlebot I signed it!

deymo (approved):
Thanks for your contribution. Please add yourself to the AUTHORS file in the same commit and reupload (force push is fine).

deymo (approved):

PR #583, opened 2021-09-14
Adapt conformance testing to 18181-3.

veluca93:
Also implement metadata and color comparison when no colorspace
conversion is required.


jonsneyers (approved):

tools/conformance/generator.py
```diff
@@ -20,100 +19,111 @@
...
+    parser.add_argument('--peak_error',
+                        metavar='PEAK_ERROR',
+                        type=float,
+                        required=True,
+                        help='peak error for each testcase')
+    parser.add_argument('--rmse',
+                        metavar='RMSE',
+                        type=float,
+                        required=True,
+                        help='max RMSE for each testcase')
```
deymo, line 119: shouldn't this be different per test-case?

veluca93: It could also be different per-frame - I think we can manually modify it when needed, and merge multiple corpora otherwise.

tools/conformance/djxl_conformance.cc
```diff
@@ -236,16 +233,79 @@ class JSONArray : public JSONField {
...
-  uint32_t events = JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING;
+  uint32_t events =
+      JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING | JXL_DEC_PREVIEW_IMAGE;
   if (opts.pixel_prefix) events |= JXL_DEC_FRAME | JXL_DEC_FULL_IMAGE;
-  if (opts.jpeg_path) events |= JXL_DEC_JPEG_RECONSTRUCTION;
-  if (opts.preview) events |= JXL_DEC_PREVIEW_IMAGE;
   // We need to output the frame header info in the metadata.
   if (opts.metadata_path) events |= JXL_DEC_FRAME;
 
+  if (opts.jpeg_path) {
```
deymo, line 305: Why not do `events = JXL_DEC_JPEG_RECONSTRUCTION` in this case if this is what you want? (if jpeg_path then nothing else) I think it should be possible to get both from the API.

veluca93: @lvandeve AFAIU you cannot get both JPEG and pixels, right?

deymo: I was proposing to rollback the part that creates a new function with a new decoder and decoder loop and instead set `events = JXL_DEC_JPEG_RECONSTRUCTION` here. You'd need to handle whether jpeg is written after the loop. That's not the same as getting both JPEG and pixels (which should be allowed IMHO but that's a separate question, at least basic info should be allowed).

veluca93: I'm not sure this would be helpful - `djxl_conformance` needs to output both JPEG and pixels anyway, are you suggesting we use a single function that will do one or the other and then change the set of options passed to it?

jonsneyers (approved):

jonsneyers (approved):

PR #595, opened 2021-09-16
Support compiling GIMP plugin with glib < 2.62

deymo:
glib 2.61 and earlier versions didn't have the helper function
`g_clear_signal_handler` now used by the GIMP plugin. This makes it
compile for older distributions like Debian buster.

### DIFF
--- a/plugins/gimp/file-jxl-save.cc
+++ b/plugins/gimp/file-jxl-save.cc
@@ -5,6 +5,8 @@
 
 #include "plugins/gimp/file-jxl-save.h"
 
+#include "gobject/gsignal.h"
+
 #define PLUG_IN_BINARY "file-jxl"
 #define SAVE_PROC "file-jxl-save"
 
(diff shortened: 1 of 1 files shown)


deymo:
@xiota FYI

I don't know if this is a hack or the proper way to fix it. You never know with glib.

veluca93 (approved):

xiota:
Where is the plugin using `g_clear_signal_handler`?  What is the compiler error/warning?

If this is needed, can it be put in a separate file, like `gtk-compat.h/c`?  Or maybe there is a pre-existing compatibility package that can be used?

deymo:
There are 9 calls to `g_clear_signal_handler` in the plugin. I don't know of any compat library for this. The point is to compile the plugin with an old glib version installed, so we would need to use a *new* compat library with the old glib. Given that this is a simple helper function I though the easiest was to include it here.

deymo:
Here's an error message from the deploy pipeline in main (the one that creates the .deb files): https://github.com/libjxl/libjxl/runs/3621107836?check_suite_focus=true#step:13:2193

There is also a warning about `JpegXlSaveOpts::SetQuality` being unused BTW.

xiota:
@deymo Sorry, I was grepping a folder with an older version of the plugin earlier.  I was thinking about what if other parts of the plugin start using the same function.  Then it would be nice if all the compatibility stuff were together.  I suppose it can be split off later.

xiota:
Right now it's setting quality/distance and updating distance/quality in separate steps when needed, but since quality and distance are linked, it should be using SetQuality and SetDistance to avoid introducing bugs by forgetting the update step.  The change is in another branch I was working on.

deymo:
Cool, I'll merge this now to get the build back to green then.

PR #598, opened 2021-09-16
Add several options to the encoder api

lvandeve:
More to follow in next MR's

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -71,6 +71,51 @@ typedef enum {
 } JxlEncoderStatus;
 
 /**
+ * Id of options to set to JxlEncoderOptions.
+ */
+typedef enum {
+  /** Sets resampling option. If enabled, the image is downsampled before
+   * compression, and upsampled to original size in the decoder. Integer option,
+   * use 0 for the default behavior (resampling only applied for low quality),
+   * 1 for no downsampling (1x1), 2 for 2x2 downsampling, 4 for 4x4
+   * downsampling, 8 for 8x8 downsampling. The default value is 0.
+   */
+  JXL_ENC_OPTION_RESAMPLING = 0,
+
+  /** Similar to JXL_ENC_OPTION_RESAMPLING, but for extra channels. Integer
+   * option, use 1 for no downsampling (1x1), 2 for 2x2 downsampling, 4 for 4x4
+   * downsampling, 8 for 8x8 downsampling. The default value is 1.
+   */
+  JXL_ENC_OPTION_EXTRA_CHANNEL_RESAMPLING = 1,
+
+  /** Enables or disables noise generation. Integer option, use -1 for the
+   * encoder default, 0 to disable, 1 to enable. The
+   */
+  JXL_ENC_OPTION_NOISE = 2,
+
+  /** Enables or disables dots generation. Integer option, use -1 for the
+   * encoder default, 0 to disable, 1 to enable.
+   */
+  JXL_ENC_OPTION_DOTS = 3,
+
+  /** Enables or disables patches generation. Integer option, use -1 for the
+   * encoder default, 0 to disable, 1 to enable.
+   */
+  JXL_ENC_OPTION_PATCHES = 4,
+
+  /** Enables or disables the gaborish filter. Integer option, use -1 for the
+   * encoder default, 0 to disable, 1 to enable.
+   */
+  JXL_ENC_OPTION_GABORISH = 5,
+
+  /** Enum value not to be used as an option. This value is added to force the
+   * C compiler to have the enum to take a known size.
+   */
+  JXL_ENC_OPTION_FILL_ENUM = 65535,
+
+} JxlEncoderOptionId;
+
+/**
  * Creates an instance of JxlEncoder and initializes it.
  *
  * @p memory_manager will be used for all the library dynamic allocations made
@@ -332,14 +377,14 @@ JxlEncoderOptionsSetEffort(JxlEncoderOptions* options, int effort);
 
 /**
  * Sets the distance level for lossy compression: target max butteraugli
- *  distance, lower = higher quality. Range: 0 .. 15.
- *  0.0 = mathematically lossless (however, use JxlEncoderOptionsSetLossless to
- *  use true lossless).
- *  1.0 = visually lossless.
- *  Recommended range: 0.5 .. 3.0.
- *  Default value: 1.0.
- *  If JxlEncoderOptionsSetLossless is used, this value is unused and implied
- *  to be 0.
+ * distance, lower = higher quality. Range: 0 .. 15.
+ * 0.0 = mathematically lossless (however, use JxlEncoderOptionsSetLossless to
+ * use true lossless).
+ * 1.0 = visually lossless.
+ * Recommended range: 0.5 .. 3.0.
+ * Default value: 1.0.
+ * If JxlEncoderOptionsSetLossless is used, this value is unused and implied
+ * to be 0.
  *
  * @param options set of encoder options to update with the new mode.
  * @param distance the distance value to set.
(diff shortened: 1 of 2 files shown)


veluca93 (commented):
Just wondering, but why not

`JxlEncoderSetFloatOption(options, ENUM_VALUE, float_value`

and similar for the other types?

jonsneyers:
I am personally in favor of a very generic and simple encode option api that is just
```
JxlEncoderStatus JxlEncoderSetOption(const char* key, const char* value)
```
where the list of keys and their valid values are documented separately. Return type says whether the option is supported or not and if the value was valid.

That way, it will be really easy to ensure that cjxl and benchmark_xl but also imagemagick convert or libvips can use the same syntax for setting options — or at least the (key, value) pairs could be made the same across tools, even if they have different syntax for setting them: `cjxl --key=value` or `benchmark_xl --codec=jxl:key=value` or `convert foo.png -set jxl:key=value foo.jxl` or something like that.
Also the way they would implement option setting would be super easy that way: just pass any unknown options to the jxl encoder, choke on it if it says unsupported or error, otherwise fine and go on to the next option.

This also means we can add encoder options without changing the api, and we can have applications that can use fancy new encode options but still compile against older versions of the encoder too. It also means you can update libjxl and automatically get the new encode options in all applications that allow arbitrary option passing.

There are some disadvantages too of this approach, but overall I think that for encode options (only the optional options) the advantages outweigh the disadvantages.

lvandeve:
For certain types of encoder options (those that don't affect the user's code, but affect internal encoder decisions, especially for more experimental options that may change with future encoder technologies), I agree with the premise of the above suggestions: having an individual function that must be maintained for everything the encoder can do can become a maintanance problem full of deprecated functions and functions that change behavior as new encoder algorithms are implemented or removed.

The string idea can be very flexible, but I'd say too flexible for low-level API, and I'd be cautious with parsing numeric values and not having data types. The string idea is useful to expose jxl encoder parameters to users of other libraries like imagemagick, it could directly translate command line flags to jxl encoder settings. However, I don't think the low level jxl encoder library needs to have this functionality.

The enum idea may be a good middle ground, as it's easier to add more enum values and deprecate existing ones, than whole functions.

Another middle ground may be string names, but the values are fixed types, such as:

```
JxlEncoderStatus JxlEncoderSetFloatOption(options, const char* key, float value)
```

This option still requires the programmer to know the type, but the string names are easy to add and deprecate.

However I think the enum idea is still the more long-term stable idea, the strings may end up creating a too chaotic ecosystem later without header documentation that documents what's deprecated, as it could do with enum values.

jonsneyers:
Enum keys have the disadvantage that _something_ will still need to map string to enum, and it also requires updating all those string to enum tables in all applications that want to expose optional encode param setting, each time an option gets added. It's better than having one function per option, but it still doesn't really allow future-proof parameter passing (unless you want to encourage applications to allow passing raw numeric enum values and leave it up to users to remember that option 42 is for configuring the foobar heuristic).

As for the option values: I get that parsing numeric values from strings does seem a bit risky and out of scope for an encode library, but OTOH it is likely that in most cases, that needs to be done somewhere anyway, and then it might be better to implement it robustly in one place than ask all applications to do it, which will have the effect of causing more bugs overall but technically those are then bugs in the applications and not in libjxl so it's not our problem.

In some cases though, you might have the values in a number type already, and of course it would be silly to ask applications to print them to a string then.

So maybe an approach could be to have
```
JxlEncoderStatus JxlEncoderSetStringOption(options, const char* key, const char* value);
JxlEncoderStatus JxlEncoderSetFloatOption(options, const char* key, float value);
JxlEncoderStatus JxlEncoderSetUintOption(options, const char* key, uint32_t value);
```

where the SetStringOption does the relevant parsing depending on `key` (or returns without trying to parse anything if `key` is unknown) and calls the corresponding numeric type function?

lvandeve:
There could be a separate function that maps a string to an enum value or returns unknown/deprecated if not known or deprecated

Then the setting of the option is still with enum value, but there exists a method to map a string to one for those applications where you'd want to expose encoder settings to external users

deymo:
I don't the `char* key` is a good idea. There doesn't always need to be *something* that maps strings to enum values; that only needs to happen if you want to map an API like a command like generic key=value; but that's not required. For code using our c API and C functions they have access to the enum values.

It is also not necessarily true that setting new options doesn't require updating anything else. What if we set an option that lets you offset the x0,y0 of your frame? or telling that the data comes in some other format (like XYB instead of RGB) for optimization purposes?

`char* key` has the disadvantage that you can misspell keys and get them ignored. In the context of the C API I don't think a char* key and a char* value are a good idea (unless the setting is actually a string). Converting strings to numbers is its own problem; specially when you don't know if you expect a float or an int, depending on the setting.

lib/include/jxl/types.h
```diff
@@ -109,6 +109,20 @@ typedef struct {
...
+  /** Leaves the option to the default, enabled or disabled.
+   */
+  JXL_OVERRIDE_DEFAULT = -1,
+  /** Force-disables the option.
+   */
+  JXL_OVERRIDE_FALSE = 0,
+  /** Force-enables the option.
+   */
+  JXL_OVERRIDE_TRUE = 1,
+} JxlOptionOverride;
```
deymo, line 124: I would love to avoid having this "override" concept in the API.

lvandeve: A very rigid alternative is a separate enum for each setting, but I am reluctant to make it that rigid because the encoder settings themselves are not very rigid: there are many encoder settings, they will change, they will get new levels/strengths added, new ones will be added, old ones will get deprecated.

Integer values and booleans can be ok.

lvandeve:
I have updated this MR with an enum to represent the option id, and the function JxlEncoderOptionsSetIntegerOption to set an option using the enum as option id.

This MR only introduces this for the new options. A next MR will also use this for the existing options (effort, distance, ...).

PTAL

jonsneyers:
I think it would be useful to also include a mapping between strings and enum values, e.g. with a function that converts a string to an enum and a function that converts an enum to a string (returning an error status in both cases if it gets an unknown input).

The reason is that otherwise we are encouraging applications to make their own mapping if they need it (e.g. to do passthrough of command line options to encoder option settings), which means two things:

- they have to come up with their own strings, and will likely all pick different strings (e.g. cjxl calls it `resampling`, benchmark_xl calls it `rs`, imagemagick calls it `resample`, ffmpeg calls it `upsampling`, libvips calls it `sampling-factor` etc)
- they need to update their mapping each time something changes or gets added

If we provide the mapping, it is more likely that we get uniform parameter names (reducing confusion and frustration) and that applications that want to allow parameter passing can implement it once and have something that keeps working with newer encoders.

deymo:
The "strings" for the options won't be localized to any language, so not very user friendly as a solution to allow pass-through of options from command line. I don't think that encouraging a pass-through of options from a higher level to the API is a good idea; depending on the option may result in unexpected things happening or misuse of the API/bugs/exploits.

For applications that already have settings that map to one of our enums it would make more sense for those applications to use their name rather than our proposed one.

lib/include/jxl/encode.h
```diff
@@ -349,6 +394,25 @@ JxlEncoderOptionsSetEffort(JxlEncoderOptions* options, int effort);
...
+ * @param options set of encoder options to update with the new mode.
+ * @param option ID of the option to set.
+ * @param value Integer value to set for this option.
+ * @return JXL_ENC_SUCCESS if the operation was successful, JXL_ENC_ERROR in
+ * case of an error, such as invalid or unknown option id, or invalid integer
+ * value for the given option. If an error is returned, the state of the
+ * JxlEncoderOptions object is still valid and is the same as before this
+ * function was called.
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderOptionsSetIntegerOption(
```
deymo, line 413: I'd suggest JxlEncoderOptionsSetInteger or JxlEncoderOptionsSetAsInteger to make it a bit less verbose; but this is fine.

lib/include/jxl/encode.h
```diff
@@ -70,6 +70,51 @@ typedef enum {
 
 } JxlEncoderStatus;
 
+/**
+ * Id of options to set to JxlEncoderOptions.
+ */
+typedef enum {
```
deymo, line 76: Please add a comment here explaining (for future developers) what the options should be numbered like since it seems that you are proposing some sparse numbering for a reason.

deymo (approved):

deymo (approved):

lib/include/jxl/encode.h
lvandeve: Done

lib/include/jxl/encode.h
lvandeve: Removed the numeric gaps for now.

Will insert more options in upcoming MRs, numbers can still change until next release

jonsneyers:
> The "strings" for the options won't be localized to any language, so not very user friendly as a solution to allow pass-through of options from command line. I don't think that encouraging a pass-through of options from a higher level to the API is a good idea; depending on the option may result in unexpected things happening or misuse of the API/bugs/exploits.
> 
> For applications that already have settings that map to one of our enums it would make more sense for those applications to use their name rather than our proposed one.

Is localization a thing for the command line options in CLI tools? Are there any relevant tools that do that? (ImageMagick? ffmpeg? libvips? GIMP scripts?)

I just think it will be annoying to have to update things in many places if a new optional encoder param gets exposed in the api. It means we have to keep cjxl, benchmark_xl, imagemagick convert, ffmpeg, libvips, etc. all in sync and have mappings from strings to enum values in all of them.

In terms of exploits: I only suggest to do this for optional encoder parameters, not basic ones like setting the source image or color profile. What kind of exploits could it lead to? A security-critical application can always choose not to allow unknown option passing and only use fixed enum values. Even without string maps, an application could implement arbitrary option passing by allowing to set options by their integer enum value, which is possibly what they will do if there is no other way to future-proof an integration.

deymo (approved):

PR #599, opened 2021-09-17
do not set default squeeze params in InvSqueeze

jonsneyers:
The normal decode flow is:
- input image has N channels
- it does Squeeze with default (empty) parameters
- at header decode, MetaSqueeze is called which will fill in the Squeeze params with default values, to get O(N log N) channels
- after channel decode, InvSqueeze is called which undoes the Squeezes (it now has actual parameters, since MetaSqueeze replaced the empty params with default params)

So there is no reason to set default squeeze params in InvSqueeze — if the params are still empty at that point, it's a bug.

### DIFF
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -77,7 +77,7 @@ TEST(JxlTest, HeaderSize) {
     io.Main().SetAlpha(std::move(alpha), /*alpha_is_premultiplied=*/false);
     AuxOut aux_out;
     Roundtrip(&io, cparams, dparams, pool, &io2, &aux_out);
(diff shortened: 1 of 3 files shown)


jonsneyers:
Correction: if the parameters are still empty at the point of InvSqueeze, it's not necessarily a bug: it could also be that the image is very small (less than 8x8) and the default parameters end up being just empty (no-op).

So the check can just be removed, and in case of empty params it just doesn't do anything.

Also changed the encoder to just not do Squeeze if it is a no-op Squeeze (saves 1 byte).

Updated the header size test with a tighter bound for the 1-pixel image with alpha.

PR #605, opened 2021-09-17
Generate readthedocs.io sphinx documentation for the C API reference.

deymo:
This is first step to generate the library API documentation in HTML
format for publication in readthedocs.io. At the moment only the
reference documentation is included (functions, symbols, enums, etc and
their definition) but no other introductory documentation.

Split things a bit between encoder, decoder, common structs, butteraugli
and threads for easier navigation of the documentation.

HTML pages are not automatically build with cmake on local builds, but
can be triggered locally with `ninja -C rtd-html`.

PR pipelines now check that `rtd-html` builds without errors and
readthedocs.io automatically triggers a build of the documentation
whenever we push to main.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -100,6 +100,11 @@ jobs:
       with:
         submodules: true
         fetch-depth: 2
+    - name: Sphinx dependencies
+      # Dependencies for sphinx HTML documentation
+      if: matrix.name == 'release'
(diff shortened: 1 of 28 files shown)


deymo:
You can view the documentation generated by this PR at: https://libjxl--605.org.readthedocs.build/en/605/

veluca93 (approved):

PR #606, opened 2021-09-17
decode_progressive: example application to demo/simulate/test progressive decoding

jonsneyers:
Adds an optional argument to `decode_oneshot` to do the decode in chunks of `N` bytes, saving the intermediate image when flushing after every chunk. This is useful to test/debug progressive decoding with the API (which is not quite the same thing as decoding a truncated image with `djxl --allow_partial_files --allow_more_progressive_steps`, since djxl doesn't use the API yet, and decoding a truncated image once is not really the same thing as doing multiple flushes during a single decode).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -19,6 +19,9 @@ pkg_check_modules(JxlThreads REQUIRED IMPORTED_TARGET libjxl_threads)
 add_executable(decode_oneshot decode_oneshot.cc)
 target_link_libraries(decode_oneshot PkgConfig::Jxl PkgConfig::JxlThreads)
 
+add_executable(decode_progressive decode_progressive.cc)
+target_link_libraries(decode_progressive PkgConfig::Jxl PkgConfig::JxlThreads)
+
 add_executable(encode_oneshot encode_oneshot.cc)
 target_link_libraries(encode_oneshot PkgConfig::Jxl PkgConfig::JxlThreads)
 
--- a/examples/decode_progressive.cc
+++ b/examples/decode_progressive.cc
@@ -1,0 +1,238 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This C++ example decodes a JPEG XL image progressively (input bytes are
+// passed in chunks). The example outputs the intermediate steps to PAM files.
+
+#include <inttypes.h>
+#include <limits.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <vector>
+
+#include "jxl/decode.h"
+#include "jxl/decode_cxx.h"
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/resizable_parallel_runner_cxx.h"
+
+bool WritePAM(const char* filename, const uint8_t* buffer, size_t w, size_t h) {
+  FILE* fp = fopen(filename, "wb");
+  if (!fp) {
+    fprintf(stderr, "Could not open %s for writing", filename);
+    return false;
+  }
+  fprintf(fp,
+          "P7\nWIDTH %" PRIu64 "\nHEIGHT %" PRIu64
+          "\nDEPTH 4\nMAXVAL 255\nTUPLTYPE "
+          "RGB_ALPHA\nENDHDR\n",
+          static_cast<uint64_t>(w), static_cast<uint64_t>(h));
+  fwrite(buffer, 1, w * h * 4, fp);
+  if (fclose(fp) != 0) {
+    return false;
+  }
+  return true;
+}
+
+/** Decodes JPEG XL image to 8-bit integer RGBA pixels and an ICC Profile, in a
+ * progressive way, saving the intermediate steps.
+ */
+bool DecodeJpegXlProgressive(const uint8_t* jxl, size_t size,
+                             const char* filename, size_t chunksize) {
+  std::vector<uint8_t> pixels;
+  std::vector<uint8_t> icc_profile;
+  size_t xsize = 0, ysize = 0;
+
+  // Multi-threaded parallel runner.
+  auto runner = JxlResizableParallelRunnerMake(nullptr);
+
+  auto dec = JxlDecoderMake(nullptr);
+  if (JXL_DEC_SUCCESS !=
+      JxlDecoderSubscribeEvents(dec.get(), JXL_DEC_BASIC_INFO |
+                                               JXL_DEC_COLOR_ENCODING |
+                                               JXL_DEC_FULL_IMAGE)) {
+    fprintf(stderr, "JxlDecoderSubscribeEvents failed\n");
+    return false;
+  }
+
+  if (JXL_DEC_SUCCESS != JxlDecoderSetParallelRunner(dec.get(),
+                                                     JxlResizableParallelRunner,
+                                                     runner.get())) {
+    fprintf(stderr, "JxlDecoderSetParallelRunner failed\n");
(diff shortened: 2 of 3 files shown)


examples/decode_oneshot.cc
```diff
@@ -19,6 +19,8 @@
 #include "jxl/resizable_parallel_runner.h"
 #include "jxl/resizable_parallel_runner_cxx.h"
 
+bool WritePFM(const char*, const float*, size_t, size_t);
```
deymo, line 22: Just move the function here.

deymo (commented):
decode_oneshot is the only straightforward example of using our decoder API we have; and the intention is to show how to do a "one shot" decoding. While having progressive decoding is a nice use case to show I think this makes it harder to understand what's going on and what's actually needed for the oneshot if you are not doing progressive. I would suggest writing a progressive example instead.

lvandeve:
What would you think about making a separate example for progressive decoding?

jonsneyers:
> What would you think about making a separate example for progressive decoding? To have one simple "just decode to pixels" example, and another example for handling progressive

Makes sense to keep the simple oneshot decode as simple as possible. The difference between the two would be rather small though (i.e. it would be mostly duplicated code, only real difference being how many bytes it feeds to the decoder and calling flush). Though maybe the differences could be made a bit larger (e.g. in the non-simple one, depend on extras to output to PNG, which would be useful to test progressive output in case of alpha).

jonsneyers:
Modified this pull request to make a new separate example, as requested by @lvandeve and @deymo.

examples/examples.cmake
```diff
@@ -5,6 +5,8 @@
 
 add_executable(decode_oneshot ${CMAKE_CURRENT_LIST_DIR}/decode_oneshot.cc)
 target_link_libraries(decode_oneshot jxl_dec jxl_threads)
+add_executable(decode_progressive ${CMAKE_CURRENT_LIST_DIR}/decode_progressive.cc)
```
deymo, line 8: Add this to the CMakeLists.txt too; CMakeLists.txt shows how to write an example project that uses libjxl.

examples/decode_progressive.cc
```diff
@@ -0,0 +1,307 @@
...
+    } else if (status == JXL_DEC_NEED_MORE_INPUT || status == JXL_DEC_SUCCESS ||
+               status == JXL_DEC_FULL_IMAGE) {
+      seen += remaining - JxlDecoderReleaseInput(dec.get());
+      printf("Flushing after %zu bytes\n", seen);
+      if (status == JXL_DEC_NEED_MORE_INPUT &&
+          JXL_DEC_SUCCESS != JxlDecoderFlushImage(dec.get())) {
+        printf("flush error (no preview yet)\n");
+      } else {
+        char png_filename[1024];
+        sprintf(png_filename, "%s-%zu.png", filename, seen);
```
deymo, line 132: `snprintf` and check the result.

examples/decode_progressive.cc
```diff
@@ -0,0 +1,307 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This C++ example decodes a JPEG XL image progressively (input bytes are passed
+// in chunks). The example outputs the intermediate steps to PNG files.
+
+#include <limits.h>
+#include <lodepng.h>
```
deymo, line 10: I would avoid the extra complexity in the example of writing png files.

jonsneyers: It's the easiest way to do RGBA output to easily viewable files though.

deymo: The purpose of this example is to show the API usage; not to create useful tools.

jonsneyers: Sure, but outputting to 8-bit RGBA is nice to have as an additional example besides the one we already have (the one that outputs to float).

deymo: If you install libjxl from a binary distribution and want to try out an example usage, you would need to also install lodepng source/dev files to save PNGs, which would be a blocker for trying out the example.

I don't think that passing num_channels=4 and JXL_TYPE_UINT8 makes much of a difference in the example source code.

Again, the goal of the examples/ directory is to show the API usage. The less lines you can use for that the better.

examples/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ pkg_check_modules(JxlThreads REQUIRED IMPORTED_TARGET libjxl_threads)
 add_executable(decode_oneshot decode_oneshot.cc)
 target_link_libraries(decode_oneshot PkgConfig::Jxl PkgConfig::JxlThreads)
 
+add_executable(decode_progressive decode_progressive.cc)
+target_link_libraries(decode_progressive PkgConfig::Jxl PkgConfig::Jxl lodepng)
```
deymo, line 23: "lodepng" is not defined in this context. This is an example project that a user may have. To compile this example with lodepng as a libjxl user we would need to also include a way to depend on lodepng from the example project. lodepng in particular is not a good project to depend on because it is not available in many places, and there's also no support on the package for easily include the libraries/headers paths (like a pkg-config .pc or a .cmake file would provide). This complexity outweighs any nice to have of an example that writes .png.

jonsneyers: I see. Maybe write .pam instead (can be done without dependencies, doesn't store icc though) and leave the png writing code there but commented out?

jonsneyers: Modified it to write .pam. The code to write png with lodepng is in a comment, for those who want to do that instead, but this way there is no dependency besides libjxl itself.

jonsneyers:
@lvandeve could you take a look? It is now a separate example application.

examples/decode_progressive.cc
```diff
@@ -0,0 +1,338 @@
...
+
+#include "jxl/decode.h"
+#include "jxl/decode_cxx.h"
+#include "jxl/resizable_parallel_runner.h"
+#include "jxl/resizable_parallel_runner_cxx.h"
+
+/*
+// Example code for outputting to PNG instead (requires lodepng)
+
+#include <lodepng.h>
```
mo271, line 25: Did you rebase after https://github.com/libjxl/libjxl/pull/923? Should here too libpng be used?

mo271 (dismissed):
What is the plan here once cjxl actually uses the c-api; will this example still be relevant?

examples/decode_progressive.cc
jonsneyers: Maybe I should just drop that big comment with a png example. The non-commented-out code outputs .pam, which is fine for ImageMagick but it's a bit silly that it's a format we don't actually support ourselves (maybe we should). I wanted something with alpha so progressive alpha can also be demo'ed with this.

jonsneyers:
> What is the plan here once cjxl actually uses the c-api; will this example still be relevant?

Well it's an example of non-oneshot decoding, so I suppose it might be useful. Once djxl uses the api, we could also have an option in it to simulate decoding of partial files

jonsneyers:
Oops accidentally closed - I shouldn't be writing comments on my phone. Anyway, I think it might be a useful example for showing how you can Flush, and to visually debug/illustrate progressive decode results.

examples/decode_progressive.cc
mo271: I see. Perhaps either do a libpng example or drop the example altogether, but I'm also fine with leaving the lodepng one as an example. It made a bit more sense when we still depended on it..

jonsneyers: I removed the commented out png encoding example, it's not very useful and it's not a good practice to have big chunks of commented out code like that.

mo271 (approved):

PR #612, opened 2021-09-20
Add brotli to deps.sh

StephanTLavavej:
Attempting to build `libjxl` from a [tarball](https://github.com/libjxl/libjxl/releases/tag/v0.5), on a system without `brotli` installed, fails with:

```
CMake Error at third_party/CMakeLists.txt:171 (message):
  Brotli not found, install brotli-dev or download brotli source code to
  third_party/brotli from https://github.com/google/brotli.  You can use
  /PATH/TO/deps.sh to download this dependency.
```

`deps.sh` doesn't download `brotli`, which this PR fixes.

The listed revision is what the submodule is currently set to, see https://github.com/libjxl/libjxl/tree/main/third_party .


deymo:
Brotli is available in most places nowadays and it is preferable to use the dependency from your system. Is installing `brotli-dev` or equivalent not an option for you?

StephanTLavavej:
It’s an option, but deps.sh seemed to be offering to do it for me, so I was confused when it didn’t work, and its omission seemed to be an oversight.

deymo:
`deps.sh` doesn't download dev packages for libgif, libjpeg-turbo, libpng, openexr either. For the ones `deps.sh` download, there's no easy "oh just install X which surely is available" which is the reason why we decided to include them from source. Maybe we can remove the brotli submodule already.

mo271 (approved):

mo271:
Thanks for you contribution, @StephanTLavavej

PR #617, opened 2021-09-21
add custom squeeze encode option (just for testing atm)

jonsneyers:
Currently the encoder always uses the default squeeze parameters.

This adds an encode option (only in cparams, not exposed in cjxl or api) to use custom squeeze parameters instead.

At the moment, it just does some random fixed custom squeeze, just for testing (improving code coverage a bit, checking that things roundtrip also with custom params). Perhaps at some point alternatives to the default squeeze params can be explored, e.g. to achieve better trade-offs between density and progressiveness.

### DIFF
--- a/lib/jxl/enc_params.h
+++ b/lib/jxl/enc_params.h
@@ -17,6 +17,7 @@
 #include "lib/jxl/butteraugli/butteraugli.h"
 #include "lib/jxl/frame_header.h"
 #include "lib/jxl/modular/options.h"
+#include "lib/jxl/modular/transform/transform.h"
 
 namespace jxl {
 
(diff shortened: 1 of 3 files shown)


lib/jxl/enc_params.h
```diff
@@ -215,6 +215,7 @@ struct CompressParams {
   // modular mode options below
   ModularOptions options;
   int responsive = -1;
+  bool customsqueeze = false;
```
veluca93, line 218: Why don't we just put a SqueezeParams here too?
(I'd say a std::optional but c++11..)

jonsneyers: A std::vector<SqueezeParams>? I guess that is the most flexible way to do it, yes. We likely will not really want to expose it like that in the encoder api though :)

veluca93: right, a vector of params would work too - empty means default.

jonsneyers:
Changed the cparam to a param vector and moved the test to modular_test.cc.

veluca93 (approved):

PR #618, opened 2021-09-21
fuzzer: integer overflow in MA decoding

jonsneyers:
Should fix a fuzzer-found integer overflow when a corrupt MA tree has a wildly out-of-range property value.


deymo (approved):
Looks harmless fuzzer bug.

PR #619, opened 2021-09-21
Remove Fields::Name() strings from non-debug builds.

deymo:
The const string returned by Name() ends up in the library and it is
only used in debug builds when Fields tracing is enabled. This patch
removes the Name() virtual method and replaces it with an empty string
in non-debug build so these strings don't end up in the library.

### DIFF
--- a/lib/jxl/field_encodings.h
+++ b/lib/jxl/field_encodings.h
@@ -19,11 +19,22 @@
 
 namespace jxl {
 
+// Macro to define the Fields' derived class Name when compiling with debug
+// names.
+#if JXL_IS_DEBUG_BUILD
+#define JXL_FIELDS_NAME(X) \
+  const char* Name() const override { return #X; }
+#else
+#define JXL_FIELDS_NAME(X)
+#endif  // JXL_IS_DEBUG_BUILD
+
 class Visitor;
 class Fields {
  public:
   virtual ~Fields() = default;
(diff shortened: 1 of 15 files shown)


lib/jxl/field_encodings.h
```diff
@@ -19,11 +19,24 @@
...
+#endif  // JXL_IS_DEBUG_BUILD
+
 class Visitor;
 class Fields {
  public:
   virtual ~Fields() = default;
+#if JXL_IS_DEBUG_BUILD
   virtual const char* Name() const = 0;
+#else
+  const char* Name() const { return ""; };
```
jonsneyers, line 38: Would it be possible to just not have this method at all in non-debug builds?
(and in that case also not compile any of the code that calls it, of course)

deymo: It is called from modular_image.cc:22 but the call is likely optimized away in non-debug builds because of an `if (false)`, however it still needs to compile. The else case is a non-virtual so the compiler doesn't need to devirtualize the method and it will likely be optimized away by the linker.

jonsneyers: Feel free to remove that call from modular_image, it is not useful anyway.

jonsneyers: (because all transforms have the name Transform)

jonsneyers (approved):

PR #620, opened 2021-09-22
deduplicate PrintTree and move enc-only debug stuff to a new file

jonsneyers:
PrintTree was duplicated in jxl_from_tree.cc, and some debug stuff that is only used in the encoder was in common files. This moves things around to make more sense.

### DIFF
--- a/lib/jxl/modular/encoding/enc_encoding.cc
+++ b/lib/jxl/modular/encoding/enc_encoding.cc
@@ -15,7 +15,6 @@
 #include <unordered_map>
 #include <unordered_set>
 
-#include "lib/jxl/base/os_macros.h"
 #include "lib/jxl/base/status.h"
 #include "lib/jxl/common.h"
 #include "lib/jxl/dec_ans.h"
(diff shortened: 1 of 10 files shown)


lib/jxl/modular/encoding/enc_encoding.cc
```diff
@@ -26,24 +25,45 @@
...
+      return {{0, 0, 255}};
+    case Predictor::Average4:
+      return {{192, 128, 128}};
+    case Predictor::Select:
+      return {{255, 255, 0}};
+    case Predictor::Gradient:
+      return {{255, 0, 255}};
+    case Predictor::Weighted:
+      return {{0, 255, 255}};
+      // TODO
```
veluca93, line 61: Might do that while we're at it :) (i.e. pick colors for all predictors)

jonsneyers: It's probably a good idea to make all code related to producing debug stuff compiled conditionally instead of using constexpr conditions that are false (which requires manually editing the code to get it to actually produce stuff). Maybe we need a new build flag for that - it's a bit orthogonal to debug asserts and symbols etc, since these things are not really about debugging the code but rather about seeing what the encoder is doing. @deymo wdyt?

deymo: constexpr has the advantage that the code still compiles even if it is optimized away, so it tends to bit rot less. For dump_image in AuxOut the current setup uses dependency injection so the code that dumps the stuff is not in the library (in the dump_image case it was a bit more critical to not have it in the library). I think this PR is fine, the debug stuff is now more contained and easier to move out of the enc lib or disable when not needed. Enc library size is not that big of a concern.

veluca93 (approved):

PR #622, opened 2021-09-22
add test for delta-palette

jonsneyers:
There was no test for encode/decode of delta palette yet.

### DIFF
--- a/lib/jxl/modular_test.cc
+++ b/lib/jxl/modular_test.cc
@@ -104,6 +104,33 @@ TEST(ModularTest, RoundtripLosslessCustomWP_PermuteRCT) {
             0.0);
 }
 
+TEST(ModularTest, RoundtripLossyDeltaPalette) {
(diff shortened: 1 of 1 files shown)


deymo:
Is there a way we can test delta-palette without doing a roundtrip? I mean, the BA distance doesn't tell us much about whether delta-palette worked as *intended* other than no crashes and maybe not totally off values.

jonsneyers:
> Is there a way we can test delta-palette without doing a roundtrip? I mean, the BA distance doesn't tell us much about whether delta-palette worked as _intended_ other than no crashes and maybe not totally off values.

I'm open to suggestions, but no crashes, relatively low distance and some kind of compression density are a good start, I think. It should at least catch major regressions in that code.

veluca93 (approved):

deymo:
FYI: This is giving a larger value in msan builds.

Issue #624, opened 2021-09-22
Documentation regaring JxlDecoderRewind API call.


novomesk:
Hello,
I started to use `JxlDecoderRewind` call. I observed slower performance after the rewind. It was OK when I set `JxlDecoderSetParallelRunner` again.

I suggest to add to the documentation information that JxlDecoderSetParallelRunner should be used again, otherwise use gets single-thread performance only.

PR #626, opened 2021-09-22
Keep parallel runner on decoder rewind

lvandeve:
The decoder should not reset the parallel runner when using rewind, only
when using reset.

Improved the way code between rewind and reset is shared.

In addition, changed the behavior of JxlDecoderSetParallelRunner to
match the documentation: the documentation says "may only set before
starting decoding", while the implementation instead allowed setting at
any time, but only once. Now it allows setting it before decoding, after
reset, and after rewind. It's not required to do so after rewind since
it keeps it now.

This addresses https://github.com/libjxl/libjxl/issues/624, by changing
the behavior rather than the documentation

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -537,8 +537,8 @@ JxlDecoderStatus JxlDecoderDefaultPixelFormat(const JxlDecoder* dec,
   return JXL_DEC_SUCCESS;
 }
 
-void JxlDecoderReset(JxlDecoder* dec) {
-  dec->thread_pool.reset();
+// Resets the state that must be reset for both Rewind and Reset
+void JxlDecoderRewindDecodingState(JxlDecoder* dec) {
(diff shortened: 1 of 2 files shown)


lib/jxl/decode.cc
```diff
@@ -537,8 +537,8 @@ JxlDecoderStatus JxlDecoderDefaultPixelFormat(const JxlDecoder* dec,
   return JXL_DEC_SUCCESS;
 }
 
-void JxlDecoderReset(JxlDecoder* dec) {
-  dec->thread_pool.reset();
+// Resets the state that must be reset for both Rewind and Reset
+void JxlDecoderResetDecodingState(JxlDecoder* dec) {
```
deymo, line 541: Maybe call this rewind too?

deymo (approved):

lib/jxl/decode.cc

deymo (approved):

PR #627, opened 2021-09-23
allow changing predictor used in delta palette

jonsneyers:
Currently, the predictor used for delta palette (`cjxl --lossy-palette`) is hardcoded to `Average4` (which is the one that works best). This changes that to use that one by default, but allows changing it by setting `-P`. The predictor used to encode the indices themselves is now hardcoded to `Zero`, which is the one that should work best.

Before: (the predictor choice affects how the indices are encoded, not the predictor used in delta-palette)
```
bees.png
Encoding          kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
----------------------------------------------------------------------------------------------------------------
jxl:m:lp:p0           114    37800    2.6398492   0.084   7.107   1.32610476   0.63799224  1.684203267014      0
jxl:m:lp:p0:P0        114    36055    2.5179831   0.085   7.566   1.32610476   0.63799224  1.606453671751      0
jxl:m:lp:p0:P3        114    37938    2.6494867   0.085   7.352   1.32610476   0.63799224  1.690351945608      0
jxl:m:lp:p0:P5        114    37709    2.6334940   0.085   7.183   1.32610476   0.63799224  1.680148703593      0
jxl:m:lp:p0:P6        114    44107    3.0803129   0.087   7.785   1.32610476   0.63799224  1.965215701010      0
Aggregate:            114    38628    2.6976411   0.085   7.394   1.32610476   0.63799224  1.721074060306      0
```

After: (the predictor choice determines what is used in delta-palette)
```
bees.png
Encoding          kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
----------------------------------------------------------------------------------------------------------------
jxl:m:lp:p0           114    36055    2.5179831   0.088   6.455   1.32610476   0.63799224  1.606453671751      0
jxl:m:lp:p0:P0        114    89544    6.2535093   0.079   5.307   7.13514614   3.36888807 21.067372977656      0
jxl:m:lp:p0:P3        114    39080    2.7292409   0.088   6.271   1.52486873   0.68404738  1.866930078909      0
jxl:m:lp:p0:P5        114    44648    3.1180948   0.087   6.474   1.34442711   0.68350368  2.131229294410      0
jxl:m:lp:p0:P6        114    36474    2.5472449   0.085   5.137  16.08805084   6.04508743 15.398318252508      0
Aggregate:            114    45978    3.2109660   0.086   5.899   3.15392071   1.43451906  4.606191845894      0
```

As expected, the other predictors are worse than Average4, and in particular the Weighted Predictor (P6) works very poorly with the current encoder for delta-palette - there might be an encoder bug there, or maybe that predictor just leads to bad error accumulation - the images look bad but not completely buggy.

Also gets rid of a specialized codepath for the Gradient predictor in delta-palette, which was unreachable before (since that predictor was never used), and is not worth specializing for (since Gradient is not a good predictor for delta-palette; something that averages values works better).

Finally a test is added for the case of the Weighted predictor, which has a separate code path that wasn't covered by tests yet. Note that the BA distance is quite high in this case, so it might be worth investigating if there's an encoder bug in that case.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -391,8 +391,8 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
       // TODO(veluca): allow all predictors that don't break residual
       // multipliers in lossy mode.
       cparams.options.predictor = Predictor::Variable;
-    } else if (cparams.responsive) {
-      // zero predictor for Squeeze residues
+    } else if (cparams.responsive || cparams.lossy_palette) {
+      // zero predictor for Squeeze residues and lossy palette
       cparams.options.predictor = Predictor::Zero;
     } else if (quality < 100) {
       // If not responsive and lossy. TODO(veluca): use near_lossless instead?
@@ -407,6 +407,9 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
       // just gradient predictor in thunder mode
       cparams.options.predictor = Predictor::Gradient;
     }
+  } else {
+    delta_pred = cparams.options.predictor;
+    if (cparams.lossy_palette) cparams.options.predictor = Predictor::Zero;
   }
   tree_splits.push_back(0);
   if (cparams.modular_mode == false) {
@@ -639,7 +642,7 @@ Status ModularFrameEncoder::ComputeEncodingData(
       maybe_palette.lossy_palette =
           (cparams.lossy_palette && maybe_palette.num_c == 3);
       if (maybe_palette.lossy_palette) {
-        maybe_palette.predictor = Predictor::Average4;
+        maybe_palette.predictor = delta_pred;
       }
       // TODO(veluca): use a custom weighted header if using the weighted
       // predictor.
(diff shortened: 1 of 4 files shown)


jyrkialakuijala (approved):
Nice!

I'll look into why weighted predictor is so bad. My original hypothesis (and why I designed Average4) was that dithering messes up the directionality assumptions.

jyrkialakuijala:
In general we could use longer benchmark codewords for palette (than lp, p and P). Only about 5 people in the world are using this I think and having more readability/clarity would help them and the rest of the people.

jonsneyers:
> Nice!
> 
> I'll look into why weighted predictor is so bad. My original hypothesis (and why I designed Average4) was that dithering messes up the directionality assumptions.

Yes, I also suspect that the self-correcting nature of WP interacts poorly with a delta approach that often ends up alternating between overshooting and undershooting, but mostly I think the main issue is that each channel will have its own WP state and evolves subpredictor weights independently. I suspect the predictor feedback loop can diverge chaotically and becomes hard to 'tame'  when effectively each channel potentially uses very different predictors.

jonsneyers:
> In general we could use longer codewords in benchmark codec specification for palette (than lp, p and P). Only about 5 people in the world are using this I think and having more readability/clarity would help them and the rest of the people.
> 
> Something like jxl:lp:palette_size=7:delta_palette_size=11:delta_palette_predictor=4:palette_index_predictor=1 would not be excessive to me.
> 
> This does not need to be addressed in this MR.

Once the encode api is more mature, and cjxl/benchmark_xl can use the api, I think we should revise all encode options that are currently not shown in `cjxl -h` but are only shown in `cjxl -v -h` or `cjxl -v -v -h`. I agree that longer names are fine for more exotic options; only for common options like -q/-d and -s/-e and maybe -p it is nice to have short versions.

PR #628, opened 2021-09-23
Store fewer fields in SplineSegment

sboukortt:


lib/jxl/splines.cc
```diff
@@ -82,8 +82,12 @@ void DrawSegment(DF df, const SplineSegment& segment, bool add, size_t y,
 
 void DrawSegment(const SplineSegment& segment, bool add, size_t y, ssize_t x0,
                  ssize_t x1, float* JXL_RESTRICT rows[3]) {
-  ssize_t x = std::max<ssize_t>(x0, segment.xbegin);
-  x1 = std::min<ssize_t>(x1, segment.xend);
+  const ssize_t xbegin =
```
sboukortt, line 85: (I can perhaps inline these, as well as maybe remove the max(0, ·).)

veluca93: I'd say we can get rid of the max(0, ), it should never do anything since x0 >= 0 and we are storing in a ssize_t anyway..

sboukortt: Thanks, that matches my reasoning. Done and inlined.

veluca93 (approved):

Issue #513, opened 2021-08-30
Overall feedback for app integration


rsebbe:
Hi,

we want to support JXL in our photo app (iOS). It is a full featured format that brings unusual yet useful capabilities like float encoding, especially for HDR, but not only. We've built it and internally integrated it (v0.5), and we have this feedback so far. (If this is not the right place for this feedback, please let me know)

- Memory footprint
Memory usage seems pretty steep when encoding a 12MP photo with JXL. I measured spikes > 800MB in Xcode when doing so, which cause crash on older iOS device lacking some memory. Is it expected? Will that stay that way in later release? Are there options that let you control that aspect?

- Encoding duration
It looks like it takes over 10s to encode a 12MP on iPhone 6S, about 5s on iPhone XS (compiled with optimisations). Also, is that expected?

- Data input API
Currently, we have to provide data as float RGB triplets, or RGBA quadruplets. It would be interesting to get an RGBX option (ignored alpha) if that makes sense in the input API, as rendering format very often is aligned on 4 floats per pixel. Currently it is required to re-layout data (consumes CPU / mem) prior to pass it to `libjxl`

- Line-based encoder API for large images
In some situations, we generate very large picture that typically can't be stored in memory (too large). We generate them band by band, and we encode each band as they are made through `libjpeg` line-feeding. This approach does not seem possible at this time with `libjxl`. Is that something that will possible at some point in the future, or is it inherently impossible to achieve that because of the method used?

- Metadata passing
We'd be very interested to pass along all the EXIF / TIFF / etc. metadata to the generated image file. Currently, this does not seems possible. Even orientation seems to be ignored.

- Color space support
Could you consider adding linear ROMM / ProPhoto color space to the built-in API ? This color space makes a lot of sense for photo editing workflows.

Thank you.

veluca93:
> Hi,
> 
> we want to support JXL in our photo app (iOS). It is a full featured format that brings unusual yet useful capabilities like float encoding, especially for HDR, but not only. We've built it and internally integrated it (v0.5), and we have this feedback so far. (If this is not the right place for this feedback, please let me know)

Hi! Thank you for the feedback, and for integrating JPEG XL! This is indeed the (a?) right place for it :)

I will answer to some of your questions, and let others deal with the rest.

> 
> * Memory footprint
[...]

PR #630, opened 2021-09-23
Allow passing alpha channel to an ImageBundle without alpha.

deymo:
When de-interleaving a RGBA or GA image to an ImageBundle object that
doesn't have an alpha channel the code was crashing. This patch ignores
the alpha channel in that case. This should help with situations like
issue #513 where passing RGBX data to an RGB image.


jonsneyers (approved):

deymo:
@lvandeve  FYI

PR #631, opened 2021-09-23
adjust threshold to make msan test pass

jonsneyers:
For some reason, the msan build created a file that is slightly larger, possibly because of different rounding in entropy coding heuristics.

### DIFF
--- a/lib/jxl/modular_test.cc
+++ b/lib/jxl/modular_test.cc
@@ -124,7 +124,7 @@ TEST(ModularTest, RoundtripLossyDeltaPalette) {
   io.ShrinkTo(300, 100);
 
   compressed_size = Roundtrip(&io, cparams, dparams, pool, &io_out);
-  EXPECT_LE(compressed_size, 6700u);
+  EXPECT_LE(compressed_size, 6800u);
   cparams.ba_params.intensity_target = 80.0f;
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

veluca93:
We should probably investigate this more in-depth...

jonsneyers:
> We should probably investigate this more in-depth...

Do you think there's something non-harmless going on? It's just a few bytes difference, which is not that unusual to have between different compilers/platforms...

veluca93:
> > We should probably investigate this more in-depth...
> 
> Do you think there's something non-harmless going on? It's just a few bytes difference, which is not that unusual to have between different compilers/platforms...

I do think that - MSAN ought to just add extra checks, not to change numbers.

jonsneyers:
Investigated it, turns out to be caused by `std::geometric_distribution` behaving differently in msan builds than in regular builds, causing the pixel/diff sampling to be different that determines the prequantization in the actual sampling used as input to tree learning, causing a different tree to be constructed, leading to different entropy coding performance. The encoded data is identical, the way it is entropy coded isn't though.

Conclusion: it's harmless, but it might be worth getting rid of `std::geometric_distribution` to make encoding more deterministic between builds/platforms, to avoid surprising behavior like this.

PR #634, opened 2021-09-24
Load from other codecs to an interleaved image format.

deymo:
This patch defines a new set of classes like PackedImage that represent
the image concepts as define in the external C API. These classes
represent a jxl decoded image that will be used as the output from the
the decoder API and input to the jxl encoder API.

Migrated most codec parsers (all excepts PSD) to use this format to load
still and animation images without extra channels (besides alpha).

Manually tested .gif and .exr files. .png and .jpg are extensively
tested by the unittests.


jonsneyers:
Nice!

Maybe we should also define our own uncompressed file format that is basically just a dump of a PackedImage: some simple header for the format and dimensions, followed by a memory dump of the pixel data (little endian), followed by uncompressed metadata blobs. It would be very easy now to add a codec to read and write that, and I think it would be pretty useful.

Something like that doesn't exist yet — ppm/pfm/pam come close but they cannot store icc profiles which limits their practical utility (and they also cannot do the combination of float and alpha).

lib/extras/packed_image.h
```diff
@@ -0,0 +1,196 @@
...
+
+  ~PackedImage() {
+    if (pixels) free(pixels);
+  }
+
+  // The image size in pixels.
+  size_t xsize;
+  size_t ysize;
+
+  // Whether the y coordinated is flipped (y=0 is the last row).
```
veluca93, line 89: coordinate, not coordinated

Also, do we need this? Couldn't we undo this in the pfm reader?

lib/extras/packed_image.h
```diff
@@ -0,0 +1,196 @@
...
+    if (format.align > 1) {
+      stride = jxl::DivCeil(stride, format.align) * format.align;
+    }
+    return stride;
+  }
+};
+
+// Helper class representing a frame, as seen from the API. Animations will have
+// multiple frames, but a single frame can have a color/grayscale channel and
+// multiple
```
veluca93, line 132: ... and multiple?

lib/extras/packed_image.h
deymo: We can probably get rid of it after updating the pfm reader/writer.

lib/extras/packed_image.h
deymo: fixed

lib/extras/packed_image.h
```diff
@@ -0,0 +1,196 @@
...
+  PackedImage(size_t xsize, size_t ysize, const JxlPixelFormat& format,
+              size_t stride)
+      : xsize(xsize),
+        ysize(ysize),
+        stride(stride),
+        format(format),
+        pixels(malloc(ysize * stride)),
+        pixels_size(ysize * stride) {}
+  // Construct the image using the passed pixel buffer. The buffer is owned by
+  // this object and released with free().
```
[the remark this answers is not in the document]
deymo: changed to unique_ptr

lib/extras/packed_image.h
veluca93: I'm OK with doing it later.

lib/extras/packed_image.h
```diff
@@ -0,0 +1,173 @@
...
+
+  // Whether the y coordinate is flipped (y=0 is the last row).
+  bool flipped_y = false;
+
+  // The number of bytes per row.
+  size_t stride;
+
+  // The interleaved pixel as defined in the storage format.
+  JxlPixelFormat format;
+  std::unique_ptr<void, decltype(free)*> pixels;
```
veluca93, line 73: I'm not quite sure how happy I feel about this being public...

deymo: ok, I made it confidential.

lib/extras/packed_image.h
```diff
@@ -0,0 +1,173 @@
...
+        pixels(pixels, free),
+        pixels_size(pixels_size) {
+    JXL_ASSERT(pixels_size >= stride * ysize);
+  }
+
+  PackedImage& operator=(PackedImage& other) = delete;
+  PackedImage& operator=(PackedImage&& other) = default;
+
+  PackedImage(PackedImage& other) = delete;
+  PackedImage(PackedImage&& other) = default;
```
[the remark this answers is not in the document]
deymo: It gives you a nicer error message but it is not necessary. I removed them.

veluca93 (approved):

PR #640, opened 2021-09-24
Allow delta palette in >8-bit images.

iulia-m-comsa:
tools/benchmark_xl --input='jyrki31-depth16/*.png' --codec=jxl:m:p0:lp:9 --error_pnorm=1 
benchmark_xl v0.7.0 e0f4c38 [AVX2,SSE4,Scalar]
6 cores, 61 tasks, 6 threads, 0 inner threads

Before:
```
Encoding         kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
---------------------------------------------------------------------------------------------------------------
jxl:m:p0:lp:9      16443  5346967    2.6013328   0.077   8.516   2.28837752   0.53577100  1.393718668969      0
Aggregate:         16443  5346967    2.6013328   0.077   8.516   2.28837752   0.53577100  1.393718668969      0
```
After:
```
Encoding         kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
---------------------------------------------------------------------------------------------------------------
jxl:m:p0:lp:9      16443  5363900    2.6095708   0.055   8.615   2.26338792   0.53316436  1.391330134341      0
Aggregate:         16443  5363900    2.6095708   0.055   8.615   2.26338792   0.53316436  1.391330134341      0
```

Co-authored-by: Jyrki Alakuijala <jyrki@google.com>

### DIFF
--- a/lib/jxl/modular/transform/enc_palette.cc
+++ b/lib/jxl/modular/transform/enc_palette.cc
@@ -110,17 +110,21 @@ struct PaletteIterationData {
 
   // Populates `frequent_deltas` with items from `deltas` based on frequencies
   // and color distances.
-  void FindFrequentColorDeltas(int num_pixels) {
+  void FindFrequentColorDeltas(int num_pixels, int bitdepth) {
     using pixel_type_3d = std::array<pixel_type, 3>;
     std::map<pixel_type_3d, double> delta_frequency_map;
+    pixel_type bucket_size = 3 << std::max(0, bitdepth - 8);
     // Store frequency weighted by delta distance from quantized value.
     for (size_t i = 0; i < deltas[0].size(); ++i) {
       pixel_type_3d delta = {
-          {RoundInt(deltas[0][i], 3), RoundInt(deltas[1][i], 3),
-           RoundInt(deltas[2][i], 3)}};  // a basic form of clustering
+          {RoundInt(deltas[0][i], bucket_size),
+           RoundInt(deltas[1][i], bucket_size),
+           RoundInt(deltas[2][i], bucket_size)}};  // a basic form of clustering
       if (delta[0] == 0 && delta[1] == 0 && delta[2] == 0) continue;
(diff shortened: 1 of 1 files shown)


jonsneyers:
What is "Before" here? Because lossy-palette in > 8-bit wasn't supported there yet, right?

iulia-m-comsa:
> What is "Before" here? Because lossy-palette in > 8-bit wasn't supported there yet, right?

"Before" is current result on 16-bit images (the "jyrki31" set converted to 16-bit), which is indeed without the delta palette.

jonsneyers:
> > What is "Before" here? Because lossy-palette in > 8-bit wasn't supported there yet, right?
> 
> "Before" is current result on 16-bit images (the "jyrki31" set converted to 16-bit), which is indeed without the delta palette.

But how are the results so close to the new results?
Also: maybe it would be good to try it on 10/12-bit content that actually has that bitdepth, as opposed to 8-bit images converted to 16-bit (which keep having only 256 distinct values per channel).

veluca93 (approved):

iulia-m-comsa:
> > > What is "Before" here? Because lossy-palette in > 8-bit wasn't supported there yet, right?
> > 
> > 
> > "Before" is current result on 16-bit images (the "jyrki31" set converted to 16-bit), which is indeed without the delta palette.
> 
> But how are the results so close to the new results?
> Also: maybe it would be good to try it on 10/12-bit content that actually has that bitdepth, as opposed to 8-bit images converted to 16-bit (which keep having only 256 distinct values per channel).

Most images in the set don't actually end up using any deltas - only a couple of them do, in both the 8-bit and 16-bit versions. This is why the improvement is very small. I've also tried 10/12-bit content, there is no difference there. 
We also want to make other further improvements to the delta palette, this is just one more iteration.

jonsneyers (approved):

jyrkialakuijala (commented):
LGTM

PR #641, opened 2021-09-24
Allow setting more boxes after decoder finished

lvandeve:
The box based container format doesn't indicate which box is the last
(having an unbounded size last box is possible but not required), so the
decoder cannot know for sure when the file finished.

If there would be metadata boxes after codestream boxes, the decoder
could return JXL_DEC_SUCCESS when it decoded all codestream boxes and
the user input happened to end exactly at the last codestream box
boundary.

The API did not allow calling JxlDecoderProcessInput again after it
returned success, so it would be impossible to decode those further
metadata boxes.

This MR changes that by allowing to call JxlDecoderSetInput and then
process the further input again.

It will still give errors when calling JxlDecoderProcessInput without
setting any new input, and also when it's not the container format, or a
codestream box is encountered even though all codestream boxes were already
decoded.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -422,7 +422,15 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_orientation);
  * requires more JxlDecoderProcessInput calls to continue.
  *
  * @param dec decoder object
- * @return JXL_DEC_SUCCESS when decoding finished and all events handled.
+ * @return JXL_DEC_SUCCESS when decoding finished and all events handled. If you
+ * still have more unprocessed input data anyway, then you can still continue
+ * by using JxlDecoderSetInput and calling JxlDecoderProcessInput again, similar
+ * to handling JXL_DEC_NEED_MORE_INPUT. JXL_DEC_SUCCESS can occur instead of
+ * JXL_DEC_NEED_MORE_INPUT when, for example, the input data ended right at
+ * the boundary of a box of the container format, all essential codestream boxes
+ * were already decoded, but extra metadata boxes are still present in the next
+ * data. JxlDecoderProcessInput cannot return success if all codestream boxes
+ * have not been seen yet.
  * @return JXL_DEC_ERROR when decoding failed, e.g. invalid codestream.
  * TODO(lode) document the input data mechanism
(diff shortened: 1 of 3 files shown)


lib/jxl/decode.cc
```diff
@@ -381,6 +381,7 @@ struct JxlDecoderStruct {
   std::unique_ptr<jxl::ThreadPool> thread_pool;
 
   DecoderStage stage;
+  bool codestream_finished;
```
deymo, line 384: I don't think we need this boolean. The decoder stage could be kept at kFinished; or have an intermediate stage before kFinished if you want to allow calling ProcessInput only if new input was added. Adding this boolean basically duplicates the "stage" as now the stage is the tuple of these two values.

lib/include/jxl/decode.h
```diff
@@ -422,7 +422,15 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_orientation);
  * requires more JxlDecoderProcessInput calls to continue.
  *
  * @param dec decoder object
- * @return JXL_DEC_SUCCESS when decoding finished and all events handled.
+ * @return JXL_DEC_SUCCESS when decoding finished and all events handled. If you
```
deymo, line 425: Could we continue to return `JXL_DEC_SUCCESS` if no new input data is provided and there are no more new events (and we happen to be at a box boundary)? That would simplify the logic a bit and make `kFinished` stage mean that the jxl codestream reached the end (but other boxes are still ok).

deymo (commented):
hmm.. this is a bit of an oversight. Another option is to provide a function to close the input. I think we talked about that long time ago when designing this.

lib/jxl/decode.cc
lvandeve: Done thanks to the other comment

lib/include/jxl/decode.h
lvandeve: Good idea, done

lvandeve:
A function to close input would indeed make sense. It would break compatibility with existing code unfortunately: the decoder would keep returning NEED_MORE_INPUT until you close it. I'm still open to add this feature as it is a more logical flow, in a different MR, since this one fixes the current way of using it. I think the fix here is also acceptable without adding close, since it allows all the same things, it's just a bit more surprising

jonsneyers (approved):

Issue #638, opened 2021-09-24
GIMP plug-in use too much memory during image loading


novomesk:
Hello @xiota ,

I was able to open this animation http://188.121.162.14/jxl/animation.jxl via GIMP 2.10 plug-in:
![gimp-jxl_frames](https://user-images.githubusercontent.com/52529860/134706397-92476d44-0457-4732-961c-1f39aef0f01f.png)

However it look lot of time and consumed lot of memory.

Animation has 330 full-hd frames. The plug-in first load all the frames into memory with FLOAT precision and at the end, it shrinks all the data 4-times by converting to U8 precision. Temporarily >10GB of memory is used.

Isn't it better to load the frames in 8bit depth directly? Or if you load data in JXL_TYPE_FLOAT format, it can be converted to 8bit after each frame, so less memory is occupied during loading.

xiota:
@novomesk Thank you for letting me know.  I'm using float because JXL uses float internally, and using float on the GIMP side seemed to resolve some ICC profile profile issues.  #642 converts the layers to the destination precision before they're inserted into the image, as you suggest.

novomesk:
@xiota I tried the new code in your `gp-memory-use` branch.

The big animation now loads significantly faster.

There is one different thing I am not sure about.

The plug-in from gp-memory-use branch (built under MSYS2) loads a [JXL file](https://github.com/libjxl/conformance/blob/master/testcases/progressive/input.jxl) this way:
![gimp210](https://user-images.githubusercontent.com/52529860/134886070-a65f510b-c1ef-4e89-b53d-8b011a702fc8.png)

but in GIMP 2.99 on Linux using my plug-in the picture is brighter:
![gimp299](https://user-images.githubusercontent.com/52529860/134886282-9cba0557-6f6e-474f-a053-c3bf906837ca.png)

veluca93:
Looking at the images, it looks like the gp-memory-use branch has some kind of colorspace handling issue... I can confirm the brighter image is correct.

xiota:
@novomesk Thanks for letting me know.  I'll take a look at it.  I've been having difficulties with the ICC profiles.  When one set of images loads correctly, another doesn't, and vice versa.

xiota:
[...]

PR #642, opened 2021-09-24
GIMP Plugin: Reduce memory use when loading (and other changes)

xiota:
Convert layers to destination precision as they're decoded, before they're inserted into the image.  Also includes some minor cleanup.  Addresses #638

### DIFF
--- a/plugins/gimp/file-jxl-save.cc
+++ b/plugins/gimp/file-jxl-save.cc
@@ -116,8 +116,7 @@ bool JpegXlSaveGui::GuiOnChangeQuality(GtkAdjustment* adj_qual,
   g_clear_signal_handler(&self->handle_toggle_lossless, self->toggle_lossless);
 
   GtkAdjustment* adj_dist = self->entry_distance;
-  jxl_save_opts.quality = gtk_adjustment_get_value(adj_qual);
-  jxl_save_opts.UpdateDistance();
+  jxl_save_opts.SetQuality(gtk_adjustment_get_value(adj_qual));
   gtk_adjustment_set_value(adj_dist, jxl_save_opts.distance);
 
   self->handle_toggle_lossless = g_signal_connect(
@@ -140,8 +139,7 @@ bool JpegXlSaveGui::GuiOnChangeDistance(GtkAdjustment* adj_dist,
   g_clear_signal_handler(&self->handle_entry_quality, self->entry_quality);
   g_clear_signal_handler(&self->handle_toggle_lossless, self->toggle_lossless);
 
-  jxl_save_opts.distance = gtk_adjustment_get_value(adj_dist);
-  jxl_save_opts.UpdateQuality();
+  jxl_save_opts.SetDistance(gtk_adjustment_get_value(adj_dist));
   gtk_adjustment_set_value(adj_qual, jxl_save_opts.quality);
 
   if (!(jxl_save_opts.distance < 0.001)) {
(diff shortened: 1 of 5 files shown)


plugins/gimp/file-jxl-save.cc
```diff
@@ -789,14 +779,24 @@ bool SaveJpegXlImage(const gint32 image_id, const gint32 drawable_id,
...
+  jxl_save_opts.pixel_format.data_type = JXL_TYPE_FLOAT;
+  jxl_save_opts.SetBablType("float");
+  const Babl* destination_format =
+      babl_format(jxl_save_opts.babl_format_str.c_str());
 
+  // convert precision to float if needed
+  if (jxl_save_opts.basic_info.bits_per_sample < 32) {
+    gimp_image_convert_precision(duplicate, GIMP_PRECISION_FLOAT_LINEAR);
+  } else if (!jxl_save_opts.is_linear) {
+    gimp_image_convert_precision(duplicate, GIMP_PRECISION_FLOAT_LINEAR);
```
jonsneyers, line 798: `if (jxl_save_opts.basic_info.bits_per_sample < 32 || !jxl_save_opts.is_linear) {` ?

xiota: Yes.  Thank you.  Done.

jonsneyers (approved):

jonsneyers (changes requested):
I think it's better to ignore the original profile and just always use the target profile with float precision. That should simplify things a lot.

jonsneyers:
We don't have CI tests for plugins, but I assume you tried this with various images / encode settings?

There seems to be a merge conflict so please rebase.

xiota:
@jonsneyers  Just rebased.  I tested with the images that were causing problems before.

I noticed when saving with distance = 0 (8-bit images), the jxl files are nearly double the size of those produced by `cjxl`.  Is that because I'm sending float data to the encoder?  (even though basic info is set to 8-bit)

When not *trying* to save ICC profiles, the advanced features become unnecessary, so the save dialog is much simplified...  Problems were... GIMP doesn't accurately report whether the TRC is linear (but the profile description correctly states it).  Color encoding info cannot be changed when an ICC is attached, yet the encoder does not parse the ICC, so the TRC and other values were wrong.  (I think either the encoder needs to parse the ICC to set the correct values or the API should allow the color encoding to be changed, and the ICC is just attached for "informational" purposes.)

I don't know how to check whether lcms2 is working in "unbounded" mode.  The docs state it's capable, but I don't know how to enable it or check whether the conditions are met.  Using lcms2 because GEGL didn't seem to work properly and lcms2 is a GIMP dependency, so *should* be available whenever the plugin is compiled.

screenshot

xiota:
I just tried with a 16-bit image converted from raw.  The file `cjxl` filesize is 91.7M.  The gimp file size is 37.5M  That's about expected since the gimp plugin has `uses_original_profile = false`, which approximately doubles file sizes.

jonsneyers:
Use_original_profile = false is not lossless, and it is much larger because things get converted to XYB internally and then get saved 'losslessly' with about 12 bits per channel (the bits_per_sample does not have influence on what gets encoded when XYB is used). When you want to do lossless, you always want to use the original profile and not convert to XYB.

xiota:
@jonsneyers 

> Use_original_profile = false is not lossless and it is much larger because things get converted to XYB internally and then get saved 'losslessly' with about 12 bits per channel (the bits_per_sample does not have influence on what gets encoded when XYB is used).

The files I've saved with the API when `uses_original_profile = false` are usually smaller (by half) than with `uses_original_profile = true`.

> When you want to do lossless, you always want to use the original profile and not convert to XYB.

The version of the gimp plugin in this PR doesn't save "lossless" (that's why I wrote "d=0").  It can use `d=0` and `uses_original_profile = false`.  The problem with `uses_original_profile` are related to the ICC issues I've been having.

If the encoder won't parse and use the ICC profile, there needs to be a way to set correct color encoding values.  Currently, the API blocks changes to color encoding if an ICC is attached.  And if color encoding is set, an ICC cannot be attached.  Everything works fine if sticking with sRGB, but in that case, there isn't any need to attach a "real" ICC profile.

jonsneyers:
For lossless, the encoder doesn't need to parse the icc profile, since it doesn't need to know the meaning of the pixel data. So I recommend to set the icc profile and uses_original_profile=true in case of lossless, and to convert to linear unclamped sRGB and uses_original_profile=false in case of lossy.

@lvandeve / @sboukortt does that make sense?

xiota:
@novomesk Thank you.  I'm looking at it to see what went wrong...

mo271:
any updates on this pr?

mo271:
Closing for now, please fell to re-open once this cleanly rebases and improves the plugin.

PR #651, opened 2021-09-27
try RCTs and custom WP also for single-group images

jonsneyers:
In the case that the whole image fits in the GlobalModular section, the encoder wasn't trying different RCTs and different WP params, since that code was only being run for the ModularDC and ModularAC sections. This fixes that, which should lead to slightly denser lossless compression of images smaller than 256x256.

A side effect of this change seems to be that a test related to partial codestream boxes (`jxlp`) is now failing — likely that is unrelated to this change, so for now I have manually disabled those tests.
@lvandeve PTAL at the failing test (see the change in `decode_test.cc` in this PR). I tried finding the problem but I'm not familiar with the relevant code.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -794,6 +794,12 @@ Status ModularFrameEncoder::ComputeEncodingData(
           mrect, minShift, maxShift, ModularStreamId::ModularAC(group_id, i)});
     }
   }
+  // if there's only one group, everything ends up in GlobalModular
+  // in that case, also try RCTs/WP params for the one group
+  if (stream_params.size() == 2) {
+    stream_params.push_back(GroupParams{Rect(0, 0, xsize, ysize), 0, 1000,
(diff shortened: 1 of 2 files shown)


lvandeve:
@jonsneyers thanks for finding this bug, it's indeed a bug and should be fixed with https://github.com/libjxl/libjxl/pull/664

I'm also working on further simplifying the box handling for future PRs

jonsneyers:
Rebased and re-enabled the test that was failing.

veluca93 (approved):

tbonfort:
This is great news :) Here's the old/new sizes for a 13000x12000 tiff container using 256x256 jxl tiles, using different speed tiers:
```
444028210 jxl-3-new.tif
444028210 jxl-3.tif

427696856 jxl-5-new.tif
522358734 jxl-5.tif

427629507 jxl-7-new.tif
522965125 jxl-7.tif

425612715 jxl-9-new.tif
522582647 jxl-9.tif
```

PR #654, opened 2021-09-28
exclude more test stuff from coverage report

jonsneyers:
No need to report coverage of the test files themselves.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -551,10 +551,12 @@ cmd_coverage_report() {
   local gcovr_args=(
     -r "${real_build_dir}"
     --gcov-executable "${LLVM_COV} gcov"
-    # Only print coverage information for the jxl and fuif directories. The rest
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

ci.sh
```diff
@@ -551,10 +551,12 @@ cmd_coverage_report() {
   local gcovr_args=(
     -r "${real_build_dir}"
     --gcov-executable "${LLVM_COV} gcov"
-    # Only print coverage information for the jxl and fuif directories. The rest
+    # Only print coverage information for the libjxl directories. The rest
     # is not part of the code under test.
     --filter '.*jxl/.*'
     --exclude '.*_test.cc'
+    --exclude '.*_test-only..*'
```
deymo, line 558: This should be `.*_testonly..*`

jonsneyers: right, I was wondering why those still showed up in the report

PR #657, opened 2021-09-28
cleanup dead/unused code in butteraugli

jonsneyers:
There are some parts of butteraugli.cc that are just never used (anymore), like fast paths for kernel sizes that aren't used, and the code for the `approximate_border == true` case that is never used. There are also a few ifs that have a condition that is constrained by the surrounding code to have a fixed outcome.


jonsneyers:
Pinging @jyrkialakuijala to check that this is not removing something we'll want to start using again.

PR #659, opened 2021-09-28
remove uint24

jonsneyers:
We have code to handle (big-endian and little-endian) uint24 buffers. Nobody will use that — you can use 32-bit if you need more than 16-bit, but handling 24-bit data will be inconvenient for anyone and we will not have it in the api either (also now the code does what the documentation says it does: bits_per_sample has to be 8, 16 or 32.


deymo (approved):

jonsneyers:
> We can add this back if we ever add UINT24 to the API

Which is unlikely, imo.

PR #663, opened 2021-09-28
drop ==, != from dec_params.h

jonsneyers:
The comparison operators for dec_params were only used in one spot in benchmark_xl, no need to have that in libjxl itself.

### DIFF
--- a/tools/benchmark/benchmark_codec_jxl.cc
+++ b/tools/benchmark/benchmark_codec_jxl.cc
@@ -266,7 +266,8 @@ class JxlCodec : public ImageCodec {
                     ThreadPoolInternal* pool, CodecInOut* io,
                     jpegxl::tools::SpeedStats* speed_stats) override {
     io->frames.clear();
-    if (dparams_ != DecompressParams{}) {
+    if (dparams_.max_passes != DecompressParams().max_passes ||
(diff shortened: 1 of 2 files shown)


tools/benchmark/benchmark_codec_jxl.cc
```diff
@@ -266,7 +266,7 @@ class JxlCodec : public ImageCodec {
                     ThreadPoolInternal* pool, CodecInOut* io,
                     jpegxl::tools::SpeedStats* speed_stats) override {
     io->frames.clear();
-    if (dparams_ != DecompressParams{}) {
+    if (dparams_.max_passes < 100 || dparams_.max_downsampling != 1) {
```
sboukortt, line 269: Optionally, we wouldn’t have to know what the defaults are if we wrote this as:

```suggestion
    if (dparams_.max_passes != DecompressParams().max_passes || dparams_.max_downsampling != DecompressParams().max_downsampling) {
```

sboukortt (dismissed):

tools/benchmark/benchmark_codec_jxl.cc
jonsneyers: Yes, that's nicer than putting numeric values there. Thanks for the suggestion!

sboukortt (approved):

PR #664, opened 2021-09-28
Fix issue with partial codestream boxes

lvandeve:
Fixes issue discovered in https://github.com/libjxl/libjxl/pull/651/:

It was saving previous box data to the codestream buffer when a next
partial codestream box is encountered. However, if there were other
unrelated boxes in between, it already updated position pointers and no
longer included the previous box data.

Now it instead saves the codestream data to the internal buffer
immediately if it's a partial codestream, rather than waiting for the
next one. It has to do this anyway to concatenate the separate buffers,
so the memory usage is about the same.

Also removed variable box_begin that was never read from.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -403,7 +403,6 @@ struct JxlDecoderStruct {
   // Position of next_in in the original file including box format if present
   // (as opposed to position in the codestream)
   size_t file_pos;
-  size_t box_begin;
   size_t box_end;
   bool skip_box;
   // Begin and end of the content of the current codestream box. This could be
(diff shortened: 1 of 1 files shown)


veluca93 (dismissed):

lib/jxl/decode.cc
```diff
@@ -1604,31 +1601,19 @@ JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec) {
...
           dec->codestream_begin = dec->file_pos + pos;
           dec->codestream_end =
               (box_size == 0) ? 0 : (dec->codestream_begin + codestream_size);
           size_t avail_codestream_size =
               (box_size == 0)
                   ? (size - pos)
                   : std::min<size_t>(size - pos, box_size - pos + box_start);
-          // If already appending codestream, append what we have here too
-          if (!dec->codestream.empty()) {
+          // If this is not the last codestream, which means there are mutliple
```
jonsneyers, line 1611: multiple

jonsneyers (approved):

lib/jxl/decode.cc

veluca93 (approved):

PR #669, opened 2021-09-30
don't report coverage of debug-printing code

jonsneyers:
No need to have files in the coverage report that only contain verbose debug printing functions.

Also remove an unused function from `aux_out.cc`.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -558,6 +558,7 @@ cmd_coverage_report() {
     --filter '.*jxl/.*'
     --exclude '.*_test.cc'
     --exclude '.*_testonly..*'
+    --exclude '.*_debug.*'
     --exclude '.*test_utils..*'
     --object-directory "${real_build_dir}"
   )
--- a/lib/jxl/aux_out.cc
+++ b/lib/jxl/aux_out.cc
@@ -59,23 +59,6 @@ void AuxOut::Print(size_t num_inputs) const {
   }
 }
 
-void AuxOut::DumpCoeffImage(const char* label,
-                            const Image3S& coeff_image) const {
-  JXL_ASSERT(coeff_image.xsize() % 64 == 0);
-  Image3S reshuffled(coeff_image.xsize() / 8, coeff_image.ysize() * 8);
-  for (size_t c = 0; c < 3; c++) {
-    for (size_t y = 0; y < coeff_image.ysize(); y++) {
-      for (size_t x = 0; x < coeff_image.xsize(); x += 64) {
-        for (size_t i = 0; i < 64; i++) {
-          reshuffled.PlaneRow(c, 8 * y + i / 8)[x / 8 + i % 8] =
-              coeff_image.PlaneRow(c, y)[x + i];
-        }
-      }
-    }
-  }
-  DumpImage(label, reshuffled);
-}
-
 void ReclaimAndCharge(BitWriter* JXL_RESTRICT writer,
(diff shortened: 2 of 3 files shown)


veluca93 (dismissed):

ci.sh
```diff
@@ -556,6 +556,8 @@ cmd_coverage_report() {
     --filter '.*jxl/.*'
     --exclude '.*_test.cc'
     --exclude '.*_testonly..*'
+    --exclude '.*_debug.*'
```
deymo, line 559: this ships in the library, not sure it is fair to ignore it.

jonsneyers: The _debug stuff is behind an `if (constexpr false)`, so I don't think that really counts as shipping.

The aux_out stuff: I suppose some of the bookkeeping code (attribution of encoded bits to bitstream segments) does perhaps ship with the encoder, but most of the code there is for printing verbose output and dumping debug images, and those functions should only be called from cjxl and benchmark_xl, so they shouldn't end up in the library or be accessible in any way from the API.

deymo: Shouldn't attribution of bits be tested?

jonsneyers: Bit attribution is only used to show verbose encoder output, showing how many bits went to header, toc, DC, etc. I don't know if we should keep that functionality - and if we do, it's probably best put behind a macro that can be disabled in regular builds. In any case, I don't think this is info we want to expose in the api.

deymo: ok with not having it in the API; but I don't follow from there that it shouldn't be in the coverage report. What if the bit attribution is buggy?

If this PR is just to improve the % value in the coverage then please don't merge this PR.

Not reporting test infrastructure is one thing, because we don't test the test infrastructure and we don't intend to do that. On the other hand `lib/extras` for example has some tests (not reported here); and bugs parsing a .gif file in cjxl would be reported in this repo as issues when people try cjxl with a .gif, so it is fair to get that coverage even if codec_gif is not something in our API.

If the bit attribution in the encoder is something we want remove, then removing it is the right solution, not hiding it in the coverage report. I think that the bit attribution can be useful information when optimizing the encoder so it should stay (and have some test while at that).

jonsneyers: Fair enough. I'll change this PR to keep aux_out in the coverage report.

Bit attribution and dumping debug images can certainly be useful when debugging/optimizing the encoder, but it should not be compiled into normal builds of libjxl imo. Also if/when we eventually migrate cjxl/benchmark_xl to the api, we'll likely have to do it in a different way anyway. But you're right that hiding it from the coverage report doesn't solve anything, and it is better to resolve the underlying problem rather than just hiding it.

I'm not sure what the best approach is here, but I think it might be best to put all the bit attribution calls behind a macro (like we do with profiling, which is also something that inherently requires a presence in many source files), and turn it into no-op on normal builds and only make it do something in debug builds.

jonsneyers:
Can this be re-approved? It's the same as the original PR, except the aux_out files are still included in the coverage report (since some of that code is actually compiled in release builds of the library).
(ping @veluca93)

veluca93 (approved):

jonsneyers:
Sorry, had to rebase to get the new CI stuff (windows builds) to run, which is a blocking condition for merging but it cannot be manually triggered, you have to do a new push, which auto-dismisses approvals again, so could @veluca93 or anyone else give this an approve again?

veluca93 (approved):

PR #670, opened 2021-09-30
Refactor FastXYBToSRGB8 to use an intermediate int16 buffer.

veluca93:
This is in preparation of running more computation in 16-bit fixpoint.

Slight slowdown on fd4 (22.2MP/s -> 21.7MP/s).

### DIFF
--- a/lib/jxl/dec_cache.h
+++ b/lib/jxl/dec_cache.h
@@ -86,6 +86,10 @@ struct PassesDecoderState {
   // One row per thread
   std::vector<std::vector<float>> pixel_callback_rows;
 
+  // Buffer for 16bit fixpoint fastpath: one ImageS per thread, containing one
+  // row per channel.
+  std::vector<ImageS> fixpoint_srgb_buffer;
+
   // Seed for noise, to have different noise per-frame.
   size_t noise_seed = 0;
 
@@ -203,6 +207,12 @@ struct PassesDecoderState {
                                        kGroupDim + 2 * kGroupDataYBorder);
         ycbcr_out_images.emplace_back(kGroupDim + 2 * kGroupDataXBorder,
                                       kGroupDim + 2 * kGroupDataYBorder);
+      }
+    }
+    if (fast_xyb_srgb8_conversion) {
+      for (size_t _ = fixpoint_srgb_buffer.size(); _ < num_threads; _++) {
+        fixpoint_srgb_buffer.emplace_back(
+            ImageS(kApplyImageFeaturesTileDim * 8, 4));
       }
     }
     if (rgb_output || pixel_callback) {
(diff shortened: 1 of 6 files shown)


lib/jxl/dec_xyb-inl.h
```diff
@@ -86,10 +86,34 @@ static inline HWY_MAYBE_UNUSED bool HasFastXYBTosRGB8() {
 #endif
 }
 
-static inline HWY_MAYBE_UNUSED void FastXYBTosRGB8(
-    const Image3F& input, const Rect& input_rect, const Rect& output_buf_rect,
-    const ImageF* alpha, const Rect& alpha_rect, bool is_rgba,
-    uint8_t* JXL_RESTRICT output_buf, size_t xsize, size_t output_stride) {
+template <size_t shift>
```
deymo, line 89: Would be good to explain the requirements of this function; such as `in` needs xsize rounded up to 4 floats; and `out`needs round up to 8 int16_t, right?

lib/jxl/dec_xyb-inl.h
```diff
@@ -86,10 +86,34 @@ static inline HWY_MAYBE_UNUSED bool HasFastXYBTosRGB8() {
...
+static inline HWY_MAYBE_UNUSED void ConvertToFixpoint(const float* in,
+                                                      int16_t* out,
+                                                      size_t xsize) {
+#if HWY_TARGET == HWY_NEON
+  for (size_t x = 0; x < xsize; x += 8) {
+    // TODO(veluca): consider different approaches here, like vld1q_f32_x2.
+    float32x4_t flt_left = vld1q_f32(in + x);
+    int16x4_t int16_left = vqmovn_s32(vcvtq_n_s32_f32(flt_left, shift));
+    float32x4_t flt_right = vld1q_f32(in + x + (x + 4 < xsize ? 4 : 0));
```
deymo, line 98: 🤯

deymo (approved):

lib/jxl/dec_xyb-inl.h
veluca93: No need for that for out - no more than xsize outputs are written.

lib/jxl/dec_cache.h
```diff
@@ -205,6 +209,12 @@ struct PassesDecoderState {
                                       kGroupDim + 2 * kGroupDataYBorder);
       }
     }
+    if (fast_xyb_srgb8_conversion) {
+      for (size_t _ = fixpoint_srgb_buffer.size(); _ < num_threads; _++) {
```
sboukortt, line 213: Would it make sense to `reserve`?

sboukortt (approved):

lib/jxl/dec_cache.h
veluca93: I'd say probably not -- but I also think I should actually wait to refactor things a bit before I submit this....

PR #671, opened 2021-09-30
Special-case 16-bit "Gaborish only" path for ARM.

veluca93:
About 10% speedup of faster_decoding=3 case. It may be possible to
recover some extra performance by reducing the total amount of border
storage.

```
faster decoding 4: geomean: 22.0309 MP/s [21.1923, 22.2131]
before: geomean: 16.3272 MP/s [16.1619, 16.4523]
skip gaborish: geomean: 19.7544 MP/s [19.6237, 19.9077]
after: geomean: 18.2605 MP/s [18.1625, 18.3798]

float output, bpp*pnorm: 0.843556372436
uint8 output before, bpp*pnorm: 0.853447943526
uint8 output after, bpp*pnorm: 0.853247854260
```

Based on https://github.com/libjxl/libjxl/pull/670

### DIFF
--- a/lib/jxl/dec_16bit_reconstruct.cc
+++ b/lib/jxl/dec_16bit_reconstruct.cc
@@ -1,0 +1,129 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "lib/jxl/dec_16bit_reconstruct.h"
+
+#undef HWY_TARGET_INCLUDE
+#define HWY_TARGET_INCLUDE "lib/jxl/dec_16bit_reconstruct.cc"
+#include <hwy/foreach_target.h>
+#include <hwy/highway.h>
+
+#include "lib/jxl/dec_xyb-inl.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace jxl {
+namespace HWY_NAMESPACE {
+void FinalizeImageRect16(Image3F* input_image, const Rect& input_rect,
+                         ImageF* alpha, const Rect& alpha_rect,
+                         PassesDecoderState* dec_state, size_t thread,
+                         const Rect& frame_rect) {
+  // This function is very NEON-specific. As such, it uses intrinsics directly.
+#if HWY_TARGET == HWY_NEON
+  // WARNING: doing fixed point arithmetic correctly is very complicated.
+  // Changes to this function should be thoroughly tested.
+  JXL_ASSERT(SameSize(input_rect, frame_rect));
+  if (alpha) JXL_ASSERT(SameSize(alpha_rect, input_rect));
+
+  const FrameDimensions& frame_dim = dec_state->shared->frame_dim;
+
+  ImageS& storage = dec_state->fixpoint_buffer[thread];
+
+  size_t x0 = frame_rect.x0();
+  size_t x1 = std::min(x0 + frame_rect.xsize(), frame_dim.xsize_upsampled);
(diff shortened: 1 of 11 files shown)


lib/jxl/dec_16bit_reconstruct.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "lib/jxl/dec_cache.h"
+
+namespace jxl {
+void FinalizeImageRect16(Image3F* input_image, const Rect& input_rect,
```
deymo, line 9: document meeeee plz. For example, explain why is this unreachable in other than NEON? when should this be used?

lib/jxl/dec_16bit_reconstruct.cc
```diff
@@ -0,0 +1,129 @@
...
+                     y + frame_rect.y0() < frame_dim.ysize_upsampled + 2;
+       y++) {
+    // Mirroring
+    size_t yfrom =
+        y + frame_rect.y0() == 0
+            ? 0
+            : (y + frame_rect.y0() == frame_dim.ysize_upsampled + 1 ? y - 2
+                                                                    : y - 1);
+
+    ConvertToFixpoint<18>(input_rect.ConstPlaneRow(*input_image, 0, yfrom) - 1,
```
deymo, line 53: Explain me like I'm 5 why is this 18 but the the others are 15.

lib/jxl/dec_16bit_reconstruct.cc
```diff
@@ -0,0 +1,129 @@
...
+  // This function is very NEON-specific. As such, it uses intrinsics directly.
+#if HWY_TARGET == HWY_NEON
+  // WARNING: doing fixed point arithmetic correctly is very complicated.
+  // Changes to this function should be thoroughly tested.
+  JXL_ASSERT(SameSize(input_rect, frame_rect));
+  if (alpha) JXL_ASSERT(SameSize(alpha_rect, input_rect));
+
+  const FrameDimensions& frame_dim = dec_state->shared->frame_dim;
+
+  ImageS& storage = dec_state->fixpoint_buffer[thread];
```
deymo, line 31: Explain the way this storage is meant to be. 3 rows per channel?

lib/jxl/dec_16bit_reconstruct.cc
```diff
@@ -0,0 +1,129 @@
...
+#define HWY_TARGET_INCLUDE "lib/jxl/dec_16bit_reconstruct.cc"
+#include <hwy/foreach_target.h>
+#include <hwy/highway.h>
+
+#include "lib/jxl/dec_xyb-inl.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace jxl {
+namespace HWY_NAMESPACE {
+void FinalizeImageRect16(Image3F* input_image, const Rect& input_rect,
```
deymo, line 18: Any chance we can test that applying gaborish with this is ~the same as applying it with the filters functions in a unittest only enabled when NEON is compiled/supported?

deymo (approved):
This is such a complicated mess. LGTM

lib/jxl/dec_16bit_reconstruct.cc
veluca93: I admit this is not obvious here - there's a comment about this in dec_xyb-inl.h that states that X needs extra precision, so we use X*8 instead of X. As gaborish is linear, this is OK...

PR #674, opened 2021-09-30
remove debug printing instrumentation on header fields

jonsneyers:
I don't think we still need these — and if we do, it currently requires manually enabling printing by editing the code anyway, so you can just as well just fetch a version of `fields.*` from before this commit.

Not all the printed strings are detected by the compiler to be in unreachable code, so they did end up in the binary. So this is a (very minor) binary size improvement.

This instrumentation was useful when debugging the (quite complicated) header field mechanism with its various visitors, but I think it's safe to say that it now just works and we don't need the debug instrumentation anymore.

### DIFF
--- a/lib/jxl/fields.cc
+++ b/lib/jxl/fields.cc
@@ -62,20 +62,12 @@ class ExtensionStates {
 
 class VisitorBase : public Visitor {
  public:
-  explicit VisitorBase(bool print_bundles = false)
-      : print_bundles_(print_bundles) {}
+  explicit VisitorBase() {}
   ~VisitorBase() override { JXL_ASSERT(depth_ == 0); }
 
-  // This is the only call site of Fields::VisitFields. Adds tracing and
-  // ensures EndExtensions was called.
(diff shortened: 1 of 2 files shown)


deymo:
What's the impact on binary size of removing this?

jonsneyers:
> What's the impact on binary size of removing this?

On my non-size optimized x86 build, very minimal (-16 bytes), but I suspect that's because there is some kind of padding going on — the total size of the strings that got removed was more than that: I saw a bunch of strings like `U64(%zu) = %` in my stripped libjxl_dec.so. Also I assume that if the strings are there, then the code to do the printf calls is probably also there.

Maybe it would be good to check if there's any impact on the chrome build - probably that one is optimized better, but you never know.

lvandeve (approved):

lvandeve:
I'm ok with removing this, this has been helpful on occasion, but I agree that it doesn't need to exist when not in use, there are manual ways to add it back in rare occasions

Issue #673, opened 2021-09-30
Bad progressive decode of transcoded jpeg.


wwww-wwww:
**Describe the bug**
Progressive decode of a lossless transcode of chroma subsampled jpeg does not render correctly.
LQIP is gray, MQIP's chroma is at its original subsampled size.

**To Reproduce**
Progressively load attached files:
4:2:0
<img src="https://user-images.githubusercontent.com/19401176/135532520-75b67b82-eb2d-4f84-8d5e-a23fe3a97b22.jpg"  width="200" height="200" />
4:2:2
<img src="https://user-images.githubusercontent.com/19401176/135532577-96aaf9b4-5c77-4fb1-acaf-d47dd49575ca.jpg"  width="200" height="200" />
(renamed to .jpg to upload to github)

**Expected behavior**
Image used here is a transcode of 4:4:4 jpeg
<img src="https://user-images.githubusercontent.com/19401176/135532806-587ffcec-97b7-4448-9889-61855e1faeef.png"  width="400" height="400" />

**Screenshots**
4:2:0 at 50/384 kb
<img src="https://user-images.githubusercontent.com/19401176/135533186-56a736c3-ef37-4b1a-b8d9-b2b304357c36.png"  width="400" height="400" />
4:2:2 at 50/381 kb
<img src="https://user-images.githubusercontent.com/19401176/135533198-c425bd8e-d787-4259-8953-06b15a8b3f79.png"  width="400" height="400" />

**Environment**
Patched Firefox on libjxl commit [702cd54420513b754edb60e33b9a4438f7e9ca00](https://github.com/libjxl/libjxl/commit/702cd54420513b754edb60e33b9a4438f7e9ca00)

TheDecryptor:
I can also see this using the resampling option in djxl

Using [this image](https://github.com/libjxl/libjxl/blob/c2266c4bacaf355f3e8fe63c8d3a63f83b3cecfe/third_party/testdata/imagecompression.info/flower_foveon.png.im_q85_420.jpg) as a source with these options: `djxl.exe .\flower_foveon.png.im_q85_420.jxl -s 8 .\flower_foveon.png.im_q85_420.jxl.png` results in this output.

![flower_foveon png im_q85_420 jxl](https://user-images.githubusercontent.com/1516741/135592598-5b175c6e-072b-440e-9373-a441e49599d1.png)

PR #675, opened 2021-10-01
Fix progressive previews with chroma subsampling.

veluca93:
Fixes https://github.com/libjxl/libjxl/issues/673

Also adds a test.

### DIFF
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -1466,6 +1466,31 @@ TEST(JxlTest, JXL_TRANSCODE_JPEG_TEST(RoundtripJpegRecompressionToPixels420)) {
 }
 
 TEST(JxlTest,
+     JXL_TRANSCODE_JPEG_TEST(RoundtripJpegRecompressionToPixels420EarlyFlush)) {
+  ThreadPoolInternal pool(8);
+  const PaddedBytes orig =
+      ReadTestData("imagecompression.info/flower_foveon.png.im_q85_420.jpg");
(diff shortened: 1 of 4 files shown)


lib/jxl/jxl_test.cc
```diff
@@ -1465,6 +1465,31 @@ TEST(JxlTest, JXL_TRANSCODE_JPEG_TEST(RoundtripJpegRecompressionToPixels420)) {
...
+  ASSERT_TRUE(SetFromBytes(Span<const uint8_t>(orig), &io, &pool));
+
+  CodecInOut io2;
+  ASSERT_TRUE(SetFromBytes(Span<const uint8_t>(orig), &io2, &pool));
+
+  CompressParams cparams;
+  cparams.color_transform = jxl::ColorTransform::kYCbCr;
+
+  DecompressParams dparams;
+  dparams.max_downsampling = 8;
```
jonsneyers, line 1484: If it's not too much trouble, it might be worthwhile to have this test in `decode_test.cc` instead (or in addition), and decode a recompressed jpeg while flushing at various random bitstream truncation points (not just after the DC like it is here), to check that nothing breaks in case there are e.g. a few AC groups already available too.

jonsneyers (approved):

lib/jxl/jxl_test.cc
veluca93: AC groups don't use a specialized code path for the "flush" case though, so I don't think it's too useful...

PR #676, opened 2021-10-01
Simplify box handling in decode.cc

lvandeve:
This simplifies the box handling loop, so that it will be possible to
implement support for decoder output of metadata boxes, and handling the
metadata of reconstructed jpeg


veluca93 (dismissed):

lvandeve:
I updated this PR after the fuzzer found a bug (it was returning success after parsing a JXL file that didn't have any codestream boxes at all), please take another look

veluca93 (approved):

PR #679, opened 2021-10-01
tweaks to lossless jpeg recompression

jonsneyers:
Some small tweaks to the lossless jpeg recompression encoder choices:
- slightly adjust AC context model (use fewer contexts for smaller images)
- allow more effort to be spent on DC at -e 8, -e 9
- don't call ClusterGroups() at -e 9, it makes -e 9 extremely slow for little benefit

Tested on a corpus of the (large) CTC images + jyrki's testset (of mostly smaller images), converted to jpeg with imagemagick -quality 70 and 90. Times and sizes are totals for the whole set. Speed was not measured very accurately, but it gives an idea.

At q70:

| effort | size before | size after | time before | time after |
|-------|------------|-----------|----------|--------|
| e3 | 7,464,028 | 7,460,659  | 2.4s | 2.4s | 
| e7 | 7,460,245 | 7,456,938 | 2.5s | 2.5s |
| e8 | 7,449,599 | 7,434,265  | 4.0s | 5.1s |
| e9 | 7,449,426 | 7,430,898 | 4m39s | 22.7s |

At q90:

| effort | size before | size after | time before | time after |
|-------|------------|-----------|----------|--------|
| e3 | 17,737,365 | 17,719,048 | 4.8s | 4.6s | 
| e7 | 17,719,499 | 17,702,678 | 4.9s | 4.9s |
| e8 | 17,679,099 | 17,636,227  | 7.9s | 10.7s |
| e9 | 17,675,316 | 17,632,400 | 11m36s | 1m6s |

Some per-image statistics to indicate how consistent the results are:
At default speed (e7), for q70, the worst case was 0.34% larger (after this change compared to before this change), the best case was 0.73% smaller, and the median case was 0.2% smaller. For q90, the worst case was 0.02% larger, the best case was 0.81% smaller, and the median case was 0.43% smaller.

At slowest speed (e9), which becomes about 10x faster with this change, for q70, the worst case was 0.27% larger, the best case was 1.48% smaller, and the median case was 0.35% smaller. At q90, the worst case was 0.08% larger, the best case 1.49% smaller, and the median case 0.6% smaller.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -1408,8 +1408,8 @@ int QuantizeGradient(const int32_t* qrow, size_t onerow, size_t c, size_t x,
 }
 
 void ModularFrameEncoder::AddVarDCTDC(const Image3F& dc, size_t group_index,
-                                      bool nl_dc,
-                                      PassesEncoderState* enc_state) {
+                                      bool nl_dc, PassesEncoderState* enc_state,
+                                      bool jpeg_transcode) {
   const Rect r = enc_state->shared.DCGroupRect(group_index);
   extra_dc_precision[group_index] = nl_dc ? 1 : 0;
   float mul = 1 << extra_dc_precision[group_index];
@@ -1420,6 +1420,13 @@ void ModularFrameEncoder::AddVarDCTDC(const Image3F& dc, size_t group_index,
   stream_options[stream_id].wp_tree_mode = ModularOptions::TreeMode::kWPOnly;
   if (cparams.speed_tier >= SpeedTier::kSquirrel) {
     stream_options[stream_id].tree_kind = ModularOptions::TreeKind::kWPFixedDC;
+  }
+  if (cparams.speed_tier < SpeedTier::kSquirrel && jpeg_transcode) {
+    stream_options[stream_id].predictor =
+        (cparams.speed_tier < SpeedTier::kKitten ? Predictor::Variable
+                                                 : Predictor::Best);
+    stream_options[stream_id].wp_tree_mode = ModularOptions::TreeMode::kDefault;
+    stream_options[stream_id].tree_kind = ModularOptions::TreeKind::kLearn;
   }
   if (cparams.decoding_speed_tier >= 1) {
     stream_options[stream_id].tree_kind =
--- a/lib/jxl/enc_cache.cc
+++ b/lib/jxl/enc_cache.cc
@@ -165,7 +165,7 @@ void InitializePassesEncoder(const Image3F& opsin, ThreadPool* pool,
           dc, group_index,
           enc_state->cparams.butteraugli_distance >= 2.0f &&
               enc_state->cparams.speed_tier < SpeedTier::kFalcon,
-          enc_state);
+          enc_state, /*jpeg_transcode=*/true);
     };
     RunOnPool(pool, 0, shared.frame_dim.num_dc_groups, ThreadPool::SkipInit(),
               compute_dc_coeffs, "Compute DC coeffs");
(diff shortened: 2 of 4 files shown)


lib/jxl/enc_modular.cc
```diff
@@ -1421,6 +1421,13 @@ void ModularFrameEncoder::AddVarDCTDC(const Image3F& dc, size_t group_index,
   if (cparams.speed_tier >= SpeedTier::kSquirrel) {
     stream_options[stream_id].tree_kind = ModularOptions::TreeKind::kWPFixedDC;
   }
+  if (cparams.speed_tier < SpeedTier::kSquirrel && !nl_dc) {
```
veluca93, line 1424: Is nl_dc false IFF we are doing JPEG recompression? I'd be slightly surprised...

jonsneyers: Oh you're right, it's false IF we do JPEG recompression but it can also be false in other cases (when distance < 2).

So the way I wrote it also makes regular VarDCT spend more time on DC at effort 8 and 9. That was not intended, but it's probably not a bad idea, actually. At those slow speeds, it's doing a lot of stuff anyway, so spending a bit more time on DC is not going to be a huge problem, I think.

Some quick results: (timing is not accurate at all, laptop was running hot after doing the 'before')

Before:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:8           57482 11026557    1.5345979   0.304  19.021   2.80633187   0.62708369  0.962321348304      0
Aggregate:      57482 11026557    1.5345979   0.304  19.021   2.80633187   0.62708369  0.962321348304      0
```

After:

```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:8           57482 11006224    1.5317681   0.266  15.616   2.80633187   0.62708369  0.960546825216      0
Aggregate:      57482 11006224    1.5317681   0.266  15.616   2.80633187   0.62708369  0.960546825216      0
```

So for e8 that's a 0.18% overall density improvement for d1, at the cost of slower encode and decode. Not sure if that's worth it, but it may be (after all, if you really care about encode/decode speed more than about compression, you'll likely not use -e 8 and --faster_decoding 0).

veluca93:
I'd prefer to keep that in a separate change though :)
I'm not sure it is universally a good idea...

jonsneyers:
> I'd prefer to keep that in a separate change though :) I'm not sure it is universally a good idea...

Yes, it was not the intention of this PR to change from-pixels VarDCT, and if we're going to change that, I agree it should be in a separate PR with proper benchmarking of impact on density and speed.

jonsneyers:
Changed it so the changes in DC encoding apply only to the jpeg recompression case, not the general VarDCT case where it accidentally also applied.

veluca93 (approved):

lib/jxl/enc_modular.cc
jyrkialakuijala: the impact on vardct is likely much more at high distance where dc is more bytes (can be up to half at high distances like d32 range), i.e., this may help to catch up a little with AVIF at super low qualities

jonsneyers: Currently vardct uses three types of DC encoding:

- d<2: it just quantizes the DC and stores the quantized DC losslessly <-- this is where spending more effort on that lossless encoding does help for compression (though it hurts for enc/dec speed when it uses non-specialized code paths)
- 2 <= d < 4.5: it quantizes the DC in a way that is near-lossless: basically quantized predictor residuals -1, 0, 1 are kept as is (to have enough precision to avoid banding in slow gradients) and other residuals get their lsb set to zero.
- d > 4.5: a lossy modular DC frame is used, using quantization of Squeeze residuals

For the d > 4.5 case, we could look at ways to tune the DC frame encoding — currently it just does the same thing as if it was just any image to be encoded lossily with modular, which is likely not the best thing to do.

I'm personally not very interested in d > 8 because I think it has no practical applications whatsoever except pleasing people who like to evaluate image codecs at ridiculously low bitrates.

PR #693, opened 2021-10-06
Implement outputting box data (non-brob)

lvandeve:
This implements outputting box data to user buffers. In this PR, only
non-brob boxes are supported, the decompression with brotli is not yet
implemented.

Also added list of boxes to the jxlinfo.c example, and fixed the
decode_exif_metadata.cc example to ensure it works.

### DIFF
--- a/examples/jxlinfo.c
+++ b/examples/jxlinfo.c
@@ -25,9 +25,9 @@ int PrintBasicInfo(FILE* file) {
 
   JxlDecoderSetKeepOrientation(dec, 1);
 
-  if (JXL_DEC_SUCCESS !=
-      JxlDecoderSubscribeEvents(
-          dec, JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING | JXL_DEC_FRAME)) {
+  if (JXL_DEC_SUCCESS != JxlDecoderSubscribeEvents(
+                             dec, JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING |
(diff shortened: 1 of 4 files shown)


examples/jxlinfo.c
```diff
@@ -25,9 +25,9 @@ int PrintBasicInfo(FILE* file) {
 
   JxlDecoderSetKeepOrientation(dec, 1);
 
-  if (JXL_DEC_SUCCESS !=
-      JxlDecoderSubscribeEvents(
-          dec, JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING | JXL_DEC_FRAME)) {
+  if (JXL_DEC_SUCCESS != JxlDecoderSubscribeEvents(
+                             dec, JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING |
+                                      JXL_DEC_FRAME | JXL_DEC_BOX)) {
```
veluca93, line 30: I find the indentation odd here

veluca93 (approved):

examples/jxlinfo.c
lvandeve: I agree with that, but unfortunately ./ci.sh lint | patch -p1 keeps producing this indentation even when starting with the whole statement on one line

I prefer to keep the automated output anyway

veluca93: Agreed, very peculiar output from clang-format...

veluca93 (approved):

Issue #699, opened 2021-10-07
error printed by jxlinfo


novomesk:
[jxl_info.zip](https://github.com/libjxl/libjxl/files/7303233/jxl_info.zip) contains **bucuresti2.jxl** file.

It was encoded via following command:

`cjxl 10921852336_28ebe2249c_o.jpg bucuresti2.jxl -p --centerfirst -d 10 --num_threads=2 --strip -e 5`

When I use **jxlinfo**, I get some error:
```
 jxlinfo bucuresti2.jxl
dimensions: 10916x9985
have_container: 0
uses_original_profile: 0
bits_per_sample: 8
have_preview: 0
have_animation: 0
orientation: 1 (Normal)
num_color_channels: 3
num_extra_channels: 0
color profile:
  format: JPEG XL encoded color profile
  color_space: 0 (RGB color)
  white_point: 1 (D65)
  primaries: 1 (sRGB)
  transfer_function: 13 (sRGB)
  rendering_intent: 1 (Relative)
frame:
  still frame, unnamed
/var/tmp/portage/media-libs/libjxl-9999/work/libjxl-9999/lib/jxl/passes_state.cc:55: JXL_FAILURE: kUseDcFrame specified for dc_level 0, but no frame was decoded with level 1
/var/tmp/portage/media-libs/libjxl-9999/work/libjxl-9999/lib/jxl/dec_frame.cc:289: JXL_RETURN_IF_ERROR code=1: InitializePassesSharedState(frame_header_, &dec_state_->shared_storage)
Decoder error
```

novomesk:
When don't use ParallelRunner, it is more stable (but very slow).

PR #700, opened 2021-10-07
return early (before init/allocation) if no actual decode will happen

jonsneyers:
When no actual decode will happen, just parsing of frameheader/TOC, there is no need to actually initialize things and allocate internal stuff that will be used during decode. Also it's not a good idea to check things like the presence of DC frames since they will not have been decoded either, but that's OK since the frame itself will not be decoded. So just return early (after reading the frameheader and TOC, but before the actual frame decoder initialization).

This fixes https://github.com/libjxl/libjxl/issues/699 and somewhat reduces the memory footprint of `jxlinfo` and other applications that just want to get info about frames without doing actual decode.


lvandeve:
Looks correct to me from the API point of view

PR #703, opened 2021-10-07
Remove Robust and Descriptive statistics.

veluca93:
They were for the most part unused in the main library, and caused some
interop problems because of the use of RNGs. This gets rid of a
significant amount of LoC (~1k), as well as the only uses of
std::normal_distribution and std::poisson_distribution.

Also converts the benchmarks in gauss_blur_test to use gbench.

No BPP or pnorm changes.

```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:d0.7        13270  3401128    2.0503469   3.860  23.290   1.47008669   0.48388256  0.992127114835      0
jxl:d1          13270  2682333    1.6170262   4.023  24.737   1.78285837   0.61856733  1.000239584869      0
jxl:d2          13270  1693780    1.0210838   4.286  25.686   3.12266779   0.98560314  1.006383363499      0
jxl:d4          13270  1017240    0.6132362   3.840  19.237   5.74765110   1.58872606  0.974264363420      0
jxl:d8          13270   580162    0.3497467   3.691  18.627   9.35825157   2.58073121  0.902602273036      0
Aggregate:      13270  1555941    0.9379884   3.935  22.126   3.37857946   1.03878140  0.974364851568      0

Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:d0.7        13270  3401128    2.0503469   3.714  24.224   1.47008669   0.48388256  0.992127114835      0
jxl:d1          13270  2682333    1.6170262   3.867  25.641   1.78285837   0.61856733  1.000239584869      0
jxl:d2          13270  1693780    1.0210838   4.263  24.652   3.12266779   0.98560314  1.006383363499      0
jxl:d4          13270  1017240    0.6132362   3.731  20.015   5.74765110   1.58872606  0.974264363420      0
jxl:d8          13270   580162    0.3497467   3.645  18.540   9.35825157   2.58073121  0.902602273036      0
Aggregate:      13270  1555941    0.9379884   3.838  22.434   3.37857946   1.03878140  0.974364851568      0
```

Fixes #698.

Issue #698, opened 2021-10-07
RobustStatisticsTest failing on Alpine aarch64, ppc64le, s390x


Hello71:
**Describe the bug**
I am packaging libjxl for Alpine Linux. Some tests failing, but I am unsure of the cause.

**To Reproduce**
Build libjxl on Alpine Linux and run tests.

**Expected behavior**
Successful build and tests.

**Environment**
 - OS: Alpine Linux
 - Compiler version: gcc 10.3.1_git20210921-r1 (see https://git.alpinelinux.org/aports/tree/main/gcc for details), musl 1.2.2
 - CPU type: aarch64, ppc64le, s390x, x86 (32-bit, 64-bit is ok)
 - libjxl version: 0.6

**Additional context**
For x86, the test suite requires `-msse -mfpmath=sse` in order to pass. For aarch64, ppc64le, and s390x, three tests fail in identical ways: RobustStatisticsTest.TestMode, RobustStatisticsTest.CleanLine, and DecodeTest.PixelTestWithICCProfileLossy. The last one affects all archs and is tracked as #500 already.

<details><summary>aarch64, ppc64le, and s390x failing test outputs (identical except for duration)</summary>

```
[ RUN      ] RobustStatisticsTest.TestMode
===========================frac 1
hsm 10.0 hrm 9 argmax 9.0 median 9.999786
1510 1761 1805 1810 1611
===========================frac 2
hsm 20.0 hrm 19 argmax 19.0 median 19.999693
1206 1227 1335 1310 1222
===========================frac 3
hsm 30.0 hrm 29 argmax 29.0 median 29.999731
1079 1092 1109 1110 1004
===========================frac 4
hsm 40.0 hrm 40 argmax 40.0 median 39.999756
895 937 956 916 850
===========================frac 5
hsm 49.0 hrm 49 argmax 49.0 median 49.999603
771 884 870 844 791
===========================frac 6
hsm 60.0 hrm 60 argmax 60.0 median 59.999840
768 756 785 768 723
===========================frac 7
/builds/alxu/aports/testing/libjxl/src/libjxl-0.6/lib/jxl/robust_statistics_test.cc:75: Failure
Value of: is_near(hsm, argmax) || is_near(hsm, median)
  Actual: false
Expected: true
hsm 69.0 hrm 69 argmax 70.0 median 69.999565
708 703 736 691 670
===========================frac 8
hsm 79.0 hrm 79 argmax 79.0 median 79.999634
655 692 700 700 616
[...]

PR #706, opened 2021-10-07
Remove (almost) all uses of C++ distributions/RNGs.

veluca93:
C++ distribution implementations may be different between libstdc++ and
libc++, causing inconsistencies in tests under say msan and possibly
different (Linux) distributions.

This PR implements a few common distributions and a RNG and uses it
throughout the codebase, except in a couple of places (fuzzer_corpus
mostly) where consistency is not as important.

Diff-based on https://github.com/libjxl/libjxl/pull/703

### DIFF
--- a/lib/extras/codec_test.cc
+++ b/lib/extras/codec_test.cc
@@ -9,13 +9,13 @@
 #include <stdio.h>
 
 #include <algorithm>
-#include <random>
 #include <utility>
 #include <vector>
 
 #include "gtest/gtest.h"
 #include "lib/extras/codec_pgx.h"
 #include "lib/extras/codec_pnm.h"
+#include "lib/jxl/base/random.h"
 #include "lib/jxl/base/thread_pool_internal.h"
 #include "lib/jxl/color_management.h"
 #include "lib/jxl/image.h"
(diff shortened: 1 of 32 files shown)


veluca93:
It shouldn't: slightly different results here should not diverge as the
inner computation is still done with integers and always the same number of
times. The previous problems were because of the sequence of random numbers
diverging...

veluca93:
> fwiw, alpine uses libstdc++ from gcc like the vast majority of distros. it's only the libc that's different (glibc vs musl), plus some gcc patches which are mostly backports. some distros use libc++ and llvm toolchain but alpine does not. which is not to say that this is necessarily a bad change, just that the issues in #698 are not related to libstdc++ vs libc++.

That is indeed correct - the changes in this PR are mostly related to our own testing infrastructure (in particular msan). The issues in #698 *may* be related to the libc change (I am not sure if the implementation of RNGs in libc++/libstdc++ depends on libc in some ways), but they are "fixed" in #703 simply by removing the failing tests and the code they were testing, since they were not really used anyway :)

Hello71:
> That is indeed correct - the changes in this PR are mostly related to our own testing infrastructure (in particular msan). The issues in #698 _may_ be related to the libc change (I am not sure if the implementation of RNGs in libc++/libstdc++ depends on libc in some ways), but they are "fixed" in #703 simply by removing the failing tests and the code they were testing, since they were not really used anyway :)

ah, i see. good to know :)

PR #721, opened 2021-10-12
adjust MA tree size limits

jonsneyers:
The global MA tree limit was computed without taking extra channels into account. Also added a hard cap of 1<<22 for global trees and 1<<20 for local ones.


deymo (dismissed):

jonsneyers:
CI was stuck in the Windows builds (waiting for response for 12 days) and I had no way to restart it, so I rebased. Please re-approve.

veluca93 (approved):

PR #724, opened 2021-10-13
Windows: add clang++ warning flags only when clang compiler is used

vtorri:


vtorri:
@googlebot I signed it!

PR #746, opened 2021-10-19
Use PRIuS/PRIdS macros to printf size_t/ssize_t.

### LINKED ISSUE #718: Windows: %z is not recognized
With some (old) Visual Studio versions, %z is not supported (as well as on MSYS2 with the macro __USE_MINGW_ANSI_STDIO set to 0). This leads to huge number of warnings. It can be fixed with a macro, like:

```
#ifdef _WIN32
# define FMT_ZU "%Iu"
#else
# define FMT_ZU "%zu"
#endif
```

and replace "foo %zu bar" with "foo " FMT_ZU " bar"

deymo:
PRIuS is defined to either "zu" or "Iu" depending on the platform
allowing to print size_t values in MSYS2. This macro is meant to be used
similarly to PRIu32, so that the "%" is not part of the macro allowing
expressions like `"%3" PRIuS` instead of `"%3zu"`. Similarly, `PRIdS` is
used for ssize_t.

This patch replaces all the occurrences and adds a test to bash_test.sh.

Fixes #718.

### DIFF
--- a/bash_test.sh
+++ b/bash_test.sh
@@ -84,6 +84,17 @@ test_copyright() {
   return ${ret}
 }
 
+# Check that we don't use "%zu" or "%zd" in format string for size_t.
+test_printf_size_t() {
+  if grep -n -E '%[0-9]*z[udx]' \
+      $(git ls-files | grep -E '(\.c|\.cc|\.cpp|\.h)$'); then
+    echo "Don't use '%zu' or '%zd' in a format string, instead use " \
+      "'%\" PRIuS \"' or '%\" PRIdS \"'." >&2
+    return 1
+  fi
+  return 0
+}
+
 # Check that "dec_" code doesn't depend on "enc_" headers.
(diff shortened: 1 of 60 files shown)


vtorri:
@deymo plenty of warnings like:

```
C:/Documents/msys2/home/vincent.torri/gitroot_64/libjxl_main/lib/jxl/aux_out.h:124:16: warning: unknown conversion type character 'I' in format [-Wformat=]
  124 |         printf("   [c/i:%6.2f | hst:%8" PRIdS " | ex:%8" PRIdS
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  125 |                " | h+c+e:%12.3f",
      |                ~~~~~~~~~~~~~~~~~

```

with __USE_MINGW_ANSI_STDIO set to 0 or 1

deymo:
almost correct

PR #751, opened 2021-10-20
double precision slow IDCT for experiments

veluca93:


vtorri:
On Windows, you must define _USE_MATH_DEFINES before including math.h (or cmath) so that M_PI and other math constants are defined. See https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

This will fix Windows compilation

jonsneyers:
Thanks but this is not a PR that will ever be merged, it is just for experimenting, in particular to check how much error we get from using single precision floats.

I'll close it now, we don't need to keep it open for it to serve its purpose.

PR #780, opened 2021-10-27
Use enum for encoder distance and effort options

lvandeve:
All options use the enum now.

The JxlEncoderOptionsSetLossless will be handled in a different MR,
since it's actually a combination of a modular, quality_pair and
color_transform option that still need to be added. The
JxlEncoderUseContainer option may also need some rework when
API to add boxes is added.

EDIT: also did this for the forgotten JxlEncoderOptionsSetDecodingSpeed option

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -73,42 +73,58 @@ typedef enum {
 } JxlEncoderStatus;
 
 /**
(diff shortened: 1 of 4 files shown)


deymo:
LGTM, @jonsneyers  ?

jonsneyers (approved):

jonsneyers:
LGTM.

PR #792, opened 2021-10-28
make djxl_conformance return all extra channels

jonsneyers:
Builds on https://github.com/libjxl/libjxl/pull/791.

Makes djxl_conformance produce a npy array that contains all extra channels, not just alpha like it is now.

### DIFF
--- a/tools/conformance/djxl_conformance.cc
+++ b/tools/conformance/djxl_conformance.cc
@@ -61,19 +61,38 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
 struct ImageArray {
   uint32_t xsize, ysize;
   uint32_t num_channels;
+  uint32_t num_color_channels;
+  uint32_t num_extra_channels;
 
   // An array of "frames", where each frame is a 3D array of samples with
   // dimensions (ysize, xsize, channel).
   std::vector<std::vector<uint8_t>> frames;
+  // Extra channels, as an array of frames, where each frame is an array of 2D
+  // arrays of samples
+  std::vector<std::vector<std::vector<uint8_t>>> ec_frames;
 };
 
 // Saves an ImageArray as a numpy 4D ndarray in binary format.
 bool SaveNPYArray(const char* filename, const ImageArray& arr) {
-  size_t image_size = sizeof(float) * arr.xsize * arr.ysize * arr.num_channels;
+  size_t image_size =
+      sizeof(float) * arr.xsize * arr.ysize * arr.num_color_channels;
+  size_t ec_size = sizeof(float) * arr.xsize * arr.ysize;
   for (const auto& frame : arr.frames) {
     if (frame.size() != image_size) {
       fprintf(stderr, "Invalid frame size\n");
       return false;
+    }
+  }
+  for (const auto& frame : arr.ec_frames) {
+    if (frame.size() != arr.num_extra_channels) {
+      fprintf(stderr, "Invalid extra channel count\n");
+      return false;
+    }
+    for (const auto& ch : frame) {
+      if (ch.size() != ec_size) {
+        fprintf(stderr, "Invalid extra channel size\n");
+        return false;
+      }
     }
   }
 
@@ -106,8 +125,18 @@ bool SaveNPYArray(const char* filename, const ImageArray& arr) {
     WRITE_TO_FILE(ss.str().data(), ss.str().size());
   }
 
-  for (const auto& frame : arr.frames) {
-    WRITE_TO_FILE(frame.data(), frame.size());
+  // interleave the samples from color and extra channels
+  for (size_t f = 0; f < arr.frames.size(); ++f) {
+    size_t pos = 0;
+    for (size_t y = 0; y < arr.ysize; ++y) {
+      for (size_t x = 0; x < arr.xsize; ++x, pos += sizeof(float)) {
+        WRITE_TO_FILE(arr.frames[f].data() + pos * arr.num_color_channels,
+                      arr.num_color_channels * sizeof(float));
+        for (size_t i = 0; i < arr.num_extra_channels; i++) {
+          WRITE_TO_FILE(arr.ec_frames[f][i].data() + pos, sizeof(float));
+        }
+      }
+    }
   }
 
   return fclose(file) == 0;
@@ -316,6 +345,7 @@ bool DecodeJXL(const DecodeOptions& opts) {
   }
 
   EXPECT_SUCCESS(JxlDecoderSubscribeEvents(dec.get(), events));
+  EXPECT_SUCCESS(JxlDecoderSetRenderSpotcolors(dec.get(), JXL_FALSE));
 
   // TODO(deymo): Consider using a multi-threading decoder for conformance
   // testing as well.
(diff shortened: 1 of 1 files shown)


lvandeve (dismissed):

tools/conformance/djxl_conformance.cc
```diff
@@ -61,21 +61,40 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
...
   uint32_t xsize, ysize;
   uint32_t num_channels;
+  uint32_t num_color_channels;
+  uint32_t num_extra_channels;
 
   // An array of "frames", where each frame is a 3D array of samples with
   // dimensions (ysize, xsize, channel).
   std::vector<std::vector<uint8_t>> frames;
+  // Extra channels, as an array of frames, where each frame is an array of 2D
+  // arrays of samples
```
deymo, line 71: what are the dimensions of these 2D arrays? explain how is this interleaved in the .npy

tools/conformance/djxl_conformance.cc
```diff
@@ -61,21 +61,40 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
 struct ImageArray {
   uint32_t xsize, ysize;
   uint32_t num_channels;
+  uint32_t num_color_channels;
+  uint32_t num_extra_channels;
 
   // An array of "frames", where each frame is a 3D array of samples with
   // dimensions (ysize, xsize, channel).
```
deymo, line 68: This dimension comment needs updating to mention which channels.

tools/conformance/djxl_conformance.cc
```diff
@@ -61,21 +61,40 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
 struct ImageArray {
```
deymo, line 61: Before, it was clear what this ImageArray represented because there wasn't much option. Now it requires some explanation of what the color_channels and num_extra channels are. Do we need three numbers?

tools/conformance/djxl_conformance.cc
```diff
@@ -106,8 +125,18 @@ bool SaveNPYArray(const char* filename, const ImageArray& arr) {
     WRITE_TO_FILE(ss.str().data(), ss.str().size());
   }
 
-  for (const auto& frame : arr.frames) {
-    WRITE_TO_FILE(frame.data(), frame.size());
+  // interleave the samples from color and extra channels
+  for (size_t f = 0; f < arr.frames.size(); ++f) {
+    size_t pos = 0;
+    for (size_t y = 0; y < arr.ysize; ++y) {
+      for (size_t x = 0; x < arr.xsize; ++x, pos += sizeof(float)) {
```
deymo, line 132: it might greatly improve performance (and clarity) to interleave in a std::vector<float> of num_color_channels + num_extra_chanells and then write that, even maybe a whole row.

deymo (dismissed):

lvandeve:
I tried "merge conflicts" in this pull request, and this affected the "PR / authors" check. I don't know how to revert this

deymo:
The commit that you are trying to merge has like author your user, but with a hidden email address.

deymo:
This needs to be rebased instead of merging main into it.

lvandeve:
So the question is how that wrong Merge commit I created with the online github interface can be removed again

deymo:
@lvandeve there I rebased it locally and push again

The PR says:

> Add more commits by pushing to the conformance_extrachannels branch on jonsneyers/libjxl.

`git push git@github.com:jonsneyers/libjxl.git HEAD:conformance_extrachannels` lets you push your HEAD to this PR. just make sure that when you rebase you rebase the right branch, not like I did the first time :-p

tools/conformance/djxl_conformance.cc
lvandeve: I'd say this can stay as-is for now, since the frames and ec_frames array layout makes sense from the input side, the API outputs the data in that form. So the API writes to the interleaved color and separate extra channel form, then this loop converts it into the numpy array form, I think this is good to stay as-is.

tools/conformance/djxl_conformance.cc
lvandeve: I updated the code comments in Jon's MR, documented num_channels, as well as improved the documentation of the frames and ec_frames array layout. PTAL

tools/conformance/djxl_conformance.cc
lvandeve: Updated the documentation

lvandeve:
I updated Jon's MR with more comments. I also verified that the spot image test now runs for the level 5 conformance tests.

veluca93 (approved):

tools/conformance/djxl_conformance.cc
deymo: I mean to do the interleave in a buffer here in this for loop, instead of calling write() all the time. Before we had 1 call to write() now we have one per pixel per ec channel.

tools/conformance/djxl_conformance.cc
```diff
@@ -60,22 +60,52 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
 
 struct ImageArray {
   uint32_t xsize, ysize;
+  // total amount of channels: num_color_channels + num_extra_channels
```
deymo, line 63: can we either assert this in this code or remove the member?

tools/conformance/djxl_conformance.cc
```diff
@@ -60,22 +60,52 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
...
+  // amount of color channels: 1 for grayscale, 3 for RGB
+  uint32_t num_color_channels;
+  // amount of extra channels, including alpha channels, spot colors, ...
+  uint32_t num_extra_channels;
+
+  // Both frames and ec_frames are filled in by the JXL decoder, and will be
+  // converted into a numpy array of the form (frame, ysize, xsize, channel)
+
+  // Array of the color channels of the frames. This is an array of frames,
+  // where each frame is an array of scanlines, where each scanline contains
```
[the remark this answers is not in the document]
lvandeve: Right, updated the documentation to reflect this, also mentioned they're little endian 32-bit float

tools/conformance/djxl_conformance.cc
lvandeve: Removed it

tools/conformance/djxl_conformance.cc
```diff
@@ -60,22 +60,52 @@ bool SaveFile(const char* filename, std::vector<uint8_t> data) {
...
+  // the pixels with the color channels interleaved: the order of indices is
+  // (frame, ysize, xsize, color_channel_index)
   std::vector<std::vector<uint8_t>> frames;
+
+  // Array of the extra channels of the frames. This is an array of frames,
+  // where each frame is an array of extra channels, where each extra channel
+  // is an array of scanlines of pixels: the order of indices is
+  // (frame, extra_channel_index, ysize, xsize)
```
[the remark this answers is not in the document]
lvandeve: Fixed

deymo (approved):

PR #804, opened 2021-11-01
Check ftyp validity

lvandeve:
Check if the ftyp box is present, at the correct position, and only
once, and the major brand matches "jxl "

This makes the decoder more stringent about this, which aids in
conformance.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -192,6 +192,7 @@ enum class FrameStage : uint32_t {
 enum class BoxStage : uint32_t {
   kHeader,      // Parsing box header of the next box, or start of non-container
                 // stream
+  kFtyp,        // The ftyp box
   kSkip,        // Box whose contents are skipped
   kCodestream,  // Handling codestream box contents, or non-container stream
   kPartialCodestream,  // Handling the extra header of partial codestream box
@@ -462,6 +463,7 @@ struct JxlDecoderStruct {
   // Fields for reading the basic info from the header.
   size_t basic_info_size_hint;
   bool have_container;
+  size_t box_count;
 
   // Whether the preview out buffer was set. It is possible for the buffer to
   // be nullptr and buffer_set to be true, indicating it was deliberately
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

lib/jxl/decode.cc
```diff
@@ -1743,6 +1746,23 @@ JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec) {
                sizeof(dec->box_decoded_type));
       }
 
+      // Box order validity checks
+      // The signature box at box_count == 1 is not checked here since that's
+      // already done at the beginning.
+      dec->box_count++;
+      if (dec->box_count == 2) {
```
deymo, line 1753: can't there be other, non-jxl-related boxes?

jonsneyers: No, ftyp has to be the first box after the signature box.

lvandeve:
I updated this pull request, to make the check less strict: it now no longer requires the exact same 20 bytes

Even though the specification requires those exact 20 bytes, there may be reasons to do some changes to the ftyp box in the future and still have them compatible with existing decoders

veluca93 (dismissed):

lib/jxl/decode.cc
```diff
@@ -1743,6 +1746,17 @@ JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec) {
                sizeof(dec->box_decoded_type));
       }
 
+      // Box order validity checks
+      // The signature box at box_count == 1 is not checked here since that's
+      // already done at the beginning.
+      dec->box_count++;
+      if (dec->box_count == 2 && memcmp(dec->box_type, "ftyp", 4) != 0) {
+        return JXL_API_ERROR("there second box must be the ftyp box");
```
jonsneyers, line 1754: "the second box"

jonsneyers (approved):

lib/jxl/decode.cc
lvandeve: fixed :)

veluca93 (approved):

PR #808, opened 2021-11-02
Add modular setting to encoder

lvandeve:
Add setting to the encoder to force modular or vardct mode, or use the
default (where in future implementations, encoder is allowed to e.g. use
heuristics on pixel content to choose this)

Also lets the SetDistance option set the quality_pair setting for
modular too. As dicussed with Jon, it's ok to translate distance to
a quality_pair value and expose only one setting in the API, rather than
expose the distance setting for vardct, and the quality pair setting for
modular, separately. The conversion formulas come from cjxl.cc and can
be tweaked if necessary.

Also alters the documentation of the SetLossless setting to specify what
this setting actually does as implemented, and do not be as strict in the
documentation which settings exactly this overrides since this can be
future implementation dependent.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -129,6 +129,12 @@ typedef enum {
    */
   JXL_ENC_OPTION_GABORISH = 7,
 
+  /** Enables modular encoding. Integer option, use -1 for default (encoder
+   * chooses), 0 to enforce VarDCT mode (e.g. for photographic images), 1 to
+   * enforce modular mode (e.g. for lossless images).
+   */
+  JXL_ENC_OPTION_MODULAR = 8,
+
   /** Enum value not to be used as an option. This value is added to force the
    * C compiler to have the enum to take a known size.
    */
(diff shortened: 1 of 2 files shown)


lib/include/jxl/encode.h
```diff
@@ -405,10 +411,15 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetCodestreamLevel(JxlEncoder* enc,
                                                          int level);
 
 /**
- * Sets lossless/lossy mode for the provided options. Default is lossy.
+ * Enables lossless encoding options override. When enabled, overrides all
```
deymo, line 414: nit: I don't think the concept of "override" is the most clear:  "this setting over here is the distance; this setting over there overrides other settings" is more confusing IMHO than scoping the "distance" setting with the lossless: "this is the lossy distance setting when lossless is disabled". Somehow I just don't find this sentence very clear: "Enables lossless encoding options override."

deymo (dismissed):

lib/include/jxl/encode.h
lvandeve: I've updated the comment to be more clear.

We could also think about replacing this function with one without argument (that then directly sets all relevant options rather than be a separate state, since that's much easier to explain), but I'd do that in a separate MR since that's not related to the new modular mode option here (updating the comment is related)

lvandeve:
Updated the comment, PTAL

jonsneyers (approved):

PR #811, opened 2021-11-04
Add JxlDecoderCloseInput to decoder API

lvandeve:
This adds a function to mark input explicitely as closed.

This is required to be used for correctness when box metadata
is gotten, because from streaming input it is not possible to
determine whether the current box is the last box, or the
current bytes of an unbounded last box are the final bytes.
Before, the decoder could return success when the user still
had more input (when requesting box content only), which is
not entirely straightforward.

It should also be used in general from this point on for
consistency, though for backwards compatibility with the
previous released decoder API versions, is not required
to be used when JXL_DEC_BOX is not used.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -467,11 +467,14 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec);
  * Sets input data for JxlDecoderProcessInput. The data is owned by the caller
  * and may be used by the decoder until JxlDecoderReleaseInput is called or
  * the decoder is destroyed or reset so must be kept alive until then.
+ * Cannot be called if JxlDecoderSetInput was already called and
+ * JxlDecoderReleaseInput was not yet called, and cannot be called after
+ * JxlDecoderCloseInput indicating the end of input was called.
  * @param dec decoder object
  * @param data pointer to next bytes to read from
  * @param size amount of bytes available starting from data
- * @return JXL_DEC_ERROR if input was already set without releasing,
- * JXL_DEC_SUCCESS otherwise.
+ * @return JXL_DEC_ERROR if input was already set without releasing or
+ * JxlDecoderCloseInput was already called, JXL_DEC_SUCCESS otherwise.
  */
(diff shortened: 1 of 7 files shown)


lib/include/jxl/decode.h
```diff
@@ -496,6 +496,23 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetInput(JxlDecoder* dec,
...
+ * Marks the input as finished, indicates that no more JxlDecoderSetInput will
+ * be called. This function allows the decoder to determine correctly if it
+ * should return success, need more input or error in certain cases. For
+ * backwards compatibility with a previous version of the API, using this
+ * function is optional when not using the JXL_DEC_BOX event (the decoder is
+ * able to determine the end of the image frames without marking the end), but
+ * using this function is required when using JXL_DEC_BOX for getting metadata
+ * box contents. This function does not replace JxlDecoderReleaseInput, that
+ * function should still be called if its return value is needed.
+ * JxlDecoderSetInput should be called as soon as all known input bytes are set
```
deymo, line 509: You meant JxlDecoderCloseInput here?

deymo (approved):

deymo:
This likely warrants a mention in the changelog.md

lib/include/jxl/decode.h
lvandeve: Fixed

lib/jxl/decode.cc
```diff
@@ -1524,7 +1526,12 @@ JxlDecoderStatus JxlDecoderProcessCodestream(JxlDecoder* dec, const uint8_t* in,
 
 JxlDecoderStatus JxlDecoderSetInput(JxlDecoder* dec, const uint8_t* data,
                                     size_t size) {
-  if (dec->next_in) return JXL_DEC_ERROR;
+  if (dec->next_in) {
```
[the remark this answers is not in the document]
lvandeve: Updated documentation in the header

lvandeve:
> This likely warrants a mention in the changelog.md

Done, also added the box decoding feature in there

deymo (approved):

PR #815, opened 2021-11-05
Add API to encode boxes, header only (no implementation)

lvandeve:
Adds the function JxlEncoderAddBox, which can be used either before or
after encoding image frames.

Updates the JxlEncoderUseContainer documentation to noet say
"force" after all since for metadata boxes you need to call it
explicitely (the encoder cannot know on its own if boxes will be added
at the end only)

Adds documentation about the box format and individual box types so that
it is clear what types of metadata boxes this function can be used for
and which ones the encoder automatically adds.

Pair programmed with Jyrki, who also proposed to add good documentation
about the boxes and a prototype test.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -379,8 +379,128 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelBuffer(
     const JxlEncoderOptions* options, const JxlPixelFormat* pixel_format,
     const void* buffer, size_t size, uint32_t index);
 
+/** Adds a metadata box to the file format. JxlEncoderProcessOutput must be used
+ * to effectively write the box to the output. @ref JxlEncoderUseContainer must
+ * be enabled before using this function.
+ *
+ * Background information about the container format and boxes follows here:
+ *
+ * For users of libjxl, boxes allow inserting application-specific data and
+ * metadata (Exif, XML, JUMBF and user defined boxes).
+ *
+ * The box format follows ISO BMFF and shares features and box types with other
+ * image and video formats, including the Exif, XML and JUMBF boxes. The box
+ * format for JPEG XL is specified in ISO/IEC 18181-2.
+ *
+ * Boxes in general don't contain other boxes inside, except a JUMBF superbox.
+ * Boxes follow each other sequentially and are byte-aligned. If the container
+ * format is used, the JXL stream exists out of 3 or more concatenated boxes.
+ * It is also possible to use a direct codestream without boxes, but in that
+ * case metadata cannot be added.
+ *
+ * Each box generally has the following byte structure in the file:
+ * - 4 bytes: box size including box header (Big endian. If set to 0, an
+ *   8-byte 64-bit size follows instead).
+ * - 4 bytes: type, e.g. "JXL " for the signature box, "jxlc" for a codestream
+ *   box.
+ * - N bytes: box contents.
+ *
+ * Only the box contents are provided to the contents argument of this function,
+ * the encoder encodes the size header itself.
+ *
+ * Box types are given by 4 characters. A list of known types follows:
+ * - "JXL ": mandatory signature box, must come first, 12 bytes long including
+ *   the box header
+ * - "ftyp": a second mandatory signature box, must come second, 20 bytes long
+ *   including the box header
+ * - "jxll": A JXL level box. This indicates if the codestream is level 5 or
+ *   level 10 compatible. If not present, it is level 5. Level 10 allows more
+ *   features such as very high image resolution and bit-depths above 16 bits
+ *   per channel. Added automatically by the encoder when
+ *   JxlEncoderSetCodestreamLevel is used
+ * - "jxlc": a box with the image codestream, in case the codestream is not
+ *   split across multiple boxes. The codestream contains the JPEG XL image
+ *   itself, including the basic info such as image dimensions, ICC color
+ *   profile, and all the pixel data of all the image frames.
+ * - "jxlp": a codestream box in case it is split across multiple boxes. The
+ *   encoder will automatically do this if necessary. The contents are the same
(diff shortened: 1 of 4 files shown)


lib/include/jxl/encode.h
```diff
@@ -264,8 +264,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ *   box.
+ * - N bytes: box contents.
+ *
+ * Only the box contents are provided to the contents argument of this function,
+ * the encoder encodes the size header itself.
+ *
+ * Box types are given by 4 characters. A list of known types follows:
+ * - "JXL ": mandatory signature box, must come first, 12 bytes long including
+ *   the box header
+ * - "ftyp": a second mandatory signature box, must come second, 20 bytes long
```
jonsneyers, line 299: Theoretically it could be > 20 bytes long too.

lib/include/jxl/encode.h
```diff
@@ -264,8 +264,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * header offset, which may be 4 zero-bytes. The ISO BMFF box header must not
+ * be included, only the contents.
+ * @param size size of the box contents.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error, such as when
+ * using this function without JxlEncoderUseContainer, or adding a box type
+ * that would result in an invalid file format.
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderAddBox(JxlEncoder* enc, JxlBoxType type,
+                                             const uint8_t* contents,
+                                             size_t size);
```
jonsneyers, line 379: Maybe add an argument `use_brob`? It's fine if it has to be false in the first implementation, but I think that would be the most convenient way to select whether a box is added uncompressed or compressed — an alternative would be to make it an option that is part of the `enc` state, but since you may want to do different things for each box you add, I think it's more convenient to have it as an argument in `JxlEncoderAddBox`.

lib/include/jxl/encode.h
```diff
@@ -264,8 +264,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * JxlEncoderAddImageFrame call, and/or after the last JxlEncoderAddImageFrame
+ * call and JxlEncoderCloseInput. Support for adding boxes in-between the
+ * codestream, and/or in-between image frames may be added later, and would
+ * cause the encoder to use jxlp boxes for the codestream.
+ *
+ * @param enc encoder object.
+ * @param type the box type, e.g. "Exif" for EXIF metadata, "XML " for XMP or
+ * IPTC metadata, "jumb" for JUMBF metadata.
+ * @param contents the full contents of the box, for example EXIF
+ * data. For an "Exif" box, the EXIF data must be prepented by a 4-byte tiff
```
jonsneyers, line 369: prepended

lib/include/jxl/encode.h
```diff
@@ -264,8 +264,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ *
+ * Most boxes are automatically added by the encoder and should not be added
+ * with JxlEncoderAddBox. Boxes that one may wish to add with JxlEncoderAddBox
+ * are: Exif and XML (but not when using JPEG reconstruction since if the
+ * JPEG has those, these boxes are already added automatically), jumb, and
+ * application-specific boxes.
+ *
+ * Adding metadata boxes increases the filesize. When adding Exif metadata, the
+ * data must be in sync with what is encoded in the JPEG XL codestream,
+ * specifically the image orientation.
```
jonsneyers, line 349: Could add something like: `In case of conflicting metadata, the JPEG XL codestream takes precedence.`

lib/include/jxl/encode.h
lvandeve: The specification currently specifies those exact 20 bytes, so for now this makes sense I'd say

lib/include/jxl/encode.h
lvandeve: Indeed, that's a good suggestion, done and named it "compress_box"

lib/include/jxl/encode.h
lvandeve: Fixed

lib/include/jxl/encode.h
lvandeve: Done, but pointed out it's not recommended to do that, encoders should be encouraged to keep them in sync, it's more a decoder feature

jyrkialakuijala (dismissed):

lib/include/jxl/encode.h
```diff
@@ -345,25 +465,29 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
...
+ * when metadata, level 10 features or JPEG reconstruction is used.
  *
  * If enabled, the encoder always uses the container format, even if not
- * necessary. If disabled, the encoder only uses the container format if
- * required (such as for JPEG metadata @ref JxlEncoderStoreJPEGMetadata), and
- * otherwise writes a direct codestream.
+ * necessary. If disabled, the encoder will still use the container format if
+ * required (such as for JPEG metadata @ref JxlEncoderStoreJPEGMetadata).
+ *
+ * This setting must be explicitely enabled before using @ref JxlEncoderAddBox.
```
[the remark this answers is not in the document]
lvandeve: For JPEG reconstruction, you have to call JxlEncoderStoreJPEGMetadata which you call at the beginning as well. While in theory if you call AddBox soon enough the encoder still knows it in time to be able to output the box header, if you call AddBox after already processing frames it's too late. The requirement to call it is then mainly for consistency, to have the same API whether you add boxes at the beginning or at the end. So I believe it's fine to keep as-is and require it.

jyrkialakuijala (approved):

PR #816, opened 2021-11-05
Addes more missing options to the encoder API

lvandeve:
Also added TODO's for the others that cjxl supports that need to be
supported through the API, some of which require more thought.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -109,31 +109,52 @@ typedef enum {
    */
   JXL_ENC_OPTION_EXTRA_CHANNEL_RESAMPLING = 3,
 
-  /** Enables or disables noise generation. Integer option, use -1 for the
-   * encoder default, 0 to disable, 1 to enable.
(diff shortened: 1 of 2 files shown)


lib/include/jxl/encode.h
```diff
@@ -109,31 +109,57 @@ typedef enum {
...
+   */
+
+  /** TODO(lode): photon_noise option, which is float but may be usable as
+   * an integer option instead.
+   */
+
+  /** TODO(lode): center_first, center_x and center_y options.
+   */
+
+  /** TODO(lode): tone mapping settings (intensity_target, min_nits,
```
veluca93, line 159: Aren't those in the basic_info already?

veluca93 (dismissed):

lib/include/jxl/encode.h
lvandeve: Yes they are and the encoder uses them too, great, that's done :)

veluca93 (approved):

PR #820, opened 2021-11-08
Add photon noise setting to encoder API

lvandeve:
Also adds comment that the other noise setting is not recommended, since
the photon noise works better in practice


jyrkialakuijala (dismissed):

jyrkialakuijala (approved):

jyrkialakuijala:
We fix this in a future PR, to simplify our merging efforts.

PR #821, opened 2021-11-08
Add palette options to encoder API

lvandeve:


jyrkialakuijala (dismissed):

lib/jxl/encode.cc
```diff
@@ -393,6 +393,36 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
...
+        options->values.cparams.channel_colors_percent = 80.0f;
+      } else {
+        options->values.cparams.channel_colors_percent =
+            static_cast<float>(value);
+      }
+      return JXL_ENC_SUCCESS;
+    case JXL_ENC_OPTION_PALETTE_COLORS:
+      if (value < -1 || value > 70913) return JXL_ENC_ERROR;
+      if (value == -1) {
+        options->values.cparams.palette_colors = 95.0f;
```
jonsneyers, line 417: Probably want a higher default value here - at least 500 or so tends to be a good idea (also it's an integer, not a float number here, or at least int would make sense).

lib/jxl/encode.cc
```diff
@@ -393,6 +393,36 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
...
+      } else {
+        options->values.cparams.palette_colors = value;
+      }
+      return JXL_ENC_SUCCESS;
+    case JXL_ENC_OPTION_LOSSY_PALETTE:
+      if (value < -1 || value > 1) return JXL_ENC_ERROR;
+      options->values.cparams.lossy_palette = (value == 1);
```
jonsneyers, line 424: If lossy palette is enabled, probably the default max palette size should be something else than what it is normally.

In general there might be a problem that sensible encoder defaults for some settings might depend on the value of other options, which is a bit annoying because it could mean (if we're not careful) that the order in which you set options could matter...

lib/jxl/encode.cc
lvandeve: Fixed, this was supposed to be 1 << 10 just like in enc_params.h, thanks for spotting this!

lib/jxl/encode.cc
lvandeve: When encoding, you set the options at the beginning of the frame, and then with `JxlEncoderProcessOutput` the actual encoding occurs, so the decoder can know the full picture of the combination of settings at that point

Some of the logic can be done in cjxl, the encoder binary itself, as it does not from the command line flags.

For now I'd keep it as-is now, and directly set the settings. Many options support "-1" as default, and are implementation dependent, but not yet implemented that way now. This can be done as part of the transitioning cjxl to use the API, either in cjxl, or with more logic in the API at that point.

So resolving this for now.

lvandeve: Added a TODO here now

jonsneyers (dismissed):

lvandeve:
Had to merge so need re-approval, PTAL

jyrkialakuijala (approved):

PR #823, opened 2021-11-08
encoder api: error codes

jonsneyers:
~~Builds on top of https://github.com/libjxl/libjxl/pull/358 ~~

~~Adds an encoder api return code for the case where the `jbrd` cannot be stored (e.g. because there is too much tail data), but the jpeg image data itself can be recompressed. This is not a fatal error (can still produce a useful jxl that has all the image data and any exif/xmp if there was any), but it's also not just OK (the result is not bit-exact-reconstructible to the original jpeg file).~~

~~Applications that do not know this new return code will likely treat it as an (unknown) encoder error, which is fine (this is what would happen before, when failure to write the `jbrd` would be a fatal encode error).~~

Adds `JxlEncoderGetError` which can be used to check the reason why JXL_ENCODE_ERROR was returned: it could be either bad API usage, so an application bug, or an error condition that is not caused by an application bug and an application might want to react to it, e.g. by retrying JPEG recompression without storing the `jbrd` or by informing the user that the input is invalid.

Also check the tail data size and return an error slightly earlier if it is too large, and catch this error in cjxl, so instead of a somewhat mysterious error like this:

```
./lib/jxl/fields.cc:729: JXL_FAILURE: No feasible selector for 9936393
```

you now get a somewhat more useful error like this:

```
./lib/jxl/jpeg/jpeg_data.cc:367: JXL_FAILURE: Tail data too large (max size = 4260096, size = 9936393).
```

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -73,15 +73,60 @@ typedef enum {
 
   /** DEPRECATED: the encoder does not return this status and there is no need
    * to handle or expect it.
+   * Instead, JXL_ENC_ERROR is returned with error condition
+   * JXL_ENC_ERR_NOT_SUPPORTED.
    */
   JXL_ENC_NOT_SUPPORTED = 3,
 
 } JxlEncoderStatus;
 
 /**
- * Id of encoder options for a frame. This includes options such as the
- * image quality and compression speed for this frame. This does not include
- * non-frame related encoder options such as for boxes.
+ * Error conditions:
+ * API usage errors have the 0x80 bit set to 1
+ * Other errors have the 0x80 bit set to 0
+ */
+typedef enum {
+  /** No error
+   */
+  JXL_ENC_ERR_OK = 0,
+
+  /** Generic encoder error due to unspecified cause
+   */
+  JXL_ENC_ERR_GENERIC = 1,
+
+  /** Out of memory
+   *  TODO(jon): actually catch this and return this error
+   */
+  JXL_ENC_ERR_OOM = 2,
+
+  /** JPEG bitstream reconstruction data could not be
+   *  represented (e.g. too much tail data)
+   */
+  JXL_ENC_ERR_JBRD = 3,
+
+  /** Input is invalid (e.g. corrupt JPEG file or ICC profile)
+   */
+  JXL_ENC_ERR_BAD_INPUT = 4,
+
+  /** The encoder doesn't (yet) support this. Either no version of libjxl
+   * supports this, and the API is used incorrectly, or the libjxl version
+   * should have been checked before trying to do this.
+   */
+  JXL_ENC_ERR_NOT_SUPPORTED = 0x80,
+
+  /** The encoder API is used in an incorrect way.
+   *  In this case, a debug build of libjxl should output a specific error
+   * message. (if not, please open an issue about it)
+   */
+  JXL_ENC_ERR_API_USAGE = 0x81,
+
+} JxlEncoderError;
+
+/**
+ * Id of encoder options for a frame. This includes options such as setting
+ * encoding effort/speed or overriding the use of certain coding tools, for this
+ * frame. This does not include non-frame related encoder options such as for
+ * boxes.
  */
 typedef enum {
   /** Sets encoder effort/speed level without affecting decoding speed. Valid
@@ -236,9 +281,12 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_COLOR_TRANSFORM = 24,
 
-  /** Color space for modular encoding: -1=default, 0-35=reverse color transform
+  /** Reversible color transform for modular encoding: -1=default, 0-48=RCT
    * index, e.g. index 0 = none, index 6 = YCoCg.
-   * The default behavior is to try several, depending on the speed setting.
+   * If this option is set to a non-default value, the RCT will be globally
+   * applied to the whole frame.
+   * The default behavior is to try several RCTs locally per modular group,
+   * depending on the speed and distance setting.
    */
   JXL_ENC_FRAME_SETTING_MODULAR_COLOR_SPACE = 25,
 
@@ -353,6 +401,15 @@ JXL_EXPORT void JxlEncoderSetCms(JxlEncoder* enc, JxlCmsInterface cms);
 JXL_EXPORT JxlEncoderStatus
 JxlEncoderSetParallelRunner(JxlEncoder* enc, JxlParallelRunner parallel_runner,
                             void* parallel_runner_opaque);
+
+/**
+ * Get the (last) error code in case JXL_ENC_ERROR was returned.
+ *
+ * @param enc encoder object.
+ * @return the JxlEncoderError that caused the (last) JXL_ENC_ERROR to be
+ * returned.
+ */
+JXL_EXPORT JxlEncoderError JxlEncoderGetError(JxlEncoder* enc);
 
 /**
  * Encodes JPEG XL file using the available bytes. @p *avail_out indicates how
@@ -715,6 +772,7 @@ JXL_EXPORT void JxlEncoderCloseInput(JxlEncoder* enc);
  * is an alternative to JxlEncoderSetICCProfile and only one of these two must
  * be used. This one sets the color encoding as a @ref JxlColorEncoding, while
  * the other sets it as ICC binary data.
+ * Must be called after JxlEncoderSetBasicInfo.
  *
  * @param enc encoder object.
  * @param color color encoding. Object owned by the caller and its contents are
@@ -730,6 +788,7 @@ JxlEncoderSetColorEncoding(JxlEncoder* enc, const JxlColorEncoding* color);
  * ICC color profile. This is an alternative to JxlEncoderSetColorEncoding and
  * only one of these two must be used. This one sets the color encoding as ICC
  * binary data, while the other defines it as a @ref JxlColorEncoding.
+ * Must be called after JxlEncoderSetBasicInfo.
  *
  * @param enc encoder object.
  * @param icc_profile bytes of the original ICC profile
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -27,13 +27,21 @@
 // Debug-printing failure macro similar to JXL_FAILURE, but for the status code
 // JXL_ENC_ERROR
 #ifdef JXL_CRASH_ON_ERROR
-#define JXL_API_ERROR(format, ...)                                           \
+#define JXL_API_ERROR(enc, error_code, format, ...)                          \
+  (enc->error = error_code,                                                  \
+   ::jxl::Debug(("%s:%d: " format "\n"), __FILE__, __LINE__, ##__VA_ARGS__), \
+   ::jxl::Abort(), JXL_ENC_ERROR)
+#define JXL_API_ERROR_NOSET(format, ...)                                     \
   (::jxl::Debug(("%s:%d: " format "\n"), __FILE__, __LINE__, ##__VA_ARGS__), \
    ::jxl::Abort(), JXL_ENC_ERROR)
 #else  // JXL_CRASH_ON_ERROR
-#define JXL_API_ERROR(format, ...)                                             \
-  (((JXL_DEBUG_ON_ERROR) &&                                                    \
+#define JXL_API_ERROR(enc, error_code, format, ...)                            \
+  (enc->error = error_code,                                                    \
+   ((JXL_DEBUG_ON_ERROR) &&                                                    \
     ::jxl::Debug(("%s:%d: " format "\n"), __FILE__, __LINE__, ##__VA_ARGS__)), \
+   JXL_ENC_ERROR)
+#define JXL_API_ERROR_NOSET(format, ...)                                     \
+  (::jxl::Debug(("%s:%d: " format "\n"), __FILE__, __LINE__, ##__VA_ARGS__), \
    JXL_ENC_ERROR)
 #endif  // JXL_CRASH_ON_ERROR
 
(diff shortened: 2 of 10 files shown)


deymo:
I think that if `store_jpeg_metadata` is true, this is just an encoder error (we can't represent that JPEG in a JXL) so it should just return JXL_ENC_ERROR, I don't see the need for its own error code.

It is a somewhat recoverable error (call the encoder differently and get something else), sure, but so are many others. This goes in the direction of returning more specific error codes in case of error.

jonsneyers:
> I think that if `store_jpeg_metadata` is true, this is just an encoder error (we can't represent that JPEG in a JXL) so it should just return JXL_ENC_ERROR, I don't see the need for its own error code.
> 
> It is a somewhat recoverable error (call the encoder differently and get something else), sure, but so are many others. This goes in the direction of returning more specific error codes in case of error.

Right. It would be useful to return a more specific error code for a case like this. Do you propose to do that in a different way? (e.g. return JXL_ENC_ERROR but have some way to inspect the detailed error?)

I agree that if `store_jpeg_metadata` is true, it is better to return error than to automatically (conceptually) 'try again' with `store_jpeg_metadata` false.
It might also be useful though to have a new value for `store_jpeg_metadata` that means "true if possible, false otherwise", so you don't have to do the error catching and retrying at the application side, and more importantly, so the encoder isn't doing most of the work twice in that case.

Maybe this kind of 'recover from non-fatal errors' behavior could be configured with a separate option, so you can choose whether you want it to fail whenever it cannot do exactly what you ask, or if you want it to recover from some errors and still produce something (and you can inspect the detailed error state to see if it did do something like that).

deymo:
The question here is whether the "try JPEG recompression if possible otherwise ignore jbrd" is a common use case. I don't think it is, you either need lossless reconstruction, in which case you either compress it to .jxl if possible, otherwise send the original .jpg; or you don't need byte-lossless recompression, in which case the jbrd box might be waste even in the cases where we can generate it.

We talked about this for the decoder case; have a function to get a more specific error code. This didn't rank high in the priorities compared to missing api blocking moving djxl/cjxl to use the C API.

jonsneyers:
I think for the use case of converting a personal photo album of jpeg images to jxl, the possibility to reconstruct to jpeg is typically desirable, but not critical: preserving the image data itself, together with exif/xmp, is also OK.

Anyway, I agree that this is low priority.

Without detailed encode error code, it is impossible to distinguish recoverable errors (by using different encode settings) from fatal errors (e.g. the input is not a jpeg at all). I think it is somewhat more important to have at least that distinction in the encode errors, than it is to have detailed decode errors (which are all fatal anyway, afaik, at least currently). If there is no option to do this by default, then I think it has to be at least possible for an application to retry without `jbrd` without having to blindly _always_ retry without `jbrd` even if it failed for completely unrelated reasons.

lvandeve:
I'm also on the side of having a single error code JXL_ENC_ERROR, and using compiler-flag-enabled debug messages to let developers see the underlying error. I even think that JXL_ENC_NOT_SUPPORTED should be removed and replaced by JXL_ENC_ERROR

If programmatic access to a certain error is needed, a separate function could be called after JXL_ENC_ERROR to check for this specific condition.

WDYT?

jonsneyers:
In this case, programmatic access to the error is needed because different errors might require different application reactions, e.g.:

- corrupt input: give up and maybe show a message that the input is bad
- out of memory: maybe retry with cheaper encode settings
- jbrd write failure: maybe retry without storing jbrd
- encode error due to invalid usage (incompatible encode options, doing things in the wrong order, or something like that): no programmatic access needed, but a bug needs to be fixed somewhere

All of the above are "encode errors", but maybe it is useful to at least distinguish between "api usage errors" (in which case debug messages only shown in debug builds are OK for application developers to figure out what they're doing wrong) and "correct api usage but an error condition occurred" (in which case you may want access to an actual specific error code, also in release builds of libjxl).

JXL_ENC_NOT_SUPPORTED is a specific case of "api usage error" that indeed might be redundant - either it's something that will never be supported (and then it's a real api usage error) or it is something that is not supported yet in older versions but is supported in newer versions (and then the application should check the libjxl version before trying to do it, so it's also a kind of api usage error).

For the "correct api usage but error condition occurred" case, I think it might be good to either have a separate JxlEncoderStatus, or to also just return JXL_ENC_ERROR in that case, but have an extra function to get details on the error that tells you that it's not an api usage error but something else happened.

jonsneyers:
Updated this pull request: removing JXL_ENC_NOT_SUPPORTED, instead always returning JXL_ENC_ERROR in case there is an error. The new function JxlEncoderGetError can be used to get a more specific error code, e.g. JXL_ENC_ERR_API_USAGE or JXL_ENC_ERR_BAD_INPUT.

lib/include/jxl/encode.h
```diff
@@ -66,11 +66,49 @@ typedef enum {
...
+
+  /** Input is invalid (e.g. corrupt JPEG file or ICC profile)
+   */
+  JXL_ENC_ERR_BAD_INPUT = 4,
+
+  /** The encoder doesn't (yet) support this. Either no version of libjxl
+   * supports this, and the API is used incorrectly, or the libjxl version
+   * should have been checked before trying to do this.
+   */
+  JXL_ENC_ERR_NOT_SUPPORTED = 0x80,
```
lvandeve, line 103: I think there isn't anything you can programatically do to handle this error, so it's not needed to be able to handle it and it could be merged into JXL_ENC_ERR_GENERIC.

It would best be avoided to rely on encoder error values, except for a very specific programmatic use case, such as jbrd. This because if users of the API do rely on many different possible error values for behavior, this behavior needs to maintained in the library. For many errors, there's nothing you can programmatically do to recover from them, any API usage error shouldn't be returned, but can be printed in the console when compiling with the debug options only.

It could also be argued that JXL_ENC_ERR_BAD_INPUT is an API usage error, and checking of their validity should be done elsewhere.

WDYT?

jonsneyers:
> > In this case, programmatic access to the error is needed because different errors might require different application reactions,
> 
> Your own words explain it. In this case different errors are _not needed_ because the application reaction "might" be different, but it also might not; actually in most cases I can think of it won't be different (just bubble up the error). Different error codes would be **needed** if the application **must** handle them differently. Just because in some use case it might make sense to handle it different means that for every other use case you need to check for all those other error codes that all of them mean "the image was not encoded as requested".

Yes, if you only want to know if the image was encoded as requested or not, you can just check if the return status is OK or ERROR. This pull requests simplifies the return status by no longer having a different case for NOT_SUPPORTED. But if you want to do something meaningful with the ERROR case, it will be useful for many applications to have some way to get a more detailed error. E.g. in cjxl, I would like to show a different error message when recompressing a jpeg file if the input jpeg is corrupt than if the encoder ran out of memory.

thebombzen:
I believe an API misuse error should be returned if `pixel_format->num_channels` is inconsistent with `info->num_color_channels` and possible alpha channel. I accidentally ran into this issue when I set `pixel_format->num_channels` to 1, and `info->num_color_channels` to 3 and it crashed with SIGILL (illegal instruction) which is my fault for screwing that up, but it would be nice if we got an error message saying "don't do that" rather than just have it crash down the line.

lib/include/jxl/encode.h
```diff
@@ -66,11 +66,49 @@ typedef enum {
...
+
+/**
+ * Error conditions:
+ * API usage errors have the 0x80 bit set to 1
+ * Other errors have the 0x80 bit set to 0
+ */
+typedef enum {
+  /** No error
    */
-  JXL_ENC_NOT_SUPPORTED = 3,
```
deymo, line 71: Deprecate this value (stop using it) but don't remove it. Reusing value 3 for the next return code could cause problems.

lib/jxl/encode.cc
```diff
@@ -130,11 +133,12 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
 JxlEncoderStatus JxlEncoderSetColorEncoding(JxlEncoder* enc,
                                             const JxlColorEncoding* color) {
   if (enc->color_encoding_set) {
-    // Already set
-    return JXL_ENC_ERROR;
+    enc->error = JXL_ENC_ERR_API_USAGE;
+    return JXL_API_ERROR("Color encoding is already set");
```
deymo, line 137: In brotli we had a macro/function that sets the enc->error and returns the JXL_ENC_ERROR value. Using such a macro (extending JXL_API_ERROR) instead of manually setting enc->error before returning  would avoid missing places where we should have set enc->error but we forgot. Something like `#define JXL_API_ERROR(enc, error_code, message) ....`

lib/include/jxl/encode.h
jonsneyers: Updated.

lib/jxl/encode.cc
jonsneyers: Updated.

jonsneyers:
> I believe an API misuse error should be returned if `pixel_format->num_channels` is inconsistent with `info->num_color_channels` and possible alpha channel. I accidentally ran into this issue when I set `pixel_format->num_channels` to 1, and `info->num_color_channels` to 3 and it crashed with SIGILL (illegal instruction) which is my fault for screwing that up, but it would be nice if we got an error message saying "don't do that" rather than just have it crash down the line.

Makes sense to check consistency between basic_info, color_encoding and pixel_format as early as possible instead of just allowing anything and letting it crash when it finally hits a check when actual encoding starts.

Updated JxlEncoderSetColorEncoding to check if the color encoding matches the basic info (e.g. return error if setting a grayscale color encoding when num_color_channels is set to 3), and JxlEncoderAddImageFrame to check if the pixel format makes sense too (e.g. return error if the basic info has alpha but the pixel format doesn't). That should help to catch problems earlier. It might be a bit unnecessarily strict now though - in principle we _could_ allow e.g. passing a grayscale, no alpha pixel_format buffer as input to an image with RGBA basicinfo (and just assume alpha is fully-opaque and R=G=B=gray). I'm not sure what would be best here, or the most convenient option for application developers. I think stricter checks might be most convenient in the end, since making assumption and silently making things work that shouldn't work will just cause weird things to happen if people rely on that and then get surprised if other things that shouldn't work actually don't work.

novomesk:
On the decoding side, apps often request 4-channels pixel data (RGBA) even in situations when opened JXL is without alpha channel.

One example is here: https://github.com/libjxl/libjxl/blob/main/examples/decode_oneshot.cc#L52

So far it works (I can read RGB images as RGBA or RGBX) but if it is going to be discouraged in the future, I'd like to know.

jonsneyers:
> On the decoding side, apps often request 4-channels pixel data (RGBA) even in situations when opened JXL is without alpha channel.
> 
> One example is here: https://github.com/libjxl/libjxl/blob/main/examples/decode_oneshot.cc#L52
> 
> So far it works (I can read RGB images as RGBA or RGBX) but if it is going to be discouraged in the future, I'd like to know.

As far as I am concerned, decode side I would certainly not want to discourage that. In a decoder/viewer, it is convenient to 'normalize' everything to RGBA and not have to worry about grayscale, no-alpha, etc, because you don't know what you're getting and it is practical to have only one case to deal with.

In an encoder however, I assume you know what you are inputting, and I see fewer reasons to e.g. allow using RGBA buffers to encode grayscale images.
Maybe there are cases when that is convenient though (e.g. because you already have the buffer as RGBA), but then maybe what is really needed is an encode option that allows the encoder to drop trivial alpha channels and turn images where R=G=B for all pixels into grayscale images, instead of requiring you to check that yourself (so you can set basicinfo/color encoding correctly). If you have to check all pixels anyway anyway, you can just as well copy the pixels to a new buffer, it's not much extra effort at that point.

(The only complication with a "drop trivial alpha / turn R=G=B into grayscale" encode option is that this is image info, not frame info - so if the first frame has only trivial alpha but the second doesn't, there's a problem...)

mo271:
What is the status with this?

jonsneyers:
> What is the status with this?

If there's agreement that something like this is useful to have in the encode api, then I'll make an effort to update this old pull request to the current code.

jonsneyers:
Rebased it, which was quite a bit of work considering it's a 6 month old PR that made changes just about everywhere in encode.cc. I hope nothing got messed up in the process.

jonsneyers:
During the rebasing I also fixed a bug in the handling of JXL_ENC_FRAME_SETTING_MODULAR_COLOR_SPACE (manually setting it to default -1 would cause it to do YCoCg instead of doing the default thing), and changed its behavior to be like in the spec everywhere (which makes cjxl -C 1 do something different than before, but I think that's fine).

Also the documentation for the frame settings enum was misleading, it said these options can be used to change things like image quality but actually image quality is one of the only things that cannot be set with an enum option. So fixed that too.

lib/jxl/encode.cc
```diff
@@ -1477,7 +1652,8 @@ JxlEncoderStatus JxlEncoderSetFrameName(JxlEncoderFrameSettings* frame_settings,
                                         const char* frame_name) {
   std::string str = frame_name ? frame_name : "";
   if (str.size() > 1071) {
-    return JXL_API_ERROR("frame name can be max 1071 bytes long");
+    return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_API_USAGE,
+                         "frame name can be max 1071 bytes long");
```
mo271, line 1656: perhaps mention what str.size() is?

mo271 (approved):

lib/jxl/encode.cc
jonsneyers: Sorry, missed this comment. Yes, could include that in the error message, but if you don't mind I'm not going to update this PR just for that since it's probably not _that_ important.

Issue #459, opened 2021-08-17
Resampled images are being decoded partially (top-left) by libjxl


ziemek99:
**Describe the bug**
Images that use `--resampling=2` (other resampling factors weren't tested) are being shown partially (just the top-left part) in libjxl-dependent apps, e.g. Chrome Canary, Firefox Nightly, nomacs etc.

**To Reproduce**
1. Encode an image with `--resampling=2` option.
2. Decode JXL with `djxl`.
3. Load JXL file to an app that uses libjxl.

**Expected behavior**
Both images, no matter if decoded with `djxl` or shown by libjxl-dependent app, should be shown in full.

**Screenshots**
Expected result (PNG file from `djxl`):
![after](https://user-images.githubusercontent.com/15923635/129734542-a45f45c4-4c6c-4b73-a589-1b8cf665b01e.png)
Chrome Canary (buggy):
![chrome](https://user-images.githubusercontent.com/15923635/129734859-e893d79b-dede-4ce7-9cb0-754fb3369e51.png)
Firefox Nightly (buggy):
![firefox](https://user-images.githubusercontent.com/15923635/129734979-c28b8dbb-3827-48bc-b921-f6b7a7edff76.png)

nomacs (buggy) (different image, submitted by @ScopeBurst on Discord):
![nomacs](https://user-images.githubusercontent.com/15923635/129735326-fb0433af-feb1-4290-b38c-78d0dbd765dc.png)

**Environment**
 - OS: Windows
 - Compiler version: msvc+vcpkg (GitHub artifacts)
 - CPU type: x86_64
 - cjxl/djxl version string: v0.6.0 7193782 [SSE4,Scalar]

xiota:
I see the same behavior with an older commit: 2ebc6ec

jonsneyers:
Looks like a bug in upsampling that depends on decoding config - maybe it works when decoding to full float buffer, but not when decoding with the pixel callback or uint8 output?

ziemek99:
> Looks like a bug in upsampling that depends on decoding config - maybe it works when decoding to full float buffer, but not when decoding with the pixel callback or uint8 output?

That might be correct, I've just loaded the image into GIMP (via `file-jxl.exe` plugin) and it worked perfectly!
![gimp-jxl-resampling-correct](https://user-images.githubusercontent.com/15923635/135104650-d6ef817c-c05d-491e-8682-321529bb3ecd.png)

jonsneyers:
[...]

PR #825, opened 2021-11-08
add test for upsampling and fix uint8 upsampling bug

jonsneyers:
Adds upsampling to decode_test. ~~This test currently fails for uint8 output, due to an actual bug in that codepath. I haven't tried fixing the bug yet, but at least there's now an easy way to check if it is fixed :)~~

Fixes the upsampling bug in the uint8 codepath, thanks to @veluca93.

### DIFF
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -1474,6 +1474,7 @@ struct PixelTestConfig {
   // Exif orientation, 1-8
   JxlOrientation orientation;
   bool keep_orientation;
+  size_t upsampling;
 };
 
 class DecodeTestParam : public ::testing::TestWithParam<PixelTestConfig> {};
@@ -1495,6 +1496,8 @@ TEST_P(DecodeTestParam, PixelTest) {
                                 0};
   jxl::CompressParams cparams;
   cparams.SetLossless();  // Lossless to verify pixels exactly after roundtrip.
+  cparams.resampling = config.upsampling;
+  cparams.ec_resampling = config.upsampling;
   jxl::PaddedBytes compressed = jxl::CreateTestJXLCodestream(
       jxl::Span<const uint8_t>(pixels.data(), pixels.size()), config.xsize,
       config.ysize, orig_channels, cparams, config.add_container,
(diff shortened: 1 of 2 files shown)


lib/jxl/decode_test.cc
```diff
@@ -1540,9 +1543,19 @@ TEST_P(DecodeTestParam, PixelTest) {
...
+                                format_orig, format));
+  } else {
+    // resampling is of course not lossless, so as a rough check:
+    // count pixels that are more than off-by-25 in the 8-bit value of one of
+    // the channels
+    EXPECT_LE(
+        ComparePixels(
+            pixels.data(), pixels2.data(), xsize, ysize, format_orig, format,
+            50.0 * (config.data_type == JXL_TYPE_UINT8 ? 1.0 : 256.0)),
+        300u);
```
lvandeve, line 1557: Maybe this amount could be made dependent on the xsize and ysize rather than hardcoded 300? Some % of pixels allowed to differ

lib/jxl/decode_test.cc
```diff
@@ -1540,9 +1543,19 @@ TEST_P(DecodeTestParam, PixelTest) {
...
   }
-
-  EXPECT_EQ(0u, ComparePixels(pixels.data(), pixels2.data(), xsize, ysize,
-                              format_orig, format));
+  if (config.upsampling == 1) {
+    EXPECT_EQ(0u, ComparePixels(pixels.data(), pixels2.data(), xsize, ysize,
+                                format_orig, format));
+  } else {
+    // resampling is of course not lossless, so as a rough check:
+    // count pixels that are more than off-by-25 in the 8-bit value of one of
```
lvandeve, line 1551: It says 25 in the comment but 50 in the code

jonsneyers: I think that factor is twice the error: if it is 2.0, then the error threshold is 1/255, so an off by one in uint8 terms.

lib/jxl/decode_test.cc
jonsneyers: It also depends on the upsampling factor: iirc about 150 differed with 2x, about 200 with 4x, and about 270 with 8x.

The bounds can be tightened but to find the upsampling bug they are good enough: in the buggy codepath, there are 9000 pixels wrong or so (basically at least 3/4 of the image remains just black...)

jonsneyers:
Might want to cherry pick this one for 0.6.x, unless we're doing a 0.7 soonish.

lvandeve (approved):

ziemek99:
I guess #459 is fixed now...

PR #826, opened 2021-11-08
encoder api: function headers for extra channels

mo271:
Header only change, implementation to follow.

Pair with Lode.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -13,6 +13,7 @@
 #ifndef JXL_ENCODE_H_
 #define JXL_ENCODE_H_
 
+#include "jxl/codestream_header.h"
 #include "jxl/decode.h"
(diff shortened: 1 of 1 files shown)


mo271:
This is a proposal for the new encoder API function. Comments/suggestions welcome.

deymo (approved):
Somewhere we need to explain better how passing the alpha channel in JxlEncoderAddImageFrame works. Is that alpha channel counted in `num_extra_channels`? (so far it wasn't).

veluca93 (approved):

mo271:
> Somewhere we need to explain better how passing the alpha channel in JxlEncoderAddImageFrame works. Is that alpha channel counted in `num_extra_channels`? (so far it wasn't).

Good point! If it wasn't counted so far, perhaps it is best to keep it that way.

PR #827, opened 2021-11-08
Remove CPU topology detection and CPU pinning

### LINKED ISSUE #168: MSYS2 build fails to detect architecture (and number of threads)
```
JPEG XL decoder v0.3.7 [AVX2,SSE4,Scalar]
/C/msys64/home/eustas/clients/libjxl/tools/cpu/cpu.cc:415: JXL_FAILURE: Unable to detect processor topology
Failed to choose default num_threads; you can avoid this error by specifying a --num_threads N argument.
```

### LINKED ISSUE #734: Simplify num_threads detection
Problems with current system:

- works only on x86_64, Mac, Haiku
- asm not portable to other archs, even in theory: CPUID-type instructions are usually not available to userspace
- on unsupported systems, always returns 1
- doesn't handle affinity on Mac or Haiku
- mishandles "mismatched" affinities on x86_64: e.g. set 0.0, 1.0, 1.1, libjxl will pick 4 threads (2 packages * 2 cores per package)
- too complicated especially considering it doesn't actually accomplish the intended calculation (see previous point)

What to replace it with though?

First, we need to decide what exactly the goal is. Is it necessary to handle affinity or filter out SMT siblings? If neither, libjxl can simply use `std::thread::hardware_concurrency`. libstdc++ (except on glibc 2.34 exactly) and libc++ don't currently handle affinity, but I think they would accept patches to do so.

SMT is likely outside of the scope of C++ standard functions though. If it is necessary to exclude SMT siblings, then AFAICT, it is necessary to use OS-specific APIs. I found the library [hwloc](https://www.open-mpi.org/projects/hwloc/) to wrap these APIs, but it is quite heavy and seems to be primarily intended for large applications on "big compute", including many features such as NUMA, PCI, and GL detection, as well as Cairo, SVG, and XML output, among many others. I think it would be overkill to add it as a dependency solely for detecting SMT. The APIs for detecting SMT on Windows and Linux are not especially complex, but they are not trivial.

So, the question is: how much benefit is there from excluding SMT? Based on my benchmarks of encoding and decoding some random 4k-ish image using libjxl ~20211010 at efforts 1..9 and num_threads 0,1,2,ncpus/2,ncpus on my Ryzen 1600 desktop and i5-8250U laptop, it seems that 0,1,2 are much slower (as expected), and ncpus is between 5% slower and 15% faster in real time compared to ncpus/2, but roughly double cpu time. Doing some quick perf stat checks, the rate of cache-misses (L3 I think) on Ryzen and LLC-misses on Intel is quite high, which supports the theory that memory bandwidth is a significant limiting factor. Therefore, the results are mixed: SMT is generally faster, but not by much, and it may be overall more efficient for the user to limit the number of threads to allow other programs to run, since CPU time spent waiting on disk can be yielded to other programs, but CPU time spent with both SMTs waiting on RAM is effectively lost.

In my opinion, the best option is to default to `std::thread::hardware_concurrency`, and document that libjxl may not use SMT effectively on some platforms, so it may be more efficient to reduce the number of threads.

### LINKED ISSUE #824: Use of uninitialized value in tools/cpu/cpu.cc - MSan
**Describe the bug**
When djxl is compiled with MSan it produces a stack trace when provided with any input file:
```
JPEG XL decoder v0.7.0 3dcc100 [SSE4,Scalar]
==12274==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55e1bea80e7d in std::_Rb_tree<unsigned int, unsigned int, std::_Identity<unsigned int>, std::less<unsigned int>, std::allocator<unsigned int> >::_M_erase(std::_Rb_tree_node<unsigned int>*) /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_tree.h:1911:7
    #1 0x55e1bea80df1 in std::_Rb_tree<unsigned int, unsigned int, std::_Identity<unsigned int>, std::less<unsigned int>, std::allocator<unsigned int> >::_M_erase(std::_Rb_tree_node<unsigned int>*) /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_tree.h:1913:4
    #2 0x55e1bea7fc46 in std::_Rb_tree<unsigned int, unsigned int, std::_Identity<unsigned int>, std::less<unsigned int>, std::allocator<unsigned int> >::~_Rb_tree() /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_tree.h:995:9
    #3 0x55e1bea7fc46 in std::set<unsigned int, std::less<unsigned int>, std::allocator<unsigned int> >::~set() /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_set.h:281:22
    #4 0x55e1bea7fc46 in jpegxl::tools::cpu::(anonymous namespace)::X64_Topology::Field::~Field() /home/fuzzing/libjxl/tools/cpu/cpu.cc:352:9
    #5 0x55e1bea7fc46 in jpegxl::tools::cpu::(anonymous namespace)::X64_Topology::Detect(jpegxl::tools::cpu::ProcessorTopology*) /home/fuzzing/libjxl/tools/cpu/cpu.cc:174:3
    #6 0x55e1bea7fc46 in jpegxl::tools::cpu::DetectProcessorTopology(jpegxl::tools::cpu::ProcessorTopology*) /home/fuzzing/libjxl/tools/cpu/cpu.cc:402:7
    #7 0x55e1be0e0db2 in jpegxl::tools::DecompressArgs::ValidateArgs(jpegxl::tools::CommandLineParser const&) /home/fuzzing/libjxl/tools/djxl.cc:159:10
    #8 0x55e1be0d2a2f in jpegxl::tools::(anonymous namespace)::DecompressMain(int, char const**) /home/fuzzing/libjxl/tools/djxl_main.cc:64:13
    #9 0x55e1be0d2a2f in main /home/fuzzing/libjxl/tools/djxl_main.cc:209:10
    #10 0x7f3cebe400b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #11 0x55e1be0567fd in _start (/home/fuzzing/libjxl/build/tools/djxl+0x2b7fd)

SUMMARY: MemorySanitizer: use-of-uninitialized-value /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_tree.h:1911:7 in std::_Rb_tree<unsigned int, unsigned int, std::_Identity<unsigned int>, std::less<unsigned int>, std::allocator<unsigned int> >::_M_erase(std::_Rb_tree_node<unsigned int>*)
Exiting
```

**To Reproduce**

Just run djxl with the following input file (or any other input file) and without additional arguments:
[075bb593caad6f86133d8ad630e4fa6d.png.jxl.txt](https://github.com/libjxl/libjxl/files/7492881/075bb593caad6f86133d8ad630e4fa6d.png.jxl.txt)

**Expected behavior**
djxl should not make use of an uninitialized value and probably add some additional assertions or similar. Looking at past reported issues it is possible that this is just a false alert, however.

**Environment**
 - OS: Ubuntu 20.04
 - Compiler version: clang 9.0
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL decoder v0.7.0 3dcc100 [SSE4,Scalar]

**Additional context**
This bug was found while trying to fuzz libjxl with MSan instrumentation as part of a university project.

deymo:
CPU pinning by djxl/cjxl wasn't compatible with the asymmetric CPUs like
those found in most phones, since we might want to pin to the big cores
or to the small cores, depending on other factors. Newest Intel CPUs are
also asymmetric adding to this problem.

Instead, this patch removes all the CPU topology detection and pinning,
leaving it up to the OS scheduler. Defaulting num_threads to the number
of cores instead of the number of threads (which is higher with HT) has
a mixed performance impact, it may be better or worse depending on the
platform but in any case it is not too significant.

Fixes #824, fixes #734, fixes #168.


veluca93 (approved):

jonsneyers (approved):

fdwr:
Ah, nice, this PR appears to remove the error I get with Release 0.6.1 Oct 29th on Windows 10 - one shouldn't *have* to explicitly specify a number of threads (default to 1 if nothing else). So I'll defer opening an issue and either build myself and wait for release++. Meanwhile the main readme.md is inaccurate because it says to just type "cjxl input.png output.jxl", but you actually need to type "cjxl input.png output.jxl  --num_threads 1".

```
o:\>cjxl.exe screenshots\cs519-shaders-ripples.jpg out.jxl
JPEG XL encoder v0.6.1 a205468 [SSE4,Scalar]
Failed to choose default num_threads; you can avoid this error by specifying a --num_threads N argument.
Use 'cjxl.exe -h' for more information
```

PR #829, opened 2021-11-09
Add several modular options to enc API

lvandeve:


jyrkialakuijala (approved):

lib/jxl/encode.cc
```diff
@@ -433,6 +433,28 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
...
+      // TODO(lode): also add color transform option (xyb, none, ycbcr)
+      if (value < -1 || value > 37) return JXL_ENC_ERROR;
+      options->values.cparams.colorspace = value;
+      return JXL_ENC_SUCCESS;
+    case JXL_ENC_OPTION_MODULAR_GROUP_SIZE:
+      if (value < -1 || value > 3) return JXL_ENC_ERROR;
+      // TODO(lode): the default behavior of this parameter for cjxl is
+      // to choose 1 or 2 depending on the situation. This behavior needs to be
+      // implemented either in the C++ library by allowing to set this to -1, or
+      // kept in cjxl and set it to 1 or 2 using this API.
```
jonsneyers, line 446: My preference: do it in the library.
In general, I would make cjxl as simple/dumb as possible and put all the heuristic decisions in the library, in the case that default -1 options are used.

PR #831, opened 2021-11-09
API proposal: JxlDecoderSetCoalescing

jonsneyers:
Proposed new decode option to disable frame coalescing.

This allows extracting layers from a multi-layer image (in case of Gimp/Photoshop-like use cases), and also possibly to save memory when decoding animations with lots of small cropped frames — at the cost of having to do the blending in the application instead of having the decoder handle it.
Obviously, default behavior doesn't change and still returns coalesced frames as before.

It is partially implemented, but more checks need to be added, and tests and an example application should still be added. Also documentation needs to be improved. This option does significantly change the behavior, e.g. you can no longer set the output buffer immediately after decoding the image header, but you have to do it after decoding the frame header (since without coalescing, frame dimensions are not related to image dimensions). There are probably some places left in the decoder implementation that have an implicit assumption that all frames are canvas-sized.

Only the most basic blend info is currently returned; this is not enough information to do the blending yourself. If and how exactly to expose the more exotic blend info is something that requires some more thought: likely in an application you only want to deal with the common case of normal blending of a simple stack of layers, not weird cases of blending using the non-first alpha channel using a nontrivial source frame etc.

In general, this approach also in principle just doesn't work, since a frame could require patches that are taken from a coalesced frame, which would be a problem because with coalescing disabled, we wouldn't have that source frame. That is currently not a problem in practice because the encoder never does that, but it is in principle a problem (could be solved, but it would be hard to do without just always _also_ doing the coalescing in case there's a patch reference in the future — and that would of course kill any memory benefits).

This feature (if it also gets added to the encode api) would make it possible to effectively use JPEG XL for layered images, e.g. as an option in the Gimp plugin. As far as I know, currently the only image formats that can be used in practice for layered images are TIFF, PSD, XCF and MIFF (ImageMagick's uncompressed internal format). The compression options in any of these formats are quite weak (essentially only uncompressed, RLE, and simple general-purpose compression). JPEG XL would be the first one to have good compression, which could be a killer feature for some use cases, since often near-lossless compression would be good enough but merging the layers is not OK. Compared to e.g. PSD, probably file size savings in the order of 90-95% could be achieved.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -15,7 +15,7 @@ set(JPEGXL_LIBRARY_VERSION
 # It is important to update this value when making incompatible API/ABI changes
 # so that programs that depend on libjxl can update their dependencies. Semantic
 # versioning allows 0.y.z to have incompatible changes in minor versions.
-set(JPEGXL_SO_MINOR_VERSION 6)
+set(JPEGXL_SO_MINOR_VERSION 7)
 if (JPEGXL_MAJOR_VERSION EQUAL 0)
 set(JPEGXL_LIBRARY_SOVERSION
     "${JPEGXL_MAJOR_VERSION}.${JPEGXL_SO_MINOR_VERSION}")
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -289,7 +289,69 @@ typedef struct {
   uint64_t extensions;
 } JxlHeaderExtensions;
 
-/** The header of one displayed frame. */
+/** Frame blend modes.
+ * If coalescing is enabled (default), this can be ignored.
+ */
+typedef enum {
+  JXL_BLEND_REPLACE = 0,
+  JXL_BLEND_ADD = 1,
+  JXL_BLEND_BLEND = 2,
+  JXL_BLEND_MULADD = 3,
+  JXL_BLEND_MUL = 4,
+} JxlBlendMode;
+
+/** The information about blending the color channels or a single extra channel.
+ * If coalescing is enabled (default), this can be ignored.
+ */
+typedef struct {
+  /** Blend mode.
+   * Always equal to JXL_BLEND_REPLACE if coalescing is enabled.
+   */
+  JxlBlendMode blendmode;
+  /** Reference frame ID to use as the 'bottom' layer (0-3).
+   */
+  uint32_t source;
+  /** Which extra channel to use as the 'alpha' channel for blend modes
+   * JXL_BLEND_BLEND and JXL_BLEND_MULADD.
+   */
+  uint32_t alpha;
+  /** Clamp values to [0,1] for the purpose of blending.
+   */
+  JXL_BOOL clamp;
+} JxlBlendInfo;
+
+/** The information about layers.
+ * If coalescing is enabled (default), this can be ignored.
+ */
+typedef struct {
+  /** Horizontal offset of the frame (can be negative, always zero if coalescing
+   * is enabled)
+   */
+  int32_t crop_x0;
+  /** Vertical offset of the frame (can be negative, always zero if coalescing
+   * is enabled)
+   */
+  int32_t crop_y0;
+  /** Width of the frame (number of columns, always equal to image width if
+   * coalescing is enabled)
+   */
+  uint32_t xsize;
+  /** Height of the frame (number of rows, always equal to image height if
+   * coalescing is enabled)
+   */
+  uint32_t ysize;
+  /** The blending info for the color channels. Blending info for extra channels
+   * has to be retrieved separately using JxlDecoderGetExtraChannelBlendInfo.
+   */
+  JxlBlendInfo blend_info;
+  /** After blending, save the frame as reference frame with this ID (0-3).
+   * Special case: if the frame duration is nonzero, ID 0 means "will not be
+   * referenced in the future".
+   */
+  uint32_t save_as_reference;
+} JxlLayerInfo;
+
+/** The header of one displayed frame or non-coalesced layer. */
 typedef struct {
   /** How long to wait after rendering in ticks. The duration in seconds of a
    * tick is given by tps_numerator and tps_denominator in JxlAnimationHeader.
@@ -314,6 +376,10 @@ typedef struct {
   /** Indicates this is the last animation frame.
    */
   JXL_BOOL is_last;
+
+  /** Information about the layer in case of no coalescing.
+   */
+  JxlLayerInfo layer_info;
 } JxlFrameHeader;
 
 #if defined(__cplusplus) || defined(c_plusplus)
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -206,12 +206,17 @@ typedef enum {
    * JxlDecoderGetFrameHeader can be used at this point. A note on frames:
    * a JPEG XL image can have internal frames that are not intended to be
    * displayed (e.g. used for compositing a final frame), but this only returns
-   * displayed frames. A displayed frame either has an animation duration or is
-   * the only or last frame in the image. This event occurs max once per
-   * displayed frame, always later than JXL_DEC_COLOR_ENCODING, and always
-   * earlier than any pixel data. While JPEG XL supports encoding a single frame
-   * as the composition of multiple internal sub-frames also called frames, this
-   * event is not indicated for the internal frames.
+   * displayed frames, unless JxlDecoderSetCoalescing was set to JXL_FALSE: in
+   * that case, the individual layers are returned, without blending. Note that
+   * even when coalescing is disabled, only frames of type kRegularFrame are
+   * returned; frames of type kReferenceOnly and kLfFrame are always for
+   * internal purposes only and cannot be accessed. A displayed frame either has
+   * an animation duration or is the only or last frame in the image. This event
+   * occurs max once per displayed frame, always later than
+   * JXL_DEC_COLOR_ENCODING, and always earlier than any pixel data. While JPEG
+   * XL supports encoding a single frame as the composition of multiple internal
+   * sub-frames also called frames, this event is not indicated for the internal
+   * frames.
    */
   JXL_DEC_FRAME = 0x400,
 
@@ -228,11 +233,12 @@ typedef enum {
    */
   JXL_DEC_DC_IMAGE = 0x800,
 
-  /** Informative event by JxlDecoderProcessInput: full frame decoded.
-   * JxlDecoderSetImageOutBuffer must be used after getting the basic image
-   * information to be able to get the image pixels, if not this return status
-   * only indicates we're past this point in the codestream. This event occurs
-   * max once per frame and always later than JXL_DEC_DC_IMAGE.
+  /** Informative event by JxlDecoderProcessInput: full frame (or layer, in case
+   * coalescing is disabled) is decoded. JxlDecoderSetImageOutBuffer must be
+   * used after getting the basic image information to be able to get the image
+   * pixels, if not this return status only indicates we're past this point in
+   * the codestream. This event occurs max once per frame and always later than
+   * JXL_DEC_DC_IMAGE.
    */
   JXL_DEC_FULL_IMAGE = 0x1000,
 
@@ -417,6 +423,22 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_orientation);
  */
 JXL_EXPORT JxlDecoderStatus
 JxlDecoderSetRenderSpotcolors(JxlDecoder* dec, JXL_BOOL render_spotcolors);
+
+/** Enables or disables coalescing of zero-duration frames. By default, frames
+ * are returned with coalescing enabled, i.e. all frames have the image
+ * dimensions, and are blended if needed. When coalescing is disabled, frames
+ * can have arbitrary dimensions, a non-zero crop offset, and blending is not
+ * performed. For display, coalescing is recommended. For loading a multi-layer
+ * still image as separate layers (as opposed to the merged image), coalescing
+ * has to be disabled.
+ *
+ * @param dec decoder object
+ * @param coalescing JXL_TRUE to enable coalescing (default), JXL_FALSE to
+ * disable it.
+ * @return JXL_DEC_SUCCESS if no error, JXL_DEC_ERROR otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetCoalescing(JxlDecoder* dec,
+                                                    JXL_BOOL coalescing);
 
 /**
  * Decodes JPEG XL file using the available bytes. Requires input has been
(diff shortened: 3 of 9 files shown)


jonsneyers:
Added a test consisting of an animation where some frames have two layers, frames have various dimensions and offsets, and decode is done with and without coalescing, with and without frame skipping.

Frame skipping is also useful in the still image case, so you can efficiently extract specific layers instead of having to decode the whole image. Currently this works but has the performance bug that it will still decode the whole thing because these layers depend on one another (in the coalescing case), though they actually don't necessarily depend on one another in the no-coalescing case.

lib/include/jxl/codestream_header.h
```diff
@@ -314,6 +314,32 @@ typedef struct {
...
+  int32_t crop_y0;
+  /** Width of the frame (number of columns)
+   */
+  uint32_t xsize;
+  /** Height of the frame (number of rows)
+   */
+  uint32_t ysize;
+  /** Blend modes: 0=kReplace, 1=kAdd, 2=kBlend, 3=kMulAdd, 4=kMul
+   */
+  uint32_t blendmode;
```
lvandeve, line 339: this one would make sense as an enum

lib/include/jxl/codestream_header.h
```diff
@@ -314,6 +314,32 @@ typedef struct {
   /** Indicates this is the last animation frame.
    */
   JXL_BOOL is_last;
+
+  /** If coalescing is enabled, the following can be ignored; crop_x0/crop_y0 is
```
lvandeve, line 318: How does this comment render in the doxygen? I'm not sure how it handles this loose comment. Maybe a short comment to each individual field affected by this can be added, like "this field is only used if coalescing is disabled"

lib/jxl/decode.cc
```diff
@@ -2413,15 +2433,22 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderImageOutBufferSize(
   if (format->num_channels < 3 && !dec->metadata.m.color_encoding.IsGray()) {
     return JXL_API_ERROR("Grayscale output not possible for color image");
   }
-
-  size_t row_size =
-      jxl::DivCeil(dec->metadata.oriented_xsize(dec->keep_orientation) *
-                       format->num_channels * bits,
-                   jxl::kBitsPerByte);
+  size_t oriented_xsize = dec->metadata.oriented_xsize(dec->keep_orientation);
```
lvandeve, line 2436: The comment in the header of this function could be updated to say that in case of no coalescing, it can only be called after the frame event (as opposed to after basic info as it says now), and will return the size required for the possibly cropped frame in that case, rather than the full image

lib/include/jxl/codestream_header.h
jonsneyers: Agreed, or a struct with an enum in it, depending on how much we want to expose. If the goal is to expose enough info to let the application do all the blending itself, then we also need to expose source, clamp, alpha channel, and blendmodes for extra channels - but I'm not sure if it makes sense to expose all that; it might be better to make no-coalescing only work for "simple" layers (the kind of layers you could get with GIF/APNG/XCF/PSD as input), and make it return a decode error if too exotic stuff is happening.

lib/jxl/decode.cc
```diff
@@ -2549,7 +2576,23 @@ JxlDecoderStatus JxlDecoderGetFrameHeader(const JxlDecoder* dec,
   }
   header->name_length = dec->frame_header->name.size();
   header->is_last = dec->frame_header->is_last;
-
+  if (!dec->coalescing && dec->frame_header->custom_size_or_origin) {
+    header->crop_x0 = dec->frame_header->frame_origin.x0;
```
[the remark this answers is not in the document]
jonsneyers: Right!

These combinations of having layers and nontrivial orientation are not likely to occur in the wild, since at the moment jxl is the only image format that supports both layers and orientation :) - but of course we do have to make it do the right thing. How orientation affects crop positions makes me feel dizzy btw.

lvandeve:
> you can no longer set the output buffer immediately after decoding the image header, but you have to do it after decoding the frame header

To me that looks fine, I like having one right way to do things, and always setting the out buffer after JXL_DEC_JPEG_NEED_MORE_OUTPUT or JXL_DEC_FRAME rather than being able to do it before a frame is reasonable

jonsneyers:
> > you can no longer set the output buffer immediately after decoding the image header, but you have to do it after decoding the frame header
> 
> To me that looks fine, I like having one right way to do things, and always setting the out buffer after JXL_DEC_JPEG_NEED_MORE_OUTPUT or JXL_DEC_FRAME rather than being able to do it before a frame is reasonable

As it is now in this PR, setting the buffer earlier is still possible, just not when coalescing is disabled. So no existing applications will break.

lib/include/jxl/decode.h
```diff
@@ -418,6 +424,22 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_orientation);
...
+ * performed. For display, coalescing is recommended. For loading a multi-layer
+ * still image as separate layers (as opposed to the merged image), coalescing
+ * has to be disabled.
+ *
+ * @param dec decoder object
+ * @param coalescing JXL_TRUE to enable coalescing (default), JXL_FALSE to
+ * disable it.
+ * @return JXL_DEC_SUCCESS if no error, JXL_DEC_ERROR otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetCoalescing(JxlDecoder* dec,
```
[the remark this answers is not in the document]
jonsneyers: It is an interesting point that it might be desirable that some of the extra channels are semantically "internal only" in a hypothetical encoder (which does not yet exist, to be clear). However, that is a problem that is completely orthogonal to coalescing or not: also in the current (always-coalescing) API, there is at the moment no way to differentiate between "internal" extra channels and "to be exposed" extra channels.

A potential solution for that problem could be to reserve a specific channel name to mean "internal only", and then in the normal (coalescing) case such channels would be hidden by the api. In the non-coalescing case, if we want to allow applications to do their own blending in all cases, then such "internal only" extra channels would have to be exposed.

I think for applications it is hard and error-prone to actually do their own blending in all cases, since the way frame blending works in jxl is very flexible (the source frame can be chosen, the alpha channel can be chosen, all extra channels can have their own blend mode, etc). I think it might be better to only support the basic Gimp/Photoshop use case of a linear chain of layers in the non-coalescing decode api, and return a decode error if more exotic stuff is happening (i.e. something created by a hypothetical future encoder), effectively forcing the application to try again with coalescing enabled (letting the decoder figure out the exotic stuff).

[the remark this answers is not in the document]
jonsneyers: For the "internal only" extra channels, probably it would be good to define a new ExtraChannelType for it (we have values 7-14 still available and reserved for future ec types, with values 15 and 16 (`kNonOptional` and `kOptional`) as generic catch-all values for future ec types - where the idea is that then the ec name will have to indicate the real type). We could add a channel type `kInternalOnly`, which like the frame type `kReferenceOnly` indicates that it is something that shouldn't be exposed in the API since it is only an auxiliary thing used for better compression, not a semantic element of the actual image/animation.

I would then argue that we should specify that `kInternalOnly` channels can only be used for blending in Patches, not for frame blending. That way, we can ensure that `kRegularFrame` frames can always be exposed as semantic elements of a multi-layer image/animation, while the API can still hide `kInternalOnly` channels and `kReferenceOnly` frames.

This is all mostly orthogonal to the question of SetCoalescing though.

Also: yes, adding new features has a cost and can introduce bugs. I do think this is an important feature though, and I think it's better to introduce it sooner than later, because otherwise we might end up making design choices that will make it nearly impossible to still add this feature without breaking existing applications.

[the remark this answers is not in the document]
deymo: This problem is not orthogonal. The plan was to not expose these "internal" extra channels. The fact that you now need to use these other extra channels for blending is part of setting this value to false, not an orthogonal problem, so this should be addressed here. This new feature is already introducing a bug that is not being considered here, it is not a hypothetical. This PR needs to explain how these "internal" extra channels are handled with the default (true) and how that changes with the false value.

For example, an alternative implementation of this would be to add another function to get "internal" data. Right now you get the FrameHeader, but a different implementation of this feature (which can be done in the future without API design changes) could let you get the internal frame header info (number of internal extra channels used for blending, blending modes, etc).

lib/include/jxl/codestream_header.h
deymo: You can look at how this looks like in the readthedocs "check" in the PR.

lib/include/jxl/codestream_header.h
```diff
@@ -314,6 +314,32 @@ typedef struct {
...
+     always zero, xsize/ysize is always equal to the image dimensions, and
+     blendmode is always 0 (kReplace). If coalescing is disabled, the layer has
+     dimensions xsize x ysize (which can be smaller or larger than the image
+     dimensions) and its top left corner is at position (crop_x0,crop_y0) -
+     these offsets can be negative.
+   */
+  /** Horizontal offset of the frame (can be negative)
+   */
+  int32_t crop_x0;
```
deymo, line 327: This is ABI incompatible in the decoder so that would need to be noted and handled too (SOVERSION needs to be bumped).

crop_x0, crop_x1, xsize and ysize are useful even in the coalescing case with animation (GIF use-cases would be the obvious example). It would make sense to coordinate this change with that feature too, which is a feature required for djxl to transcode a gif (without blending all the frames to full-size frames).

Have you consider having these in a separate struct which you only get using these "advanced" modes?

lib/include/jxl/codestream_header.h
deymo: I don't see why this would need to be a partial implementation of the feature. What do you do if coalescence is false but the .jxl file contains non-"simple" layers? fail when the coalescing version would have succeeded?? That's very odd behavior.

lib/jxl/decode.cc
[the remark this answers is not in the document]
jonsneyers: OK, fixed the crop offsets in case of orientation, so in the case of !keep_orientation and non-coalescing decoding, the layers are returned with sizes and crop offsets normalized so the application can just assume that things are always identity-oriented. I added a test for this (OrientedCroppedFrameTest) to verify that in all 8 orientations, manually blending non-coalesced frames (here kept easy by using only kReplace) based on the returned offset and dimensions results in the same image as letting the default coalescing decoder do it.

lib/include/jxl/codestream_header.h
jonsneyers: For GIF/APNG transcoding, you would use non-coalescing decoding, because otherwise you'd only get full-sized frames (and even if you would still get the crop offset and 'real' frame size, the pixel data would be after blending which is not what you want).

The crop coords/offset could still be useful in case of coalesced animation if you want to use the libjxl blending but still only want to refresh part of the screen. For that to work though, in principle what you need is not the cropbox of the last nonzero-duration frame, but a bounding box around all frames since the last nonzero-duration frame.

Separate struct could make sense, it might help for conceptual clarity but also if we want to do a lightweight libjxl_dec that doesn't include stuff like non-coalescing that isn't needed for viewing images/animations, it might make it easier to put these things behind a compile flag. Would you make this struct part of the JxlFrameHeader struct, or make it an entirely separate struct that can be accessed with a new function?

deymo: For GIF/APNG animations you still need to get the coalesced frames, not the non-coalesced. Each frame can have layers, but that's not part of what you want to display.

Indeed cropped x0/y0 here would mean the crop in the coalesced API frames, not the internal zero-duration ones; that's the pixel data we are returning. The frame would not be "after blending", it would the frame that was meant to be as a second animation frame for example, even if it has alpha. That second frame could have been created with multiple layers, but that can still be all internal. Getting cropped frames when the output outside a given rectangle didn't change has applications in decoder applications (only refresh the portion of the animation that changed), not just in gimp/photoshop type of use-cases like the non-coalescing feature.

We want to support the coalesced use case for sure. Mixing "layers" and frames in the API is making all this more complex, which is why I don't think this is a good time to add the feature.

lib/include/jxl/codestream_header.h
jonsneyers: Yes, we could expose all relevant FrameHeader fields (`blending_info`, `ec_blending_info`, and `save_as_reference`), and then in theory an application could manually keep track of the 3 reference slots (the fourth one can only be used by Patches so wouldn't be exposed) and do all the blending. Maybe that's a better approach.

I was just assuming that in most applications, the only thing they will care about is RGBA with kReplace or kBlend blending (as that is what is needed for transcoding GIF/APNG/XCF/PSD, and what is already implemented in image libraries like ImageMagick). So there's a risk that they will only implement that. From a practical point of view, it doesn't make a big difference if the application fails because libjxl returns a not_supported error, or if it fails because libjxl returns the actual exotic blendinfo and then the application gives up.

The advantage of returning the full blendinfo is that it leaves open the possibility for applications to actually implement exotic blending itself, as well as leaving open the possibility to extract layers even if the application does not know how to blend them (which could be useful e.g. for a jxltran application that can convert a lossless jxl to a lossy jxl without making all frames canvas-sized in the process).

The disadvantage of returning the full blendinfo is that it's quite a bit of extra stuff to add to the API (not a big issue if we can put it behind a compile flag), and that there's a risk that applications will not actually inspect all the details of how the blending should be done, make assumptions instead, and produce incorrect results (which would of course be an application bug, not a libjxl bug, but still...).

I don't have a strong opinion either way; I'm leaning towards returning the full blendinfo but I would also be fine returning a "not supported" error when there is exotic blendinfo and exposing only a simplified subset of the info that is easier for applications to deal with (and then of course document well that setting coalescing to false can have that consequence).

jonsneyers: Updated this part. It is now a full implementation (exposes all blending info).

lib/include/jxl/codestream_header.h
jonsneyers: This part was modified and now no longer has a loose comment.

lib/include/jxl/codestream_header.h
jonsneyers: Bumped the SOVERSION and moved the info to a separate struct. For the color channels it is always returned as part of the frame header, for the extra channels the blend info for those has to be fetched separately with a new function (to keep things fixed-size).

lib/jxl/decode.cc
jonsneyers: Added a comment.

jonsneyers:
I _think_ this now fully implements all that is needed for a non-coalescing decode API. PTAL.

If this one can land, I can get started on the encode side of this. The encode side will be needed if we want to migrate cjxl to the encode api for the GIF, APNG and PSD cases (where we will be need to have a way to queue cropped frames).
(Also for the gimp plugin it would be nice to have a "keep layers" option)

jyrkialakuijala (dismissed):

jyrkialakuijala (approved):

lib/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ set(JPEGXL_LIBRARY_VERSION
 # It is important to update this value when making incompatible API/ABI changes
 # so that programs that depend on libjxl can update their dependencies. Semantic
 # versioning allows 0.y.z to have incompatible changes in minor versions.
-set(JPEGXL_SO_MINOR_VERSION 6)
+set(JPEGXL_SO_MINOR_VERSION 7)
```
malaterre, line 18: @jonsneyers I've inspected the change at the library level. All I can find is _new_ symbols being added to the library. In this case there is no need to bump the SONAME of the library. Could you confirm this is the case ?
Here is the documentation, I've used:
* https://docs.oracle.com/cd/E19253-01/817-1984/appendixb-45356/index.html#6mhm7plcq
Here is what I see on my side (git/master as of this morning):
```
 % readelf -d ./obj-x86_64-linux-gnu/libjxl.so | grep soname
 0x000000000000000e (SONAME)             Library soname: [libjxl.so.0.7]
```

jonsneyers: The non-opaque JxlFrameHeader struct changes so I think that requires a SONAME bump. Please correct me if I'm wrong though.

malaterre: Oh right, totally missed that change! Thanks for the quick answer.

PR #836, opened 2021-11-10
Benchmark code for fixed-point IDCT on ARM.

veluca93:


erikandre:
Wow, this is a great perf improvement @veluca93 !

I integrated the benchmark in one of our Android testing apps, and with a arm64 release build of it I get the following numbers:

**Pixel 4** 
BenchmarkFloatIDCT32x32: 92 ms
BenchmarkFastIDCT32x32: 41 ms

**Samsung Galaxy S10**

BenchmarkFloatIDCT32x32: 174 ms
BenchmarkFastIDCT32x32: 53 ms

The more than x3 speedup on the S10 is a good indication that there's a lot of devices in the wild which will benefit significantly from this optimization. As for the next step, we should gather some real world numbers from one of our main apps.

I'm still waiting for binary size numbers, but will post them in the next few hours once available.

erikandre:
Got some preliminary Android binary size measurements:
```
Change in Compressed Size: +4.0 KiB 
Change in Uncompressed Size: +10.6 KiB
```

Size wise, I think this looks quite good. These numbers were not from a proper release build due to some infra issues on my end but from a debug build. I suspect that means that for a full release build, the compressed size may be slightly smaller.

sboukortt (approved):

PR #837, opened 2021-11-10
Add color transform and already_downsampled enc options

lvandeve:
Also added test for already_downsampled. Also added TODO for issue found
with already_downsampled on odd output sizes (something that you can do
through the API and should be possible, but cjxl can't do)

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -110,127 +110,146 @@ typedef enum {
    */
   JXL_ENC_OPTION_EXTRA_CHANNEL_RESAMPLING = 3,
 
+  /** Indicates the frame added with @ref JxlEncoderAddImageFrame is already
+   * downsampled by the downsampling factor set with @ref
+   * JXL_ENC_OPTION_RESAMPLING. The input frame must then be given in the
+   * downsampled resolution, not the full image resolution. The downsampled
+   * resolution is given by ceil(xsize / resampling), ceil(ysize / resampling)
+   * with xsize and ysize the dimensions given in the basic info, and resampling
+   * the factor set with @ref JXL_ENC_OPTION_RESAMPLING.
+   * Use 0 to disable, 1 to enable. Default value is 0.
+   */
+  JXL_ENC_OPTION_ALREADY_DOWNSAMPLED = 4,
+
   /** Adds noise to the image emulating photographic film noise, the higher the
    * given number, the grainier the image will be. As an example, a value of 100
    * gives low noise whereas a value of 3200 gives a lot of noise. The default
    * value is 0.
    */
-  JXL_ENC_OPTION_PHOTON_NOISE = 4,
+  JXL_ENC_OPTION_PHOTON_NOISE = 5,
 
   /** Enables adaptive noise generation. This setting is not recommended for
    * use, please use JXL_ENC_OPTION_PHOTON_NOISE instead. Use -1 for the default
    * (encoder chooses), 0 to disable, 1 to enable.
    */
-  JXL_ENC_OPTION_NOISE = 5,
+  JXL_ENC_OPTION_NOISE = 6,
 
   /** Enables or disables dots generation. Use -1 for the default (encoder
    * chooses), 0 to disable, 1 to enable.
    */
-  JXL_ENC_OPTION_DOTS = 6,
+  JXL_ENC_OPTION_DOTS = 7,
 
   /** Enables or disables patches generation. Use -1 for the default (encoder
    * chooses), 0 to disable, 1 to enable.
    */
-  JXL_ENC_OPTION_PATCHES = 7,
+  JXL_ENC_OPTION_PATCHES = 8,
 
   /** Edge preserving filter level, -1 to 3. Use -1 for the default (encoder
    * chooses), 0 to 3 to set a strength.
    */
-  JXL_ENC_OPTION_EPF = 8,
+  JXL_ENC_OPTION_EPF = 9,
 
   /** Enables or disables the gaborish filter. Use -1 for the default (encoder
    * chooses), 0 to disable, 1 to enable.
    */
-  JXL_ENC_OPTION_GABORISH = 9,
+  JXL_ENC_OPTION_GABORISH = 10,
 
   /** Enables modular encoding. Use -1 for default (encoder
    * chooses), 0 to enforce VarDCT mode (e.g. for photographic images), 1 to
    * enforce modular mode (e.g. for lossless images).
    */
-  JXL_ENC_OPTION_MODULAR = 10,
+  JXL_ENC_OPTION_MODULAR = 11,
 
   /** Enables or disables preserving color of invisible pixels. Use -1 for the
    * default (1 if lossless, 0 if lossy), 0 to disable, 1 to enable.
    */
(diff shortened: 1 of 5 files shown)


lib/include/jxl/encode.h
```diff
@@ -110,127 +110,143 @@ typedef enum {
...
 
   /** Enables or disables delta palette. Use -1 for the default (encoder
    * chooses), 0 to disable, 1 to enable. Used in modular mode.
    */
-  JXL_ENC_OPTION_LOSSY_PALETTE = 22,
+  JXL_ENC_OPTION_LOSSY_PALETTE = 23,
+
+  /** Color transform for internal encoding: -1 = default, 0=XYB, 1=none (RGB),
+   * 2=YCbCr.
+   */
```
jonsneyers, line 232: Note: the way this is currently implemented in the encoder, for XYB it will do the fwd transform, for None it will do nothing, and for YCbCr it will (surprisingly) also do nothing (except signal in the frame header that the data is YCbCr). The idea is that you only use YCbCr if your input data happens to be YCbCr already, and you want to preserve it losslessly (e.g. to do y4m recompression, or to recompress video stills as losslessly as possible). This should be documented well if we expose this option, and there should also be a way to pass 422/420 data because ycbcr source material is often not 444.

lib/include/jxl/encode.h
```diff
@@ -110,127 +110,143 @@ typedef enum {
...
+  JXL_ENC_OPTION_COLOR_TRANSFORM = 24,
 
   /** Color space for modular encoding: 0=RGB, 1=YCoCg, 2-37=RCT, -1=default:
    * try several, depending on speed.
    */
-  JXL_ENC_OPTION_MODULAR_COLOR_SPACE = 23,
+  JXL_ENC_OPTION_MODULAR_COLOR_SPACE = 25,
```
jonsneyers, line 238: This is how it is in cjxl, but those values are a bit of a historically evolved mess. Basically 0 and 1 are special values and 2+n gets mapped to spec-RCT n. This is redundant because 0 is the same as n=0 and 1 is the same as n=6.

Before we put this in the API, maybe it is best to sanitize this and get rid of the special values 0 and 1. For cjxl we can see if we want to keep the old behavior or also sanitize it (it's an option you should rarely set except maybe to force it to 0, but for 0 the new saner behavior would be identical to the old behavior).

lib/jxl/encode.cc
```diff
@@ -323,6 +323,9 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
       options->values.cparams.resampling = value;
       return JXL_ENC_SUCCESS;
     case JXL_ENC_OPTION_EXTRA_CHANNEL_RESAMPLING:
+      // TOOD(lode): the jxl codestream allows choosing a different resampling
+      // factor for each extra channel, move this to a per-extra-channel setting
```
jonsneyers, line 327: It's per frame and per extra channel, so cannot be moved to extrachannelinfo (which is global for the whole image).
We should also check that the total upsampling factor of extra channels (ec_resampling << dim_shift) is at least as large as the upsampling factor of the color image, since that is a spec requirement and the encoder and decoder will choke if that constraint is not satisfied.

lib/include/jxl/encode.h
lvandeve: I updated the comment, please take another look.

Please check if the updated comment is valid in all cases (both vardct and modular), or let me know if parts of this applies to modular, or vardct, only.

lib/include/jxl/encode.h
lvandeve: I've updated the meaning of the API to match the spec. So the cjxl 0 and 1 can now be reached using 0 and 6. It's best if the API follows the specification, so with this updated comment and implementation it should match. It converts it to the internal C++ codec representation internally now.

lib/include/jxl/encode.h
lvandeve: 422/420 support can be done in a different MR than this one. The setting can already be used for 444 now.

lib/jxl/encode.cc
lvandeve: Updated the TODO to make clear it's per frame and per extra channel and how to implement it

lib/include/jxl/encode.h
jonsneyers: LGTM. Agreed that 422/420 doesn't have to happen in this PR.

lib/jxl/encode.cc
```diff
@@ -433,10 +444,18 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
       // alternatively, in the cjxl binary like now)
       options->values.cparams.lossy_palette = (value == 1);
       return JXL_ENC_SUCCESS;
+    case JXL_ENC_OPTION_COLOR_TRANSFORM:
+      if (value < -1 || value > 2) return JXL_ENC_ERROR;
+      if (value == -1) {
+        options->values.cparams.color_transform = jxl::ColorTransform::kXYB;
```
jonsneyers, line 450: As discussed elsewhere, default setting probably shouldn't happen here in the first place (but only when actual encoding starts). In this case, for example, kXYB is a good default for lossy while for lossless kNone has to be used and it would not make sense to use kXYB by default.

lib/jxl/encode_test.cc
```diff
@@ -367,7 +367,10 @@ TEST(EncodeTest, OptionsTest) {
               JxlEncoderOptionsSetInteger(
                   options, JXL_ENC_OPTION_MODULAR_PREDICTOR, 14));
     VerifyFrameEncoding(enc.get(), options);
-    EXPECT_EQ(30, enc->last_used_cparams.colorspace);
+    // It was set to 30, but becomes 32 because in the C++ implementation, the
+    // numerical RCT values are shifted 2 compared to the specification. The
+    // API uses the values seen in the JXL specification.
+    EXPECT_EQ(32, enc->last_used_cparams.colorspace);
```
jonsneyers, line 373: It might make sense to also make the C++ implementation just follow the spec. The weird numbering is a historical thing from a time when YCoCg was still a separate modular transform and RCT didn't exist yet.

lvandeve: I'd do that in a separate MR, the goal here is to add all cjxl options as options to the API so that cjxl can use it. The internal C++ implementation is separate and mixing changing behavior with the API makes it more difficult to revert MR's and find out which one causes an issue.

lib/jxl/encode.cc
lvandeve: In CompressParams, color_transform is initialized as `ColorTransform color_transform = ColorTransform::kXYB`, so if the implementation here sets it to kXYB, it does exactly the same as before. It's not possible that this MR can introduce the problem of wrong default color transform value.

So this MR does not change any behavior. If we need a refactoring for default values, let's do it in a different MR. What do you think?

jonsneyers: Yes, let's do it in a follow-up MR. I think it's best to set things to -1 by default in the api options, and to initialize cparams just before actual compression starts, so defaults can be made sensible (and we can e.g. avoid the problem that applications think they're doing lossless but it's actually XYB).

jonsneyers (approved):

jonsneyers (approved):

PR #839, opened 2021-11-10
lib/jxl/encode.cc: add XYB fallback check to AddImageFrame

thebombzen:
The encoder will no longer refuse AddImageFrame if the
color encoding is set to XYB via JxlBasicInfo, even if it
is not set directly.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -632,7 +632,7 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
   }
 
   jxl::ColorEncoding c_current;
-  if (options->enc->metadata.m.xyb_encoded) {
+  if (!options->enc->color_encoding_set) {
     if ((pixel_format->data_type == JXL_TYPE_FLOAT) ||
         (pixel_format->data_type == JXL_TYPE_FLOAT16)) {
       c_current =
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -607,7 +607,12 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(const JxlEncoderOptions* options,
 JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
                                          const JxlPixelFormat* pixel_format,
                                          const void* buffer, size_t size) {
-  if (!options->enc->basic_info_set || !options->enc->color_encoding_set) {
+  if (!options->enc->basic_info_set ||
+      (!options->enc->color_encoding_set &&
+       !options->enc->metadata.m.xyb_encoded)) {
+    // Basic Info must be set, and color encoding must be set directly,
+    // or set to XYB via JxlBasicInfo.uses_original_profile = JXL_FALSE
+    // Otherwise, this is an API misuse.
     return JXL_ENC_ERROR;
   }
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,4 @@ Samuel Leong <wvvwvvvvwvvw@gmail.com>
 Alex Xu (Hello71) <alex_y_xu@yahoo.ca>
 Vincent Torri <vincent.torri@gmail.com>
 Artem Selishchev
+Leo Izen <leo.izen@gmail.com>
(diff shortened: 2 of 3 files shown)


thebombzen:
@googlebot I signed it!

mo271:
please add yourself to the AUTHORS file

thebombzen:
@mo271 done, although it wasn't sorted so I added my name at the bottom

lib/jxl/encode.cc
```diff
@@ -607,7 +607,11 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(const JxlEncoderOptions* options,
 JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
                                          const JxlPixelFormat* pixel_format,
                                          const void* buffer, size_t size) {
-  if (!options->enc->basic_info_set || !options->enc->color_encoding_set) {
+  if (!options->enc->basic_info_set ||
+    !options->enc->color_encoding_set && !options->enc->metadata.m.xyb_encoded) {
```
mo271, line 611: adding parentheses here for clarify will fix the "'&&' within '||'" we see in some tests.

mo271 (changes requested):
Other than the comment about clarifying the if condition it looks good to me!

mo271:
> @mo271 done, although it wasn't sorted so I added my name at the bottom

Thanks, I can fix the order later...

lib/jxl/encode.cc
thebombzen: I thought it was decently clear since I had a newline, but I'll add parentheses anyway just to make it extra clear.

thebombzen:
@mo271 pushed the requested change

mo271:
Thanks for fixing the lint! There is still 1 failing test, namely `AddFrameBeforeColorEncodingTest` in `encode_test.cc`. I think it would be best to fix this by changing the line 82 there to
`basic_info.uses_original_profile = false;` 
to 
`basic_info.uses_original_profile = true;`.
Could you also provide some more information on why the change you propose is necessary? Thanks!

thebombzen:
> Could you also provide some more information on why the change you propose is necessary? Thanks!

The docs, and the behavior both say that `basic_info.uses_original_profile = false` causes libjxl to ignore the color information passed to it. If you don't call `JxlEncoderSetColorEncoding` before `JxlEncoderAddImageFrame` then `JxlEncoderAddImageFrame` will fail at this point. If you *do* call `JxlEncoderSetColorEncoding` then it won't fail here, but it will completely ignore the information that you pass to it with `JxlEncoderSetColorEncoding`. It shouldn't require you to call a function to specify information only to ignore that information.

thebombzen:
Anyway, I pushed a change, let's hope the encode test succeeds now.

mo271:
> > Could you also provide some more information on why the change you propose is necessary? Thanks!
> 
> The docs, and the current behavior, both say that `basic_info.uses_original_profile = false` causes libjxl to ignore the color information passed to it. If you don't call `JxlEncoderSetColorEncoding` before `JxlEncoderAddImageFrame` then `JxlEncoderAddImageFrame` will fail at this point. If you _do_ call `JxlEncoderSetColorEncoding` then it won't fail here, but it will completely ignore the information that you pass to it with `JxlEncoderSetColorEncoding`. It shouldn't require you to call a function to specify information only to ignore that information.

That makes sense, thanks for adding the explanation!

jonsneyers:
Wait, also in the XYB case _something_ gets signalled in the jxl image header to be the nominal colorspace of the image. I assume that's what JxlSetColorEncoding is used for in that case, no?

Also you are passing the RGB data in _some_ colorspace, and I would assume it is the same one as the one set with JxlSetColorEncoding.

I think the documentation needs to be clarified and we also need to make sure that the encode api actually does something that makes sense here...

thebombzen:
The encode documentation says that if you set `basic_info.uses_original_profile = false`, it instead uses SRGB for integer input and Lienar SRGB for float input. This is actually incorrect, what happens is it sets the color transform to XYB, not to SRGB (i.e. no color transform).

The documentation does correctly state though that it ignores the information passed in `JxlEncoderSetColorEncoding`. Personally, I think what it should do, if you set uses_original_profile to false is what it *says* it does, but I don't know what's intended.

thebombzen:
Basically the problem with the current behavior is as follows:

If you set `basic_info.uses_original_profile = false`, the docs say that it assumes Linear SRGB input for float samples and SRGB input for integer samples.

Then, if you call `JxlEncoderImageAddFrame`, it will still fail, telling you that you did not set the color encoding. If you *do* set the color encoding, it does not fail, but ignores what you set, which means the current boilerplate is just an extra call to `SetColorEncoding` (that does nothing) every time you set `uses_original_profile = false` with no benefit. This PR simply removes that requirement of this extra pointless call.

mo271 (approved):

mo271:
Thanks for the fix!

PR #844, opened 2021-11-11
Update Highway to 0.15.0, add version requirement

jan-wassenberg:


third_party/CMakeLists.txt
```diff
@@ -88,7 +88,7 @@ if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/highway/CMakeLists.txt" AND
   configure_file("${CMAKE_CURRENT_SOURCE_DIR}/highway/LICENSE"
                  ${PROJECT_BINARY_DIR}/LICENSE.highway COPYONLY)
 else()
-  pkg_check_modules(HWY libhwy)
+  pkg_check_modules(HWY libhwy>=0.15.0)
```
deymo, line 91: ahh, this doesn't work on ubuntu 18.04 because the cmake there is too old. let's remove this line for now.

deymo (approved):

third_party/CMakeLists.txt
jan-wassenberg: OK, turned it into a comment.

deymo (approved):

PR #838, opened 2021-11-10
encoder api: functions to handle extra channels

mo271:
Filling in the functions:

 - `JxlEncoderInitExtraChannelInfo`
 - `JxlEncoderSetExtraChannelInfo`
 - `JxlEncoderSetExtraChannelName`
 
Pair with Lode

PR #845, opened 2021-11-11
encoder api: extra channels

mo271:
This adds `JxlEncoderSetExtraChannelBuffer` and
modifies the a roundtrip test to test it.

Pair with Lode

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -742,7 +742,9 @@ JXL_EXPORT void JxlEncoderInitBasicInfo(JxlBasicInfo* info);
  *
  * If the JxlBasicInfo contains information of extra channels beyond an alpha
  * channel, then @ref JxlEncoderSetExtraChannelInfo must be called between
- * JxlEncoderSetBasicInfo and @ref JxlEncoderAddImageFrame.
+ * JxlEncoderSetBasicInfo and @ref JxlEncoderAddImageFrame. In order to indicate
+ * extra channels, the value of `info.num_extra_channels` should be set to the
+ * number of extra channels, also counting the alpha channel if present.
  *
  * @param enc encoder object.
  * @param info global image metadata. Object owned by the caller and its
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -8,8 +8,10 @@
 #include <brotli/encode.h>
 
 #include <algorithm>
+#include <cstddef>
 #include <cstring>
 
+#include "jxl/codestream_header.h"
 #include "lib/jxl/aux_out.h"
 #include "lib/jxl/base/span.h"
 #include "lib/jxl/codec_in_out.h"
@@ -441,6 +443,7 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
   if (info->alpha_bits > 0 && info->alpha_exponent_bits > 0) {
     return JXL_ENC_NOT_SUPPORTED;
   }
+
   switch (info->alpha_bits) {
     case 0:
       break;
@@ -454,6 +457,10 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
     default:
       return JXL_ENC_ERROR;
   }
+  // The number of extra channels includes the alpha channel, so for example and
+  // RGBA with no other extra channels, has exactly num_extra_channels == 1
+  enc->metadata.m.num_extra_channels = info->num_extra_channels;
+  enc->metadata.m.extra_channel_info.resize(enc->metadata.m.num_extra_channels);
   enc->metadata.m.xyb_encoded = !info->uses_original_profile;
   if (info->orientation > 0 && info->orientation <= 8) {
     enc->metadata.m.orientation = info->orientation;
@@ -992,6 +999,7 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
   }
   queued_frame->frame.SetFromImage(std::move(*io.Main().color()),
                                    io.Main().c_current());
+  // TODO(firsching) add extra channels here
   queued_frame->frame.jpeg_data = std::move(io.Main().jpeg_data);
   queued_frame->frame.color_transform = io.Main().color_transform;
   queued_frame->frame.chroma_subsampling = io.Main().chroma_subsampling;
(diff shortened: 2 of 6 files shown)


mo271:
This is WIP and does not work as indented yet.

lib/jxl/encode.cc
```diff
@@ -220,6 +222,15 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
     default:
       return JXL_ENC_ERROR;
   }
+  // Adjusting the number of extra channels; SetAlphaBits sets this to 1
+  // if alpha is present, so info->num_extra_channels should be provided
+  // as at least one in that case.
+  // if (enc->metadata.m.num_extra_channels > info->num_extra_channels){
+  //  return JXL_API_ERROR("Invalid value for number of extra channel");
+  //}
+  enc->metadata.m.num_extra_channels += info->num_extra_channels;
```
jonsneyers, line 231: I would consider alpha to be an extra channel like any other, even if it has the special property that there are special functions for it and it can be passed to the encoder / received from the decoder together with the color image using RGBA buffers. In the spec (and in the C++ code), an alpha channel is treated like any other extra channel. So let's try to avoid doing +1 or -1 on num_extra_channels in the API.

lib/include/jxl/encode.h
```diff
@@ -573,7 +573,13 @@ JXL_EXPORT void JxlEncoderInitBasicInfo(JxlBasicInfo* info);
...
  * If the JxlBasicInfo contains information of extra channels beyond an alpha
  * channel, then @ref JxlEncoderSetExtraChannelInfo must be called between
- * JxlEncoderSetBasicInfo and @ref JxlEncoderAddImageFrame.
+ * JxlEncoderSetBasicInfo and @ref JxlEncoderAddImageFrame. In order to indicate
+ * extra channels, the value of `info.num_extra_channels` should be set to the
+ * number of extra channels, also counting the alpha channel if present. If an
+ * alpha channel is indicated by `info.alpha_bits` being non-zero and no other
+ * extra channel is present, then `info.num_extra_channels` should be set to
+ * `1`. However in this case we also allow a value of `0` for
+ * `info.num_extra_channels`.
```
jonsneyers, line 582: It feels a bit wrong to have this behavior where num_extra_channels is allowed to be incorrect in some specific (but common) case.

I think backwards compatibility shouldn't be a huge concern yet for the encode api (we already broke a few things anyway), so I prefer breaking things now and having a clean API (even if it breaks some of the early adopters) rather than already having a wonky API with some 'legacy' idiosyncrasies in it before we reach 1.0.

See also https://github.com/libjxl/libjxl/pull/823#issuecomment-970448469

lib/jxl/encode.cc
lvandeve: To me it looks ok to require setting num_extra_channels in the info to 1. Note that this is not backwards compatible, in any application using the encoder where info.alpha_bits is used, num_extra_channels will have to be set correctly as well.

It would be nice, though, to avoid having to call JxlEncoderSetExtraChannelInfo for a default alpha channel. You should be able to call it if you want to, but not be required to: if you set alpha_bits in the basic info, it's implied that there is an alpha channnel.

We decided that for simplicity, this has to be the extra channel with index 0.

WDYT?

lib/include/jxl/encode.h
mo271: I'm all for a clean API. 
So even when forgetting about backwards compatibility, the question remains if it would be more convenient not to have to set 
`info.num_extra_channels` to 1 in the case that you have an interleaved images with alpha. Do we want to be strict here or not. In any case, I think we should document very clearly what is needed for the encoder API to work.

lib/jxl/encode.cc
jonsneyers: Sounds good to me to have a special case for a default alpha channel, configured from the basicinfo.

I would prefer to be strict on `num_extra_channels`, so we have the invariant that the total number of channels is always equal to `num_color_channels + num_extra_channels`. This is something that you'll likely often want to know, and if we're not strict on this, then the shortest expression that gives you the total number of channels would be something like `num_color_channels + num_extra_channels + (num_extra_channels == 0 && alpha_bits > 0 ? 1 : 0)` which feels clumsy.

lib/jxl/encode.cc
```diff
@@ -870,6 +882,7 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(const JxlEncoderOptions* frame_settings,
   }
   queued_frame->frame.SetFromImage(std::move(*io.Main().color()),
                                    io.Main().c_current());
+  // TODO: maybe add extra channels? if JPEG-1 recompression allows to add any.
```
jonsneyers, line 885: The spec allows adding alpha or other extra channels to a recompressed jpeg - from the Part 1 point of view, there is no difference between 'jxl' VarDCT and 'recompressed-jpeg' VarDCT.

From the Part 2 point of view (where `jbrd` reconstruction is described), it's also no problem: extra channels will simply be ignored when reconstructing the JPEG.

From an end-user perspective, it is maybe a bit confusing to have jxl files that can be reconstructed to a jpeg but the jpeg is actually only part of the jxl (e.g. the alpha channel gets stripped in the process). But I suppose that's a matter of djxl and other applications to e.g. show a warning in such a case indicating that there's more info in the jxl than what is represented in the reconstructed jpeg.

lib/jxl/encode.cc
mo271: Make sense, now I have changed to code such that it is required to always set `num_extra_channels` correctly.

lib/jxl/encode.cc
mo271: Agreed, clarified TODO

lib/include/jxl/encode.h
mo271: changed to the more strict behavior, so that `num_extra_channels` always has to be correct, i.e. at least 1 if alpha channel is present.

mo271:
For now, this implements the proposed function from #838. 
What is missing is to move the place where the alpha channels are handled to a different place/make their treatment compared to other extra channels more uniform.

mo271:
...rebasing manually to resolve conflicts with #939

jonsneyers (approved):

lvandeve (approved):

PR #846, opened 2021-11-11
JPG codec: use num_components instead of output_components...

eustas:
... while it is not populated

We do check that those are equal, so it is ok to use one instead another.


veluca93 (dismissed):

eustas:
Fixed formatting. Please rubberstamp.

veluca93 (approved):

PR #849, opened 2021-11-15
check alphabetic order in AUTHORS

mo271:
In addition to checking if the author of the commit is in AUTHORS we
also check the orderof the individual authors.


mo271:
As expected the authors check now fails, I will add sorting of the authors, so that it doesn't fail anymore.

sboukortt (approved):

PR #852, opened 2021-11-15
Add use-boxes and close-boxes functions to encoder API

lvandeve:
Add two new functions for boxes to the encoder API, to require user to
indicate they plan to add metadata boxes, and close boxes.

The reason is: it is more efficient for the encoder, if it knows no more
boxes will follow, to output a final codestream box of unbounded size.
On the other hand, if the possibility exists that the user wants to add
a box after the final codestream box, the encoder will have to output a
box with size indicated in the header, so it must hold on to the entire
output before passing it to the user to be able to write the size in the
box header.

Without the use-boxes and close-boxes functions, the encoder would
always have to operate under the assumption a box after the codestream
could follow.

In theory it's possible to have only JxlEncoderCloseBoxes, but then the
user would always have to call it, even when not intending to use any
boxes at all, so instead only assuming there can be boxes after calling
JxlEncoderUseBoxes has less such traps.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -396,7 +396,7 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelBuffer(
     const void* buffer, size_t size, uint32_t index);
 
 /** Adds a metadata box to the file format. JxlEncoderProcessOutput must be used
- * to effectively write the box to the output. @ref JxlEncoderUseContainer must
+ * to effectively write the box to the output. @ref JxlEncoderUseBoxes must
  * be enabled before using this function.
  *
  * Background information about the container format and boxes follows here:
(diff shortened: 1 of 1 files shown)


lib/include/jxl/encode.h
```diff
@@ -514,6 +514,29 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddBox(JxlEncoder* enc, JxlBoxType type,
                                              size_t size,
                                              JXL_BOOL compress_box);
 
+/**
+ * Indicates the intention to add metadata boxes. This allows @ref
+ * JxlEncoderAddBox can be used. When using this function, then it is required
```
sboukortt, line 519: ?     ↓
```suggestion
 * Indicates the intention to add metadata boxes. This allows @ref
 * JxlEncoderAddBox to be used. When using this function, then it is required
```

lvandeve: Fixed, thanks!

jonsneyers (approved):

PR #853, opened 2021-11-15
Add JxlEncoderCloseFrames

lvandeve:
Edited MR description after Jon's idea to keep JxlEncoderCloseInput:

Adds the separate JxlEncoderCloseFrames to close only frames.
JxlEncoderCloseInput now closes both frames and boxes.

The individual closing can also be necessary as it helps user declare
encoder can be more efficient e.g. if more frames will be added but no
more metadata boxes.

The combined JxlEncoderCloseInput is both backwards compatible and
convenient.

(Old description below:

Rename JxlEncoderCloseInput to JxlEncoderCloseFrames

This to distinguish it from CloseBoxes. The encoder needs to know
separately when each type of input ends, to allow to correctly generate
the correct box header for the final box. The CloseInput name is quite
confusing when it only closes frames and other input like boxes can
still be added.

See https://github.com/libjxl/libjxl/pull/852 for the addition of the
close boxes function.

Given that this is a rename of an always needed function, this is a
backwards incompatible change.

I believe it's still possible to do this without affecting a large
amount of usages (unlike decoder API which would require updating several
usages). The encoder API has also not been marked as frozen.

It's also OK to not submit this MR. There'll then be a function with a
bad and possibly confusing name in the encoder API in that case, but
it remains backwards compatible.)

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -491,7 +491,7 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelBuffer(
  * For now metadata boxes can only be added before or after the codestream with
  * all frames, so using JxlEncoderAddBox is only possible before the first
  * JxlEncoderAddImageFrame call, and/or after the last JxlEncoderAddImageFrame
- * call and JxlEncoderCloseInput. Support for adding boxes in-between the
+ * call and JxlEncoderCloseFrames. Support for adding boxes in-between the
  * codestream, and/or in-between image frames may be added later, and would
  * cause the encoder to use jxlp boxes for the codestream.
  *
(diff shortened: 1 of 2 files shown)


jonsneyers:
I'm in favor of this rename, but it might make sense to also keep a CloseInput function that is equivalent to calling _both_ CloseFrames and CloseBoxes, i.e. as a convenience function to declare that everything has been passed to the encoder.

That way the names can make sense while backwards compatibility is not an issue in practice, since applications that now use CloseInput after encoding typically won't add boxes afterwards.

So there would be a change in behavior of CloseInput: it would close everything. If you only want to close the frame writing or the metadata box writing, you can use the new functions.

Also don't forget to update the Changelog.

lvandeve:
> I'm in favor of this rename, but it might make sense to also keep a CloseInput function that is equivalent to calling _both_ CloseFrames and CloseBoxes, i.e. as a convenience function to declare that everything has been passed to the encoder.
> 
> That way the names can make sense while backwards compatibility is not an issue in practice, since applications that now use CloseInput after encoding typically won't add boxes afterwards.
> 
> So there would be a change in behavior of CloseInput: it would close everything. If you only want to close the frame writing or the metadata box writing, you can use the new functions.
> 
> Also don't forget to update the Changelog.

That's a good idea, done. This also allowed to keep using closeinput in all the examples, tests, etc.. so much less files affected by this MR

jonsneyers (approved):

PR #860, opened 2021-11-16
Refactor enc_color_management as the implementation of a C interface

sboukortt:
This will eventually allow other implementations of that interface to be
passed to the C encoding API. The actual plumbing is still to be done.

### DIFF
--- a/lib/include/jxl/cms_interface.h
+++ b/lib/include/jxl/cms_interface.h
@@ -1,0 +1,220 @@
+/* Copyright (c) the JPEG XL Project Authors. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file.
+ */
+
+/** @file cms_interface.h
+ * @brief Interface to allow the injection of different color management systems
+ * (CMSes, also called color management modules, or CMMs) in JPEG XL.
+ *
+ * A CMS is needed by the JPEG XL encoder and decoder to perform colorspace
+ * conversions. This defines an interface that can be implemented for different
+ * CMSes and then passed to the library.
+ */
+
+#ifndef JXL_CMS_INTERFACE_H_
+#define JXL_CMS_INTERFACE_H_
+
+#include "jxl/color_encoding.h"
+#include "jxl/memory_manager.h"
+
+#if defined(__cplusplus) || defined(c_plusplus)
+extern "C" {
+#endif
+
+/** Represents an input or output colorspace to a color transform, as a
+ * serialized ICC profile. */
+typedef struct {
+  /** The serialized ICC profile. This is guaranteed to be present and valid. */
+  struct {
+    const uint8_t* data;
+    size_t size;
+  } icc;
+
+  /** Structured representation of the colorspace, if applicable. If all fields
+   * are different from their "unknown" value, then this is equivalent to the
+   * ICC representation of the colorspace. If some are "unknown", those that are
+   * not are still valid and can still be used on their own if they are useful.
+   */
+  JxlColorEncoding color_encoding;
+
+  /** Number of components per pixel. This can be deduced from the other
+   * representations of the colorspace but is provided for convenience and
+   * validation. */
+  size_t num_channels;
+} JxlColorProfile;
+
+/** Allocates and returns the data needed for @p num_threads parallel transforms
+ * from the @p input colorspace to @p output, with up to @p pixels_per_thread
+ * pixels to transform per call to JxlCmsInterface::run. @p init_data comes
+ * directly from the JxlCmsInterface instance. Since @c run only receives the
+ * data returned by @c init, a reference to @p init_data should be kept there
+ * if access to it is desired in @c run. Likewise for JxlCmsInterface::destroy.
+ *
+ * The ICC data in @p input and @p output is guaranteed to outlive the @c init /
+ * @c run / @c destroy cycle.
+ *
+ * @param init_data JxlCmsInterface::init_data passed as-is.
+ * @param num_threads the maximum number of threads from which
+ *        JxlCmsInterface::run will be called.
+ * @param pixels_per_thread the maximum number of pixels that each call to
+ *        JxlCmsInterface::run will have to transform.
+ * @param input_profile the input colorspace for the transform.
+ * @param output_profile the colorspace to which JxlCmsInterface::run should
+ *        convert the input data.
+ * @param intensity_target for colorspaces where luminance is relative
+ *        (essentially: not PQ), indicates the luminance of (1, 1, 1). This is
+ *        only useful for conversions between a relative luminance colorspace
+ *        and an absolute luminance one, in either direction.
+ *        @p intensity_target cd/m² in the absolute colorspace should map to and
+ *        from (1, 1, 1) in the relative one.
+ * @return The data needed for the transform, or @c NULL in case of failure.
+ *         This will be passed to the other functions as @c user_data.
+ */
+typedef void* (*jpegxl_cms_init_func)(void* init_data, size_t num_threads,
+                                      size_t pixels_per_thread,
+                                      const JxlColorProfile* input_profile,
+                                      const JxlColorProfile* output_profile,
+                                      float intensity_target);
+
+/** Returns a buffer that can be used by callers of the interface to store the
+ * input of the conversion or read its result, if they pass it as the input or
+ * output of the @c run function.
+ * @param user_data the data returned by @c init.
+ * @param thread the index of the thread for which to return a buffer.
+ * @return A buffer that can be used by the caller for passing to @c run.
+ */
+typedef float* (*jpegxl_cms_get_buffer_func)(void* user_data, size_t thread);
+
+/** Executes one transform and returns true on success or false on error. It
+ * must be possible to call this from different threads with different values
+ * for @p thread, all between 0 (inclusive) and the value of @p num_threads
+ * passed to @c init (exclusive). It is allowed to implement this by locking
+ * such that the transforms are essentially performed sequentially, if such a
+ * performance profile is acceptable. @p user_data is the data returned by
+ * @c init.
+ * The buffers each contain @p num_pixels × @c num_channels interleaved floating
+ * point (0..1) samples where @c num_channels is the number of color channels of
+ * their respective color profiles. It is guaranteed that the only case in which
+ * they might overlap is if the output has fewer channels than the input, in
+ * which case the pointers may be identical.
+ * For CMYK data, 0 represents the maximum amount of ink while 1 represents no
+ * ink.
+ * @param user_data the data returned by @c init.
+ * @param thread the index of the thread from which the function is being
+ *        called.
+ * @param input_buffer the buffer containing the pixel data to be transformed.
+ * @param output_buffer the buffer receiving the transformed pixel data.
+ * @param num_pixels the number of pixels to transform from @p input to
+ * @p output.
+ * @return JXL_TRUE on success, JXL_FALSE on failure.
+ */
+typedef JXL_BOOL (*jpegxl_cms_run_func)(void* user_data, size_t thread,
+                                        const float* input_buffer,
+                                        float* output_buffer,
+                                        size_t num_pixels);
+
+/** Performs the necessary clean-up and frees the memory allocated for user
+ * data.
+ */
+typedef void (*jpegxl_cms_destroy_func)(void*);
+
+/**
+ * Interface for performing colorspace transforms. The @c init function can be
+ * called several times to instantiate several transforms, including before
+ * other transforms have been destroyed.
+ *
+ * The call sequence for a given colorspace transform could look like the
+ * following:
+ * @dot
+ * digraph calls {
+ *   newrank = true
+ *   node [shape = box, fontname = monospace]
+ *   init [label = "user_data <- init(\l\
+ *     init_data = data,\l\
+ *     num_threads = 3,\l\
+ *     pixels_per_thread = 20,\l\
+ *     input = (sRGB, 3 channels),\l\
+ *     output = (Display-P3, 3 channels),\l\
+ *     intensity_target = 255\l\
+ *   )\l"]
+ *   subgraph cluster_0 {
+ *   color = lightgrey
+ *   label = "thread 1"
+ *   labeljust = "c"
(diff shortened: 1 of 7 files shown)


lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+  struct {
+    const uint8_t* data;
+    size_t size;
+  } icc;
+
+  /** Structured representation of the colorspace, if applicable. */
+  JxlColorEncoding color_encoding;
+
+  /** Number of components per pixel. */
+  size_t num_channels;
```
lvandeve, line 48: What are the possible values of this field and their meaning?

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+ * @param num_threads the maximum number of threads from which
+ *        JxlCmsInterface::run will be called.
+ * @param pixels_per_thread the maximum number of pixels that each call to
+ *        JxlCmsInterface::run will have to transform.
+ * @param input the input colorspace for the transform.
+ * @param output the colorspace to which JxlCmsInterface::run should convert the
+ *        input data.
+ * @param intensity_target for colorspaces where luminance is relative
+ *        (essentially: not PQ), indicates the luminance of (1, 1, 1).
+ * @return The data needed for the transform, or @c null in case of failure.
```
lvandeve, line 71: This is called user_data in the function below, perhaps this name could be integrated in this comment as well for clarity?

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+/** Represents an input or output colorspace to a color transform, as a
+ * serialized ICC profile. */
+typedef struct {
+  /** The serialized ICC profile. */
+  struct {
+    const uint8_t* data;
+    size_t size;
+  } icc;
+
+  /** Structured representation of the colorspace, if applicable. */
```
lvandeve, line 44: How can one know whether color_encoding, icc, or both are being used? Should there be a flag for one, or for each?

lib/include/jxl/cms_interface.h
lvandeve, line 71: NULL is more appropriate for C

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @file cms_interface.h
+ * @brief Interface to allow the injection of different color management systems
+ * (also called color management modules, or CMMs) in JPEG XL.
```
jonsneyers, line 18: A bit strange to introduce the acronym CMM here while everywhere else we're going with CMS.

lib/include/jxl/cms_interface.h
jonsneyers: I imagine that for an external CMS, the convenient way to do things is to always use icc.

For the internal partial CMS that we have in libjxl (which is needed to do decoding in case of nontrivial frame blending), only color_encoding is implemented.

What is going to implement this C interface, and will it be acceptable/possible to have an implementation that is only partial, e.g. that only works with color_encoding but not with icc profiles?

lib/include/jxl/cms_interface.h
jonsneyers: For implementing all the color conversions needed to show a jxl image on a screen, we need at least the following:

- 1: Grayscale
- 3: RGB
- 4: CMYK

It might also be useful to support other things like L*a*b* or RGBA/RGBX, but it's probably best to keep this minimal so it's not too much work to implement this interface (3 cases is already enough).

Is the pixel format just implicitly floats with the components interleaved and a nominal range of [0..1] where 0 is black (i.e. no light or max ink)?

It is useful to spell out such conventions, especially for the CMYK case where [0..100] is also used, and 0 can either mean black (max ink) or white (no ink).

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+ * for @p thread, all between 0 (inclusive) and the value of @p num_threads
+ * passed to @c init (exclusive). It is allowed to implement this by locking
+ * such that the transforms are essentially performed sequentially, if such a
+ * performance profile is acceptable. @p user_data is the data returned by
+ * @c init.
+ * @param user_data the data returned by @c init.
+ * @param thread the index of the thread from which the function is being
+ *        called.
+ * @param input the buffer containing the pixel data to be transformed.
+ * @param output the buffer receiving the transformed pixel data.
```
jonsneyers, line 99: It's perhaps slightly confusing that the variable names `input` and `output` are used for both JxlColorProfiles (in init_func) and for buffers (here). Maybe `in_profile`, `out_profile`, `in_buffer`, `out_buffer` are clearer names.

It would be good to clarify that `in_buffer` is an array of `num_pixels * in_profile.num_channels` floats and `out_buffer` is an array of `num_pixels * out_profile.num_channels` floats.

It would also be good to clarify whether it is allowed that `in_buffer` overlaps with `out_buffer` (e.g. to do conversions in-place). Probably we want to allow that, but we want to restrict it to cases where you can just read a pixel from `in_buffer` and then write it to `out_buffer` without having to worry about overwriting pixel values you will still need to read later.

E.g. for converting CMYK to RGB or RGB to Grayscale it is OK to have `in_buffer` and `out_buffer` point to the same address, but for converting Grayscale to RGB it would not be OK.

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+ * @param init_data JxlCmsInterface::init_data passed as-is.
+ * @param num_threads the maximum number of threads from which
+ *        JxlCmsInterface::run will be called.
+ * @param pixels_per_thread the maximum number of pixels that each call to
+ *        JxlCmsInterface::run will have to transform.
+ * @param input the input colorspace for the transform.
+ * @param output the colorspace to which JxlCmsInterface::run should convert the
+ *        input data.
+ * @param intensity_target for colorspaces where luminance is relative
+ *        (essentially: not PQ), indicates the luminance of (1, 1, 1).
```
jonsneyers, line 70: How is `intensity_target` supposed to be interpreted / taken into account by a CMS? Does it apply to the input or output colorspace?

lib/include/jxl/cms_interface.h
```diff
@@ -0,0 +1,210 @@
...
+ * @c run / @c destroy cycle.
+ *
+ * @param init_data JxlCmsInterface::init_data passed as-is.
+ * @param num_threads the maximum number of threads from which
+ *        JxlCmsInterface::run will be called.
+ * @param pixels_per_thread the maximum number of pixels that each call to
+ *        JxlCmsInterface::run will have to transform.
+ * @param input the input colorspace for the transform.
+ * @param output the colorspace to which JxlCmsInterface::run should convert the
+ *        input data.
```
jonsneyers, line 68: It might be useful to restrict the allowed combinations of (input, output).

E.g. I think it is OK to restrict the output colorspace to be an RGB one (not CMYK or Grayscale).

lib/include/jxl/cms_interface.h
sboukortt: Good points, how is this?

lib/include/jxl/cms_interface.h
sboukortt: I am not sure about that, as the plan would be to also reuse this for a high-level decoding API used by `djxl`, so we may want to convert to a grayscale colorspace if that was the original one. What do you think?

lib/include/jxl/cms_interface.h
sboukortt: It is only relevant if one of the profiles is relative and the other is absolute, in which case it applies to relative one, whether that is the input or the output. Do you think this should be made somehow clearer?

lib/include/jxl/cms_interface.h
lvandeve: I also meant that a comment for the user to know what to do with this field would be helpful

lib/include/jxl/cms_interface.h
lvandeve: To me this looks good, and CMS is more common I think so I'm ok with CMS anywhere else outside of this mention

lib/include/jxl/cms_interface.h
lvandeve: AFAIK, the user will implement a function that receive this JxlColorProfile, right (two instances)? So the user is the one who is able to use the color_encoding field if they want to. But how can they know that they can use it, if there's no indication whether it's unused, or filled in?

The JxlColorProfile could contain "unknown" but still have valid other values, in theory, the jxl file format allows this

lib/include/jxl/cms_interface.h
jonsneyers: For CMYK, 0=max ink is how it is done in jxl and in skcms, but not how it is done in lcms. See also https://github.com/libjxl/libjxl/pull/237

It's somewhat redundant if this is the same as the number of channels implied by the ICC profile, but it might be convenient to have it so implementations don't have to figure it out themselves, and it could be a good sanity check which might help to avoid buffer overrun bugs caused by someone thinking it's 3 components while in reality there is only 1 etc.

lib/include/jxl/cms_interface.h
jonsneyers: Yes, introducing both acronyms and then using one of them looks fine, it was just weird to introduce CMM but not CMS when CMS is the one that gets used.

lib/include/jxl/cms_interface.h
jonsneyers: It should be documented what is guaranteed to be present, what is maybe there, and what takes precedence in case of conflicts. Maybe we also need a boolean indicating whether it is safe to use the struct colorspace instead of the ICC profile.

In the CMYK case, the spec requires that to be done always with an explicit CMYK ICC profile, a kBlack extra channel, and ColorSpace in the image header set to RGB (since we don't have an enum value there for CMYK, only for RGB, Gray, XYB and UNKNOWN). This is maybe a bit confusing and it might be clearer to add a JXL_COLOR_SPACE_CMYK enum value in the api, which gets returned in case there is a kBlack channel. Not that the struct colorspace can ever be used in the CMYK case, so maybe it doesn't really matter.

lib/include/jxl/cms_interface.h
jonsneyers: If this is a parameter that gets ignored in the case of absolute to absolute or relative to relative, that should be documented.

I think @lvandeve's suggestion might make sense to have this field in JxlColorProfile, maybe with a special value for "relative". Conceptually it makes more sense to me to have "fully defined"  colorspaces (not necessarily actually fully defined in the case of relative luminance, but at least all available info is in the struct that way).

sboukortt: The input intensity target matters if the input is in a relative colorspace and the output is absolute, for example (1, 1, 1) in sRGB would be converted to (0.6046, 0.6046, 0.6046) in PQ if `intensity_target = 255`, or to (0.8274, 0.8274, 0.8274) if `intensity_target = 2000`.

Conversely, (0.6046, 0.6046, 0.6046) in PQ would be converted to (1, 1, 1) in sRGB if `intensity_target = 255`, or to (0.3922, 0.3922, 0.3922) if `intensity_target = 2000`.

Regarding making it part of the colorspace vs. the image, the way I see it is:

1) currently, in the rest of the codebase, we consider it a property of the image. It is the maximum luminance that can possibly be found in the image, and corresponds to (1, 1, 1) when the image is encoded using a relative colorspace. So there is some logic in having it that way, and it also makes it clear that relative<->relative conversions won’t need to concern themselves about this as it will be the same on both sides.

2) Alternatively, we could consider it a property of the colorspace, but then it would deviate from the meaning that the term “intensity_target” currently has so we should probably change it in this interface to make that clear, for example to “white_luminance” as in EXR. The invoker of the CMS would then set that to `intensity_target` if the colorspace is relative, or to 10 000 for PQ. Then, there would be less need to treat PQ specially in the CMS: the condition for the luminance rescaling would instead be `input_white_luminance ≠ output_white_luminance`. So there is indeed logic in support of this option as well although it would require more changes to the current code.

jonsneyers: Option 2 seems most attractive to me.

In general, it would also be good to clarify the semantics of intensity_target in the bitstream, in the encoder, the decoder, the butteraugli metric, and here. It's indeed a good idea to use different names when it means different things (or has different implications).

lib/include/jxl/cms_interface.h
jonsneyers: Converting to grayscale might make sense, yes. Converting from RGB to CMYK should not be needed.

lib/include/jxl/cms_interface.h
jonsneyers: LGTM!

lib/include/jxl/cms_interface.h
lvandeve: It should be clear to the user how to use the field or parameter, the user has to implement this function, and e.g. if you implement it using lcms or skcms, it'd be good if the intensity_target parameter is well documented (or alternatively there is a good example showing how to use it) so that the user knows if they have to use the parameter before or after calling the CMS, or it's a setting the CMS itself supports

lib/include/jxl/cms_interface.h
sboukortt: Makes sense, thanks. Added, as well as a comment clarifying the CMYK representation (but the latter, I added below, to the documentation for `run`).

lib/include/jxl/cms_interface.h
sboukortt: I would favor sticking with 1, at least for now. I think it brings fewer unnecessary complications. I have added some clarification of how to interpret the `intensity_target` parameter, how does it look?

lvandeve (dismissed):

lib/include/jxl/cms_interface.h
lvandeve: Given the meaning of "brightest luminance present in the image", having one seems fine

veluca93 (approved):

Issue #859, opened 2021-11-16
fclose() / JxlDecoderCloseInput() not called in the jxlinfo.c example


vtorri:
afaics, fclose() / JxlDecoderCloseInput() are not called in jxlinfo.c example, is it normal ?

jonsneyers:
If it's not done, it's probably because we were lazy and didn't bother with it since the program gets terminated anyway. But that's bad style and not a good example, so we should fix that :)

PR #862, opened 2021-11-17
call JxlDecoderCloseInput and fclose in jxlinfo

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/859

In principle, jxlinfo could return an error when the file ends with a not-until-eof box, since then the decoder would keep asking for more input (since there might be more boxes after that one).

@lvandeve could you check that it's calling JxlDecoderCloseInput at the correct time?


lvandeve:
@jonsneyers yes the JxlDecoderCloseInput is at the correct time, given that it happens once, when the end of the file is reached, right after setting this final input

lvandeve (approved):

PR #863, opened 2021-11-17
Refector encoder for box output, and support jxlp

lvandeve:
Various refactorings in the encoder to prepare for adding the
implementation of the add-boxes functions.

This also makes the encoder support jxlp boxes (partial codestream
boxes): it'll now add each frame in its own jxlp box, and possibly
an as early as possible jxlp box with the basic info, because it's
much simpler in the encoder to insert other boxes between frames when
using jxlp rather than a single jxlc. This is more efficient for the
decoder too since no more boxes with unbounded size are used.

Also made the decoder more strict about final jxlp being indicated:
experimenting with this new encoder feature showed that the decoder
ignored the error of not having a jxlp box marked as last.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -488,20 +488,14 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelBuffer(
  * multiple boxes of type jxlp are used. This is handled by the encoder when
  * needed.
  *
- * For now metadata boxes can only be added before or after the codestream with
- * all frames, so using JxlEncoderAddBox is only possible before the first
- * JxlEncoderAddImageFrame call, and/or after the last JxlEncoderAddImageFrame
- * call and JxlEncoderCloseFrames. Support for adding boxes in-between the
- * codestream, and/or in-between image frames may be added later, and would
- * cause the encoder to use jxlp boxes for the codestream.
- *
  * @param enc encoder object.
  * @param type the box type, e.g. "Exif" for EXIF metadata, "XML " for XMP or
  * IPTC metadata, "jumb" for JUMBF metadata.
  * @param contents the full contents of the box, for example EXIF
  * data. For an "Exif" box, the EXIF data must be prepended by a 4-byte tiff
  * header offset, which may be 4 zero-bytes. The ISO BMFF box header must not
- * be included, only the contents.
+ * be included, only the contents. Owned by the caller and its contents are
+ * copied internally.
  * @param size size of the box contents.
  * @param compress_box Whether to compress this box as a "brob" box. Requires
  * Brotli support.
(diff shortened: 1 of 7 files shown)


jonsneyers (approved):

jonsneyers:
Looks good to me. Putting each frame in its own `jxlp` is a sensible thing to do by default, but it's probably a good idea to have encoder options to guide decisions there — for maximum density, you want to avoid the container if possible and put everything in a single `jxlc`; if encoding things in chunks is desirable, there might be other useful strategies, e.g. maybe you can have some max box size and allow multiple frames per `jxlp` as long as the max jxlp box size is not exceeded, or maybe you want to have basic info + DC in the first jxlp, etc etc.

(no need to add such options right now, of course)

lvandeve:
> Looks good to me. Putting each frame in its own `jxlp` is a sensible thing to do by default, but it's probably a good idea to have encoder options to guide decisions there — for maximum density, you want to avoid the container if possible and put everything in a single `jxlc`; if encoding things in chunks is desirable, there might be other useful strategies, e.g. maybe you can have some max box size and allow multiple frames per `jxlp` as long as the max jxlp box size is not exceeded, or maybe you want to have basic info + DC in the first jxlp, etc etc.
> 
> (no need to add such options right now, of course)

I don't think an option is needed, just more complex implementation, which can be done in any MR.

The main issue is: boxes must have the size in the header, so to get a single jxlc box with size in header, encoder would have to store all encoded frames in memory, and only output data to the user once all frames were added and frame input closed.

Before this MR, the encoder used an unbounded jxlc box (final box, no size in its header) to solve this. However, that's incompatible with the user adding metadata boxes after this.

After this MR, the encoder solves the problem instead by using multiple jxlp each with the size of a single encoded frame in the header.

All of this applies to multiple frames: if there's a single frame, both before and after this MR the encoder will output a single known-size jxlc (only exception is after this MR it may split the basic info into a jxlp in some cases, which is so useful for the decoder that it's ok to be non-configurable)

Some future extra encoder logic could be (for multiple frames case):

-if user does not use metadata boxes (didn't call UseBoxes), use the old unbounded jxlc strategy again, rather than always jxlp like this MR does

-if user calls AddFrame multiple times and then calls CloseFrames, and also closed boxes (or never uses metadata boxes), and only then calls ProcessOutput, let encoder create a known-size jxlc. This would be smarter than what encoder did before this MR, as it always worked frame by frame.

In case of multiple frames, the overhead of the multiple jxlp boxes is so negligable that it's not super important though. Remember that it already outputs a jxlc now in case of single frame and no metadata boxes.

jonsneyers:
Yes, maybe we can just make the encoder do something good in all cases without requiring an option. I agree that a few bytes overhead doesn't matter enough to have an option for it.

For large still images it might still be useful to split them in multiple jxlps in some cases, e.g. one per layer, or one for the DC groups, followed by metadata, followed by the rest of the image data. But we can decide later if an option for that would be needed (it could be an Enum option if we want it, so easy to add in a backwards compatible way).

PR #864, opened 2021-11-17
Fix configuration of a static MSVC build

donny-dont:
Windows uses .lib files for its libraries so setting the suffix to .a prevents find_library from functioning. This value is already set on the platform level so there isn't any reason to override.

Also don't use unix libraries with MSVC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,12 +161,12 @@ message(STATUS
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 if(JPEGXL_STATIC)
-  set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
   set(BUILD_SHARED_LIBS 0)
   # Clang developers say that in case to use "static" we have to build stdlib
   # ourselves; for real use case we don't care about stdlib, as it is "granted",
   # so just linking all other libraries is fine.
-  if (NOT APPLE)
+  if (NOT MSVC AND NOT APPLE)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
     set(CMAKE_EXE_LINKER_FLAGS
         "${CMAKE_EXE_LINKER_FLAGS} -static -static-libgcc -static-libstdc++")
   endif()


CMakeLists.txt
```diff
@@ -158,12 +158,11 @@ message(STATUS
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 if(JPEGXL_STATIC)
-  set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
   set(BUILD_SHARED_LIBS 0)
   # Clang developers say that in case to use "static" we have to build stdlib
   # ourselves; for real use case we don't care about stdlib, as it is "granted",
   # so just linking all other libraries is fine.
-  if (NOT APPLE)
+  if (UNIX AND NOT APPLE)
```
jonsneyers, line 165: Are there any other non-UNIX targets besides MSVC that might be influenced by this change?
It might perhaps be better to do
```suggestion
  if (NOT MSVC AND NOT APPLE)
```
?

donny-dont: If you're on Windows and compiling in a Cygwin or MSYS the UNIX variable is set to `TRUE` so I believe this is what you would want.

In CMake one of the places that does `set(UNIX 1)` is in https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/UnixPaths.cmake#L15 which is included in the CYWIN module at https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/CYGWIN.cmake#L64 and MSYS includes that at https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/MSYS.cmake#L2 .

The docs for CMake aren't explicit on what platforms count as UNIX but this change would prevent MSVC and Apple builds from adding these flags and would keep the same behavior for all other platforms.

donny-dont: Since our time zones don't have much of any overlap for working hours I want to be clear that I'm fine with the proposed change being committed. I just wanted to make sure I gave enough information for you all to make an informed choice on what the values should be.

Thanks for the review!

sboukortt: The Cygwin-based MSYS subsystem sets it but MinGW (“Windows-GNU”) does not: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-GNU.cmake

So I think `NOT MSVC` would indeed be preferable here.

sboukortt (approved):

CMakeLists.txt
```diff
@@ -158,12 +158,11 @@ message(STATUS
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 if(JPEGXL_STATIC)
-  set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
```
sboukortt, line 161: I just realized that we may still want this in the `if (NOT MSVC AND NOT APPLE)` branch. Let’s perhaps move it there instead of removing it?

donny-dont: I'm happy to move it back if that's what's needed I just haven't seen a case where someone overrides that value in practice.

Its typically done in the Platform files like the Windows-GNU one you pointed at had it set like this https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-GNU.cmake#L27

sboukortt: From what I remember, the rationale is that when building statically, we want to find the static (`.a`) version of libraries specifically and not the dynamic (`.dll.a`) one.

donny-dont: Updated with fix.

jonsneyers: Do we have static builds in CI actions that can fail or succeed depending on getting this right?

I dunno why we had that line in the first place, it indeed seems like something that should not need to be manually set (cmake should know best how to do static or dynamic library search on each platform) but maybe there was a need somehow.

I am used to shared libraries being `.so` and static ones being `.a`, but I don't know how it is on non-linux platforms.

jonsneyers (approved):

CMakeLists.txt
jonsneyers: Maybe we did it like that because on linux, we don't do a 'real' static build, and still depend on stdlib.

PR #870, opened 2021-11-18
Proposed documentation wording tweaks.

fischbacher:
The biggest fix is making the documentation about handling/ignoring
embedded orientation information un-ambiguous.


lib/include/jxl/decode.h
```diff
@@ -403,7 +405,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSubscribeEvents(JxlDecoder* dec,
  * @return JXL_DEC_SUCCESS if no error, JXL_DEC_ERROR otherwise.
  */
 JXL_EXPORT JxlDecoderStatus
-JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_orientation);
+JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL keep_data_orientation);
```
jonsneyers, line 408: Also have to update the `@param` name in the documentation above, those names need to match.

jonsneyers: I'm not sure if `keep_data_orientation` is a clearer name than `keep_orientation`. The word "data" is quite generic.

Maybe `skip_auto_orientation` would be a better name here. If you go just by the param name (ignoring the rest of the documentation), it is somewhat ambiguous what "keeping" orientation means: does it mean keeping the image un-oriented, i.e. not applying the orientation (this is what it means), or does it mean 'keeping' the image orientation, i.e. applying orientation (this is what it doesn't mean, but it could be misinterpreted easily this way).

lib/include/jxl/decode.h
```diff
@@ -371,27 +372,28 @@ JXL_EXPORT size_t JxlDecoderSizeHintBasicInfo(const JxlDecoder* dec);
...
+ * applying the transformation from the orientation setting, returning
+ * the image in the as-in-bitstream pixeldata orientation.
  * This may be faster to decode since the decoder doesn't have to apply the
  * transformation, but can cause wrong display of the image if the orientation
  * tag is not correctly taken into account by the user.
  *
- * By default, this option is disabled, and the decoder automatically corrects
- * the orientation.
+ * By default, this option is disabled, and the decoder respects the
+ * orientation setting provided by the image.
```
jonsneyers, line 396: It wouldn't hurt to keep the old wording too, for maximum clarity, e.g.:
```suggestion
 * orientation setting provided by the image, i.e., it automatically orients the image correctly.
```

lib/include/jxl/decode.h
```diff
@@ -599,10 +601,10 @@ typedef enum {
  *   an unknown or xyb color space. In that case,
  *   JxlDecoderGetColorAsICCProfile is not available.
  *
- * If you wish to render the image using a system that supports ICC profiles,
- * use JxlDecoderGetColorAsICCProfile first. If you're looking for a specific
- * color space possibly indicated in the JPEG XL image, use
- * JxlDecoderGetColorAsEncodedProfile first.
+ * When rendering an image on a system that supports ICC profiles,
+ * JxlDecoderGetColorAsICCProfile should be used first. When rendering
+ * for a specicif color space, possibly indicated in the JPEG XL
```
jonsneyers, line 606: specific

lib/include/jxl/decode.h
```diff
@@ -113,9 +113,9 @@ JXL_EXPORT void JxlDecoderDestroy(JxlDecoder* dec);
 
 /**
  * Return value for JxlDecoderProcessInput.
- * The values above 0x40 are optional informal events that can be subscribed to,
- * they are never returned if they have not been registered with
- * JxlDecoderSubscribeEvents.
+ * The values from JXL_DEC_EXTENSIONS onwards are optional informal events
```
jonsneyers, line 116: It was already there, but I think the word "informal" is not correct (this is not about formal versus informal). In the JxlDecoderSubscribeEvents documentation, these are called "informational", which is not the same word, and also not really the best word (all return values are in some sense informational; the word is too vague). In the comments above the enum values themselves they are called "informative" events, which is yet another word.

Also the word "optional" is not really conveying the meaning well - I think it's more that you have to "opt in" to them, but that is not really just an optional thing to do, at least if you want to get an image decoded it's actually required, not optional.

I think the whole concept could be explained a bit better, namely that there are two different kinds of return values, some that you can always get (the ones below 0x40), and others that you can only get if you subscribe to them, and most of all: that the decoder will do different things depending on what you subscribe to: if you don't subscribe to getting metadata boxes, they will just be ignored; if you don't subscribe to getting pixel data, no actual decoding will happen, etc. So subscribing to events is not just about the granularity or frequency of the information/interruptions you get during decode, but it is also about indicating what exactly you want the decoder to do.

Maybe the wording "opt-in events" is clearer than "optional informal events" or "informative events" or "informational events"?

jonsneyers: Also: JXL_DEC_BASIC_INFO is the first opt-in event, i.e. the "above 0x40" was meant inclusively.

lib/include/jxl/decode.h
fischbacher: Agree that that name is not great either. If going further astray than merely expanding the name is OK, I would propose "ignore_reorientation" - it's short, and captures what is happening here. Perhaps: ignore_reorientation_step.

What do you think?

lib/include/jxl/decode.h
fischbacher: I think the term "correct" should not be used in technical writing, since it basically gets used as an abbreviation for "you know what is meant". But if we both know what is meant (and that claim might actually be more contentious than it might look like): why not simply say what is meant here?

How about: "By default, this option is disabled, and after initial decoding, the pixel data gets re-oriented according to the image's Orientation setting".

lib/include/jxl/decode.h
jonsneyers: I like `skip` more than `ignore`, I have no preference on `orientation` vs `reorientation` vs `auto_orientation`.

Since the function name cannot be changed, I'm also fine with sticking with `keep_orientation` (to keep it the same as in the function name) and just hoping that people will read the documentation.

lib/include/jxl/decode.h
jonsneyers: Yes, that looks good too. I'd drop the "after initial decoding" because it's an implementation detail whether we first decode pixels and then orient them, or do the orienting while decoding — what matters is what gets returned to the application.

So maybe something like this:

"By default, this option is disabled, and the returned pixel data is re-oriented according to the image's Orientation setting".

lib/include/jxl/decode.h
fischbacher: The problem is: "keep which orientation" - the one that the raw image data uses or the one that the orientation setting specifies?

Assuming that changing the function name is not an option here: "keep_data_orientation" would at least provide some more context by saying "it's the orientation that the data comes in that we want to keep here" - as opposed to using the orientation setting provided separately. And in that light, I do not see "data" as "too generic", since it's "the orientation that the image data is shipped with" - vs. "the orientation setting specified on top of that".

So, let's do `skip_reorientation` here? If the function could be named JxlDecoderSetSkipReorientation(), that would (in my view) be less ambiguous than the current name, but I see that that might not be an option here.

lib/include/jxl/decode.h
fischbacher: Very good. Going with this last proposal.

lib/include/jxl/decode.h
fischbacher: Will be going with "informative" here for now, since this is what the other text is using, but agree that this perhaps should be re-worded throughout.

lib/include/jxl/decode.h
jonsneyers: Yes, `skip_reorientation` sounds good.

jonsneyers (approved):

mo271 (approved):
refreshing the approval from @jonsneyers.

Issue #647, opened 2021-09-27
Allow builing of libpixbufloader-jxl.dll on Windows under MSYS2


novomesk:
Hello,

right now the gdk-pixbuf plugin is not built on Windows because of the following condition:
```
if(NOT WIN32)
  add_subdirectory(gdk-pixbuf)
endif()
```
I understand that MS Visual Studio users most probably don't need it but it is useful for some applications using MSYS2 platform.

PR #875, opened 2021-11-19
build gdk-pixbuf plugin when building for MSYS

jonsneyers:
fixes https://github.com/libjxl/libjxl/issues/647

(at least I hope it does)

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -3,7 +3,7 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-if(NOT WIN32)
+if(NOT MSVC)
   add_subdirectory(gdk-pixbuf)
 endif()
 


plugins/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-if(NOT WIN32)
+if(MSYS OR NOT WIN32)
```
kmilos, line 6: `MSYS` is only defined only for 'MSYS Makefiles'... You want to test for `MINGW`, or even better/simpler, just test for `NOT MSVC`

jonsneyers: Ah thanks, I'll update it. I know very little about all those different Windows build flags, so thanks for pointing me to the right one :)

kmilos (approved):
LGTM

eustas (approved):

PR #876, opened 2021-11-19
Fix HLG<->HLG and PQ<->PQ conversions

sboukortt:
Those occur when using djxl’s `--color_space=` flag.

### DIFF
--- a/lib/jxl/color_management_test.cc
+++ b/lib/jxl/color_management_test.cc
@@ -233,5 +233,27 @@ TEST_F(ColorManagementTest, D2700ToSRGB) {
                           FloatNear(0.601, 1e-3)));
 }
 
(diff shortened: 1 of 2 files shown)


jonsneyers (approved):

jonsneyers:
Is the title of this PR accurate? This seems to also affect HLG<->anything or PQ<->anything conversions, no?

sboukortt:
> Is the title of this PR accurate? This seems to also affect HLG<->anything or PQ<->anything conversions, no?

It does apply to them but I didn’t want to imply anything about the delta in correctness for those. :p It is not yet clear to me why this didn’t work without this change (and it bothers me), and I have potentially more changes in mind for HLG<->other conversions.

PR #888, opened 2021-11-22
fix no-coalescing fuzzerbugs

jonsneyers:
~~Couldn't find any fuzzerbugs running it locally for a short while, but still good to test both coalesced and non-coalesced decoding to find potential bugs.~~

Fixes a fuzzer-found bug in the uint8 buffer case (where stride was not computed correctly in the no-coalescing case).
Also adds an ASSERT in dec_external_image to hopefully find more such bugs and to turn buffer overruns into an abort.

### DIFF
--- a/lib/jxl/dec_external_image.cc
+++ b/lib/jxl/dec_external_image.cc
@@ -315,11 +315,14 @@ Status ConvertChannelsToExternal(const ImageF* channels[], size_t num_channels,
   // First channel may not be nullptr.
   size_t xsize = channels[0]->xsize();
   size_t ysize = channels[0]->ysize();
-
   if (stride < bytes_per_pixel * xsize) {
     return JXL_FAILURE("stride is smaller than scanline width in bytes: %" PRIuS
                        " vs %" PRIuS,
                        stride, bytes_per_pixel * xsize);
+  }
+  if (!out_callback &&
+      out_size < (ysize - 1) * stride + bytes_per_pixel * xsize) {
+    return JXL_FAILURE("out_size is too small to store image");
   }
 
   const bool little_endian =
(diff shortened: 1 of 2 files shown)


lib/jxl/dec_external_image.cc
```diff
@@ -315,12 +315,12 @@ Status ConvertChannelsToExternal(const ImageF* channels[], size_t num_channels,
   // First channel may not be nullptr.
   size_t xsize = channels[0]->xsize();
   size_t ysize = channels[0]->ysize();
-
   if (stride < bytes_per_pixel * xsize) {
     return JXL_FAILURE("stride is smaller than scanline width in bytes: %" PRIuS
                        " vs %" PRIuS,
                        stride, bytes_per_pixel * xsize);
   }
+  JXL_ASSERT(out_callback || out_size >= ysize * stride);
```
veluca93, line 323: Not sure how I feel about adding an assert here, shouldn't we return a failure instead?

veluca93:
Can we split adding fuzzing and fixing a fuzzbug in two PRs or, at least, two commits?

jonsneyers:
> Can we split adding fuzzing and fixing a fuzzbug in two PRs or, at least, two commits?

Yes. Updated this one to just fix the bug. I'll make a new one to add fuzzing.

lib/jxl/dec_external_image.cc
jonsneyers: Makes sense. Changed the condition to correspond to https://github.com/libjxl/libjxl/issues/994 (don't require padding after last row), and returning failure instead of having an assert.

veluca93 (approved):

PR #893, opened 2021-11-22
Add remaining cjxl options to encoder api

lvandeve:
Also renamed the channel palette settings to what I now think are more clear names

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -209,13 +209,13 @@ typedef enum {
    * percentage of range. Use 0-100 to set an explicit percentage, -1 to use the
    * encoder default. Used for modular encoding.
    */
-  JXL_ENC_OPTION_CHANNEL_COLORS_PRE_TRANSFORM_PERCENT = 20,
+  JXL_ENC_OPTION_CHANNEL_COLORS_GLOBAL_PERCENT = 20,
 
-  /** Use Local channel palette if the amount of colors is smaller than this
-   * percentage of range. Use 0-100 to set an explicit percentage, -1 to use the
-   * encoder default. Used for modular encoding.
+  /** Use Local (per-group) channel palette if the amount of colors is smaller
+   * than this percentage of range. Use 0-100 to set an explicit percentage, -1
+   * to use the encoder default. Used for modular encoding.
    */
-  JXL_ENC_OPTION_CHANNEL_COLORS_PERCENT = 21,
+  JXL_ENC_OPTION_CHANNEL_COLORS_GROUP_PERCENT = 21,
 
   /** Use color palette if amount of colors is smaller than or equal to this
    * amount, or -1 to use the encoder default. Used for modular encoding.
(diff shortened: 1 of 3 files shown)


lib/include/jxl/encode.h
```diff
@@ -251,6 +251,24 @@ typedef enum {
    */
   JXL_ENC_OPTION_MODULAR_PREDICTOR = 27,
 
+  /** Fraction of pixels used to learn MA trees as a percentage. -1 = default,
+   * 0 = no MA and fast decode, 50 = the default, 100 = All.
+   */
+  JXL_ENC_OPTION_MODULAR_TREE_PERCENT = 28,
```
jonsneyers, line 257: These enum names are already long enough, but maybe `JXL_ENC_OPTION_MODULAR_MA_TREE_LEARNING_PERCENT` is slightly more informative.

lib/include/jxl/encode.h
```diff
@@ -251,6 +251,24 @@ typedef enum {
    */
   JXL_ENC_OPTION_MODULAR_PREDICTOR = 27,
 
+  /** Fraction of pixels used to learn MA trees as a percentage. -1 = default,
+   * 0 = no MA and fast decode, 50 = the default, 100 = All.
```
jonsneyers, line 255: Can also mention that higher values will use more encode memory.

lib/include/jxl/encode.h
```diff
@@ -251,6 +251,24 @@ typedef enum {
...
+   */
+  JXL_ENC_OPTION_MODULAR_TREE_PERCENT = 28,
+
+  /** Number of extra (previous-channel) MA tree properties to use. -1 =
+   * default, 0-11 = valid values. Recommended values are in the range 0 to 3,
+   * or 0 to amount of channels minus 1 (including all extra channels, and
+   * excluding color channels when using VarDCT mode). Higher value gives slower
+   * encoding and slower decoding.
+   */
+  JXL_ENC_OPTION_MODULAR_EXTRA_PROPERTIES = 29,
```
jonsneyers, line 265: Maybe `_NB_PREV_CHANNELS` is better. The actual number of extra properties is 4 times this number, since there are 4 extra properties per previous channel.

lib/jxl/encode.cc
```diff
@@ -606,6 +606,40 @@ JxlEncoderStatus JxlEncoderOptionsSetInteger(JxlEncoderOptions* options,
       options->values.cparams.options.predictor =
           static_cast<jxl::Predictor>(value);
       return JXL_ENC_SUCCESS;
+    case JXL_ENC_OPTION_MODULAR_TREE_PERCENT:
+      if (value < -1 || value > 100) return JXL_ENC_ERROR;
```
jonsneyers, line 610: Values above 100 (above 1.0) might be useful, since this variable also influences other things that don't get clamped.

lvandeve: Documented this, and made it unbounded. If there's a reasonable max value you're aware of, we can add that as a bound

lib/include/jxl/encode.h
lvandeve: Done

lib/include/jxl/encode.h
lvandeve: Done

lib/include/jxl/encode.h
lvandeve: nothing against the longer name in this case, done

jonsneyers (approved):

PR #855, opened 2021-11-15
Add find module for hwy

donny-dont:
The module looks for hwy using pkg-config if available otherwise it will look for the library in the `CMAKE_PREFIX_PATH`. If found a target is created that mirrors the in source build of the library.

PR #897, opened 2021-11-22
Add find module for lcms2

donny-dont:
The module looks for lcms2 using pkg-config if available otherwise it will look for the library in the CMAKE_PREFIX_PATH. If found a target is created that mirrors the in source build of the library.


donny-dont:
In https://github.com/libjxl/libjxl/pull/855 the `CMAKE_MODULE_PATH` is added to. This PR is dependent on that one landing first.

sboukortt (dismissed):

sboukortt:
Thanks!

donny-dont:
Thanks for the review @sboukortt !

In terms of merging you should order it like this
1. #864 
2. #855 
3. this one

The first PR fixes an issue where in static builds the library suffix was being overridden in Windows which prevented a successful build when using find modules. The second PR adds the cmake directory to the module search path. Then this one will build without issue when merged.

jonsneyers (approved):

PR #898, opened 2021-11-22
Add find module for brotli

donny-dont:
The module looks for brotli using pkg-config if available otherwise it will look for the library in the CMAKE_PREFIX_PATH. If found a target is created that mirrors the in source build of the library.

The brotli headers do not have versioning so a minimum version is not requested.


donny-dont:
In https://github.com/libjxl/libjxl/pull/855 the `CMAKE_MODULE_PATH` is added to. This PR is dependent on that one landing first.

eustas (approved):

jonsneyers (approved):

donny-dont:
Thank you @jonsneyers and @eustas !

PR #902, opened 2021-11-24
Support encoding compressed brob boxes

lvandeve:


lib/jxl/encode.cc
```diff
@@ -64,6 +66,42 @@ void QueueBox(JxlEncoder* enc,
   enc->input_queue.emplace_back(std::move(queued_input));
   enc->num_queued_boxes++;
 }
+
+JxlEncoderStatus BrotliCompress(int quality, const uint8_t* in, size_t in_size,
```
veluca93, line 70: Don't we do something very similar in the "jpeg/" folder? Could we refactor things to have a single implementation?

veluca93 (approved):

lib/jxl/encode.cc
lvandeve: The brotli code for JPEG data looks quite integrated within the JPEG data encoding itself, so the refactoring would be non-trivial (at least compared to the scope of this MR). I added a TODO instead

This did remind me to add msan unpoison calls, otherwise the external brotli lib causes issues with msan, PTAL

veluca93 (approved):

PR #903, opened 2021-11-24
Bake-in unpublished API.

eustas:
We still don't want it to be public, but have to "enable" it for fuzzing the wrapper libraries like libvips.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -20,27 +20,11 @@
 #include "lib/jxl/image_bundle.h"
 #include "lib/jxl/loop_filter.h"
 #include "lib/jxl/memory_manager_internal.h"
+#include "lib/jxl/sanitizers.h"
 #include "lib/jxl/toc.h"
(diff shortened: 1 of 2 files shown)


jonsneyers (approved):

lib/jxl/decode.cc
```diff
@@ -872,6 +874,15 @@ std::unique_ptr<BitReader, std::function<void(BitReader*)>> GetBitReader(
       });
 }
 
+bool CheckSizeLimit(JxlDecoder* dec, size_t xsize, size_t ysize) {
```
lvandeve, line 877: Is it necessary for this function to be visible through linking, or can it be made static or anonymous namespace?

eustas: Ooops! Fixed.

eustas:
PTAL

lvandeve (dismissed):

jonsneyers (approved):

PR #905, opened 2021-11-25
Deal with over/underflows in spline coordinate calculations.

eustas:


lib/jxl/splines.cc
```diff
@@ -129,6 +129,8 @@ void ComputeSegments(const Spline::Point& center, const float intensity,
   segment.inv_sigma = 1.0f / sigma;
   segment.sigma_over_4_times_intensity = .25f * sigma * intensity;
   segment.maximum_distance = maximum_distance;
+  // TODO(eustas): this will work incorrectly for (center.y >= 1 << 23)
+  //               we have to use double precision in that case...
   ssize_t y0 = center.y - maximum_distance + .5f;
   ssize_t y1 = center.y + maximum_distance + 1.5f;  // one-past-the-end
   for (ssize_t y = std::max<ssize_t>(y0, 0); y < y1; y++) {
```
jonsneyers, line 136: In case `y1` is for whatever reason very large, this loop can cause a lot of memory consumption.

I don't think there is any reason to keep going after y >= frame_height, right? I think it's a good idea to replace the `y < y1` with `y < std::min<ssize_t>(y1, [FRAME HEIGHT]` (which will require some plumbing to actually know the frame height here).

eustas: Overall, at some moment, we will have pipelined decoding and, hopefully, per-group spline rasterization. At that moment we need to remember to use group-origin relater coordinates, and viola - problem solved.

jonsneyers: ok, but can we for now cap `maximum_distance` and `y1` to something sensible? I think it is currently unbounded so in principle with the right input, I think you can make `y1` have arbitrary values here, which means arbitrary memory consumption.

(also `y1` could overflow causing no spline to be drawn, but I suppose that's a smaller problem)

eustas: Done. Even better - limited spline area as well.

lib/jxl/splines.cc
```diff
@@ -507,10 +549,13 @@ Status Splines::InitializeDrawCache(size_t image_xsize, size_t image_ysize,
   segment_indices_.clear();
   segment_y_start_.clear();
   std::vector<std::pair<size_t, size_t>> segments_by_y;
+  Spline spline;
+  // TODO(eustas): not in the spec; limit splines feature up to 32MPx.
+  size_t px_limit = 1 << 25;
```
jonsneyers, line 554: This limits the total rendered spline area to 32 megapixels, right?

That might be a somewhat limiting factor for jxl art :)
(e.g. you can have a 4 megapixel image with a bunch of overlapping big splines that could together maybe have a nominal area larger than 32 megapixels)

It makes sense to limit the total area though - that's the main thing that will determine memory and cpu cost to decode.

But maybe we should put a limit on it that is a bit less strict and maybe also make it depend on the image dimensions (a 100x100 image with 32MPx worth of splines is something else than a 10000x10000 images with 32 MPx worth of splines, imo).

And we should put this limit in the spec then.

I would suggest to limit the total spline area to `(1 << 16) + (xsize * ysize << 4)` or something like that (i.e. allow every pixel to be part of at most 16 overlapping splines on average).

Opinions welcome. Ping @veluca93 @lvandeve @lovell

veluca93: I like the idea of an adaptive limit

eustas: I propose just to have some preliminary limit at the moment.
Until splines are rendered group-wise size-based limit could become a CPU/memory threat.
Also not sure that this should be specced. Spline rendering should allow some quality trade-offs.

veluca93: I'm not sold of that - we have plenty of things that are not rendered group-wise and use O(image) memory, so I don't see the issue of a per-image-size limit.

jonsneyers: Quality trade-offs is one thing. Refusing to decode an image is something else. The whole point of profiles and levels is to establish some subset of the bitstream that is guaranteed to decode. If we start introducing de facto extra limits (that are actually limiting, not checks like coords being within max image dimensions range), I think it's better to be explicit about it and add it to the spec.

We can have a low limit and/or one that is not image size dependent, or one that guarantees that groups can be decoded with constant memory (so a limit of spline area per group, not per image). But if it's a limiting limit (e.g. for someone who would want to make an svg to jxl splines converter), we have to make sure the limits are well documented and described in the spec. Otherwise we're basically making a reference implementation that doesn't conform to the spec.

jyrkialakuijala: Any convergence on this?

jonsneyers (dismissed):

lib/jxl/splines.cc
jonsneyers: It looks OK to me as it is now. We'll have to see if and what exactly we put in the spec as a limitation, but for now these limits look sane to me without being restrictive.

eustas:
Rebased, please rubber-stamp.

lib/jxl/splines.cc
```diff
@@ -536,8 +586,13 @@ Status Splines::InitializeDrawCache(const size_t image_xsize,
       continue;
     }
     HWY_DYNAMIC_DISPATCH(SegmentsFromPoints)
-    (spline, points_to_draw, arc_length, segments_, segments_by_y);
+    (spline, points_to_draw, arc_length, segments_, segments_by_y, &px_limit);
+    if (px_limit == 0) {
+      // Drop the splines that does not fit the budget.
+      break;
```
sboukortt, line 592: Would it make sense to return an error instead?

sboukortt (approved):

lib/jxl/splines.cc
eustas: Going to revert to initial impl, as agreed.

sboukortt (dismissed):

eustas:
@sboukortt how did you create "spline_on_first_frame.jxl"?
(it does not fit our limits, because it is too bold...)

sboukortt:
> @sboukortt how did you create "spline_on_first_frame.jxl"? (it does not fit our limits, because it is too bold...)

I think either @jonsneyers created it using `jxl_from_tree`, or I started from his source and added the rectangle.

eustas:
Oh, do you happen to have src text file? =)

eustas:
Nevermind, will craft one on https://jxl-art.surma.technology/

sboukortt (dismissed):

eustas:
Strange. Test passes on my PC...

eustas:
Synced to TOT and got repro on my PC. I wonder, what have changed recently?

eustas:
Something in 321c0ca17e9337f0bfb36e0662306c3134fe876d

veluca93 (approved):

PR #906, opened 2021-11-25
Test fix: In conformance tests, use liblcms2.so.2 rather than liblcms…

fischbacher:
…2.so

The latter only exists on some Linux distributions.


veluca93 (approved):

veluca93:
Can you send this to libjxl/conformance too?

(eventually, this repo should use libjxl/conformance as a submodule)

PR #907, opened 2021-11-25
Add checks for codestream level features

lvandeve:


jyrkialakuijala (approved):

lib/jxl/encode.cc
```diff
@@ -117,15 +187,20 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
   // box.
 
   if (!wrote_bytes) {
+    // First time encoding any data, verify the level 5 vs level 10 settings
+    if (JXL_ENC_SUCCESS != VerifyLevelSettings(this)) {
+      return JXL_API_ERROR("Codestream level features verification failed");
```
jonsneyers, line 192: It might be more convenient to automatically bump up the level to 10 if needed, if it can still be done.

In case of multiple frames, it will in some cases be unavoidable to set the level manually in an application (since level 5 restrictions might only get violated in later frames), but in the single frame case, I _think_ it should be possible to derive the level automatically from the image and frame data.

So I'd prefer to have this something like the uses_container setting: you can optionally explicitly say you want to use a container (or set the level to 10), but if you don't do anything, it will automatically do something sensible (avoid the container if possible, stick to level 5 if possible).

Otherwise I am worried that applications might just always set the level to 10 to be sure, even if nearly always they're actually staying within level 5 limits.

lvandeve: If it automatically bumps to level 10, then people could inadvertently create a level 10 file while level 5 was intended. 

So a possible worry is that an application just always sets it to 10, but with the other behavior the worry instead is that you create a level 10 jxl file that might be incompatible without intending to.

So I'd prefer to keep as is for now. Maybe a function to check the minimum required level, given the settings, could be added in a next MR

PR #909, opened 2021-11-25
header only: proposed encoder animation API

lvandeve:
Adds function for setting the frame header containing animation
duration. Header-only, no implementation yet.

This change is open for discussion: let's find the best way to add this
to the encoder.

In this proposal, this setting is set to the JxlEncoderOptions, because
JxlEncoderOptions ara in reality frame options. The frame header
belongs to a frame, so setting its information to JxlEncoderOptions looks more
logical than setting it to JxlEncoder directly. Other opinions are
welcome.

JxlEncoderOptions is a confusing, JxlFrameOptions or similar could have
been a better name, you cannot use it to set boxes-related options to
the encoder for example. To clear this up, I also renamed it "frame_options"
wherever it was called "options" before and improved its documentation where used.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -362,6 +362,65 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
                                                     size_t* avail_out);
 
 /**
+ * Sets the frame information for this frame to the encoder. This includes
+ * animation information such as frame duration to store in the frame header.
+ * The frame header fields represent the frame as passed to the encoder, but not
+ * necessarily the exact values as they will be encoded file format: the encoder
+ * could change crop and blending options of a frame for more efficient encoding
+ * or introduce additional internal frames. Animation duration and time code
+ * information is not altered since those are immutable metadata of the frame.
+ *
+ * TODO(lode): JxlFrameHeader currently only contains animation information,
+ * also allow setting crop and blending fields.
+ *
+ * It is not required to use this function, however if have_animation is set
+ * to true in the basic info, then this function should be used to set the
+ * time duration of this individual frame. By default individual frames have a
+ * time duration of 0, making them form a composite still. See @ref
+ * JxlFrameHeader for more information.
+ *
+ * This information is stored in the JxlEncoderOptions and so is used for any
+ * frame encoded with these JxlEncoderOptions. It is ok to change between @ref
+ * JxlEncoderAddImageFrame calls, each added image frame will have the frame
+ * header that was set in the options at the time of calling
+ * JxlEncoderAddImageFrame.
+ *
+ * The is_last and name_length fields of the JxlFrameHeader are ignored, use
+ * @ref JxlEncoderCloseFrames to indicate last frame, and @ref
+ * JxlEncoderSetFrameName to indicate the name and its length instead. Calling
+ * this function will clear any name that was previously set with @ref
+ * JxlEncoderSetFrameName.
+ *
+ * @param frame_settings set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_header frame header data to set.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetFrameInfo(
+    JxlEncoderOptions* frame_settings, const JxlFrameHeader* frame_header);
+
+/**
+ * Sets the name of the animation frame. This function is optional, frames are
+ * not required to have a name. This setting is a part of the frame header, and
+ * the same principles as for @ref JxlEncoderSetFrameInfo apply. The
+ * name_length field of JxlFrameHeader is ignored by the encoder, this function
+ * determines the name length instead as the length in bytes of the C string.
+ *
+ * The maximum possible name length is 1071 bytes (excluding terminating null
+ * character).
+ *
+ * Calling @ref JxlEncoderSetFrameInfo clears any name that was previously set.
+ *
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_name name of the next frame to be encoded, as a UTF-8 encoded C
+ * string (zero terminated).
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetFrameName(
+    JxlEncoderOptions* frame_options, const char* frame_name);
+
+/**
  * Sets the buffer to read JPEG encoded bytes from for the next frame to encode.
  *
  * If JxlEncoderSetBasicInfo has not yet been called, calling
@@ -724,6 +783,8 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelInfo(
 /**
  * Sets the name for the extra channel at the given index in UTF-8. The index
  * must be smaller than the num_extra_channels in the associated JxlBasicInfo.
+ *
+ * TODO(lode): remove size parameter for consistency with JxlEncoderSetFrameName
  *
  * @param enc encoder object
  * @param index index of the extra channel to set.
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -369,11 +369,16 @@ typedef struct {
   uint32_t timecode;
 
   /** Length of the frame name in bytes, or 0 if no name.
-   * Excludes null termination character.
+   * Excludes null termination character. This value is set by the decoder.
+   * For the encoder, this value is ignored and @ref JxlEncoderSetFrameName is
+   * used instead to set the name and the length.
    */
   uint32_t name_length;
 
-  /** Indicates this is the last animation frame.
+  /** Indicates this is the last animation frame. This value is set by the
+   * decoder to indicate no further frames follow. For the encoder, it is not
+   * required to set this value and it is ignored, @ref JxlEncoderCloseFrames is
+   * used to indicate the last frame to the encoder instead.
    */
   JXL_BOOL is_last;
 


lib/include/jxl/encode.h
```diff
@@ -41,8 +41,10 @@ JXL_EXPORT uint32_t JxlEncoderVersion(void);
 typedef struct JxlEncoderStruct JxlEncoder;
 
 /**
- * Opaque structure that holds frame specific encoding options for a JPEG XL
- * encoder.
+ * The options for a single image frame to be encoded with the JXL encoder.
+ * This includes options such as compression speed, as well as frame metadata.
+ *
+ * Despite the name JxlEncoderOptions, it represents options of a frame instead.
```
jonsneyers, line 47: Maybe it would make sense to rename it (to `JxlEncoderFrameSettings` or something) and for now keep the old name as a deprecated synonym (that maybe we can get rid of when we finalize the api).

I think "EncoderOptions" is misleading if it doesn't just include encode options (like effort and quality target) but also actual frame data (like duration and crop dimensions/offset). The word "Settings" is similar to "Options" but implies that it can also be data, not just optional options. 

So I would use `frame_settings` instead of `frame_options` throughout, and also rename this struct to `JxlEncoderFrameSettings` with for now a deprecated synonym `JxlEncoderOptions` in order to not break existing encoders.

lib/include/jxl/encode.h
```diff
@@ -361,14 +363,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ * this function will clear any name that was previously set with @ref
+ * JxlEncoderSetFrameName.
+ *
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_header frame header data to set.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetFrameHeader(
+    const JxlEncoderOptions* frame_options, const JxlFrameHeader* frame_header);
```
jonsneyers, line 403: Shouldn't `frame_options` (which I would rename to `frame_settings`) be non-`const` here?
I guess the question is whether the `frame_header` info gets stored in the `enc` struct or in the `frame_options` struct. It feels a bit strange to have a `Set` function that only gets `const` arguments...

lib/include/jxl/encode.h
```diff
@@ -361,14 +363,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ *
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_name name of the frame to set.
+ * @param name_length length of the frame name, excluding null-termination char.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
+ */
+JXL_EXPORT JxlEncoderStatus
+JxlEncoderSetFrameName(const JxlEncoderOptions* frame_options,
+                       const char* frame_name, size_t name_length);
```
jonsneyers, line 423: This allows frame names to contain null chars, which might be useful and is technically allowed by the spec, but it might be better/easier to just use a C string without explicit length, so it cannot have nulls in the middle. The spec also says that the name is to be interpreted as a UTF-8 string, which kind of implies that a null byte terminates it since UTF-8 does not have multi-byte sequences containing null bytes, allowing them to be C strings.

In any case it would be good to add to the documentation that the maximum length (in bytes) of the frame_name (excluding null at the end) is 1071 bytes (largest size that can be represented is 48 + u(10)).

lib/include/jxl/encode.h
```diff
@@ -812,16 +867,18 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetCodestreamLevel(JxlEncoder* enc,
...
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
  * @param lossless whether to override options for lossless mode
  * @return JXL_ENC_SUCCESS if the operation was successful, JXL_ENC_ERROR
  * otherwise.
  */
-JXL_EXPORT JxlEncoderStatus
-JxlEncoderOptionsSetLossless(JxlEncoderOptions* options, JXL_BOOL lossless);
+JXL_EXPORT JxlEncoderStatus JxlEncoderOptionsSetLossless(
+    JxlEncoderOptions* frame_options, JXL_BOOL lossless);
```
jonsneyers, line 877: Not really related to this PR, but this one is confusing and has lead to applications getting things wrong. It only sets frame options to lossless, while the image options (in particular: using xyb or not) are still lossy by default.

By having this function like this, we give the false impression that it is possible to use lossless settings after the image header has already been written and been set to use XYB.

I think it might be better to deprecate this function and instead have a global encoder object flag for lossless that has to be set when setting the basicinfo. If you want to do multi-frame images with both vardct and 'lossless' XYB, you can do that manually.

lib/jxl/encode.cc
```diff
@@ -884,24 +902,24 @@ JxlEncoderStatus JxlEncoderAddImageFrame(const JxlEncoderOptions* options,
...
-    c_current = options->enc->metadata.m.color_encoding;
+    c_current = frame_options->enc->metadata.m.color_encoding;
   }
 
-  size_t xsize = options->enc->metadata.xsize();
-  size_t ysize = options->enc->metadata.ysize();
-  if (options->values.cparams.already_downsampled) {
-    size_t factor = options->values.cparams.resampling;
+  size_t xsize = frame_options->enc->metadata.xsize();
+  size_t ysize = frame_options->enc->metadata.ysize();
```
jonsneyers, line 909: Here we would then use the frame dimensions, not the image dimensions, and also set the frame header fields (offset, duration, blend mode etc), right?

lib/include/jxl/encode.h
```diff
@@ -361,14 +363,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
  * are copied internally.
  * @param size size of buffer in bytes.
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus JxlEncoderAddJPEGFrame(
-    const JxlEncoderOptions* options, const uint8_t* buffer, size_t size);
+    const JxlEncoderOptions* frame_options, const uint8_t* buffer, size_t size);
+
+/**
+ * Sets the frame header with animation information for this frame.
```
jonsneyers, line 377: It's not just animation information, also crop and blend info.

Not needed now, but it would eventually be good to have an encode option to allow the encoder to modify the crop and blend info that was given. I.e. what you set here, is how you pass the frame to the encoder (e.g. full-sized frame with kReplace blending), but is not necessarily how it actually gets encoded (e.g. the encoder might decide that it is better to use a cropbox instead and kAdd blending instead, for the particular pixels you gave it).

For now (since we do not have an implementation that does such things) how you pass the frame data and how it gets encoded would be the same, but it might be wise to already anticipate in the documentation that at some point this might no longer necessarily be the case, if we would add an encode option that allows e.g. frame trimming. So this function would not necessarily be one that "Sets the frame header", but rather it "specifies how the frame gets passed to the encoder". Maybe the actual frame header and what was passed will be the same thing (in case, say, `JxlEncoderSetMetaLiberty(JXL_STRICT)`), maybe it will not be (in case, say, `JxlEncoderSetMetaLiberty(JXL_ALLOW_TRIM | JXL_ALLOW_DIFFERENT_BLEND)`).

A similar concept of "how it is passed might be different from how it is encoded" might by the way be useful to allow dropping alpha channels that turn out to be fully opaque or to store an image that was passed as RGB pixels as Grayscale if R=G=B for every pixel (probably only feasible for the single-frame case but still potentially useful).

lib/include/jxl/encode.h
```diff
@@ -361,14 +363,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ * JxlEncoderSetFrameName to indicate the name and its length instead. Calling
+ * this function will clear any name that was previously set with @ref
+ * JxlEncoderSetFrameName.
+ *
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_header frame header data to set.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetFrameHeader(
```
jonsneyers, line 402: See my comment above. It might be better to call this `JxlEncoderSetFrameInfo` or something like that, to leave open the possibility that what you're setting here is not necessarily exactly what will be written in the bitstream frameheader, but rather is how you will present the frame to the encoder, which can then choose to encode it just like that, or maybe to make some changes like trimming unchanged parts.

lib/include/jxl/encode.h
lvandeve: This function only changes frame options, that are written in the frame header. AFAIK, it is possible to write both lossless and lossy frames, or any mix thereof, after the image header was written. Which setting of the image header do you mean?

lib/include/jxl/encode.h
lvandeve: Another option is to keep JxlEncoderOptions as-is (though better documented), and not use them for the frame metadata purpose. But then have a JxlEncoderSetFrameHeader that takes an encoder object, instead of a JxlEncoderOptions

But it would then be odd that JxlEncoderAddImageFrame is done with a JxlEncoderOptions, while JxlEncoderSetFrameHeader that sets the header of that frame would be done with JxlEncoder instead.

About renaming: that would break backwards compatibility of course.

lib/include/jxl/encode.h
jonsneyers: Uses_xyb is part of the image header. If it is true, then no frames can be really lossless, since all frames are in XYB then.

lib/include/jxl/encode.h
lvandeve: The frame_settings rename was done on another MR, this comment is obsolete

lib/jxl/encode.cc
lvandeve: this change in encode.cc was unintended as it's a header-only change, this comment is now obsolete

lib/include/jxl/encode.h
lvandeve: Let's handle this in a different MR if needed, maybe it should indeed become a global encoding setting, also need to check why the unit tests with JxlEncoderOptionsSetLossless are passing, given that they are not touching the xyb_encoded field...

But that's unrelated to this MR, so resolving this comment

lib/include/jxl/encode.h
lvandeve: Yes indeed it probably should, just like JxlEncoderOptionsSetInteger, done

lib/include/jxl/encode.h
lvandeve: Added the comment about 1071 bytes

Also removed the size parameter. For consistency we should also remove it from JxlEncoderSetExtraChannelName (in a different MR, added a TODO in this one)

I actually see advantages either way, giving the size has as advantage that you can leave out the null termination character which can be needed if you are getting the string from some encoded format in memory where it's not present. But for this MR I'll do it without name_length, a later MR that makes JxlEncoderSetFrameName and JxlEncoderSetExtraChannelName consistent can discuss this issue instead

lib/include/jxl/encode.h
lvandeve: Added a TODO for blend and crop info, that's indeed not for this MR. Also documented the proposed encoder behavior

lib/include/jxl/encode.h
lvandeve: Makes sense, renamed to JxlEncoderSetFrameInfo

lvandeve:
Did parts of this PR in https://github.com/libjxl/libjxl/pull/922 and https://github.com/libjxl/libjxl/pull/921, now only the addition of the new API functions in the header is done in this PR, please take another look

lib/include/jxl/encode.h
```diff
@@ -361,6 +361,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetFrameInfo(
+    JxlEncoderOptions* frame_settings, const JxlFrameHeader* frame_header);
+
+/**
+ * Sets the name of the animation frame. This function is optional, frames are
+ * not required to have a name. This setting is a part of the frame header, and
+ * the same principles as for @ref JxlEncoderSetFrameInfo apply. The
+ * name_length field of JxlFrameHeader is ignored by the encoder, this function
+ * determines the name length instead as the length of the C string in UTF-8.
```
jonsneyers, line 407:
```suggestion
 * determines the name length instead as the length in bytes of the C string.
```

lib/include/jxl/encode.h
```diff
@@ -361,6 +361,64 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ *
+ * The maximum possible name length is 1071 bytes (excluding terminating null
+ * character).
+ *
+ * Calling @ref JxlEncoderSetFrameInfo clears any name that was previously set.
+ *
+ * @param frame_options set of options and metadata for this frame. Also
+ * includes reference to the encoder object.
+ * @param frame_name name of the frame to set.
```
jonsneyers, line 416:
```suggestion
 * @param frame_name name of the next frame to be encoded, as a UTF-8 encoded C string (zero terminated).
```

lvandeve:
Suggestions applied

lvandeve:
Applying the suggestions through the github interface causes the authors check to fail, let's see how this can be fixed...

lvandeve:
Yay, the authors check issue is fixed and the suggestions now applied by amending the original commit, PTAL

lib/include/jxl/encode.h
```diff
@@ -361,6 +361,65 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
...
+ * Sets the frame information for this frame to the encoder. This includes
+ * animation information such as frame duration to store in the frame header.
+ * The frame header fields represent the frame as passed to the encoder, but not
+ * necessarily the exact values as they will be encoded file format: the encoder
+ * could change crop and blending options of a frame for more efficient encoding
+ * or introduce additional internal frames. Animation duration and time code
+ * information is not altered since those are immutable metadata of the frame.
+ *
+ * TODO(lode): JxlFrameHeader currently only contains animation information,
+ * also allow setting crop and blending fields.
```
jonsneyers, line 374: Actually since https://github.com/libjxl/libjxl/pull/831 it does already contain `JxlLayerInfo` which has crop and blend info. So I think this TODO can be dropped (though it still needs to be implemented to actually take that info into account in the encoder). Can remove the TODO later too.

jonsneyers (approved):

lib/include/jxl/encode.h
lvandeve: Will remove it later :)

Issue #910, opened 2021-11-25
CI: Cross-compiling i686-linux-gnu job is broken


eustas:
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libqt5x11extras5-dev:i386 : Depends: libqt5x11extras5:i386 (= 5.12.8-0ubuntu1) but it is not going to be installed
 qtbase5-dev:i386 : Depends: libqt5concurrent5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5core5a:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5dbus5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5gui5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5network5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5printsupport5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5sql5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5test5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5widgets5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Depends: libqt5xml5:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
                    Recommends: libqt5opengl5-dev:i386 (= 5.12.8+dfsg-0ubuntu1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

Going to "disable" this job for a while to unblock commit queue.

eustas:
```
 libpcre2-8-0 : Breaks: libpcre2-8-0:i386 (!= 10.39-1+ubuntu20.04.1+deb.sury.org+1) but 10.34-7 is to be installed
 libpcre2-8-0:i386 : Breaks: libpcre2-8-0 (!= 10.34-7) but 10.39-1+ubuntu20.04.1+deb.sury.org+1 is to be installed
```

PR #911, opened 2021-11-25
Unblock commit queue.

eustas:
See #910.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -318,8 +318,8 @@ jobs:
           libgtk2.0-dev
 
           # QT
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

eustas:
Added a real fix. PTAL

veluca93:
> Added a real fix. PTAL

Ehhh, this is not really a real fix - it just disables cross-compiling things that depends on qt5...

lvandeve (approved):

veluca93:
Let's maybe put a TODO to once again test viewers on i386/armhf at some point?

eustas:
Oh, indeed, it reports "Qt5 was not found."
I'll put back the todo.

PR #913, opened 2021-11-25
Header only: decoder progressive event api

lvandeve:


lib/include/jxl/decode.h
```diff
@@ -1137,6 +1147,20 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxType(JxlDecoder* dec,
...
+ * values mean: 0 = only trigger for the DC image, the 8x8th lower resolution
+ * image. 1 = also trigger when a full pass of groups is ready. Higher values
+ * indicate more steps but are not yet implemented. Higher values always include
+ * the events of lower values as well.
+ *
+ * @param dec decoder object
+ * @param detail at which level of detail to trigger JXL_DEC_FRAME_PROGRESSION
+ */
+JXL_EXPORT void JxlDecoderSetProgressiveDetail(JxlDecoder* dec,
+                                               uint32_t detail);
```
jonsneyers, line 1162: Not sure if this should be seen as an integer or as a bitfield. If we add more things like intermediate stages of DC (e.g. in the dc frame case), or stages of a pass (e.g. for the center-first case), then it might make sense to 'subscribe' to some detail but not to other (e.g. show dc previews but not stages of a pass, or the other way around).

jonsneyers: I can see the following steps (eventually) being exposed:

- preview frame (already exposed separately)
- first DC preview (e.g. could be a 'blurhash' type image, e.g. 16x16 pixels worth of data, upsampled a lot)
- 1:32 image (only available if DC frame is used which itself is progressive)
- 1:16 image (same)
- 1:8 image (always available in VarDCT mode)
- 1:4 image (only available if AC progression is used or lossy modular)
- 1:2 image (same)
- orthogonally to resolution steps:
  - DC group decoded (so can show DC preview of a region)
  - AC group decoded
  - full pass of groups decoded (that not necessarily corresponds to a full resolution, e.g. there could be a 1:1 pass that is not the final pass)
- also orthogonal: new layer available in the non-`kSkipProgressive` case

I can see use cases for subscribing to arbitrary subsets of these events.

lvandeve: The reason for not making it a bit flag is to not expose these underlying details to the API so that it's possible to have variability in the implementation.

It's a question of:

Do we want to support and allow control over every possible progressive step there exists, even if this means that depending on the type of image (modular, vardct, ...) settings can behave very differently. This also requires the user to have a test set of different images with different encodings to check if they get roughly as many progressive steps as desired on the different types (e.g. if a user would only test vardct images, they may completely forget to enable lossy modular progression)

Or do we want to create a way to use the API to get "less" or "more" progressive steps, and let the implementation use the best way to guarantee that detail setting X gives a certain expected amount of steps (e.g. 1 step, DC and full, for detail setting 1)

I'm more in favor of the latter approach, and thus to keep the step-based detail parameter, not individual flags

jonsneyers: I agree that we don't necessarily have to expose super fine grained control over progressive detail.

It's just that there might be two relatively different use cases for progressive decoding:

- showing a preview while the data is still loading, but the final image is full resolution/quality
- stop decoding as soon as "enough" detail is available, e.g. if there's a progressive step equivalent to the detail of a 1:2 image, and you're rendering a 20 megapixel image on a 2 megapixel screen, then you could stop/pause decoding at that point (maybe resume if the user decides to zoom in)

I'm not sure if "expected amount of steps" captures that. E.g. I can imagine that in a browser, you might want to do pre-DC (blurhash-like placeholder), DC and full, while in a local image viewer, you might want to do e.g. single-step 1:2 decoding.

lvandeve: The preview can indeed be left out of this, it's its own frame that could be independent from any image frames in theory

New layer I'd also leave out for now, in general this setting is about lower grained versions of the final image, not about versions where new data from a next composite still frame gets added

So from the list above there are then 2 types of progressive:

-steps where the entire image becomes available at higher resolution: anything from 1:32 (or 1:64 if that can be done by DC frames) to 1:8, and then only for modular also 1:4 and 1:2

-steps involving groups, where part of the image gets updated. However, when a full pass is ready, that's also a full image, so you could see this as similar to the 1:2 and 1:4 case for modular

So maybe the following values for detail could capture these well, where each next value also includes the previous values:

```
0: only DC
1: DC and any full image step above: this is 1:2 and 1:4 for modular, and any full pass for VarDCT
2: any full image step, including 1:32 and higher
3: add any AC group that's ready
4: add any DC group that's ready
```

Alternatively, if a bit mask, then one that is not too fine grained, e.g.

```
1: resolution steps lower than DC (1:32, 1:16, 1:64 if this exists)
2: DC (1:8)
4: resolution steps higher than DC (1:2, 1:4, as well as vardct passes)
8: AC groups
16: DC groups
```

Does this make sense?

jonsneyers: 1:4 and 1:2 are not just for modular, also VarDCT can do that using progressive AC passes that only send the corresponding low-freq AC; there's bitstream syntax to describe which pass will give what resolution.

I suppose in any case we'll need some way to inform the application what amount of detail is actually available when a JXL_DEC_FRAME_PROGRESSION event happens. So assuming we'll have something for that, I think it's fine to just use your first proposal:

```
0: only (full) DC
1: DC and any full image step above: this is 1:2 and 1:4 for modular, and any full pass for VarDCT
2: any full image step, including 1:32 and higher
3: add any AC group that's ready
4: add any DC group that's ready
```

and then the application can figure out itself what it wants to do (Flush or not, stop decoding or not) depending on whether it has what it wants to have.

For a way for applications to know what is currently available: I'm not sure what the best way to do that would be; maybe something like this: `uint8_t JxlDecoderGetProgressiveStatus(dec,x0,y0,w,h)`: if it returns the value `k`, that means that in the rect given by x0,y0,w,h, all pixels have an amount of detail corresponding to (at least) a 1:2^`k` version of the image. Let's say with a convenience case where you can use 0,0,0,0 as a shorthand for "the whole image", which is probably a common thing you'll want to know.
That way, you could e.g. watch the rect that is currently in the viewport, and do a flush+repaint iff its detail improves (decreases), and ignore other JXL_DEC_FRAME_PROGRESSION events.

jonsneyers: Anyway, to conclude: `uint32_t detail` is fine :)

jonsneyers (approved):

Issue #915, opened 2021-11-26
Crash when trying to save.


sati32:
JxlEncoderStatus JxlEncoderOptionsSetDistance(JxlEncoderOptions* options, float distance)

there is a line:
  if (distance < 6.56f) {
    jpeg_quality = -5.456783f * std::log(0.0256f * distance - 0.16384f);
  } else {
    jpeg_quality = -11.11111f * distance + 101.11111f;
  }
this generates a crash, because log of negative values is not possible.
It should be
  if (distance > 6.56f) {

jonsneyers:
Thanks for reporting!

PR #916, opened 2021-11-26
fix JxlEncoderOptionsSetDistance

jonsneyers:
fixes https://github.com/libjxl/libjxl/issues/915

Also fixes the case where you set distance to 0.0001 or some other too-small-but-not-zero value.


lvandeve (approved):
Thanks!

PR #920, opened 2021-11-29
Take a JxlCmsInterface explicitly at the internal C++ library boundary

sboukortt:
What follows is to expose a way to override it from the public C API.


jonsneyers:
WASM build is failing for some reason...

mo271 (approved):

PR #921, opened 2021-11-29
Make clear JxlEncoderOptions are frame settings

lvandeve:
Splitting this off from https://github.com/libjxl/libjxl/pull/909

Rename the argument to make more obvious that it are frame settings,
not settings for the entire encoder

This is a backwards compatible change. If we decide to rename the struct
itself, that's not intended for this PR.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -41,8 +41,12 @@ JXL_EXPORT uint32_t JxlEncoderVersion(void);
 typedef struct JxlEncoderStruct JxlEncoder;
 
 /**
- * Opaque structure that holds frame specific encoding options for a JPEG XL
- * encoder.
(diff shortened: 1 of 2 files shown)


jonsneyers (approved):

jonsneyers:
OK for me. I would also rename the struct itself and for now keep the old struct name around as a `typedef` synonym (in order to not break backwards compatibility) but mark the old name as deprecated so applications can already start using the new name. Then maybe in 1.0 we can drop the old name. I prefer to have a clean 1.0 API, not something that feels like it is already carrying the weight of legacy. Documentation like "Despite the name JxlEncoderOptions, it represents settings of a frame" is not so nice in that respect.

lvandeve:
I agree with that, let's fix it in a next MR

PR #922, opened 2021-11-29
Write animation in basic info in encoder

lvandeve:
This writes the animation settings to the main image header. This is
part of what is needed to fully support animation, but does not yet
allow setting the time of each individual frame.

This change inadvertedly ended up in the header-only pull request https://github.com/libjxl/libjxl/pull/909

Implemented in a separate pull request now, and added test for this (and
most other basic info as well)


lvandeve:
This solves part of https://github.com/libjxl/libjxl/pull/509

jonsneyers (approved):

PR #923, opened 2021-11-29
use libpng for all png decoding

jonsneyers:
Currently, cjxl/djxl are using both libpng and lodepng: lodepng is used for decoding/encoding still PNGs, while libpng is used for decoding APNG.

To simplify things, we'll try to use libpng for everything. We're choosing libpng over lodepng because for distributions like Debian, libpng will already be available and is a better choice in terms of security and packaging considerations.

This PR is a first step in that direction. It makes the APNG decoding more complete: it adds support for ICC/gAMA/cHRM/sRGB, Exif/XMP metadata, and 16-bit images. This makes the code path using lodepng to decode PNG files redundant, so it is removed. This also makes it possible to correctly load APNG animations with a color profile, metadata, or 16 bits per sample (something that didn't work correctly before).

Next step will be to implement libpng encoding, so we can remove the dependency on lodepng completely. That will be for a different PR though.


lib/extras/codec_apng.cc
```diff
@@ -187,6 +369,15 @@ int processing_finish(png_structp png_ptr, png_infop info_ptr) {
   }
 
   png_process_data(png_ptr, info_ptr, footer, 12);
+  // before destroying: check if we encountered any metadata chunks along the
+  // way
+  png_textp text_ptr;
+  int num_text;
+  png_get_text(png_ptr, info_ptr, &text_ptr, &num_text);
+  for (int i = 0; i < num_text; i++) {
+    (void)BlobsReaderPNG::Decode(text_ptr[i], metadata);
```
lvandeve, line 378: Perhaps the return value could be used, it was used like this:

```
  if (!BlobsReaderPNG::Decode(state.s.info_png, &ppf->metadata)) {
    JXL_WARNING("PNG metadata may be incomplete");
  }
```

lib/extras/codec_apng.cc
```diff
@@ -380,23 +585,69 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
             delete[] chunk.p;
             break;
           }
+        } else if (id == kId_iCCP) {
+          if (processing_data(png_ptr, info_ptr, chunk.p, chunk.size)) {
+            JXL_WARNING("Corrupt iCCP chunk");
+            delete[] chunk.p;
```
lvandeve, line 591: Not for this MR since this was already used before, but it could be a good idea to use RAII for these resources

lvandeve (approved):

lib/extras/codec_apng.cc
veluca93: https://en.cppreference.com/w/cpp/language/raii#:~:text=Resource%20Acquisition%20Is%20Initialization%20or,in%20limited%20supply)%20to%20the

(i.e. have destructors free the memory)

jonsneyers: Ah, yes, that is indeed a nicer way of doing it. I didn't know that acronym, and I read it as camelcase RALL (Remove All?), which confused me. Sans serif fonts have that downside...

petrdiblik: IMHO Homo sapiens should have improved their fonts already to avoid these ambiguities. But that's for another discussion… :-)

lib/extras/codec_apng.cc
jonsneyers: Yes, but that function always returned `true` before, so that warning was never actually shown.

I changed the function to return `false` if it encounters some text chunk it doesn't understand, but that happens quite often it seems, so it results in a lot of warnings (e.g. png files often have simple text comments containing creation date, modify date, software used for editing, and things like that).

jonsneyers:
A side effect of this is that the wasm build of cjxl/benchmark_xl will no longer have png support (since we don't look for shared libraries in that build). I don't know if that's a problem or not.

sboukortt:
The new behavior seems a bit different for a few PNGs from http://www.schaik.com/pngsuite/, for example `basi3p01.png`:

basi3p01

It used to roundtrip properly through `cjxl -m`:

basi3p01

But apparently not quite anymore:

basi3p01

jonsneyers:
Yes, that one is an easy fix (was incorrectly setting bitdepth in case of palette).

There's another one that causes trouble with the chunk pointer freeing, I might end up replacing that with something without `new` and `delete`.

jonsneyers:
I have identified a problem in case the tRNS chunk is used to add alpha to an image that doesn't have alpha according to the IHDR. I'll fix it tomorrow, but for now do not merge this, it can segfault on such images (allocates a buffer without alpha and libpng writes too much in it).

jonsneyers:
OK, fixed a bunch of stuff (that was probably broken already, but we just never noticed since that code was only used in the animated case, as the lodepng code was tried first). Now all of the non-corrupt examples from PngSuite roundtrip OK (except for the ones with a weird gamma of 2.5, which is not supported by jxl and also didn't work before).

Also got rid of the ugly `delete[]` stuff, and added interpretation of the `sBIT` chunk, so we can properly encode e.g. 10-bit PNGs as 10-bit jxl files, if the PNG contains this optional chunk to indicate that only 10 bits are significant.

PTAL

lvandeve (approved):

PR #925, opened 2021-11-29
Use FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION to disable ANS checksums.

veluca93:
Note that this does not enable the flag yet.

### DIFF
--- a/lib/jxl/dec_ans.h
+++ b/lib/jxl/dec_ans.h
@@ -235,7 +235,11 @@ class ANSSymbolReader {
     return ReadSymbolWithoutRefill(histo_idx, br);
   }
 
-  bool CheckANSFinalState() { return state_ == (ANS_SIGNATURE << 16u); }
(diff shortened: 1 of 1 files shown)


lib/jxl/dec_ans.h
```diff
@@ -235,7 +235,11 @@ class ANSSymbolReader {
     return ReadSymbolWithoutRefill(histo_idx, br);
   }
 
-  bool CheckANSFinalState() { return state_ == (ANS_SIGNATURE << 16u); }
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
sboukortt, line 238: `#ifndef ... #else` is a bit surprising, I read it the wrong way around the first time. I wonder if it wouldn’t be clearer with `#ifdef`, but then maybe it’s just me.

sboukortt (approved):

PR #926, opened 2021-11-30
Add a function to the C API to allow replacing the default CMS

sboukortt:


lib/include/jxl/encode.h
```diff
@@ -311,6 +311,18 @@ JXL_EXPORT void JxlEncoderReset(JxlEncoder* enc);
...
+ * Sets the color management system (CMS) that will be used for color conversion
+ * (if applicable) during encoding. May only be set before starting encoding. If
+ * left unset, the default CMS implementation will be used.
+ *
+ * @param enc encoder object.
+ * @param cms structure representing a CMS implementation. See JxlCmsInterface
+ * for more details.
+ */
+JXL_EXPORT void JxlEncoderSetColorManagementSystem(JxlEncoder* enc,
+                                                   JxlCmsInterface cms);
```
jonsneyers, line 324: Maybe call it just `JxlEncoderSetCMS`? I don't mind the longer name but I don't think it's really useful to spell it out like that either. Maybe `JxlEncoderSetCmsInterface` is also an option, that makes it more consistent with `JxlCmsInterface`.

sboukortt: Fair point, how is this?

jonsneyers (approved):

lib/include/jxl/encode.h
jonsneyers: LGTM.

PR #928, opened 2021-11-30
Use unaligned store in SingleToSingle

eustas:
In ModularImageToDecodedRect group origin is right shifted...
So alignement might become too small for the current architecture.


jonsneyers:
Have you observed a case where this happens, or is it a theoretical concern?

If it's only a theoretical concern, then I would keep the stores aligned and add a JXL_ASSERT somewhere to make sure that the x0 is actually aligned.

If I'm not mistaken, the maximum right shift should be 3 (I think that's the largest dim_shift we allow). The smallest group size is 128x128. So I think in principle the offset should be aligned to 128>>3 pixels, which is 16*4 bytes. That should still be just OK for avx-512, no?

If unaligned stores can happen here, it would be good to understand how/why. If this only happens in exotic and probably not very useful cases, it might be worthwhile to reject such cases (and update the spec to not allow it).

I don't know what the speed cost is of changing the aligned operations to unaligned ones, probably small in the bigger picture, but still, might be good to check if you can measure a speed difference in say `jxl:m:2` decoding.

eustas:
The case is discovered by oss-fuzz. In the testcase x0 = 768 >> 6 = 12. I'll attach transforms dump soon.

jonsneyers:
Question is if we want to support a dim_shift of 6 (or > 3 in general). If we're planning to make that work, then yes we should use unaligned stores. If we're not planning to support more than 8x upsampling (so max dim_shift 3) we should probably fail early on such inputs and still use aligned stores.

@veluca93 any opinions on this?

veluca93:
> Question is if we want to support a dim_shift of 6 (or > 3 in general). If we're planning to make that work, then yes we should use unaligned stores. If we're not planning to support more than 8x upsampling (so max dim_shift 3) we should probably fail early on such inputs and still use aligned stores.
> 
> @veluca93 any opinions on this?

I'd avoid anything >3 and I wonder how we ended up with that value... That ought not to be possible AFAIU.

eustas:
`maximum_distance` could be wildly large...

jonsneyers:
> > Question is if we want to support a dim_shift of 6 (or > 3 in general). If we're planning to make that work, then yes we should use unaligned stores. If we're not planning to support more than 8x upsampling (so max dim_shift 3) we should probably fail early on such inputs and still use aligned stores.
> > @veluca93 any opinions on this?
> 
> I'd avoid anything >3 and I wonder how we ended up with that value... That ought not to be possible AFAIU.

Well in the ExtraChannelInfo spec we have `U32(0, 3, 4, 1 + u(3))` to encode the `dim_shift`, so certainly values > 3 are possible (and one of them even has a short code).

I have no strong preference on this: I don't think limiting it to 3 is very limiting in practice, and if that helps to keep the implementation simpler/faster, I think it's a reasonable thing to do. We might then want to change the spec to `U32(0, 3, 1, 2)` then though (that shouldn't break any existing files, afaik), or at least constrain the value to be <= 3.

There's also `ec_upsampling` which is cumulative with the `dim_shift`; I suppose we want to restrict `ec_upsampling << dim_shift` to be at most 8, then?

veluca93:
> > > Question is if we want to support a dim_shift of 6 (or > 3 in general). If we're planning to make that work, then yes we should use unaligned stores. If we're not planning to support more than 8x upsampling (so max dim_shift 3) we should probably fail early on such inputs and still use aligned stores.
> > > @veluca93 any opinions on this?
> > 
> > 
> > I'd avoid anything >3 and I wonder how we ended up with that value... That ought not to be possible AFAIU.
> 
> Well in the ExtraChannelInfo spec we have `U32(0, 3, 4, 1 + u(3))` to encode the `dim_shift`, so certainly values > 3 are possible (and one of them even has a short code).
> 
> I have no strong preference on this: I don't think limiting it to 3 is very limiting in practice, and if that helps to keep the implementation simpler/faster, I think it's a reasonable thing to do. We might then want to change the spec to `U32(0, 3, 1, 2)` then though (that shouldn't break any existing files, afaik), or at least constrain the value to be <= 3.
> 
> There's also `ec_upsampling` which is cumulative with the `dim_shift`; I suppose we want to restrict `ec_upsampling << dim_shift` to be at most 8, then?

I mean in the sense that AFAIU we error out whenever we hit >3 :)

jonsneyers:
Ah right, we are already checking that. Then I wonder how we can get a value of 6 there.

There might be a way to reach 4 atm: if the color `upsampling` is 8, and YCbCr with chroma subsampling is used in modular mode, it could be 4. I think it would be reasonable to not allow that (only allow chroma subsampling if the `upsampling <= 4`).

I have no clue how we're getting to 6 there, so it would be good to take a look at that case and see what case we missed, since it's likely a bug.

eustas:
As promised going to provide more info soon (asap)

eustas:
It seems that this is caused by the same problem as https://github.com/libjxl/libjxl/pull/936
At least, it is fixed by that PR. (i.e. outrageous shift is just not undone squeeze).

jonsneyers:
Maybe in that case just add a check that shift is <= 3 at that point, and maybe also that there are no not-undone transforms.

eustas:
PTAL

lib/jxl/dec_modular.cc
```diff
@@ -470,6 +470,7 @@ Status ModularFrameDecoder::ModularImageToDecodedRect(
     return true;
   }
   JXL_DASSERT(rect.IsInside(decoded));
+  JXL_DASSERT(gi.transform.empty());
```
jonsneyers, line 473: Please make these regular JXL_ASSERT instead of JXL_DASSERT, or even JXL_CHECK. There's no reason to only have these safety checks in msan/asan/debug builds but not in release builds; the impact on speed and binary size will be very close to zero and the improvement in security is worth it, imo.

eustas: Done.

jonsneyers (approved):

PR #929, opened 2021-11-30
apng output

jonsneyers:
Builds on https://github.com/libjxl/libjxl/pull/923

Removes the lodepng-based `codec_png` implementation and replaces it with a libpng implementation (`codec_apng`). This also does animated PNG encoding. Updated `djxl` to produce animated png output instead of separate frames if it is decoding an animation.


ziemek99:
Is there anything else left that blocks removing lodepng from `third_party/`?

jonsneyers:
> Is there anything else left that blocks removing lodepng from `third_party/` and `deps.sh`?

Probably not. Once the PR this one depends on has landed, I'll rebase this one and try removing lodepng.

lib/extras/codec_apng.cc
```diff
@@ -347,7 +350,7 @@ int processing_data(png_structp png_ptr, png_infop info_ptr, unsigned char* p,
 
   if (setjmp(png_jmpbuf(png_ptr))) {
     png_destroy_read_struct(&png_ptr, &info_ptr, 0);
-    return 1;
+    recounturn 1;
```
ziemek99, line 353: Are you sure it's not a typo? What is it supposed to do?

jonsneyers: it was a typo :)

lib/extras/codec_apng.cc
```diff
@@ -658,5 +661,218 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
...
+
+    png_ptr = png_create_write_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
+
+    if (!png_ptr) return JXL_FAILURE("Could not init png encoder");
+
+    info_ptr = png_create_info_struct(png_ptr);
+    if (!info_ptr) return JXL_FAILURE("Could not init png info struct");
+
+#ifndef __GNUC__
+    // gcc produces spurious "variable might be clobbered by ‘longjmp’" errors
```
veluca93, line 753: I'm not sure this is a spurious warning - we have types with nontrivial destructors on the same stack frame as where we setjmp, which is *never* a good idea...

veluca93:
How well tested is this?

lib/extras/codec_apng.cc
jonsneyers: Right. Maybe we can just skip all the setjmp setting in codec_apng.cc (both for encode and decode), which makes libpng abort on error instead of us catching it and returning failure, but the code is not part of libjxl, only used in tools, so maybe that's OK?

jonsneyers:
> How well tested is this?

I'll do a roundtrip test on the pngsuite test set to see if that catches any problems.

lib/extras/codec_apng.cc
veluca93: Why don't we just put the setjmp in an outer function?

jonsneyers:
> > How well tested is this?
> 
> I'll do a roundtrip test on the pngsuite test set to see if that catches any problems.

All the non-corrupt images in the pngsuite test set roundtrip (except for the gamma=2.5 ones, but that's a jxl limitation that is not related to png input/output). I did `cjxl -m in.png out.jxl` followed by `djxl out.jxl out.png` and `compare -metric pae in.png out.png null:` to check for differences. One had an off-by-one in 16-bit caused by having sBIT being set to 13, but I think that's because ImageMagick is rounding differently than we are when discarding the 3 lsb (it's in any case not a png writing issue but an issue of how we interpret the input png, so not related to this PR).

jonsneyers:
As for animated png output: I've only tested it on two or three animated jxls, and it produced output that looked OK.

lib/extras/codec_apng.cc
jonsneyers: I don't see an easy way to do that, but I think it's fine to use the default libpng error handling instead of catching them. It shouldn't happen that you get png _encode_  errors, and if you do, it's not going to make much of a difference if the error gets caught or not. So I'm just removing the setjmp in the encode path.

jonsneyers:
Adding CI:full label to see if that reveals any issues.

lib/extras/codec_apng.cc
veluca93: I'm not 100% sure I like this but I am fine with it until proven otherwise :)

veluca93 (approved):

PR #930, opened 2021-11-30
Define FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION when appropriate

sboukortt:


CMakeLists.txt
```diff
@@ -231,12 +231,6 @@ if(MSVC)
   add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 endif()
 
-if("${JPEGXL_ENABLE_FUZZERS}" OR "${JPEGXL_ENABLE_COVERAGE}")
```
veluca93, line 234: Was this removal on purpose?

sboukortt: Yes, a quick `git grep` suggests that we don’t actually use it anywhere, and it is doubtful that we should.

veluca93 (dismissed):

veluca93 (approved):

PR #931, opened 2021-11-30
Implement the simple render pipeline.

veluca93:
The pipeline is single-threaded and uses a large amount of memory.
Also modify trivial stages to test the behaviour of the pipeline, by
checking that downsampled all-0 input is upsampled to all-0.

### DIFF
--- a/lib/jxl/render_pipeline/simple_render_pipeline.cc
+++ b/lib/jxl/render_pipeline/simple_render_pipeline.cc
@@ -6,16 +6,169 @@
 #include "lib/jxl/render_pipeline/simple_render_pipeline.h"
 
 namespace jxl {
-void SimpleRenderPipeline::PrepareForThreads(size_t num) {
-  // TODO(veluca): actually allocate input buffers.
+void SimpleRenderPipeline::PrepareForThreadsInternal(size_t num) {
+  auto ch_size = [](size_t frame_size, size_t shift) {
+    return DivCeil(frame_size, 1 << shift) + kRenderPipelineXOffset * 2;
+  };
+  for (auto ch_shifts : channel_shifts_[0]) {
+    channel_data_.push_back(
+        ImageF(ch_size(frame_dimensions_.xsize_padded, ch_shifts.first),
+               ch_size(frame_dimensions_.ysize_padded, ch_shifts.second)));
+  }
 }
 
 std::vector<std::pair<ImageF*, Rect>> SimpleRenderPipeline::PrepareBuffers(
     size_t group_id, size_t thread_id) {
-  return {};
+  std::vector<std::pair<ImageF*, Rect>> ret;
+  std::pair<size_t, size_t> min_color_shifts{64, 64};
+  for (size_t c = 0; c < channel_data_.size() && c < 3; c++) {
+    min_color_shifts.first =
+        std::min(min_color_shifts.first, channel_shifts_[0][c].first);
+    min_color_shifts.second =
+        std::min(min_color_shifts.first, channel_shifts_[0][c].second);
+  }
+  for (size_t c = 0; c < channel_data_.size(); c++) {
+    const size_t gx = group_id % frame_dimensions_.xsize_groups;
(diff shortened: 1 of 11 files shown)


lib/jxl/render_pipeline/simple_render_pipeline.cc
```diff
@@ -6,16 +6,146 @@
...
-void SimpleRenderPipeline::PrepareForThreads(size_t num) {
-  // TODO(veluca): actually allocate input buffers.
+void SimpleRenderPipeline::PrepareForThreadsInternal(size_t num) {
+  for (size_t c = 0; c < channel_shifts_[0].size(); c++) {
+    channel_data_.push_back(ImageF(DivCeil(frame_dimensions_.xsize_padded,
+                                           1 << channel_shifts_[0][c].first) +
+                                       kRenderPipelineXOffset * 2,
+                                   DivCeil(frame_dimensions_.ysize_padded,
+                                           1 << channel_shifts_[0][c].second) +
+                                       kRenderPipelineXOffset * 2));
```
jonsneyers, line 16: These kind of computations of dimensions (and rects, like in PrepareBuffers) are probably better factored out into auxiliary functions. That will help to make this code easier to read and thus more 'obviously correct' :)

(also such auxiliary functions will probably be useful also in the non-simple render pipeline)

veluca93: They will definitely *not* be useful in the non-simple render pipeline :)

lib/jxl/render_pipeline/simple_render_pipeline.cc
```diff
@@ -6,16 +6,146 @@
...
+      }
+      new_channels[c] =
+          ImageF((input_sizes[c].first << stage->settings_.shift_x) +
+                     kRenderPipelineXOffset * 2,
+                 (input_sizes[c].second << stage->settings_.shift_y) +
+                     kRenderPipelineXOffset * 2);
+      output_channels[c] = &new_channels[c];
+    }
+
+    // Add padding to all kInOut channels.
```
jonsneyers, line 63: This padding is to implement Mirroring like in section 6.5 of the spec, right? It might be useful to say that in the comment. In particular, I wouldn't really call this "padding", since to me that implies the values don't matter.

lib/jxl/render_pipeline/simple_render_pipeline.cc
```diff
@@ -6,16 +6,146 @@
...
+    for (size_t c = 0; c < channel_data_.size(); c++) {
+      if (stage->GetChannelMode(c) == RenderPipelineChannelMode::kIgnored) {
+        continue;
+      }
+      JXL_ASSERT(ysize == input_sizes[c].second || ysize == kInf);
+      ysize = input_sizes[c].second;
+      JXL_ASSERT(xsize == input_sizes[c].first || xsize == kInf);
+      xsize = input_sizes[c].first;
+    }
+
```
jonsneyers, line 102: Maybe have an assert here that xsize/ysize are not `kInf` anymore at this point? Better to catch cases where all channels are somehow ignored early, and not go do 2^64 iterations of a no-op loop or something.

jonsneyers (approved):

jonsneyers (dismissed):

lib/jxl/render_pipeline/simple_render_pipeline.cc
```diff
@@ -6,16 +6,171 @@
...
+                    kRenderPipelineXOffset + frame_dimensions_.xsize_padded,
+                    kRenderPipelineXOffset + frame_dimensions_.ysize_padded);
+    ret.emplace_back(&channel_data_[c], rect);
+  }
+  return ret;
 }
 
 void SimpleRenderPipeline::ProcessBuffers(size_t group_id, size_t thread_id) {
-  // TODO(veluca): actually run the pipeline.
+  if (!IsDone()) return;
```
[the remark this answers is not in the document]
veluca93: Renamed IsDone to make things clearer :)

sboukortt (approved):

Issue #919, opened 2021-11-27
Decoder illegal instruction crash with allow_partial_files and allow_more_progressive_steps


PeterHorvath10:
**Describe the bug**
This 363 byte file [illegalInstruction.fuzz.txt](https://github.com/libjxl/libjxl/files/7612115/illegalInstruction.fuzz.txt) causes an illegal instruction crash when the decoder is used with  the --allow_partial_files and --allow_more_progressive_steps falgs. 

**To Reproduce**
djxl illegalInstruction.fuzz.txt --allow_partial_files --allow_more_progressive_steps

**Expected behavior**
The decoder should exit without crashing.

**Screenshots**
![Screenshot from 2021-11-27 18-31-56](https://user-images.githubusercontent.com/47642195/143691169-b53d3206-bd08-4ec2-8fe5-6e3017031274.png)

**Environment**
 - OS: Ubuntu 20.04
 - Compiler version: clang 10.0.0
 - CPU type: x86_64
 - cjxl/djxl version string: djxl v0.6.1 a205468  asan [AVX2,SSE4,SSSE3,Scalar]

**Additional context**
Honggfuzz was used to find the input file.

jonsneyers:
On the current version, I get the following output:

```
JPEG XL decoder v0.7.0 4861a95 [SSE4,SSSE3,Scalar]
Read 363 compressed bytes.
No output file specified.
Decoding will be performed, but the result will be discarded.
../lib/jxl/fields.cc:357: Skipping 22-bit extension(s)
../lib/jxl/fields.cc:840: JXL_FAILURE: F16 infinity or NaN are not supported
../lib/jxl/quant_weights.cc:474: JXL_RETURN_IF_ERROR code=1: F16Coder::Read(br, &dc_quant_[c])
../lib/jxl/dec_frame.cc:403: JXL_RETURN_IF_ERROR code=1: dec_state_->shared_storage.matrices.DecodeDC(br)
../lib/jxl/dec_frame.cc:221: JXL_RETURN_IF_ERROR code=1: frame_decoder.ProcessSections( section_info.data(), section_info.size(), section_status.data())
../lib/jxl/dec_file.cc:174: JXL_FAILURE: Not enough data.
../lib/jxl/dec_bit_reader.h:259: JXL_FAILURE: Read more bits than available in the bit_reader
Failed to decompress to pixels.
```

PeterHorvath10:
[...]

PR #936, opened 2021-12-01
Do not ignore fatal error on group decoding.

eustas:
Otherwise we might touch channels with not undone transforms.


jonsneyers (approved):

jonsneyers:
Is this related to https://github.com/libjxl/libjxl/issues/919?

eustas:
Perhaps. I'm currently working through oss-fuzz reports.

PR #937, opened 2021-12-01
Use the correct macro to enable memory limits in fuzz builds.

veluca93:


eustas (approved):

eustas:
This will require changes in oss-fuzz projects as well. Including libvips.

veluca93:
AFAIU that macro is enabled by default in oss-fuzz projects.

eustas:
For honggfuzz as well?

veluca93:
> For honggfuzz as well?

No idea - it's the recommended macro by LLVM's fuzzing guide though

sboukortt (approved):

PR #939, opened 2021-12-02
Rename JxlEncoderOptions to JxlEncoderFrameSettings

lvandeve:
As discussed in https://github.com/libjxl/libjxl/pull/909

The reason for this is that these settings only apply to a frame, they
include frame metadata, and the name encoder options could be taken to
mean global options of the encoder.

Identifiers that already existed in the 0.6 release have the old name
marked as deprecated: JxlEncoderOptions, JxlEncoderOptionsCreate,
JxlEncoderOptionsSetDistance, JxlEncoderOptionsSetLossless

Other functions that are already marked deprecated are kept as-is, e.g.
JxlEncoderOptionsSetDecodingSpeed

Other identifiers that changed name have not been part of a release yet
so can be renamed without keeping the old name as deprecated:
JxlEncoderSetIntegerOption, JxlEncoderFrameSettingId and the enum values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    the non-coalesced case.
  - decoder API: new function `JxlDecoderGetExtraChannelBlendInfo` to get
    the blending information for extra channels in the non-coalesced case.
+ - encoder API: added ability to set several encoder options to frames using
+   `JxlEncoderFrameSettingsSetOption`
 
 ### Changed
 - decoder API: using `JxlDecoderCloseInput` at the end of all input is required
   when using JXL_DEC_BOX, and is now also encouraged in other cases, but not
   required in those other cases for backwards compatiblity.
 - encoder API: `JxlEncoderCloseInput` now closes both frames and boxes input.
+
+### Deprecated
+- encoder API: `JxlEncoderOptions`: use `JxlEncoderFrameSettings` instead
+- encoder API: `JxlEncoderOptionsCreate`: use `JxlEncoderFrameSettingsCreate`
+  instead
+- encoder API: `JxlEncoderOptionsSetDistance`: use `JxlEncoderSetFrameDistance`
+  instead
+- encoder API: `JxlEncoderOptionsSetLossless`: use `JxlEncoderSetFrameLossless`
+  instead
+- encoder API: `JxlEncoderOptionsSetEffort`: use `JxlEncoderFrameSettingsSetOption(
+  frame_settings, JXL_ENC_FRAME_SETTING_EFFORT, effort)` instead.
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -23,12 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Changed
 - decoder API: using `JxlDecoderCloseInput` at the end of all input is required
   when using JXL_DEC_BOX, and is now also encouraged in other cases, but not
   required in those other cases for backwards compatiblity.
 - encoder API: `JxlEncoderCloseInput` now closes both frames and boxes input.
+- encoder API: renamed `JxlEncoderOptions` to `JxlEncoderFrameSettings`, and
+  renamed related functions such as `JxlEncoderOptionsCreate` to
+  `JxlEncoderFrameSettingsCreate`. The old names are kept available but are
+  marked as deprecated.
```
jonsneyers, line 37: It might be worth specifically listing all function/struct names that have become deprecated, and what the new name is. That way an application dev can update their code just by looking at the Changelog, without having to dig in in the .h files or api documentation.

Perhaps instead of putting this under `### Changed`, it makes sense to have another section `### Deprecated` for this.

CHANGELOG.md
```diff
@@ -23,12 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    the non-coalesced case.
  - decoder API: new function `JxlDecoderGetExtraChannelBlendInfo` to get
    the blending information for extra channels in the non-coalesced case.
+ - encoder API: added ability to set several encoder options to frames using
+   `JxlEncoderSetFrameSetting`
```
jonsneyers, line 27: There's also `JxlEncoderSetFrameInfo`.

Maybe those two names are a bit confusing. One is to change an encode option, the other is to update frame header info.

Perhaps the following names are clearer:

- `JxlEncoderFrameSettingsSetEncodeOption`  (instead of `JxlEncoderSetFrameSetting`)
- `JxlEncoderFrameSettingsSetHeaderInfo` (instead of `JxlEncoderSetFrameInfo`)

(or JxlEncoderSetFrameSettingsOption / JxlEncoderSetFrameSettingsInfo)

In any case I would consistently use plural `FrameSettings` even if the function is only used to change one setting.

lvandeve: Makes sense, I chose the following names, with the word "Set" later because especially for the frame name the early "Set" was confusing:

 - JxlEncoderFrameSettingsSetOption
 - JxlEncoderFrameSettingsSetInfo
 - JxlEncoderFrameSettingsSetName

CHANGELOG.md
lvandeve: Done

jonsneyers (approved):

PR #941, opened 2021-12-02
Implement frame animation settings in encoder API

lvandeve:
Writes the frame duration fields and frame name

Blending and cropping fields not yet implemented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    the blending information for extra channels in the non-coalesced case.
  - encoder API: added ability to set several encoder options to frames using
    `JxlEncoderFrameSettingsSetOption`
+ - encoder API: new function `JxlEncoderFrameSettingsSetInfo` to set animation
+   and blending parameters of the frame.
 
 ### Changed
 - decoder API: using `JxlDecoderCloseInput` at the end of all input is required
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -367,9 +367,6 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
  * or introduce additional internal frames. Animation duration and time code
  * information is not altered since those are immutable metadata of the frame.
  *
- * TODO(lode): JxlFrameHeader currently only contains animation information,
- * also allow setting crop and blending fields.
- *
  * It is not required to use this function, however if have_animation is set
  * to true in the basic info, then this function should be used to set the
  * time duration of this individual frame. By default individual frames have a
@@ -390,7 +387,8 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
  *
  * @param frame_settings set of options and metadata for this frame. Also
  * includes reference to the encoder object.
- * @param frame_header frame header data to set.
+ * @param frame_header frame header data to set. Object owned by the caller and
+ * does not need to be kept in memory, its information is copied internally.
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus
@@ -410,14 +408,14 @@ JxlEncoderFrameSettingsSetInfo(JxlEncoderFrameSettings* frame_settings,
  * Calling @ref JxlEncoderFrameSettingsSetInfo clears any name that was
  * previously set.
  *
- * @param frame_options set of options and metadata for this frame. Also
+ * @param frame_settings set of options and metadata for this frame. Also
  * includes reference to the encoder object.
  * @param frame_name name of the next frame to be encoded, as a UTF-8 encoded C
- * string (zero terminated).
+ * string (zero terminated). Owned by the caller, and copied internally.
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus JxlEncoderFrameSettingsSetName(
-    JxlEncoderFrameSettings* frame_options, const char* frame_name);
+    JxlEncoderFrameSettings* frame_settings, const char* frame_name);
 
 /**
  * Sets the buffer to read JPEG encoded bytes from for the next frame to encode.
@@ -736,6 +734,17 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetICCProfile(JxlEncoder* enc,
  * @param info global image metadata. Object owned by the caller.
  */
 JXL_EXPORT void JxlEncoderInitBasicInfo(JxlBasicInfo* info);
+/**
+ * Initializes a JxlFrameHeader struct to default values.
+ * For forwards-compatibility, this function has to be called before values
+ * are assigned to the struct fields.
+ * The default values correspond to a frame with no animation duration and the
+ * 'replace' blend mode. After using this function, For animation duration must
+ * be set, for composite still blend settings must be set.
+ *
+ * @param frame_header frame metadata. Object owned by the caller.
+ */
+JXL_EXPORT void JxlEncoderInitFrameHeader(JxlFrameHeader* frame_header);
 
 /**
  * Sets the global metadata of the image encoded by this encoder.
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    the blending information for extra channels in the non-coalesced case.
  - encoder API: added ability to set several encoder options to frames using
    `JxlEncoderFrameSettingsSetOption`
+ - encoder API: new function `JxlEncoderSetFrameInfo` to set animation and
```
jonsneyers, line 28: JxlEncoderFrameSettingsSetInfo, right?

lib/include/jxl/encode.h
```diff
@@ -410,14 +408,14 @@ JxlEncoderFrameSettingsSetInfo(JxlEncoderFrameSettings* frame_settings,
...
  * previously set.
  *
- * @param frame_options set of options and metadata for this frame. Also
+ * @param frame_settings set of options and metadata for this frame. Also
  * includes reference to the encoder object.
  * @param frame_name name of the next frame to be encoded, as a UTF-8 encoded C
  * string (zero terminated).
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus JxlEncoderFrameSettingsSetName(
```
veluca93, line 417: This function's documentation should state something about how long that string should live for ;)

veluca93 (dismissed):

lib/include/jxl/encode.h
jonsneyers: I suggest we immediately copy it so it is never an issue, but it would indeed be good to document that so you don't have to worry about it.

lib/jxl/encode.cc
```diff
@@ -418,6 +427,24 @@ void JxlEncoderInitBasicInfo(JxlBasicInfo* info) {
...
 
+void JxlEncoderInitFrameHeader(JxlFrameHeader* info) {
+  info->duration = 0;
+  info->timecode = 0;
+  info->name_length = 0;
+  info->is_last = JXL_FALSE;
+  info->layer_info.crop_x0 = 0;
+  info->layer_info.crop_y0 = 0;
+  // TODO(lode): must these be set to image size instead, or will caller do
+  // this?
```
jonsneyers, line 438: Initializing to image dimensions seems a more sensible default value to me, so you don't have to explicitly set these if you are sending full-sized frames or layers.

lib/jxl/encode.cc
```diff
@@ -418,6 +427,24 @@ void JxlEncoderInitBasicInfo(JxlBasicInfo* info) {
...
+  info->layer_info.xsize = 0;
+  info->layer_info.ysize = 0;
+  info->layer_info.blend_info.blendmode = JXL_BLEND_REPLACE;
```
jonsneyers, line 441: I would pick a combination of default values that makes sense in at least some use cases. Either a duration of one tick and kReplace blending (so fixed fps animation is the default) or a duration of zero and kBlend blending (so full-sized layers of a composite still is the default). I have no strong preference on which of these is the default, but the combination of zero duration and kReplace does not make sense so I wouldn't use that as the default.

lib/jxl/encode.cc
```diff
@@ -418,6 +427,24 @@ void JxlEncoderInitBasicInfo(JxlBasicInfo* info) {
...
+  info->layer_info.blend_info.source = 0;
+  info->layer_info.blend_info.alpha = 0;
+  info->layer_info.blend_info.clamp = 0;
+  info->layer_info.save_as_reference = 0;
```
jonsneyers, line 445: If zero duration layers are the default, then I guess source and save_as_reference should both be 1, right @veluca93 ?

If kReplace is the default and full-sized frames are the default, then it's fine to keep these at 0.

veluca93: I'd say correct.

CHANGELOG.md
lvandeve: Right

lib/include/jxl/encode.h
lvandeve: Done

lib/jxl/encode.cc
lvandeve: See other comment: After thinking about the options (see other comment) I have now chosen to follow the specification default, and that has it at 0. User should set them if they want to use crop. We should also add have_crop, but not in this MR since blending (of which cropping is part) is not for this MR

lib/jxl/encode.cc
lvandeve: See other comment: I chose JXL spec default for everything, and the default are 0. User should set up values when needed for the user's purpose

lib/jxl/encode.cc
lvandeve: I've thought about the options here, and was going to add JxlBasicInfo as input to copy the crop xsize from, but after trying out the different options, have chosen a different alternative instead.

Option 1: set up the default to be sensible for non-animation, with JXL_BLEND_BLEND: this is not a good option, because adding multiple frames for animation is more common, and with the blend default, the user has to manually undo a lot of fields to enable animation.

Option 2: set up the default to be sensible for animation. This is slightly more useful than option 1, but has the issue of giving an invalid state if the basic info does not have "have_animation" in it. And making the initialization depend on the basic info for this only makes the function more complex to use for the user (as they may still want a composite still frame in this case too). So this is also not a good option

Option 3: set the default to the default values that are in the specification. the JXL specification happens to have the combination of duration 0, and JXL_BLEND_REPLACE, as its default. This option is not so bad imho: the values are mostly set up for animation, the user just has to set duration (and possibly timecode), that's not too bad at all. When the user wants a composite still, they need some more knowledge of all the other fields, but at least they don't have to touch animation related fields (setting duration back to 0) for this.

So I went with option 3: all defaults for the spec. As for crop, xsize and ysize default to 0 in the spec, so setting them to this. To use cropping, we could add the have_crop field like the spec has (and then the user sets the desired crop xsize and ysize when the set this field to true), but that's unrelated to this MR

lvandeve:
PTAL: default initialization now follows the spec defaults

lib/jxl/encode.cc
jonsneyers: We could define have_crop to be implicitly equal to `layer_info.xsize && layer_info.ysize`, i.e. if you don't touch them then have_crop is false and the frame is implicitly canvas-sized.

lib/jxl/image_bundle.h
```diff
@@ -213,8 +213,15 @@ class ImageBundle {
   YCbCrChromaSubsampling chroma_subsampling;
 
   FrameOrigin origin{0, 0};
-  // Animation-related information. This assumes GIF- and APNG- like animation.
+
+  // Animation-related information, corresponding to the timecode and duration
+  // fields of the jxl::AnimationFrame of the jxl::FrameHeader.
+  // TODO(lode): ImageBundle is used here to carry the information from
+  // jxl::FrameHeader, consider instead passing a jxl::FrameHeader directly to
+  // EncodeFrame or having a field of that type here.
```
jonsneyers, line 221: +1 on having a FrameHeader field here. The current approach is incomplete: crop dimensions are given by whatever the frame dimensions happen to be, position is given by `origin`, `duration` is here, and blending is only specified in a simplified way (extra channel blending is ignored here, as well as clamping and the possibility to use different sources, etc).
It would be cleaner to replace the current fields by a single FrameHeader field, which would immediately make it complete.

jonsneyers: (obviously not for this PR)

jonsneyers (approved):

lib/include/jxl/encode.h
```diff
@@ -736,6 +734,17 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetICCProfile(JxlEncoder* enc,
  * @param info global image metadata. Object owned by the caller.
  */
 JXL_EXPORT void JxlEncoderInitBasicInfo(JxlBasicInfo* info);
+/**
+ * Initializes a JxlFrameHeader struct to default values.
+ * For forwards-compatibility, this function has to be called before values
+ * are assigned to the struct fields.
+ * The default values correspond to a frame with no animation and the 'blend'
+ * blend mode. For animation, the duration must be set and possibly a different
```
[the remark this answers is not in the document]
lvandeve: Updated the documentation

lib/jxl/image_bundle.h
lvandeve: Acknowledged

jonsneyers (approved):

PR #942, opened 2021-12-02
Apply the HLG OOTF when encoding to VarDCT and unapply it when decoding to HLG

sboukortt:
This effectively makes VarDCT always display-referred, instead of being
display-referred if and only if the original colorspace did not happen
to be HLG. It means that the linear output of the decoder is now always
suitable for display at a peak luminance of `intensity_target`.

JxlCmsInterface now has the responsibility of converting between scene
and display light if HLG is involved on one side and not the other. The
default implementation, JxlCms, does it and can be used for inspiration.

Images previously encoded from HLG to VarDCT with the default HLG
`intensity_target` of 1000, having not had the OOTF applied, will now be
decoded to a HLG signal brighter than the original since the OOTF will
nevertheless still be unapplied. But arguably, it was a bug of previous
encoders, since the linear signal was also too bright for display at a
peak luminance of 1000 cd/m².

### DIFF
--- a/lib/include/jxl/cms_interface.h
+++ b/lib/include/jxl/cms_interface.h
@@ -4,7 +4,9 @@
  * license that can be found in the LICENSE file.
  */
 
-/** @file cms_interface.h
+/** @addtogroup libjxl_common
+ * @{
+ * @file cms_interface.h
  * @brief Interface to allow the injection of different color management systems
  * (CMSes, also called color management modules, or CMMs) in JPEG XL.
  *
@@ -64,11 +66,19 @@ typedef struct {
  * @param output_profile the colorspace to which JxlCmsInterface::run should
  *        convert the input data.
  * @param intensity_target for colorspaces where luminance is relative
- *        (essentially: not PQ), indicates the luminance of (1, 1, 1). This is
- *        only useful for conversions between a relative luminance colorspace
- *        and an absolute luminance one, in either direction.
- *        @p intensity_target cd/m² in the absolute colorspace should map to and
- *        from (1, 1, 1) in the relative one.
+ *        (essentially: not PQ), indicates the luminance at which (1, 1, 1) will
+ *        be displayed. This is useful for conversions between PQ and a relative
+ *        luminance colorspace, in either direction: @p intensity_target cd/m²
+ *        in PQ should map to and from (1, 1, 1) in the relative one.\n
+ *        It is also used for conversions to and from HLG, as it is
+ *        scene-referred while other colorspaces are assumed to be
+ *        display-referred. That is, conversions from HLG should apply the OOTF
+ *        for a peak display luminance of @p intensity_target, and conversions
+ *        to HLG should undo it. The OOTF is a gamma function applied to the
+ *        luminance channel (https://www.itu.int/rec/R-REC-BT.2100-2-201807-I
+ *        page 7), with the gamma value computed as
+ *        <tt>1.2 * 1.111^log2(intensity_target / 1000)</tt> (footnote 2 page 8
+ *        of the same document).
  * @return The data needed for the transform, or @c NULL in case of failure.
  *         This will be passed to the other functions as @c user_data.
  */
@@ -218,3 +228,5 @@ typedef struct {
 #endif
 
 #endif /* JXL_CMS_INTERFACE_H_ */
+
+/** @} */
--- a/lib/jxl/enc_color_management.cc
+++ b/lib/jxl/enc_color_management.cc
@@ -60,6 +60,12 @@ struct JxlCms {
   void* lcms_transform;
 #endif
 
+  // These fields are used when the HLG OOTF or inverse OOTF must be applied.
+  bool apply_hlg_ootf;
+  size_t hlg_ootf_num_channels;
+  // Y component of the primaries.
+  std::array<float, 3> hlg_ootf_luminances;
+
   ImageF buf_src;
   ImageF buf_dst;
   float intensity_target;
@@ -67,6 +73,9 @@ struct JxlCms {
   ExtraTF preprocess = ExtraTF::kNone;
   ExtraTF postprocess = ExtraTF::kNone;
 };
+
+Status ApplyHlgOotf(JxlCms* t, float* JXL_RESTRICT buf, size_t xsize,
+                    bool forward);
 }  // namespace
 }  // namespace jxl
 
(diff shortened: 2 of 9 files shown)


lib/include/jxl/cms_interface.h
```diff
@@ -64,11 +64,20 @@ typedef struct {
...
- *        @p intensity_target cd/m² in the absolute colorspace should map to and
- *        from (1, 1, 1) in the relative one.
+ *        (essentially: not PQ), indicates the luminance at which (1, 1, 1) will
+ *        be displayed. This is useful for conversions between a relative
+ *        luminance colorspace and an absolute luminance one, in either
+ *        direction: @p intensity_target cd/m² in the absolute colorspace should
+ *        map to and from (1, 1, 1) in the relative one.  
+ *        It is also used for conversions to and from HLG, as it is
+ *        scene-referred while other colorspaces are assumed to be
+ *        display-referred. That is, conversions from HLG should apply the OOTF
```
lvandeve, line 74: When converting from PQ to HLG (or vice versa), is the intensity_target value then used for both purposes at the same time?

lib/include/jxl/cms_interface.h
```diff
@@ -64,11 +64,20 @@ typedef struct {
...
  *        convert the input data.
  * @param intensity_target for colorspaces where luminance is relative
- *        (essentially: not PQ), indicates the luminance of (1, 1, 1). This is
- *        only useful for conversions between a relative luminance colorspace
- *        and an absolute luminance one, in either direction.
- *        @p intensity_target cd/m² in the absolute colorspace should map to and
- *        from (1, 1, 1) in the relative one.
+ *        (essentially: not PQ), indicates the luminance at which (1, 1, 1) will
+ *        be displayed. This is useful for conversions between a relative
+ *        luminance colorspace and an absolute luminance one, in either
```
lvandeve, line 69: Given that the comment higher says "essentially: not PQ", does that mean that the sentence here then means: for conversion between PQ and another color space? It would be useful to say that in a slightly more direct way

lib/jxl/enc_color_management.cc
```diff
@@ -650,6 +670,106 @@ cmsContext GetContext() {
...
+    luminances[i] = ys[i] * (chromaticities[i][0] * white_XYZ[0] +
+                             chromaticities[i][1] * white_XYZ[1] +
+                             chromaticities[i][2] * white_XYZ[2]);
+  }
+  return true;
+}
+
+Status ApplyHlgOotf(JxlCms* t, float* JXL_RESTRICT buf, size_t xsize,
+                    bool forward) {
+  if (295 <= t->intensity_target && t->intensity_target <= 305) {
```
lvandeve, line 721: Where do the values 295 and 305 come from?

lib/include/jxl/cms_interface.h
sboukortt: Yes indeed. Taking PQ->HLG as an example: it is first used in the PQ->linear part to determine which part of the PQ range is mapped to 0-1, and then in linear->HLG to determine the gamma used in the inverse OOTF before the OETF. The end result is an HLG signal which, if displayed on a monitor with a peak luminance of `intensity_target`, would look the same as the PQ signal (although it wouldn’t scale in the same way to different peak luminances).

Conversely, HLG->PQ does: HLG ---[inverse OETF]--> 0-1 linear scene light ---[OOTF (`intensity_target`)]--> 0-1 linear display light ---[scaling (`intensity_target`) and PQ inverse EOTF]--> PQ, resulting in a PQ signal representing how the HLG signal would look on a monitor with a peak luminance of `intensity_target`.

lib/include/jxl/cms_interface.h
sboukortt: That’s fair. I think that was an attempt at “future-proofing” the comment but it might not be worth it if it hurts the present. Is this better?

lib/jxl/enc_color_management.cc
sboukortt: The cut-off is somewhat arbitrary but they are the range of values that give a gamma of approximately one with the formula below (1.2 · 1.111^log2(`intensity_target` / 1000)).

sboukortt:
Ah, of course clang-format doesn’t like Markdown’s trailing double-spaces for line breaks.

lib/include/jxl/cms_interface.h
```diff
@@ -64,11 +64,19 @@ typedef struct {
...
  * @param intensity_target for colorspaces where luminance is relative
- *        (essentially: not PQ), indicates the luminance of (1, 1, 1). This is
- *        only useful for conversions between a relative luminance colorspace
- *        and an absolute luminance one, in either direction.
- *        @p intensity_target cd/m² in the absolute colorspace should map to and
- *        from (1, 1, 1) in the relative one.
+ *        (essentially: not PQ), indicates the luminance at which (1, 1, 1) will
+ *        be displayed. This is useful for conversions between PQ and a relative
+ *        luminance colorspace, in either direction: @p intensity_target cd/m²
+ *        in PQ should map to and from (1, 1, 1) in the relative one.\n
```
jonsneyers, line 70: I was checking the generated html documentation to see how this `\n` gets rendered (I don't know if that becomes a line break or a literal "\n"). Looks like all of the documentation in this file doesn't actually get generated though, for some reason. Could you look into that?

jonsneyers (approved):

lib/include/jxl/cms_interface.h
sboukortt: From what I can tell, it does but the struct gets its own separate page. Is that also what happens for you?

`\n` is equivalent to `@n` which does generate a newline: https://www.doxygen.nl/manual/commands.html#cmdn (Not sure what it has to do with the printf function but oh, well.)

jonsneyers: Well I'm just looking at the docs generated from this PR by following the link at the bottom of the "Conversation" page: https://libjxl--942.org.readthedocs.build/en/942/
I cannot find any documentation for these functions there. Maybe something has to be added to a higher level index or something?

sboukortt: Oh, yes, I was looking at the direct Doxygen output. For Sphinx, it was indeed missing an `@addtogroup`. I have put it in `libjxl_common` given that we would like to eventually also have a high-level decoding API that would use it. How does that sound?

jonsneyers: OK for me, though maybe it would make sense to have a separate group for all color encoding / color management related things.

In general, regarding the documentation, I think we need a bit more than just documentation of every struct and function. We also need a more high-level overview, and this is especially true for the color encoding / color management stuff since it is likely that the API will be used in wrong ways if we don't carefully explain things (and part of the documentation will have to be basically reader education rather than actual libjxl documentation).

(but that can of course wait for a different PR :))

lvandeve (approved):

PR #943, opened 2021-12-02
Add an XYB conversion stage.

veluca93:


lib/jxl/render_pipeline/xyb_stage.cc
```diff
@@ -0,0 +1,166 @@
...
+                  float* JXL_RESTRICT temp) const final {
+    PROFILER_ZONE("UndoXYB");
+
+    const HWY_FULL(float) d;
+    const size_t xsize_v = RoundUpTo(xsize, Lanes(d));
+    float* JXL_RESTRICT row0 = GetInputRow(input_rows, 0, 0);
+    float* JXL_RESTRICT row1 = GetInputRow(input_rows, 1, 0);
+    float* JXL_RESTRICT row2 = GetInputRow(input_rows, 2, 0);
+    // All calculations are lane-wise, still some might require
+    // value-dependent behaviour (e.g. NearestInt). Temporary unposion last
```
ziemek99, line 88: unpoison

lib/jxl.cmake
```diff
@@ -197,6 +197,8 @@ set(JPEGXL_INTERNAL_SOURCES_DEC
   jxl/render_pipeline/simple_render_pipeline.cc
   jxl/render_pipeline/simple_render_pipeline.h
   jxl/render_pipeline/test_render_pipeline_stages.h
+  jxl/render_pipeline/xyb_stage.cc
+  jxl/render_pipeline/xyb_stage.h
```
jonsneyers, line 201: It might make sense to use filenames of the form `stage_something` instead of `something_stage`, so they'll stay together when sorted.

lib/jxl/render_pipeline/xyb_stage.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
jonsneyers, line 5: It would be useful to add some comments here or in the .cc file about what this stage is doing, and maybe what invariants/expectations/assumptions there are about the input/output.

It's also not clear to me if these implementations of stages will be shared between the simple and complicated pipeline or not. I would assume the whole point is that they can be shared (and the main difference will be that the simple pipeline calls them on full frame buffers in a simple chain while the complicated one does things per group and has to deal with borders and dependencies and keeping track of when to allocate what), but I might be wrong and it would be good to clarify this somewhere in a comment.

lib/jxl.cmake
veluca93: Doesn't make that much of a difference as they're all under render_pipeline, but done.

lib/jxl/render_pipeline/xyb_stage.h
veluca93: Added a comment about what this does.

The whole point about the pipeline is that there should be no assumptions/preconditions/expectations needed for all its stages, so I don't think we need more comments about that.

As for sharing implementations: by default I'd assume that if two things are called a different implementation of an interface, they expose the same behaviour as far as that interface is concerned, so I don't see the need to clarify that in a comment.

lib/jxl/render_pipeline/stage_xyb.cc
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+ private:
+  OpsinParams opsin_params_;
+  Op op_;
+};
+
+std::unique_ptr<RenderPipelineStage> GetXYBStage(
+    const OutputEncodingInfo& output_encoding_info) {
+  if (output_encoding_info.color_encoding.tf.IsLinear()) {
```
jonsneyers, line 124: It doesn't make sense to me that only the transfer function matters here.
I would expect the primaries to also matter, or is that part of the conversion already included in the opsin_params?

jonsneyers (approved):

lib/jxl/render_pipeline/stage_xyb.cc
veluca93: Yeah, they should be. Note that this is the same behaviour as what dec_reconstruct does today :)

jonsneyers (approved):

PR #944, opened 2021-12-02
Add a "Write to IB" stage.

veluca93:


lib/jxl/render_pipeline/write_to_ib_stage.cc
```diff
@@ -0,0 +1,52 @@
...
+          image_bundle_->extra_channels()[ec].Row(ypos) + xpos - xextra,
+          GetInputRow(input_rows, 3 + ec, 0) + kRenderPipelineXOffset - xextra,
+          sizeof(float) * (xsize + 2 * xextra));
+    }
+  }
+
+  RenderPipelineChannelMode GetChannelMode(size_t c) const final {
+    return c < 3 + image_bundle_->extra_channels().size()
+               ? RenderPipelineChannelMode::kInput
+               : RenderPipelineChannelMode::kIgnored;
```
jonsneyers, line 39: Is that really needed? Non-existing channels will always be ignored by any stage, can't we just have an invariant that GetChannelMode() does not get called with an out-of-bounds channel index?

Or is this preparing for the scenario where the output image bundle has fewer extra channels than the actual jxl image, since some extra channels are "internal only"? (or not needed to render the image, e.g. a kDepth channel)

In that case, it makes sense to keep it like this, but add a comment that the output image bundle can have fewer extra channels than the input, semantics being that only the first N are written.

lib/jxl/render_pipeline/write_to_ib_stage.cc
```diff
@@ -0,0 +1,52 @@
...
+#include "lib/jxl/image_bundle.h"
+
+namespace jxl {
+
+namespace {
+class WriteToImageBundleStage : public RenderPipelineStage {
+ public:
+  explicit WriteToImageBundleStage(ImageBundle* image_bundle)
+      : RenderPipelineStage(RenderPipelineStage::Settings()),
+        image_bundle_(image_bundle) {}
```
jonsneyers, line 18: Might be useful to have some asserts/checks here to ensure some things:

- image bundle dimensions are OK (might also be useful to clarify what is expected in terms of orientation and padding to multiples of 8)
- image bundle doesn't have _more_ extra channels than what is available
- image bundle buffers are already allocated

(or possibly it might make sense / be more convenient to not just check but also do the correct initialization of the image_bundle here)

lib/jxl/render_pipeline/write_to_ib_stage.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#ifndef LIB_JXL_RENDER_PIPELINE_XYB_STAGE_H_
```
jonsneyers, line 6: non-updated copy pasta

lib/jxl/render_pipeline/write_to_ib_stage.cc
veluca93: I added some asserts.

FWIW orientation is not yet handled by any pipeline stage, so that's a non-concern for now.

lib/jxl/render_pipeline/write_to_ib_stage.cc
veluca93: At some point, we'll have noise patterns as pipeline "channels" :)

jonsneyers (approved):

PR #946, opened 2021-12-03
make modular encoding faster (mostly lightning)

jonsneyers:
Some speed tweaks, mostly for lossless -e 1.

No change in density nor in quality.

Before:
```
Encoding [Modular, lossless, lightning], 4 threads.
Compressed to 16788658 bytes (12.222 bpp).
4064 x 2704, geomean: 25.72 MP/s [20.88, 26.89], 30 reps, 4 threads.

real	0m13.375s
user	0m28.461s
sys	0m1.183s
```

After:
```
Encoding [Modular, lossless, lightning], 4 threads.
Compressed to 16788658 bytes (12.222 bpp).
4064 x 2704, geomean: 35.07 MP/s [26.63, 40.60], 30 reps, 4 threads.

real	0m9.960s
user	0m23.619s
sys	0m1.411s
```

So with 4 threads in real time it's about 35% faster, in terms of cpu time it's about 20% faster.

For comparison: when I use ImageMagick's `convert` to encode the same image 30 times, it takes 137 seconds with default settings, and still 39 seconds with `-quality 10` (which in case of png encoding means faster encoding).

The -e 1 jxl is 16788658 bytes (and was encoded in 1/3 of a second), the default ImageMagick png is 17774160 bytes (and took about 4 seconds to encode), the fast png is 19030828 bytes (and took a bit over 1 second).

So for local use cases where you just want to quickly save an image losslessly, this might be an attractive option.

### DIFF
--- a/lib/jxl/enc_ans.cc
+++ b/lib/jxl/enc_ans.cc
@@ -1496,14 +1496,32 @@ size_t BuildAndEncodeHistograms(const HistogramParams& params,
     uint_config = HybridUintConfig(10, 0, 0);
   }
   for (size_t i = 0; i < tokens.size(); ++i) {
-    for (size_t j = 0; j < tokens[i].size(); ++j) {
-      const Token token = tokens[i][j];
-      total_tokens++;
-      uint32_t tok, nbits, bits;
-      (token.is_lz77_length ? codes->lz77.length_uint_config : uint_config)
-          .Encode(token.value, &tok, &nbits, &bits);
-      tok += token.is_lz77_length ? codes->lz77.min_symbol : 0;
-      builder.VisitSymbol(tok, token.context);
+    if (codes->lz77.enabled) {
+      for (size_t j = 0; j < tokens[i].size(); ++j) {
+        const Token& token = tokens[i][j];
+        total_tokens++;
+        uint32_t tok, nbits, bits;
+        (token.is_lz77_length ? codes->lz77.length_uint_config : uint_config)
+            .Encode(token.value, &tok, &nbits, &bits);
+        tok += token.is_lz77_length ? codes->lz77.min_symbol : 0;
+        builder.VisitSymbol(tok, token.context);
+      }
+    } else if (num_contexts == 1) {
+      for (size_t j = 0; j < tokens[i].size(); ++j) {
+        const Token& token = tokens[i][j];
+        total_tokens++;
+        uint32_t tok, nbits, bits;
+        uint_config.Encode(token.value, &tok, &nbits, &bits);
+        builder.VisitSymbol(tok, /*token.context=*/0);
+      }
+    } else {
+      for (size_t j = 0; j < tokens[i].size(); ++j) {
+        const Token& token = tokens[i][j];
+        total_tokens++;
+        uint32_t tok, nbits, bits;
+        uint_config.Encode(token.value, &tok, &nbits, &bits);
+        builder.VisitSymbol(tok, token.context);
+      }
     }
   }
 
@@ -1569,33 +1587,49 @@ size_t WriteTokens(const std::vector<Token>& tokens,
   size_t numallbits = 0;
   // Writes in *reversed* order.
   auto addbits = [&](size_t bits, size_t nbits) {
-    JXL_DASSERT(bits >> nbits == 0);
-    if (JXL_UNLIKELY(numallbits + nbits > BitWriter::kMaxBitsPerCall)) {
-      out.push_back(allbits);
-      out_nbits.push_back(numallbits);
-      numallbits = allbits = 0;
+    if (JXL_UNLIKELY(nbits)) {
+      JXL_DASSERT(bits >> nbits == 0);
+      if (JXL_UNLIKELY(numallbits + nbits > BitWriter::kMaxBitsPerCall)) {
+        out.push_back(allbits);
+        out_nbits.push_back(numallbits);
+        numallbits = allbits = 0;
+      }
+      allbits <<= nbits;
+      allbits |= bits;
+      numallbits += nbits;
     }
-    allbits <<= nbits;
-    allbits |= bits;
-    numallbits += nbits;
   };
   const int end = tokens.size();
   ANSCoder ans;
-  for (int i = end - 1; i >= 0; --i) {
-    const Token token = tokens[i];
-    const uint8_t histo = context_map[token.context];
-    uint32_t tok, nbits, bits;
-    (token.is_lz77_length ? codes.lz77.length_uint_config
-                          : codes.uint_config[histo])
-        .Encode(tokens[i].value, &tok, &nbits, &bits);
-    tok += token.is_lz77_length ? codes.lz77.min_symbol : 0;
-    const ANSEncSymbolInfo& info = codes.encoding_info[histo][tok];
-    // Extra bits first as this is reversed.
-    addbits(bits, nbits);
-    num_extra_bits += nbits;
-    uint8_t ans_nbits = 0;
-    uint32_t ans_bits = ans.PutSymbol(info, &ans_nbits);
-    addbits(ans_bits, ans_nbits);
+  if (codes.lz77.enabled || context_map.size() > 1) {
+    for (int i = end - 1; i >= 0; --i) {
+      const Token token = tokens[i];
+      const uint8_t histo = context_map[token.context];
+      uint32_t tok, nbits, bits;
+      (token.is_lz77_length ? codes.lz77.length_uint_config
+                            : codes.uint_config[histo])
+          .Encode(tokens[i].value, &tok, &nbits, &bits);
+      tok += token.is_lz77_length ? codes.lz77.min_symbol : 0;
+      const ANSEncSymbolInfo& info = codes.encoding_info[histo][tok];
+      // Extra bits first as this is reversed.
+      addbits(bits, nbits);
+      num_extra_bits += nbits;
+      uint8_t ans_nbits = 0;
+      uint32_t ans_bits = ans.PutSymbol(info, &ans_nbits);
+      addbits(ans_bits, ans_nbits);
+    }
+  } else {
+    for (int i = end - 1; i >= 0; --i) {
+      uint32_t tok, nbits, bits;
+      codes.uint_config[0].Encode(tokens[i].value, &tok, &nbits, &bits);
+      const ANSEncSymbolInfo& info = codes.encoding_info[0][tok];
+      // Extra bits first as this is reversed.
+      addbits(bits, nbits);
+      num_extra_bits += nbits;
+      uint8_t ans_nbits = 0;
+      uint32_t ans_bits = ans.PutSymbol(info, &ans_nbits);
+      addbits(ans_bits, ans_nbits);
+    }
   }
   const uint32_t state = ans.GetState();
   writer->Write(32, state);
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -529,11 +529,21 @@ Status ModularFrameEncoder::ComputeEncodingData(
         size_t xsize_shifted = DivCeil(xsize, 1 << gi.channel[c_out].hshift);
         size_t ysize_shifted = DivCeil(ysize, 1 << gi.channel[c_out].vshift);
         gi.channel[c_out].shrink(xsize_shifted, ysize_shifted);
-        for (size_t y = 0; y < ysize_shifted; ++y) {
-          const float* const JXL_RESTRICT row_in = color->PlaneRow(c, y);
-          pixel_type* const JXL_RESTRICT row_out = gi.channel[c_out].Row(y);
-          JXL_RETURN_IF_ERROR(float_to_int(row_in, row_out, xsize_shifted, bits,
-                                           exp_bits, fp, factor));
+        std::atomic<bool> has_error{false};
+        JXL_RETURN_IF_ERROR(RunOnPool(
+            pool, 0, ysize_shifted, ThreadPool::NoInit,
+            [&](const int task, const int thread) {
+              const size_t y = task;
+              const float* const JXL_RESTRICT row_in = color->PlaneRow(c, y);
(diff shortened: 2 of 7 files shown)


lib/jxl/enc_ans.cc
```diff
@@ -1496,14 +1496,33 @@ size_t BuildAndEncodeHistograms(const HistogramParams& params,
...
+        const Token& token = tokens[i][j];
+        total_tokens++;
+        uint32_t tok, nbits, bits;
+        (token.is_lz77_length ? codes->lz77.length_uint_config : uint_config)
+            .Encode(token.value, &tok, &nbits, &bits);
+        tok += token.is_lz77_length ? codes->lz77.min_symbol : 0;
+        builder.VisitSymbol(tok, token.context);
+      }
+    } else if (num_contexts == 1) {
+      Histogram& singlehisto = const_cast<Histogram&>(builder.Histo(0));
```
veluca93, line 1510: This branch feels like it could be quite dangerous... Does it give significant benefit?

lib/jxl/enc_ans.cc
```diff
@@ -1569,33 +1588,49 @@ size_t WriteTokens(const std::vector<Token>& tokens,
   size_t numallbits = 0;
   // Writes in *reversed* order.
   auto addbits = [&](size_t bits, size_t nbits) {
-    JXL_DASSERT(bits >> nbits == 0);
-    if (JXL_UNLIKELY(numallbits + nbits > BitWriter::kMaxBitsPerCall)) {
-      out.push_back(allbits);
-      out_nbits.push_back(numallbits);
-      numallbits = allbits = 0;
+    if (nbits) {
```
veluca93, line 1591: I assume this could have a JXL_UNLIKELY

lib/jxl/enc_modular.cc
```diff
@@ -529,12 +529,21 @@ Status ModularFrameEncoder::ComputeEncodingData(
...
+              const size_t y = task;
+              const float* const JXL_RESTRICT row_in = color->PlaneRow(c, y);
+              pixel_type* const JXL_RESTRICT row_out = gi.channel[c_out].Row(y);
+              if (!float_to_int(row_in, row_out, xsize_shifted, bits, exp_bits,
+                                fp, factor)) {
+                has_error = true;
+              };
+            },
+            "float2int");
+        if (has_error)
```
veluca93, line 545: braces pls

lib/jxl/enc_modular.cc
```diff
@@ -1230,12 +1246,10 @@ Status ModularFrameEncoder::PrepareStreamParams(const Rect& rect,
...
       gc.hshift = fc.hshift;
       gc.vshift = fc.vshift;
+      JXL_ASSERT((r.x0() & 0xf) == 0);
       for (size_t y = 0; y < r.ysize(); ++y) {
-        const pixel_type* const JXL_RESTRICT row_in = r.ConstRow(fc.plane, y);
-        pixel_type* const JXL_RESTRICT row_out = gc.Row(y);
-        for (size_t x = 0; x < r.xsize(); ++x) {
-          row_out[x] = row_in[x];
-        }
+        memcpy(gc.Row(y), JXL_ASSUME_ALIGNED(r.ConstRow(fc.plane, y), 64),
```
veluca93, line 1251: This assume_aligned looks like it will break if we ever make pixel_type a 16-bit type.

Do we really need it? Does it make a significant difference?

lib/jxl/modular/encoding/enc_encoding.cc
```diff
@@ -248,7 +250,7 @@ Status EncodeModularChannelMAANS(const Image &image, pixel_type chan,
         pixel_type_w topleft = (x && y ? *(r + x - 1 - onerow) : left);
         int32_t guess = ClampedGradient(top, left, topleft);
         int32_t residual = r[x] - guess;
-        tokens->emplace_back(tree[0].childID, PackSigned(residual));
+        (*tokens)[pos++].value = PackSigned(residual);
```
veluca93, line 253: Couldn't we just reserve() instead of Resize()-ing, and leave emplace_back?

lib/jxl/modular/transform/enc_rct.cc
```diff
@@ -31,40 +31,44 @@ Status FwdRCT(Image& input, size_t begin_c, size_t rct_type) {
...
-        pixel_type First = in0[x];
-        pixel_type Second = in1[x];
-        pixel_type Third = in2[x];
-        if (second == 1) {
-          Second = Second - First;
-        } else if (second == 2) {
-          Second = Second - ((First + Third) >> 1);
+  RunOnPool(
+      pool, 0, h, ThreadPool::SkipInit(),
+      [&](const int task, const int thread) {
```
veluca93, line 36: Can we put the lambda in a variable?

lib/jxl/modular/encoding/enc_encoding.cc
jonsneyers: We already did reserve(), but this is faster (probably because there's no size check).

veluca93: How much faster are we talking about?

jonsneyers: I remember having a noticeable difference, I don't remember how much. I will try it when I get the chance.

jonsneyers: Speed goes down to about 33 MP/s on that image (from about 35 MP/s), if I keep the emplace_back.

veluca93: Fair enough. Might be worth leaving a comment that this is noticeably faster in some cases.

lib/jxl/enc_modular.cc
jonsneyers: Nah it can be dropped. I just saw a lot of unaligned memmoves in perf top, but probably only a few of those originate here (adding the assume aligned doesn't seem to significantly reduce the time spent in unaligned memmoving). And the assumption seems to be wrong even for int32_t pixel_type, so I'll drop it.

lib/jxl/enc_ans.cc
jonsneyers: I'll add it, for lossless it doesn't seem to make much of a difference (since it's actually not really that unlikely), but for lossy I think it will help a little.

lib/jxl/enc_ans.cc
jonsneyers: Without that branch:
4064 x 2704, geomean: 32.15 MP/s [24.28, 34.25], 30 reps, 4 threads.

With that branch:
4064 x 2704, geomean: 35.27 MP/s [26.83, 36.66], 30 reps, 4 threads.

The thing with -e 1 is that it is spending so little overall effort that even silly things like this can make a noticeable difference.

This is what perf top looks like after this pull request:
```
  33.65%  cjxl                                [.] jxl::WriteTokens
  17.95%  cjxl                                [.] jxl::EncodeModularChannelMAANS
   9.28%  cjxl                                [.] jxl::BuildAndEncodeHistograms
   8.61%  libc-2.32.so                        [.] __memmove_avx_unaligned_erms
   6.58%  cjxl                                [.] jxl::(anonymous namespace)::float_to_int
   4.50%  cjxl                                [.] jxl::ThreadPool::RunCallState<jxl::Status (unsigned long), jxl::FwdRCT(jxl::Image&, unsigned long, unsigned long, jxl::ThreadPool*)::$_
   3.55%  cjxl                                [.] std::vector<jxl::Token, std::allocator<jxl::Token> >::_M_fill_insert
   1.71%  [kernel]                            [k] clear_page_rep
   1.16%  [kernel]                            [k] do_user_addr_fault
   0.68%  cjxl                                [.] jxl::BitWriter::Write
```

Looks like doing resize() on the token vector is still not that cheap, given that it just has to zero-initialize the memory (but it's using more generic fill code instead). The float to int and memmoves should mostly be avoidable. The top 3 functions are of course needed, but any small speedup in there will have a noticeable effect on the overall encode time.

lib/jxl/modular/transform/enc_rct.cc
jonsneyers: Done.

lib/jxl/enc_modular.cc
jonsneyers: Dropped it. Didn't make any difference, and was wrong too. Also, I hope in the future we can just avoid this copying, either by letting local modular images be rects within the global one, or by skipping the global modular image when the only global transforms are RCTs, and converting the input buffer directly to local modular images.

lib/jxl/enc_modular.cc
jonsneyers: added them

lib/jxl/enc_ans.cc
veluca93: Makes sense. Can we modify the histogram builder interface so that it's clear that something like this could happen?

jonsneyers: Maybe do that in a later pull request? There might be an opportunity to speed up the histogram building a bit more by parallelizing it: basically split the token list in N sublists (one per thread), compute N histograms, then add them together (there's already a function for that, I've seen). This is one of the steps that are currently not parallelized at all, while most of the histogram building happens globally so it would make sense to parallelize it.

veluca93: I'd rather also move adding this branch to a later PR then - I have strong feelings against breaking an interface's contract this way ;)

jonsneyers: Would it be OK to just change `HistogramBuilder.Histo()` to not return a `const` Histogram?

veluca93: yes if it goes together with a nice comment :)

jonsneyers: Did it in a different way, just using ` builder.VisitSymbol(tok, 0)` instead of `builder.VisitSymbol(tok, token.context)` and letting the compiler do a specialized inlining of that. That's cleaner than the manual specialization with the nasty `const_cast`.

veluca93 (approved):

lib/jxl/enc_ans.cc
```diff
@@ -1496,14 +1496,32 @@ size_t BuildAndEncodeHistograms(const HistogramParams& params,
...
+        tok += token.is_lz77_length ? codes->lz77.min_symbol : 0;
+        builder.VisitSymbol(tok, token.context);
+      }
+    } else if (num_contexts == 1) {
+      for (size_t j = 0; j < tokens[i].size(); ++j) {
+        const Token& token = tokens[i][j];
+        total_tokens++;
+        uint32_t tok, nbits, bits;
+        uint_config.Encode(token.value, &tok, &nbits, &bits);
+        builder.VisitSymbol(tok, 0);  // token.context == 0
```
veluca93, line 1515: usual syntax is /*context=*/0 (I believe clang-tidy will check it matches up)

jonsneyers: Right.

lib/jxl/modular/encoding/enc_encoding.cc
```diff
@@ -479,24 +485,42 @@ Status ModularEncode(const Image &image, const ModularOptions &options,
...
+  }
+  if (options.zero_tokens) {
+    tokens->resize(tokens->size() + total_tokens, {0, 0});
+  } else {
+    // Do one big allocation for all the tokens we'll need,
+    // to avoid reallocs that might require copying.
+    // Also make the allocation non-initializing.
+    size_t pos = tokens->size();
+    auto nia_tokens =
+        reinterpret_cast<std::vector<Token, no_init_alloc<Token>> *>(tokens);
```
veluca93, line 506: Uhhhh... no. This is *definitely* UB.

What's the speed difference wrt a good old reserve()?

jonsneyers: Well reserve() doesn't help because then afterwards you still need to do emplace_back() (capacity check for every token) or resize() (which will unnecessarily initialize the memory). Doing an initializing resize() is already better than having a capacity check per token, but it's still a few percent slower than avoiding that unnecessary initialization.

I suppose the cleaner way to do it is to define a type `std::vector<Token, no_init_alloc<Token>>` and use it consistently instead of a vector with the default allocator.

veluca93: Can you benchmark the various options (resize(), reserve(), this PR and current state)?

I'm not going to LGTM this change as it is. I would be OK with the unallocated vector if no reinterpret_cast is involved (and if the allocator doesn't introduce UB in interesting ways either)

jonsneyers: Yes, I agree that the reinterpret_cast is nasty. I think the best approach is to have our own class for Token vectors, which is basically just the same thing as std::vector<Token, no_init_allocator>. That'll take some significant code changes though (we have lots of stuff using std::vector<Token>) so maybe I'll change this PR to do an initializing resize and then later it will become a non-initialized resize.

jonsneyers: Updated it, the difference between this and skipping the initialization is small.

veluca93: Can you update the numbers above / add new ones?

jonsneyers: | effort | before | after (no init) | after (with init, current PR) |
|-------|--------|------|-------|
| 1 (lightning) | 25.91 MP/s | 36.57 MP/s | 36.79 MP/s |
| 2 (thunder) | 21.44 MP/s | 26.48 MP/s | 26.10 MP/s |
| 3 (falcon) | 14.55 MP/s | 17.59 MP/s | 17.47 MP/s |

The difference between no init and init is probably more noise than actual difference.

veluca93 (approved):

lib/jxl/modular/encoding/enc_encoding.cc
veluca93: Ok, so I propose we don't do anything further ;)

PR #948, opened 2021-12-03
Add a testing infrastructure for the rendering pipeline.

veluca93:
This patch adds a (not exposed) option that uses the rendering pipeline,
giving an error for all stages that are not implemented, and uses this
flag to verify the same behaviour between the slow pipeline and the
current behaviour in new tests.

### DIFF
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -53,6 +53,8 @@
 #include "lib/jxl/passes_state.h"
 #include "lib/jxl/quant_weights.h"
 #include "lib/jxl/quantizer.h"
+#include "lib/jxl/render_pipeline/stage_write_to_ib.h"
+#include "lib/jxl/render_pipeline/stage_xyb.h"
 #include "lib/jxl/sanitizers.h"
 #include "lib/jxl/splines.h"
 #include "lib/jxl/toc.h"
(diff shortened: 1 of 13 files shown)


lib/jxl/dec_frame.cc
```diff
@@ -651,6 +654,72 @@ Status FrameDecoder::ProcessACGroup(size_t ac_group_id,
...
+  }
+  if (dec_state_->shared->frame_header.CanBeReferenced()) {
+    JXL_ABORT("Not implemented: save as reference");
+  }
+
+  if (frame_header_.color_transform == ColorTransform::kYCbCr) {
+    JXL_ABORT("Not implemented: YCbCr");
+  } else if (frame_header_.color_transform == ColorTransform::kXYB) {
+    builder.AddStage(GetXYBStage(dec_state_->output_encoding_info));
+  }  // Nothing to do for kNone.
```
jonsneyers, line 706: Obviously not related to this PR, but we might actually want to do something for kNone: if the original colorspace is an enum one, there's no reason why we have to return pixels in that colorspace: we can also convert to the preferred colorspace. If we do that (also for the kYCbCr case), we can make it more likely that you'll get pixels in the preferred colorspace, which will make it more useful for applications to actually set the preferred colorspace.

lib/jxl/dec_frame.h
```diff
@@ -54,8 +54,11 @@ class FrameDecoder {
  public:
   // All parameters must outlive the FrameDecoder.
   FrameDecoder(PassesDecoderState* dec_state, const CodecMetadata& metadata,
-               ThreadPool* pool)
-      : dec_state_(dec_state), pool_(pool), frame_header_(&metadata) {}
+               ThreadPool* pool, bool use_slow_rendering_pipeline)
```
jonsneyers, line 57: It might be more convenient to not add this to the constructor but have a `SetUseRenderingPipeline` function that doesn't take a bool argument but an int/enum one (kDisabled, kSlow, kLowMemory) so later the fancy pipeline can also be tested without having to change the constructor again.

veluca93: I don't envision a state in which both pipelines will coexist with the current path - rather, once the low-memory pipeline is implemented and the slow pipeline can do both, we'll remove the current path and have just the two pipelines.

jonsneyers (approved):

PR #951, opened 2021-12-06
Instruct EMCC to link libpng.

eustas:


jonsneyers:
Does this also cause it to have JPEGXL_ENABLE_APNG set?

veluca93 (approved):

veluca93:
> Does this also cause it to have JPEGXL_ENABLE_APNG set?

I don't see how that could happen :)

eustas:
Nope, it is just a request to emcc to link "port" library (if necessary).

jonsneyers:
But it will never actually do anything then in terms of making the wasm builds of cjxl/djxl/benchmark_xl support png? Or will it?

PR #952, opened 2021-12-06
expose intrinsic size in the API

mo271:
We add fields in the `JXLBasicInfo`: a boolean `have_intrinsic_size` and
a new `JXLIntrinsicSizeHeader` struct that contains the dimensions of
the intrinsic size.
We add a parameter in the parameterized decode tests, in order to test
the intrinsic size.

### DIFF
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -241,24 +241,26 @@ typedef struct {
    */
   JxlAnimationHeader animation;
 
-  /** Indicates if an intrinsic size is present in the codestream.
-   * If JXL_TRUE, the intrinsic image size is given by intrinsic_size.
+  /** Intrinsic width of the image.
    * The intrinsic size can be different from the actual size in pixels
    * (as given by xsize and ysize) and it denotes the recommended dimensions
    * for displaying the image, i.e. applications are advised to resample the
    * decoded image to the intrinsic dimensions.
    */
-  JXL_BOOL have_intrinsic_size;
+  uint32_t intrinsic_xsize;
 
-  /** Dimensions of the intrinsic size, only used if have_intrinsic_size is
-   * JXL_TRUE.
+  /** Intrinsic heigth of the image.
+   * The intrinsic size can be different from the actual size in pixels
+   * (as given by xsize and ysize) and it denotes the recommended dimensions
+   * for displaying the image, i.e. applications are advised to resample the
+   * decoded image to the intrinsic dimensions.
    */
-  JxlIntrinsicSizeHeader intrinsic_size;
+  uint32_t intrinsic_ysize;
 
   /** Padding for forwards-compatibility, in case more fields are exposed
    * in a future version of the library.
    */
-  uint8_t padding[96];
+  uint8_t padding[100];
 } JxlBasicInfo;
 
 /** Information for a single extra channel.
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -190,14 +190,6 @@ typedef struct {
    */
   JXL_BOOL have_animation;
 
-  /** Indicates if an intrinsic size is present in the codestream.
-   * If JXL_TRUE, the intrinsic image size is given by intrinsic_size.
-   * The intrinsic size can be different from the actual size in pixels (as given by xsize and ysize)
-   * and it denotes the recommended dimensions for displaying the image,
-   * i.e. applications are advised to resample the decoded image to the intrinsic dimensions.
-   */
-  JXL_BOOL have_intrinsic_size;
-
   /** Image orientation, value 1-8 matching the values used by JEITA CP-3451C
    * (Exif version 2.3).
    */
@@ -249,6 +241,14 @@ typedef struct {
    */
   JxlAnimationHeader animation;
 
+  /** Indicates if an intrinsic size is present in the codestream.
+   * If JXL_TRUE, the intrinsic image size is given by intrinsic_size.
+   * The intrinsic size can be different from the actual size in pixels (as given by xsize and ysize)
+   * and it denotes the recommended dimensions for displaying the image,
+   * i.e. applications are advised to resample the decoded image to the intrinsic dimensions.
+   */
+  JXL_BOOL have_intrinsic_size;
+
   /** Dimensions of the intrinsic size, only used if have_intrinsic_size is
    * JXL_TRUE.
    */
@@ -257,7 +257,7 @@ typedef struct {
   /** Padding for forwards-compatibility, in case more fields are exposed
    * in a future version of the library.
    */
-  uint8_t padding[108];
+  uint8_t padding[96];
 } JxlBasicInfo;
 
 /** Information for a single extra channel.
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -72,6 +72,15 @@ typedef struct {
   uint32_t ysize;
 } JxlPreviewHeader;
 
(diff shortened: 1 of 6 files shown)


lib/include/jxl/codestream_header.h
```diff
@@ -72,6 +72,15 @@ typedef struct {
   uint32_t ysize;
 } JxlPreviewHeader;
 
+/** The intrinsic size header */
+typedef struct {
+  /** Preview width in pixels */
```
jonsneyers, line 77:
```suggestion
  /** Intrinsic width in pixels */
```

lib/include/jxl/codestream_header.h
```diff
@@ -72,6 +72,15 @@ typedef struct {
...
+  uint32_t xsize;
+
+  /** Preview height in pixels */
```
jonsneyers, line 80:
```suggestion
  /** Intrinsic height in pixels */
```

lib/include/jxl/codestream_header.h
```diff
@@ -181,6 +190,11 @@ typedef struct {
    */
   JXL_BOOL have_animation;
 
+  /** Indicates if an intrinsic size is present in the codestream.
+   * Its dimensions are not included in the basic info.
```
jonsneyers, line 194:
```suggestion
   * If JXL_TRUE, the intrinsic image size is given by intrinsic_size.
   * The intrinsic size can be different from the actual size in pixels (as given by xsize and ysize)
   * and it denotes the recommended dimensions for displaying the image,
   * i.e. applications are advised to resample the decoded image to the intrinsic dimensions.
```

lib/include/jxl/codestream_header.h
```diff
@@ -232,6 +246,11 @@ typedef struct {
...
   JxlAnimationHeader animation;
 
+  /** Dimensions of the intrinsic size, only used if have_intrinsic_size is
+   * JXL_TRUE.
+   */
+  JxlIntrinsicSizeHeader intrinsic_size;
+
   /** Padding for forwards-compatibility, in case more fields are exposed
    * in a future version of the library.
    */
```
jonsneyers, line 256: Padding should be adjusted so the struct size remains the same. I think I set it to 108 because I was anticipating 8 bytes for the intrinsic dimensions :)

lib/include/jxl/codestream_header.h
```diff
@@ -181,6 +190,11 @@ typedef struct {
...
   JXL_BOOL have_animation;
 
+  /** Indicates if an intrinsic size is present in the codestream.
+   * Its dimensions are not included in the basic info.
+   */
+  JXL_BOOL have_intrinsic_size;
```
jonsneyers, line 196: Not sure if we really need that bool. We could also just always have intrinsic dimensions - when there are none signalled in the bitstream, decoder side they are just equal to the actual dimensions. Encoder side, we can have a special case where intrinsic dimensions of 0 x 0 (which is not representable in the jxl bitstream anyway) correspond to setting no intrinsic dimensions (so implicitly setting them equal to the actual dimensions).

jonsneyers: In any case, if we add the bool, it should be at the end, so we don't break ABI compatibility (not that it matters much, we'll break it anyway between the previous released version and the next one, but still).

lib/include/jxl/codestream_header.h
mo271: Ok, adjusted it, using 12 bytes, because of the boolean.

lib/include/jxl/codestream_header.h
mo271: Moved it to the back.

Not having the bool might lead to some confusion: when the intrinsic dimensions are set exactly to the image dimension, one could interpret that as well as "not having intrinsic dimensions", but this might be precisely the intrinsic dimensions that one might want to signal sometimes. 
Going for signalling "no intrinsic dimension" with 0x0 would be fine with me, but I think it requires less explanation with that extra bool. Also it is consistent with the way we treat previews. 
What are the drawbacks of having that extra bool?

Added them to the end of the struct.

jonsneyers: Semantically, there is no difference between "not having intrinsic dimensions" and "having intrinsic dimensions equal to the image dimensions". If there are no intrinsic dimensions signaled, it has to be assumed that the intrinsic dimensions are equal to the actual dimensions, and if intrinsic dimensions are signaled that are equal to the actual dimensions, that doesn't mean anything else. It's like the difference between a JPEG without Exif orientation and a JPEG with Exif orientation set to the Identity.

There are no big drawbacks to having an extra bool, but it might be more convenient if you can just directly read `intrinsic_xsize`, `intrinsic_ysize` if you want to know the dimensions to render an image at, rather than having to do something like `if (bi.have_intrinsic_size) { render_width = bi.intrinsic_size.xsize; render_height = bi.intrinsic_size.ysize; } else { render_width = bi.xsize; render_height = bi.ysize; }`.

Previews are a bit different: not all images have a preview, and if there is none, then it doesn't make sense to talk about its dimensions.
In contrast, intrinsic size semantically always exists: if it's not explicitly signaled, then it defaults to the image dimensions, but that's just syntax (optimizing a few bytes by not having to signaling the trivial default) rather than a semantic difference.

So what I would propose is the following: no bool, and

- decoder: always sets intrinsic_xsize / intrinsic_ysize to their semantic values (equal to image dimensions if the jxl does not have explicitly signaled intrinsic dimensions)
- encoder: if intrinsic_xsize/ysize are set to zero or if they are set equal to the image dimensions, then don't explicitly signal intrinsic dimensions, otherwise set them to the given dimensions. I don't think we need to have a way to produce a jxl that explicitly (redundantly) signals the intrinsic dimensions to be equal to the image dimensions.

lib/include/jxl/codestream_header.h
```diff
@@ -72,6 +72,15 @@ typedef struct {
...
 
+/** The intrinsic size header */
+typedef struct {
+  /** Intrinsic width in pixels */
+  uint32_t xsize;
+
+  /** Intrinsic height in pixels */
+  uint32_t ysize;
+} JxlIntrinsicSizeHeader;
+
```
jonsneyers, line 83: I would either just inline this struct in the BasicInfo struct (`intrinsic_xsize`, `intrinsic_ysize`), or define a generic `JxlSize` struct that is used anywhere where we need to specify image dimensions.
Currently we have a somewhat inconsistent mix of things:

- image dimensions are inlined: JxlBasicInfo.xsize/ysize
- frame dimensions are inlined in JxlFrameHeader.layer_info
- preview dimensions are in JxlPreviewHeader (which contains nothing else than an xsize/ysize so maybe JxlPreviewSize would have been a better name)
- intrinsic dimensions are in JxlIntrinsicSizeHeader

I think maybe we should just always inline things, i.e. have JxlBasicInfo.preview_xsize/ysize and JxlBasicInfo.intrinsic_xsize/ysize instead of having extra structs for it. A struct with just two fields is not really worth it imo.

lib/include/jxl/codestream_header.h
mo271: ok, I have removed `have_intrinsic_size`. Not quite sure if this is really an improvement, because this way it is less consistent with the way it is handled internally. However I like the argument that it will be easier to use in the API, because `intrinsic_[xy]size` will always be something meaningful.

lib/include/jxl/codestream_header.h
mo271: Now I just add two fileds to `JxlBasicInfo`, thereby using only the 8 bytes you provisioned.

jonsneyers (approved):

PR #957, opened 2021-12-07
allow thermal channels

mo271:
Currently it is impossible to encode a `JXL_CHANNEL_THERMAL`, we fix
this by adding the corresponding `MakeBit`.

### DIFF
--- a/lib/jxl/image_metadata.h
+++ b/lib/jxl/image_metadata.h
@@ -65,7 +65,7 @@ static inline constexpr uint64_t EnumBits(ExtraChannel /*unused*/) {
   using EC = ExtraChannel;
   return MakeBit(EC::kAlpha) | MakeBit(EC::kDepth) | MakeBit(EC::kSpotColor) |
          MakeBit(EC::kSelectionMask) | MakeBit(EC::kBlack) | MakeBit(EC::kCFA) |
-         MakeBit(EC::kUnknown) | MakeBit(EC::kOptional);
+         MakeBit(EC::kThermal) | MakeBit(EC::kUnknown) | MakeBit(EC::kOptional);
 }
 
 // Used in ImageMetadata and ExtraChannelInfo.
--- a/lib/jxl/roundtrip_test.cc
+++ b/lib/jxl/roundtrip_test.cc
@@ -515,6 +515,7 @@ TEST(RoundtripTest, Uint16FrameRoundtripTest) {
 TEST(RoundtripTest, Uint8FrameRoundtripTest) {
   std::vector<std::vector<std::pair<JxlExtraChannelType, std::string>>>
       extra_channels_cases = {{},
+                              {{JXL_CHANNEL_THERMAL, "temperature"}},
                               {{JXL_CHANNEL_ALPHA, "my extra alpha channel"}},
                               {{JXL_CHANNEL_CFA, "my cfa channel"}},
                               {{JXL_CHANNEL_CFA, "my cfa channel"},


veluca93:
@jonsneyers Can you double-check that this does the right thing wrt spec?

jonsneyers:
Why do we have such a weird MakeBit implementation of Enums in the first place? Is there no more elegant way to do that, so such accidents cannot happen in the first place because there are no two places that have to list the same values?

veluca93:
> Why do we have such a weird MakeBit implementation of Enums in the first place? Is there no more elegant way to do that, so such accidents cannot happen in the first place because there are no two places that have to list the same values?

I don't believe that can really be done in C++ unless we just use one big macro...

jonsneyers:
> > Why do we have such a weird MakeBit implementation of Enums in the first place? Is there no more elegant way to do that, so such accidents cannot happen in the first place because there are no two places that have to list the same values?
> 
> I don't believe that can really be done in C++ unless we just use one big macro...

Why do we need to complicate Enums to have bitfields containing valid values? Just allowing any value between 0 and [last enum value] would be a lot simpler. Would there be a problem if we did that instead?

mo271:
> > > Why do we have such a weird MakeBit implementation of Enums in the first place? Is there no more elegant way to do that, so such accidents cannot happen in the first place because there are no two places that have to list the same values?
> > 
> > 
> > I don't believe that can really be done in C++ unless we just use one big macro...
> 
> Why do we need to complicate Enums to have bitfields containing valid values? Just allowing any value between 0 and [last enum value] would be a lot simpler. Would there be a problem if we did that instead?

In this particular case, there are "reserved" channels, which are not used and hence not allowed, we could just allow them.

Here I just wanted make thermal channels work, thinking about refactoring how MakeBit works with Enums should be done in a different pull request, if we decide to do so. (I think it is fine the way it is now.)

Perhaps it is a good idea to check that the missing-thermal-accident didn't creep into the spec.

jonsneyers:
The spec knows about kThermal, I'm on my phone atm so cannot check, but I remember we even specified that the channel values are to be interpreted in Kelvin.

jonsneyers (approved):

jonsneyers:
Just for the record: yes, the spec has kThermal at enum value 6, and it is described as "Infrared thermography image. Sample values are in units of Kelvin."

Btw thinking about it a bit more, this might be unclear: how is scaling done when integer bit depths are used? Are 8-bit integer values to be interpreted as having

- a nominal range from 0 to 255 Kelvin (but nothing prevents you from going outside that range in the bitstream and encode e.g. the value 273), or
- a nominal range from 0 to 1 Kelvin (which would be more consistent with how we always do scaling; and then you'd have to go way outside the nominal range, e.g. the freezing temperature of water would be represented as the value 69653).

I think the latter maybe makes more sense, because range is always unbounded anyway (well, it's effectively limited by int32_t), so to me it makes more sense to have bit depth represent precision rather than range (which is consistent with how things are interpreted for RGB or alpha or anything else). So if the bit depth is 8, then the effective unit is 1/255 Kelvin. If you only want to represent temperatures that are integers in Kelvin, you could do it by setting the bit depth to 1.

Note that compression-wise, it doesn't matter much because in lossless compression, a channel palette can and will be used to bring the effective range of the encoded values to the integer interval [0, nb_distinct_values). And in lossy compression (quantized Squeeze), the quantization amount can just be adjusted. So either representation would have similar compression performance.

To me it seems like the second option is what is implied by the spec, but it's a bit ambiguous and should be clarified.

PR #959, opened 2021-12-07
Support progressive event for DC

lvandeve:
This implements the progressive API event fo rthe DC image.

This is currently only supported for vardct without extra channels: for
modular and extra channels it's only supported if squeeze is used, and
it may not correctly work with Flush yet in that case.

### DIFF
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -361,6 +361,7 @@ Status FrameDecoder::InitFrame(BitReader* JXL_RESTRICT br, ImageBundle* decoded,
   decoded_ac_global_ = false;
   is_finalized_ = false;
   finalized_dc_ = false;
+  num_sections_done_ = 0;
   decoded_dc_groups_.clear();
   decoded_dc_groups_.resize(frame_dim_.num_dc_groups);
   decoded_passes_per_ac_group_.clear();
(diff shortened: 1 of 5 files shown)


lib/jxl/dec_frame.cc
```diff
@@ -816,6 +831,34 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
...
+      if (!decoded_->metadata()->extra_channel_info.empty()) {
+        // If extra channels are encoded with modular without squeeze, they
+        // don't support DC. If the are encoded with squeeze, DC works in theory
+        // but the implementation may not yet correctly support this for Flush.
+        // Therefore, can't correctly pause for a progressive step if there is
+        // an extra channel (including alpha channel)
+        can_return_dc = false;
+      }
+      if (frame_header_.encoding != FrameEncoding::kVarDCT) {
+        // DC is not guaranteed available in modular mode and may be a black
```
veluca93, line 850: to be available

veluca93 (dismissed):

lib/jxl/dec_frame.cc
```diff
@@ -816,6 +831,34 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
...
+        // for DC and the entire full resolution image is available at once.
+        can_return_dc = false;
+      }
+      if (!decoded_->metadata()->extra_channel_info.empty()) {
+        // If extra channels are encoded with modular without squeeze, they
+        // don't support DC. If the are encoded with squeeze, DC works in theory
+        // but the implementation may not yet correctly support this for Flush.
+        // Therefore, can't correctly pause for a progressive step if there is
+        // an extra channel (including alpha channel)
+        can_return_dc = false;
```
jonsneyers, line 847: It should be possible at this point to know whether the modular channel(s) can be shown or not, right?

Can keep it like this for this PR, but would be good to have the actual condition in here. Alpha does use squeeze by default in VarDCT mode so it's not like it's a rare case.

jonsneyers (dismissed):

lib/jxl/dec_frame.cc
lvandeve: Does the current implementation of FrameDecoder::Flush handle squeeze correctly? At least alpha didn't work in my testing (with relatively default compression settings), but something else may have been wrong

lib/jxl/dec_frame.cc
lvandeve: thanks, fixed

veluca93 (approved):

lib/jxl/dec_frame.cc
jonsneyers: When I try it with `decode_progressive` (https://github.com/libjxl/libjxl/pull/606), things do seem to work, yes. When did you do your testing? Things were broken before https://github.com/libjxl/libjxl/pull/629 was merged.

PR #963, opened 2021-12-08
Remove JXL_ENC_NOT_SUPPORTED

lvandeve:
Only use JXL_ENC_ERROR for errors. This prevents the user having to
distinguish between different error cases, and is similar to how the
decoder works. Also, JXL_ENC_NOT_SUPPORTED was not used for sufficiently
many cases to be worth its additional handling.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -71,7 +71,8 @@ typedef enum {
    */
   JXL_ENC_NEED_MORE_OUTPUT = 2,
 
-  /** The encoder doesn't (yet) support this.
+  /** DEPRECATED: the encoder does not return this status and there is no need
+   * to handle or expect it.
    */
   JXL_ENC_NOT_SUPPORTED = 3,
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - encoder API: `JxlEncoderOptionsSetDecodingSpeed`: use
   `JxlEncoderFrameSettingsSetOption(frame_settings,
   JXL_ENC_FRAME_SETTING_DECODING_SPEED, tier)` instead.
+- encoder API: deprecated `JXL_ENC_NOT_SUPPORTED`, the encoder returns
+  `JXL_ENC_ERROR` instead and there is no need to handle
+  `JXL_ENC_NOT_SUPPORTED`.
 
 ## [0.6.1] - 2021-10-29
 ### Changed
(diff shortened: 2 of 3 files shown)


lib/include/jxl/encode.h
```diff
@@ -63,18 +63,15 @@ typedef enum {
...
    */
   JXL_ENC_ERROR = 1,
 
   /** The encoder needs more output buffer to continue encoding.
    */
   JXL_ENC_NEED_MORE_OUTPUT = 2,
 
-  /** The encoder doesn't (yet) support this.
-   */
-  JXL_ENC_NOT_SUPPORTED = 3,
```
veluca93, line 76: I think this should rather be deprecated...

lvandeve: I'd rather remove it, since it has 0 usages in our code (which could be a good indication of the amount of usages outside), and the encoder has not yet been released like the decoder

lvandeve: After adding the change to the changelog I reconsidered, and added it back as deprecated.

At some point the deprecated stuff should be actually removed though

mo271 (approved):

sboukortt (approved):

jonsneyers (approved):

jonsneyers:
See also https://github.com/libjxl/libjxl/pull/823

A single error code sounds good to me for the purpose of basic error handling, but for more advanced error handling, I think it's useful to have more detailed information. Basically JXL_ENC_ERROR is what you get whenever the encoder could not do what you wanted it to do, for whatever reason, but I think applications still want to be able to distinguish different cases:

- API usage error: not really needed to catch this at runtime in non-debug builds
- something went wrong during encode for some reason (could be a libjxl bug)
- out of memory during encode: application might want to retry with less encode effort instead of just failing
- input is invalid (e.g. malformed ICC profile, corrupt JPEG): application might want to complain to the user, or decode the broken JPEG to pixels (it could be a truncated progressive JPEG that actually looks OK) and encode from there
- input JPEG is valid but `jbrd` cannot be written (e.g. because of too much tail data): application might want to retry without producing `jbrd`

Distinguishing between "API usage error" and "not supported" is probably not needed though: those are basically the same thing. In principle we could add new encoder options or change the range of accepted values to encode functions (e.g. allow higher distances) without bumping the ABI version, but if an application uses that without first checking the libjxl version, that's an API usage error. We don't know in advance what will be supported in the future, so any "API usage error" could actually be a "not supported" and vice versa. So it's pointless to distinguish them.

lib/include/jxl/encode.h
jonsneyers: I would suggest to do a 1.0-rc1 'release' (not really a release but a release candidate, so no commitment to maintain it) in which the deprecated stuff still exists, then a 1.0-rc2 in which it is removed, and in the actual 1.0 I would try to have a clean API.

lvandeve:
@jonsneyers indeed this doesn't affect the proposal for more fine grained error reporting given the usefulness of the not supported status. That said, status codes are not the ideal method to get more fine grained error information, there could be some other way to get it after you received a JXL_ENC_ERROR

PR #964, opened 2021-12-08
Simplify handling of extra channels info.

mo271:
Instead of setting the extra channel info via `SetAlphaBits`, we now use
`JxlEncoderSetExtraChannelInfo`.

Co-authored-by: Lode <lode@google.com>

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -494,23 +494,31 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
     return JXL_API_ERROR("floating point alpha not (yet) supported");
   }
 
-  switch (info->alpha_bits) {
-    case 0:
-      break;
-    case 32:
-    case 16:
-      enc->metadata.m.SetAlphaBits(16);
-      break;
-    case 8:
-      enc->metadata.m.SetAlphaBits(info->alpha_bits);
-      break;
-    default:
-      return JXL_ENC_ERROR;
+  if (!(info->alpha_bits == 0 || info->alpha_bits == 8 ||
+        info->alpha_bits == 16 || info->alpha_bits == 32)) {
+    return JXL_API_ERROR("invalid number alpha_bits");
   }
   // The number of extra channels includes the alpha channel, so for example and
   // RGBA with no other extra channels, has exactly num_extra_channels == 1
   enc->metadata.m.num_extra_channels = info->num_extra_channels;
   enc->metadata.m.extra_channel_info.resize(enc->metadata.m.num_extra_channels);
+  if (info->num_extra_channels == 0 && info->alpha_bits) {
+    return JXL_API_ERROR(
+        "when alpha_bits is non-zero, the number of channels must be at least "
+        "1");
+  }
+  // If the user provides non-zero alpha_bits, we make the channel info at index
+  // zero the appropriate alpha channel.
+  if (info->alpha_bits) {
+    JxlExtraChannelInfo channel_info;
+    JxlEncoderInitExtraChannelInfo(JXL_CHANNEL_ALPHA, &channel_info);
+    channel_info.bits_per_sample = info->alpha_bits;
+    channel_info.exponent_bits_per_sample = info->alpha_exponent_bits;
+    if (JxlEncoderSetExtraChannelInfo(enc, 0, &channel_info)) {
+      return JXL_ENC_ERROR;
+    }
+  }
+
   enc->metadata.m.xyb_encoded = !info->uses_original_profile;
   if (info->orientation > 0 && info->orientation <= 8) {
     enc->metadata.m.orientation = info->orientation;
@@ -562,6 +570,7 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelInfo(
   if (index >= enc->metadata.m.num_extra_channels) {
     return JXL_API_ERROR("Invalid value for the index of extra channel");
   }
+
   jxl::ExtraChannelInfo& channel = enc->metadata.m.extra_channel_info[index];
   channel.type = static_cast<jxl::ExtraChannel>(info->type);
   channel.bit_depth.bits_per_sample = info->bits_per_sample;


lib/jxl/encode.cc
```diff
@@ -489,23 +489,34 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
...
+  if (info->alpha_bits) {
+    JxlExtraChannelInfo channel_info;
+    JxlEncoderInitExtraChannelInfo(JXL_CHANNEL_ALPHA, &channel_info);
+    switch (info->alpha_bits) {
+      case 32:
+      case 16:
+        channel_info.bits_per_sample = 16;
+        break;
+      case 8:
+        channel_info.bits_per_sample = info->alpha_bits;
```
jonsneyers, line 512: Actually now that I'm thinking about it: why are we restricting it to be 8 or 16?

There should be no issue having an alpha channel that is, say, 10-bit.

How the pixels are passed to the encoder or from the decoder is a different thing, but that's not what this part of the API is about (we have pixelformat for that, and yes, there it has to be 8-bit, 16-bit, or I guess float).

So I think we can just set `channel_info.bits_per_sample = info->alpha_bits`.
We should also set `channel_info.exponent_bits_per_sample = info->alpha_exp_bits`, for the floating-point case.

SetExtraChannelInfo should take care of invalid values then, but you'll have to get its return status.

lib/jxl/encode.cc
```diff
@@ -489,23 +489,34 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
...
+        break;
+      default:
+        return JXL_ENC_ERROR;
+    }
+    JxlEncoderSetExtraChannelInfo(enc, 0, &channel_info);
```
jonsneyers, line 517: I suppose the aim is to avoid having to manually add the alpha channel since it's a common case, and to use e.g.  `basicinfo.alpha_bits = 8;` as a one-line way to add an alpha channel.

But if you want to, you could also do (in an application)
```
JxlExtraChannelInfo alpha;
JxlEncoderInitExtraChannelInfo(JXL_CHANNEL_ALPHA, &alpha);
alpha.bits_per_sample = 8;
JxlEncoderSetExtraChannelInfo(enc, 0, &alpha);
```

and that should also work, right? With or without setting `alpha_bits` in the basic info. Do I understand that correctly?

I'm not sure if I like having two ways to do the same thing - I think it would have been better if we didn't treat alpha as something special and have alpha_bits in the basic info. For passing pixel buffers it is convenient to have a way to interleave an extra channel with the RGB, but RGBA is not the only thing that is useful: CMYK (i.e. RGBK with an icc profile that makes RGB behave like CMY) and e.g. RGB+depth or RGB+thermal could also be useful. More generally it might also be useful to do more than 4 channels, e.g. CMYKA is a thing and I can also imagine e.g. RGB+depth+thermal to be useful, but if we don't want to go there, then "color + extra channel 0" would already be more useful than requiring that channel to represent alpha.

mo271: > I suppose the aim is to avoid having to manually add the alpha channel since it's a common case, and to use e.g. `basicinfo.alpha_bits = 8;` as a one-line way to add an alpha channel.
> 
> But if you want to, you could also do (in an application)
> 
> ```
> JxlExtraChannelInfo alpha;
> JxlEncoderInitExtraChannelInfo(JXL_CHANNEL_ALPHA, &alpha);
> alpha.bits_per_sample = 8;
> JxlEncoderSetExtraChannelInfo(enc, 0, &alpha);
> ```
> 
> and that should also work, right? With or without setting `alpha_bits` in the basic info. Do I understand that correctly?
Yes, that should work. Also `bits_per_sample` will be initialized with `8` by `JxlEncoderInitExtraChannelInfo`, so no need for that line if you have 8 bits in the alpha channel. 
 Right now `JxlEncoderSetExtraChannelInfo` does not check for validity of the `bits_per_sample` value.

I guess the question is how much boiler-plate we expect from an user of the api.
> 
> I'm not sure if I like having two ways to do the same thing - I think it would have been better if we didn't treat alpha as something special and have alpha_bits in the basic info. For passing pixel buffers it is convenient to have a way to interleave an extra channel with the RGB, but RGBA is not the only thing that is useful: CMYK (i.e. RGBK with an icc profile that makes RGB behave like CMY) and e.g. RGB+depth or RGB+thermal could also be useful. More generally it might also be useful to do more than 4 channels, e.g. CMYKA is a thing and I can also imagine e.g. RGB+depth+thermal to be useful, but if we don't want to go there, then "color + extra channel 0" would already be more useful than requiring that channel to represent alpha.
Yeah, I think the "A"-case is much more common than the "K" or "depth" or "thermal" case. As of now, 
`JxlEncoderAddImageFrame` is hard-coded to search for the first alpha channel and put the 4th (or 2nd) interleaved channel there. I'm not sure if we really want to adapt it to allow putting the interleaved 4th (or 2nd) channel to a channel of a different kind.

jonsneyers: I think it's probably easier to make the interleaving/de-interleaving function use whatever extra channel 0 happens to be, instead of checking specifically for the alpha channel. I agree that A is way more common than anything else (but this could change over time as cameras with depth sensors become more common).

Anyway, that's all not really related to this PR, so I'll close this comment thread :)

lib/jxl/encode.cc
mo271: Ok, I agree that there should be no problem having e.g. 10-bit alpha. 
(And right now this is already possible by manually defining the extra channel. 
I'm not sure what that weird switch-case was meant for. 
Setting `bits_per_sample` here to unreasonable values, eventually results in failures with meaningful error messages, coming from `VisitFields` in `image_metadata.cc`.
Should I add more checks in `setExtraChannelInfo`? If yes, which kind?

jonsneyers: The conditions on bits_per_sample and exponent_bits_per_sample are identical between the color channels and the extra channels, so it would be good to use the same check when setting it.

mo271: I doubled the check from the beginning of `JxlEncoderSetBasicInfo` to `JxlEncoderSetExtraChannelInfo`, so using the same conditions for color channels and extra channels now, please check if this looks reasonable.

lib/jxl/encode.cc
```diff
@@ -557,6 +558,17 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelInfo(
...
   }
+  if (!info->exponent_bits_per_sample) {
+    if (!(info->bits_per_sample > 0 && info->bits_per_sample <= 24)) {
+      return JXL_API_ERROR("Invalid value for bits_per_sample");
+    }
+  } else if (!(info->bits_per_sample == 32 &&
+               info->exponent_bits_per_sample == 8) &&
+             !(info->bits_per_sample == 16 &&
+               info->exponent_bits_per_sample == 5)) {
+    return JXL_API_ERROR("Invalid float description");
```
jonsneyers, line 569: This condition is overly strict. This checks if it is binary32 or binary16, which are the two cases we have a pixelformat for. But the bitstream allows anything with at most 8 exp bits and at most 24 remaining bits (mantissa+sign).

lib/jxl/encode.cc
```diff
@@ -562,6 +563,15 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderSetExtraChannelInfo(
   if (index >= enc->metadata.m.num_extra_channels) {
     return JXL_API_ERROR("Invalid value for the index of extra channel");
   }
+  if (!info->exponent_bits_per_sample) {
+    if (!(info->bits_per_sample > 0 && info->bits_per_sample <= 24)) {
+      return JXL_API_ERROR("Invalid value for bits_per_sample");
+    }
+  } else if ((info->exponent_bits_per_sample > 8) ||
+             (info->bits_per_sample > 24 + info->exponent_bits_per_sample)) {
+    return JXL_API_ERROR("Invalid float description");
```
mo271, line 572: Better like this?

lib/jxl/encode.cc
lvandeve: This is the behavior that the code had before this MR as well (the encoder was cautious), perhaps changing behavior could be done in another one?

jonsneyers: OK to split it up.

lib/jxl/encode.cc
jonsneyers: Maybe it's best to have an auxiliary function `CheckValidBitdepth(bits_per_sample, exponent_bits_per_sample)` that gets used both here (for extra channels) and when checking those fields for the color channels. For now we can keep the overly strict behavior, and then do a follow-up MR that changes the behavior to allow anything the bitstream allows (currently the code is confusing what the pixelformat allows and what the bitstream allows).

lib/jxl/encode.cc
mo271: Sure I can split it into two pull requests.

lib/jxl/encode.cc
mo271: sounds like a plan!

lib/jxl/encode.cc
mo271: first part here, leaving old behavior

lib/jxl/encode.cc
mo271: When checking those fields for the color channels, we also call `SetFooSamples`, which we do manually for the alpha, but an additional check won't hurt, I guess.

mo271: follow up here: #977

jonsneyers (approved):

PR #969, opened 2021-12-08
Add EPF stages.

veluca93:


lib/jxl/render_pipeline/stage_epf.cc
```diff
@@ -0,0 +1,491 @@
...
+}  // namespace HWY_NAMESPACE
+}  // namespace jxl
+HWY_AFTER_NAMESPACE();
+
+#if HWY_ONCE
+namespace jxl {
+
+HWY_EXPORT(GetEPFStage0);
+HWY_EXPORT(GetEPFStage1);
+HWY_EXPORT(GetEPFStage2);
```
sboukortt, line 471: Since it doesn’t seem as though `GetEPFStage` actually ever gets called with a dynamic value for `epf_stage`, would it perhaps make sense to expose these three functions directly?

sboukortt (approved):

lib/jxl/render_pipeline/stage_epf.cc
veluca93: I think I'd prefer to expose a minimal interface. No strong opinion though.

sboukortt (approved):

PR #970, opened 2021-12-08
Update AUTHORS

nebenzhal:


jonsneyers:
You didn't really have to do this as a separate pull request — you could just do a `git commit --amend` and `git push --force` to add it to the first pull request. You'll have to update that one anyway because CI will keep failing as long as it's not rebased.

PR #971, opened 2021-12-08
simplify Enums

jonsneyers:
Enums were using a somewhat complicated mechanism to validate values. This was somewhat useful for the whitepoint, primaries and transfer curves of color encodings, where the Enums are sparse (not all values are defined).

This PR simplifies things a bit by getting rid of the EnumBits and EnumName functions, and instead have the convention that every Enum has a `kLastEnumVal` that indicates the valid range of values. For whitepoint, primaries and transfer curve, the check for valid values has been made explicit in the relevant field encode/decode function (though arguably it would be more future-proof to not enforce that during field parsing, and just produce an error whenever the unknown color encoding actually needs to be used for color conversion - but I didn't want to change behavior too much with this PR).

For extra channels of a kReserved type, decoding will now not fail - I think it should be up to applications to decide if they want to show the image anyway or not. Otherwise I _think_ this PR should not change behavior.


lib/extras/codec.cc
```diff
@@ -50,10 +50,9 @@ std::string ExtensionFromCodec(Codec codec, const bool is_gray,
     case Codec::kPSD:
       return ".psd";
     case Codec::kUnknown:
+    default:
```
sboukortt, line 53: I suspect that the purpose of writing the switch in this way was to make it [exhaustive](https://abseil.io/tips/147), i.e. make it a compile-time error not to handle a new case should one be added. For codecs, perhaps it would make sense to keep it that way?

jonsneyers: Sure, in this case the Enum is not actually a bitstream field Enum anyway.

sboukortt: (It seems the default is still there.)

jonsneyers: Ah right, I missed a spot. Fixed.

mo271:
This seems potentially error prone, I'm not sure we should do it.

jonsneyers:
> This seems potentially error prone, I'm not sure we should do it.

Feel free to close this one without merging, it's not very important.

The only thing that I think we should take from this PR is to make the decoder not hard-fail when encountering extra channel types that it doesn't know yet. I think those should be handled like kUnknown: maybe produce a warning, but not hard-abort decoding.

PR #972, opened 2021-12-08
Save blending options in encoder

lvandeve:
Implement several of the frame blending options in the encoder (that
is, pass them on to the C++ encoder)

This also adds the have_crop field to JxlLayerInfo: for initialization
purposes, as well as for the documentation, it's easier to have a flag
to disable it, than to indicate no crop by setting the the layer info
sizes to the full image dimensions (making the way to indicate the
default depenent on other circumstances)

Does not yet support cropping, or multi values for save as reference.

### DIFF
--- a/lib/include/jxl/codestream_header.h
+++ b/lib/include/jxl/codestream_header.h
@@ -315,7 +315,7 @@ typedef struct {
 } JxlHeaderExtensions;
 
 /** Frame blend modes.
- * If coalescing is enabled (default), this can be ignored.
+ * When decoding, if coalescing is enabled (default), this can be ignored.
  */
 typedef enum {
   JXL_BLEND_REPLACE = 0,
@@ -326,11 +326,12 @@ typedef enum {
 } JxlBlendMode;
 
 /** The information about blending the color channels or a single extra channel.
- * If coalescing is enabled (default), this can be ignored.
+ * When decoding, if coalescing is enabled (default), this can be ignored and
+ * the blend mode is considered to be JXL_BLEND_REPLACE.
+ * When encoding, these settings apply to the pixel data given to the encoder.
  */
 typedef struct {
   /** Blend mode.
-   * Always equal to JXL_BLEND_REPLACE if coalescing is enabled.
    */
   JxlBlendMode blendmode;
   /** Reference frame ID to use as the 'bottom' layer (0-3).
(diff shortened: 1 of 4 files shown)


lib/include/jxl/codestream_header.h
```diff
@@ -346,32 +348,47 @@ typedef struct {
...
- * If coalescing is enabled (default), this can be ignored.
+ * When decoding, if coalescing is enabled (default), this can be ignored.
+ * When encoding, these settings apply to the pixel data given to the encoder,
+ * the encoder could choose an internal representation that differs.
  */
 typedef struct {
-  /** Horizontal offset of the frame (can be negative, always zero if coalescing
-   * is enabled)
+  /** Whether cropping is applied for this frame. If false, crop_x0, crop_y0
+   * are assumed to be zero, and xsize and ysize are assumed to match the main
```
jonsneyers, line 357: I would say for encoding crop_x0/y0 and xsize/ysize are ignored if have_crop is false, and for decoding they will be equal to the image dimensions in that case.

"are assumed to be" is a bit vague...

lib/include/jxl/codestream_header.h
```diff
@@ -346,32 +348,47 @@ typedef struct {
...
+   * image dimensions. When decoding, if coalescing is enabled (default), this
+   * is always false, regardless of the internal encoding in the JPEG XL
+   * codestream.
+   */
+  JXL_BOOL have_crop;
+
+  /** Horizontal offset of the frame (can be negative). Only used if have_crop
```
jonsneyers, line 364:
```suggestion
  /** Horizontal offset of the frame (can be negative). Ignored by the encoder if have_crop
    * is false; when decoding with coalescing enabled or if have_crop is false, this is always equal to 0.
```

lib/jxl/encode_test.cc
```diff
@@ -1022,6 +1023,12 @@ TEST(EncodeTest, AnimationHeaderTest) {
   // Decode to verify the boxes, we don't decode to pixels, only the boxes.
   JxlDecoderPtr dec = JxlDecoderMake(nullptr);
   EXPECT_NE(nullptr, dec.get());
+
+  // To test the blend_info fields, coalescing must be set to false in the
+  // decoder. This also assumes the encoder keeps them as-is, in theory it's
+  // permitted for the encoder to alter the blend mode as long as it gives
+  // the same result when coalescing.
```
jonsneyers, line 1030: This is debatable imo: do we expect an encoder to (visually) preserve the coalesced frames or the non-coalesced frames?

I think generally for animations you mostly care only about the coalesced frames and it's fine if the encoder chooses its own blend modes and crops, while for layered still images I think the blend mode and crop is part of the image metadata that you want to preserve. In particular, it can be useful to have layers that are larger than the canvas and I would consider it wrong for an encoder to silently trim it to canvas size (which would of course still give the same result when coalescing).

I think in the end we might need an encode option to select whether the encoder is allowed to alter crops and/or blend modes. In general there might be a few of these things where we may want the encoder to take more or less liberty in changing things, like e.g. dropping a trivial (all-opaque) alpha channel, silently convert RGB to Grayscale if R=G=B, auto-adjust the bit depth (e.g. if all pixels are 0 or 255), remove duplicate frames and add up their durations, preserve invisible pixels or not, etc etc. It is use-case dependent where the boundary lies between what an encoder is allowed to do and what it isn't allowed to do.

lib/include/jxl/codestream_header.h
lvandeve: Done

lib/include/jxl/codestream_header.h
lvandeve: I've instead removed this part, since have_crop already explains this

lib/jxl/encode_test.cc
lvandeve: Makes sense, I've removed this comment for now.

For next steps, if you don't explicitely call JxlEncoderFrameSettingsSetInfo I'd say the encoder should be able to choose (to e.g. split one frame into multiple internal frames if some specific settings are enabled), and when explicitely calling it, indeed options to enable or disable encoding techniques that do this can be added

jonsneyers: Splitting a frame into multiple internal frames that are not kRegularFrame is always fine imo - but changing what you get when doing a non-coalesced decode is something else.

jonsneyers (approved):

PR #976, opened 2021-12-09
don't do setjmp in *san builds

jonsneyers:
We are getting strange msan behavior, and it might be caused by this `setjmp`. The only purpose of that `setjmp` is to be able to catch corrupt png input without aborting, so for the purpose of msan testing it might be OK to just let libpng abort in that case.


mo271 (commented):
seems to fix non-deterministic msan failures I have experienced.

mo271 (approved):

veluca93:
> seems to fix non-deterministic msan failures I have experienced.

I propose we instead wait and see if bundling libpng fixes the issues.

jonsneyers:
Fingers crossed that we get a green pipeline again now so this one can be closed.

jonsneyers:
Yay!
No longer needed so removing.

PR #977, opened 2021-12-09
Add check for valid bit depth.

mo271:
We add a function `CheckValidBitdepth` and use it for color channels and
for extra channels.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -492,23 +492,13 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
                                             info->exponent_bits_per_sample)) {
     return JXL_ENC_ERROR;
   }
-  if (!info->exponent_bits_per_sample) {
-    if (info->bits_per_sample > 0 && info->bits_per_sample <= 24) {
-      enc->metadata.m.SetUintSamples(info->bits_per_sample);
-    } else {
-      return JXL_ENC_ERROR;
-    }
-  } else if (info->bits_per_sample == 32 &&
-             info->exponent_bits_per_sample == 8) {
-    enc->metadata.m.SetFloat32Samples();
-  } else if (info->bits_per_sample == 16 &&
-             info->exponent_bits_per_sample == 5) {
-    enc->metadata.m.SetFloat16Samples();
-  } else {
-    return JXL_API_ERROR(
-        "other exponent bits per sample combinations than IEEE binary32 and "
-        "binary16 not (yet) supported");
-  }
+  enc->metadata.m.bit_depth.bits_per_sample = info->bits_per_sample;
+  enc->metadata.m.bit_depth.exponent_bits_per_sample =
+      info->exponent_bits_per_sample;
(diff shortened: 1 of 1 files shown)


mo271:
Follow up for #964, see https://github.com/libjxl/libjxl/pull/964#discussion_r765693845

mo271:
needs rebase after #964 is merged

lib/jxl/encode.cc
```diff
@@ -181,6 +181,19 @@ int VerifyLevelSettings(const JxlEncoder* enc, std::string* debug_string) {
   // All level 5 checks passes, so can return the more compatible level 5
   return 5;
 }
+JxlEncoderStatus CheckValidBitdepth(uint32_t bits_per_sample,
+                                    uint32_t exponent_bits_per_sample) {
+  if (!exponent_bits_per_sample) {
+    if (!(bits_per_sample > 0 && bits_per_sample <= 24)) {
```
jonsneyers, line 187: We could in principle allow up to 31 for the uint bitdepth, but our code cannot handle that atm (since everything goes through float at some point, which doesn't have enough mantissa bits for that). OK to keep it like this (though technically 25 to 31 wouldn't be invalid but just not supported).

lib/jxl/encode.cc
```diff
@@ -473,6 +486,10 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
   if (!enc->metadata.size.Set(info->xsize, info->ysize)) {
     return JXL_ENC_ERROR;
   }
+  if (CheckValidBitdepth(info->bits_per_sample,
```
jonsneyers, line 489: I don't know if this is correct or flipped, but it's probably better to add an explicit `JXL_ENC_SUCCESS !=`.

lib/jxl/encode.cc
```diff
@@ -490,27 +507,28 @@ JxlEncoderStatus JxlEncoderSetBasicInfo(JxlEncoder* enc,
         "other exponent bits per sample combinations than IEEE binary32 and "
         "binary16 not (yet) supported");
```
jonsneyers, line 508: They are actually supported. In VarDCT mode, this info is just metadata. In Modular mode, it is supported.

What is not supported is setting/getting images to/from the encoder/decoder with buffers that are not binary16, binary32, uint8 or uint16. But that's unrelated: that's a limitation of JxlPixelFormat. But there's nothing stopping you from using binary32 buffers to work with e.g. fp24 images (7 exp bits, 16 mantissa bits, 1 sign bit).

So I would suggest to simplify this whole part and just set the metadata bits_per_sample and exponent_bits_per_sample directly - if the new check above didn't return error, then the values should be fine.

lib/jxl/encode.cc
mo271: ok, added a comment

lib/jxl/encode.cc
mo271: Not using the `Set***Samples` functions anymore, but setting everything directly. Functionality should be unchanged,

jonsneyers (approved):

PR #982, opened 2021-12-09
Add 2-, 4-, and 8- way interleaves.

veluca93:


lib/jxl/simd_util-inl.h
```diff
@@ -0,0 +1,257 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Misc utilities for SIMD operations
```
jonsneyers, line 6: That's a quite generic description for what is currently only specifically interleaving.

I suppose more could be added later; possibly there are already existing "simd util" things that could be moved here?

lib/jxl/simd_util-inl.h
```diff
@@ -0,0 +1,257 @@
...
+template <class DF, class V>
+HWY_INLINE Vec<DF> Concat4(const DF df, V v0, V v1, V v2, V v3) {
+  using HF = Half<DF>;
+  return Combine(DF(), Combine(HF(), v3, v2), Combine(HF(), v1, v0));
+}
+
+#endif
+
+template <class DF, class V>
+void StoreInterleaved(const DF df, V v0, V v1, float* mem) {
```
jonsneyers, line 41: It might be useful to have a comment describing what this function does. I suppose it takes two vectors of floats v0[i] and v1[i] and writes them to `mem` in an alternating order (v0[0], v1[0], v0[1], v1[1], ...)?

jonsneyers (approved):

sboukortt (approved):

PR #983, opened 2021-12-09
Upsampling stage

veluca93:
Depends on https://github.com/libjxl/libjxl/pull/982

### DIFF
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -55,6 +55,7 @@
 #include "lib/jxl/quantizer.h"
 #include "lib/jxl/render_pipeline/stage_epf.h"
 #include "lib/jxl/render_pipeline/stage_gaborish.h"
+#include "lib/jxl/render_pipeline/stage_upsampling.h"
 #include "lib/jxl/render_pipeline/stage_write_to_ib.h"
 #include "lib/jxl/render_pipeline/stage_xyb.h"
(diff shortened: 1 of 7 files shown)


lib/jxl/dec_frame.cc
```diff
@@ -699,6 +697,13 @@ void FrameDecoder::PreparePipeline() {
   if ((frame_header_.flags & FrameHeader::kSplines) != 0) {
     JXL_ABORT("Not implemented: splines");
   }
+  if (frame_header_.upsampling != 1) {
```
jonsneyers, line 700: I know extra channels aren't implemented yet anyway in the render pipeline, but it probably wouldn't hurt to add a TODO here to remember also looking at the extra channel upsampling...

lib/jxl/render_pipeline/stage_upsampling.cc
```diff
@@ -0,0 +1,152 @@
...
+        }
+        if (N == 8) {
+          StoreInterleaved(df, ups[0], ups[1], ups[2], ups[3], ups[4], ups[5],
+                           ups[6], ups[7], dst_row + x * N);
+        }
+      }
+    }
+  }
+
+ public:
```
sboukortt, line 78: How about putting this first?

sboukortt (dismissed):

sboukortt (approved):

PR #986, opened 2021-12-10
Use bionic image for cross-compilation.

eustas:


jonsneyers:
Is this ready?

eustas:
Yes, seems to work. Though gtk module is not found by cmake, it could be fixed in next PR

jonsneyers (approved):

PR #990, opened 2021-12-10
Rename JxlEncoderFrameSettingsSetInfo and JxlEncoderFrameSettingsSetE…

lvandeve:
…xtraChannelBlendInfo

Renames JxlEncoderFrameSettingsSetInfo to JxlEncoderSetFrameHeader, and
JxlEncoderFrameSettingsSetExtraChannelBlendInfo to
JxlEncoderSetExtraChannelBlendInfo.

This matches the decoder counterparts JxlDecoderGetFrameHeader and
JxlDecoderGetExtraChannelBlendInfo

It also makes the names, which are long and hard to remember, simpler.
THe fact that it uses the frame settings is already implied by taking a
FrameSettings argument. So overall this looks like the right thing to do
to keep the functions more memorable (shorter name and matching decoder)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    the blending information for extra channels in the non-coalesced case.
  - encoder API: added ability to set several encoder options to frames using
    `JxlEncoderFrameSettingsSetOption`
- - encoder API: new functions `JxlEncoderFrameSettingsSetInfo` and
-   `JxlEncoderFrameSettingsSetExtraChannelBlendInfo` to set animation
+ - encoder API: new functions `JxlEncoderSetFrameHeader` and
+   `JxlEncoderSetExtraChannelBlendInfo` to set animation
    and blending parameters of the frame, and `JxlEncoderInitFrameHeader` and
    `JxlEncoderInitBlendInfo` to initialize the structs to set.
  - decoder/encoder API: add two fields to `JXLBasicInfo`: `intrinsic_xsize`
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -399,13 +399,13 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderProcessOutput(JxlEncoder* enc,
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus
-JxlEncoderFrameSettingsSetInfo(JxlEncoderFrameSettings* frame_settings,
-                               const JxlFrameHeader* frame_header);
+JxlEncoderSetFrameHeader(JxlEncoderFrameSettings* frame_settings,
+                         const JxlFrameHeader* frame_header);
 
 /**
  * Sets blend info of an extra channel. The blend info of extra channels is set
  * separately from that of the color channels, the color channels are set with
- * @ref JxlEncoderFrameSettingsSetInfo.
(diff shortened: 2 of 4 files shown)


jonsneyers (approved):
LGTM. Best to be as symmetric as possible with the decoder in the naming. Maybe it would be a good idea to have some documentation somewhere that lists all the decoder functions and corresponding encoder functions, in the right order, and indicating which ones are optional and which ones are not. That info can be derived from the detailed documentation, but I think a compact overview (you could call it an "api cheat sheet") would be very convenient. And it could help us to spot potential naming inconsistencies that can perhaps still be fixed.

lvandeve:
> LGTM. Best to be as symmetric as possible with the decoder in the naming. Maybe it would be a good idea to have some documentation somewhere that lists all the decoder functions and corresponding encoder functions, in the right order, and indicating which ones are optional and which ones are not. That info can be derived from the detailed documentation, but I think a compact overview (you could call it an "api cheat sheet") would be very convenient. And it could help us to spot potential naming inconsistencies that can perhaps still be fixed.

Good idea, found out that there's one more name mismatch that way, will send a follow up MR for that. JxlDecoderGetFrameName vs. JxlEncoderFrameSettingsSetName

For reference, I'll put the list here for now (with the mismatched frame name as-is now):

```

JxlDecoderStatus
JxlEncoderStatus

JxlDecoderCreate
JxlEncoderCreate

JxlDecoderReset
JxlEncoderReset

JxlDecoderDestroy
JxlEncoderDestroy

JxlDecoderSetParallelRunner
JxlEncoderSetParallelRunner

JxlDecoderProcessInput
JxlEncoderProcessOutput

JxlDecoderSetInput
(encoder output done through JxlEncoderProcessOutput parameters itself)

JxlDecoderSetExtraChannelBuffer, JxlDecoderExtraChannelBufferSize
JxlEncoderSetExtraChannelBuffer

JxlDecoderCloseInput
JxlEncoderCloseInput, JxlEncoderCloseFrames, JxlEncoderCloseBoxes

JxlDecoderSetBoxBuffer, JxlDecoderReleaseBoxBuffer, JxlDecoderGetBoxType, JxlDecoderGetBoxSizeRaw
JxlEncoderAddBox

JxlDecoderSetDecompressBoxes
flag at the JxlEncoderAddBox function

JxlDecoderSetImageOutBuffer, JxlDecoderImageOutBufferSize, JxlDecoderSetImageOutCallback
JxlEncoderAddImageFrame

JxlDecoderDCOutBufferSize, JxlDecoderSetDCOutBuffer (deprecated)
N/A

JxlDecoderSetJPEGBuffer, JxlDecoderReleaseJPEGBuffer
JxlEncoderAddJPEGFrame

JxlDecoderGetBasicInfo
JxlEncoderSetBasicInfo

JxlDecoderGetFrameHeader
JxlEncoderSetFrameHeader

JxlDecoderGetExtraChannelInfo
JxlEncoderSetExtraChannelInfo

JxlDecoderGetExtraChannelName
JxlEncoderSetExtraChannelName

JxlDecoderGetFrameName
JxlEncoderFrameSettingsSetName

JxlDecoderGetExtraChannelBlendInfo
JxlEncoderSetExtraChannelBlendInfo

/
JxlEncoderSetCms

JxlDecoderGetColorAsICCProfile, JxlDecoderGetICCProfileSize, JxlDecoderGetColorAsEncodedProfile, (JxlDecoderSetPreferredColorProfile)
JxlEncoderSetICCProfile

JxlDecoderSetPreviewOutBuffer, JxlDecoderPreviewOutBufferSize
(TODO for encoder)
```

PR #992, opened 2021-12-10
Pipeline stage for splines

sboukortt:


lib/jxl/render_pipeline/render_pipeline_test.cc
```diff
@@ -178,6 +181,32 @@ std::vector<RenderPipelineTestInputSettings> GeneratePipelineTests() {
       all_tests.push_back(s);
     }
 
+    {
+      auto s = settings;
+      s.cparams_descr = "Splines";
+
+      const ColorCorrelationMap cmap;
```
veluca93, line 188: Maybe we can get this to a function?

veluca93 (dismissed):

lib/jxl/render_pipeline/render_pipeline_test.cc
sboukortt: Good idea, thanks.

PR #999, opened 2021-12-13
Add abseil-cpp github dependency (for: flag-parsing in C API based cjxl)

fischbacher:


fischbacher (commented):
Looks as expected.

third_party/CMakeLists.txt
```diff
@@ -111,6 +111,17 @@ include(lodepng.cmake)
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/lodepng/LICENSE"
                ${PROJECT_BINARY_DIR}/LICENSE.lodepng COPYONLY)
 
+# abseil-cpp
+if( NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/abseil-cpp/absl/base/options.h" )
+  message(FATAL_ERROR "Please run ${PROJECT_SOURCE_DIR}/deps.sh to fetch the "
+          "build dependencies.")
+endif()
+set(ABSL_BUILD_TESTING OFF CACHE BOOL "Build tests" FORCE)
```
veluca93, line 119: Why FORCE?

veluca93:
We should probably add an equivalent of LIBJXL_FORCE_SYSTEM_BROTLI also for absl, and probably modify the debian/ files and/or the CI configuration accordingly.

third_party/CMakeLists.txt
fischbacher: Should not be needed, removed.

veluca93:
Can we wait to submit this until we have support to system ABSL? (i.e. LIBJXL_FORCE_SYSTEM_ABSL)

mo271:
Seems like this is enough for using absl (tried with @fischbacher further changes for the new encoder)

data-man:
So big library for flag-parsing only?
https://github.com/CLIUtils/CLI11 seems much lighter.

vtorri:
@data-man i agree with you, and even more : parsing options of a program with str(n)cmp, or with std::string::compare() method, is just trivial.

mo271:
We decided to go with gflags instead. Thanks @data-man and @vtorri for participating by commenting.

PR #1002, opened 2021-12-13
Add an option to use a bundled version of libpng

sboukortt:


lib/jxl_extras.cmake
```diff
@@ -91,22 +91,17 @@ FetchContent_Declare(
 FetchContent_MakeAvailable(zlib libpng)
 endif()  # JPEGXL_EMSCRIPTEN
 
-find_package(ZLIB)  # dependency of PNG
-find_package(PNG)
-if(PNG_FOUND AND ZLIB_FOUND)
+if(NOT JPEGXL_BUNDLE_LIBPNG)
+  find_package(PNG)
```
jonsneyers, line 95: No need to find zlib? Or is that implied by png?

jonsneyers (dismissed):

lib/jxl_extras.cmake
sboukortt: Yes, the find module for PNG already looks for it.

veluca93 (dismissed):

veluca93 (approved):

PR #1004, opened 2021-12-13
reduce MSVC warnings a bit

jonsneyers:
~~Remove meaningless `const`s,~~ add some `.f` to float assignments, avoid comparisons between uint8 and bool etc.

Also makes all RunOnPool  data functions have the signature `void datafunc(uint32, size_t)` instead of the mix of int, uint32, uint64, size_t that we had there. They all get their actual arguments as uint32 and size_t anyway (casted to whatever was in the lambda declaration because the templating makes that work), so there's no point using different types.

### DIFF
--- a/lib/jxl/convolve.cc
+++ b/lib/jxl/convolve.cc
@@ -841,7 +841,7 @@ class ConvolveT {
     const int64_t stride = in.PixelsPerRow();
     RunOnPool(
         pool, ybegin, yend, ThreadPool::SkipInit(),
-        [&](const int y, int /*thread*/) HWY_ATTR {
+        [&](const uint32_t y, size_t /*thread*/) HWY_ATTR {
           RunRow<kSizeModN>(rect.ConstRow(in, y), rect.xsize(), stride,
                             WrapRowUnchanged(), weights, out->Row(y));
         },
@@ -858,7 +858,7 @@ class ConvolveT {
     const int64_t stride = in.PixelsPerRow();
     RunOnPool(
         pool, ybegin, yend, ThreadPool::SkipInit(),
-        [&](const int y, int /*thread*/) HWY_ATTR {
+        [&](const uint32_t y, size_t /*thread*/) HWY_ATTR {
           for (size_t c = 0; c < 3; ++c) {
             RunRow<kSizeModN>(rect.ConstPlaneRow(in, c, y), rect.xsize(),
                               stride, WrapRowUnchanged(), weights,
@@ -951,7 +951,7 @@ void Symmetric5(const ImageF& in, const Rect& rect,
   const size_t ysize = rect.ysize();
   RunOnPool(
       pool, 0, static_cast<uint32_t>(ysize), ThreadPool::SkipInit(),
-      [&](const int task, int /*thread*/) {
+      [&](const uint32_t task, size_t /*thread*/) {
         const int64_t iy = task;
 
         if (iy < 2 || iy >= static_cast<ssize_t>(ysize) - 2) {
@@ -971,7 +971,7 @@ void Symmetric5_3(const Image3F& in, const Rect& rect,
   const size_t ysize = rect.ysize();
   RunOnPool(
       pool, 0, static_cast<uint32_t>(ysize), ThreadPool::SkipInit(),
-      [&](const int task, int /*thread*/) {
+      [&](const uint32_t task, size_t /*thread*/) {
         const size_t iy = task;
 
         if (iy < 2 || iy >= ysize - 2) {
@@ -1154,7 +1154,7 @@ void SlowSymmetric3(const ImageF& in, const Rect& rect,
 
   RunOnPool(
       pool, 0, static_cast<uint32_t>(ysize), ThreadPool::SkipInit(),
-      [&](const int task, int /*thread*/) {
+      [&](const uint32_t task, size_t /*thread*/) {
         const int64_t iy = task;
         float* JXL_RESTRICT out_row = out->Row(static_cast<size_t>(iy));
 
@@ -1179,7 +1179,7 @@ void SlowSymmetric3(const Image3F& in, const Rect& rect,
 
   RunOnPool(
       pool, 0, static_cast<uint32_t>(ysize), ThreadPool::SkipInit(),
-      [&](const int task, int /*thread*/) {
+      [&](const uint32_t task, size_t /*thread*/) {
         const int64_t iy = task;
         const size_t oy = static_cast<size_t>(iy);
 
@@ -1237,7 +1237,7 @@ void SlowSeparable5(const ImageF& in, const Rect& rect,
   const size_t ysize = rect.ysize();
   RunOnPool(
       pool, 0, static_cast<uint32_t>(ysize), ThreadPool::SkipInit(),
-      [&](const int task, int /*thread*/) {
+      [&](const uint32_t task, size_t /*thread*/) {
         const int64_t y = task;
 
         float* const JXL_RESTRICT row_out = out->Row(y);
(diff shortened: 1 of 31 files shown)


lib/jxl/color_encoding_internal.cc
```diff
@@ -106,8 +106,8 @@ std::string ToString(RenderingIntent rendering_intent) {
             static_cast<uint32_t>(rendering_intent));
 }
 
-static double F64FromCustomxyI32(const int32_t i) { return i * 1E-6; }
-static Status F64ToCustomxyI32(const double f, int32_t* JXL_RESTRICT i) {
+static double F64FromCustomxyI32(int32_t i) { return i * 1E-6; }
```
veluca93, line 109: Why does VS complain about the const here? it *does* have an effect, so I'd rather disable the VS warning...

veluca93: (also in many other places)

lib/jxl/dec_frame.cc
```diff
@@ -454,7 +454,7 @@ Status FrameDecoder::ProcessDCGroup(size_t dc_group_id, BitReader* br) {
     FillImage(kInvSigmaNum / lf.epf_sigma_for_modular,
               &dec_state_->filter_weights.sigma);
   }
-  decoded_dc_groups_[dc_group_id] = true;
+  decoded_dc_groups_[dc_group_id] = 1;
```
veluca93, line 457: Can we maybe do `int{true}`?

lib/jxl/dec_frame.cc
```diff
@@ -855,7 +855,7 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
   if (has_error) return JXL_FAILURE("Error in DC group");
 
   if (*std::min_element(decoded_dc_groups_.begin(), decoded_dc_groups_.end()) ==
-          true &&
+          1 &&
```
veluca93, line 858: can also check without ==, probably clearer

lib/jxl/color_encoding_internal.cc
jonsneyers: Ah, I didn't check which ones it complained about and which ones it didn't.

In terms of code behavior, a const in function arguments that are passed by value doesn't mean anything, does it?

lib/jxl/dec_frame.cc
jonsneyers: Any reason that's not a std::vector<bool> if that's how it is semantically used anyway?

lib/jxl/color_encoding_internal.cc
veluca93: It doesn't mean anything on function *declarations*. On function definitions it's the same as declaring a variable const.

lib/jxl/dec_frame.cc
veluca93: I assume you meant vector of bool?
That would be because vector of bool is evil (it's a template specialization that doesn't behave the way a normal vector would, which can cause all sort of surprises and has IIRC been deprecated)

jonsneyers: Yes, I meant vector of bool. I didn't know it was deprecated, I was hoping it would have nice specializations for things like min_element but I guess I am too optimistic. Nevermind then. Why `int{true}` and not `static_cast<uint8_t>(true)`? We are using `1` in other places already, should they all be changed? (or the `== 1` removed?)

lib/jxl/color_encoding_internal.cc
jonsneyers: Yes, it could theoretically catch bugs if the intention is that the arg remains constant throughout the function body, but for simple one-statement functions like this one I don't really see the added value. For less trivial function bodies I agree it might be worthwhile to keep the const.

lib/jxl/dec_frame.cc
veluca93: if you're referring to the comment below, then yes, let's remove the `== 1`.

As for `int{true}`, it guarantees at compile time that there's no conversion loss, which static_cast doesn't (although it should be `uint8_t{true}`, of course)

lib/jxl/color_encoding_internal.cc
veluca93: Even so, I wouldn't remove it unless there's a good reason, and I don't think there's a good reason ;)

jonsneyers: Fair enough. Brought back the `const`s (and maybe added a few new ones in the process).

I think what MSVC was complaining about was just the `const`s in data_parallel.h, I probably misinterpreted the warning messages. See https://github.com/libjxl/libjxl/pull/1005

lib/jxl/dec_frame.cc
jonsneyers: Done. Also removed some similar `=! 1`.

lib/jxl/dec_frame.cc
jonsneyers: Yes, that does look nicer. Done.

lib/jxl/convolve.cc
```diff
@@ -858,7 +858,7 @@ class ConvolveT {
     const int64_t stride = in.PixelsPerRow();
     RunOnPool(
         pool, ybegin, yend, ThreadPool::SkipInit(),
-        [&](const int y, int /*thread*/) HWY_ATTR {
+        [&](uint32_t y, size_t /*thread*/) HWY_ATTR {
```
veluca93, line 861: You still lost a couple of consts here :)

lib/jxl/dec_frame.cc
```diff
@@ -1040,11 +1039,11 @@ int FrameDecoder::References() const {
   if (is_finalized_) {
     return 0;
   }
-  if ((!decoded_dc_global_ || !decoded_ac_global_ ||
-       *std::min_element(decoded_dc_groups_.begin(),
-                         decoded_dc_groups_.end()) != 1 ||
-       *std::min_element(decoded_passes_per_ac_group_.begin(),
-                         decoded_passes_per_ac_group_.end()) < max_passes_)) {
+  if (!(decoded_dc_global_ && decoded_ac_global_ &&
+        *std::min_element(decoded_dc_groups_.begin(),
```
veluca93, line 1043: I realize this is preexisting, but could we do std::any_of / std::all_of instead of min_element? it would be, I think, a lot clearer.

lib/jxl/enc_adaptive_quantization.cc
```diff
@@ -595,7 +595,7 @@ ImageF AdaptiveQuantizationMap(const float butteraugli_target,
         impl.PrepareBuffers(num_threads);
         return true;
       },
-      [&](const int tid, int thread) {
+      [&](uint32_t tid, size_t thread) {
```
veluca93, line 598: a const is gone from here too

veluca93 (approved):

lib/jxl/dec_frame.cc
jonsneyers: Sure. Do you happen to know a more elegant way to do that than with a lambda predicate like this?

```
std::all_of(decoded_dc_groups_.begin(), decoded_dc_groups_.end(), [&](uint8_t i) { return i; })
```

veluca93: I wouldn't [&] ;)

Otherwise, I'd just use that lambda tbh - perhaps naming it somewhere so it's clear what it is.

jonsneyers: How about just `!std::find(decoded_dc_groups_.begin(), decoded_dc_groups_.end(), 0)` ?

veluca93: also works (but `uint8_t{false}`)

jonsneyers: nevermind, I was thinking `find` returns nullptr if it didn't find it, but it doesn't

jonsneyers: I did something else to make things more readable, PTAL.

lib/jxl/convolve.cc
jonsneyers: Thanks for spotting, I think they're now all back.

veluca93 (approved):

Issue #933, opened 2021-12-01
More then 4000 Warnings when compiling with Visual Studio


sati32:
I always try to get my code warning free, to see if there is a problem.
When I now compile the lib, I get more then 4000 Warnings for the lib.
Are you interested in fixing this?
They are almost not so special, but if they would be gone, you can find real problems.
One example in ac_context.h
  JXL_DASSERT((1u << log2_covered_blocks) == covered_blocks);
warns, because 1u is 32 bit, the rest is 64 bit, better would be
  JXL_DASSERT(((size_t)1 << log2_covered_blocks) == covered_blocks);

or
    return num_ctxs * kNonZeroBuckets + kZeroDensityContextCount * block_ctx;
creats a 64 Bit value that is implicit converted to 32 bit. Better would be 
    return (uint32_t)(num_ctxs * kNonZeroBuckets + kZeroDensityContextCount * block_ctx);
to tell the compiler that you know that there will be no problem. The same problem in line 124 and 137.

To be continued.

jonsneyers:
Feel free to make a pull request to get rid of these. I agree that it would be good to get things warning-free (for clang and gcc we make warnings fatal for that reason).

sati32:
#935

PR #1005, opened 2021-12-14
simplify data_parallel.h

jonsneyers:
MSVC was complaining about the function pointers having `const &` qualifiers which are meaningless/implied. I hope this will reduce MSVC warnings (see https://github.com/libjxl/libjxl/issues/933).

Also simplifies things a bit by always returning `bool` (and not sometimes `Status` which gets lost anyway), and not doing the SkipInit() case by means of an explicit specialization with a dummy struct template param.


veluca93:
The change to SkipInit seems good, but why not always return a Status? Why would the Status get lost anyway? (it should have an attribute that makes it an error to do so)

Also, taking InitFunc/DataFunc by value can introduce unnecessary copies which could be a performance footgun.

jonsneyers:
> The change to SkipInit seems good, but why not always return a Status? Why would the Status get lost anyway? (it should have an attribute that makes it an error to do so)

The Status gets lost anyway because RunOnPool returns bool.

> Also, taking InitFunc/DataFunc by value can introduce unnecessary copies which could be a performance footgun.

As far as I understand (but my understanding may be wrong), functions are never passed by value, i.e. passing by reference is implicit.

veluca93:
On Tue, 14 Dec 2021, 10:50 Jon Sneyers, ***@***.***> wrote:

> The change to SkipInit seems good, but why not always return a Status? Why
> would the Status get lost anyway? (it should have an attribute that makes
> it an error to do so)
>
> The Status gets lost anyway because RunOnPool returns bool.
>
Then we could make RunOnPool return Status too (perhaps unless it gets
passed skipinit)

> Also, taking InitFunc/DataFunc by value can introduce unnecessary copies
> which could be a performance footgun.
>
> As far as I understand (but my understanding may be wrong), functions are
> never passed by value, i.e. passing by reference is implicit.
>
Yes, but that's not true for lambdas.

jonsneyers:
> Then we could make RunOnPool return Status too (perhaps unless it gets passed skipinit)

Any idea how to do that? I think the easiest way might be to just have a RunOnPoolSkipInit with one argument less (the skipinit is implicit) that returns void, and the one with explicit init returns Status. Does that sound reasonable?

> Also, taking InitFunc/DataFunc by value can introduce unnecessary copies which could be a performance footgun. As far as I understand (but my understanding may be wrong), functions are never passed by value, i.e. passing by reference is implicit.
> Yes, but that's not true for lambdas.

Ah OK. Then I hope the MSVC warnings were just about the `const` and not about the `&`.

veluca93:
On Tue, 14 Dec 2021, 11:00 Jon Sneyers, ***@***.***> wrote:

> Then we could make RunOnPool return Status too (perhaps unless it gets
> passed skipinit)
>
> Any idea how to do that? I think the easiest way might be to just have a
> RunOnPoolSkipInit with one argument less (the skipinit is implicit) that
> returns void, and the one with explicit init returns Status. Does that
> sound reasonable?
>
Yeah that should work. Can also do two overloads, one with skipinit and the
other one for other things, or even fancier things, but those probably need
too much SFINAE for comfort...

> Also, taking InitFunc/DataFunc by value can introduce unnecessary copies
> which could be a performance footgun. As far as I understand (but my
> understanding may be wrong), functions are never passed by value, i.e.
> passing by reference is implicit.
> Yes, but that's not true for lambdas.
>
> Ah OK. Then I hope the MSVC warnings were just about the const and not
> about the &.
>
Just a reference doesn't work either ;)
I suggest we just suppress the warning in msvc.

jonsneyers:
Updated it to keep the `const &`, suppressing the MSVC warnings about it.

Also added a `void RunOnPoolSkipInit` so the one where init isn't skipped can return `Status`. Updated all the places where it is called accordingly.

This will help in case we change some of those init functions to actually fail, e.g. because they couldn't allocate the per-thread storage. There were quite a few places where no-skip-init return values were ignored, and places where skip-init return values were checked (while they're by definition true).

lib/jxl/base/data_parallel.h
```diff
@@ -104,21 +97,34 @@ class ThreadPool {
...
+  }
+}
+template <class DataFunc>
+void RunOnPoolSkipInit(ThreadPool* pool, const uint32_t begin,
+                       const uint32_t end, const DataFunc& data_func,
+                       const char* caller) {
   if (pool == nullptr) {
     ThreadPool default_pool(nullptr, nullptr);
-    ret = default_pool.Run(begin, end, init_func, data_func, caller);
+    JXL_CHECK(default_pool.Run(begin, end, ThreadPool::TrivialInit, data_func,
```
veluca93, line 117: So... how do we know that the no-init version cannot fail? Does the C API for threading state that this is the case?

jonsneyers: Only the init returns something. The data function has a void return type.

I don't know what should happen if someone implements an incorrect JxlParallelRunner that returns failure even if the init function returned success. This check will then fail.

veluca93: What if the parallel runner fails for internal reasons?

jonsneyers: It's not clear to me what the actual api is for parallel runners. If it can fail for internal reasons, then we have a bunch of places where this is not handled (none of the calls to RunOnPoolSkipInit check the result). So I was assuming that this cannot happen...

veluca93: Then the correct solution would be to handle those errors, or to clarify in the documentation of the thread pool API that this should not happen.

jonsneyers: Which option do you prefer? Currently it looks like the thread pool API allows a runner to fail for other reasons than the init failing, but none of our code with SkipInit is handling that possibility.

I'm inclined to say that the thread pool API should be changed so it cannot fail if init succeeds - otherwise it would be a technically valid implementation to just always fail, or to have unreasonable requirements like only working if 256 threads are available. I think it's better to make it so that a thread runner can only return failure if init failed, even if that means it has to fall back to just running things sequentially.

veluca93:
I don't think it would be a big burden to say that a thread pool is allowed
to fail for whatever reason. Of course, if it does fail, then the whole
decoding process will fail too.

jonsneyers:
> I don't think it would be a big burden to say that a thread pool is allowed to fail for whatever reason. Of course, if it does fail, then the whole decoding process will fail too.

OK, so I add JXL_RETURN_IF_FAILURE to all the `RunOnPoolSkipInit`s then?

veluca93:
Yeah something like that.

jonsneyers:
> Yeah something like that.

OK, did that. This might be a first small step in the direction of being able to detect OOM and gracefully return an error instead of aborting - though we'll need a lot more Status plumbing for that, I guess (and checking if allocation failed).

lib/jxl/base/data_parallel.h
```diff
@@ -104,21 +97,33 @@ class ThreadPool {
...
+    ThreadPool default_pool(nullptr, nullptr);
+    return default_pool.Run(begin, end, init_func, data_func, caller);
+  } else {
+    return pool->Run(begin, end, init_func, data_func, caller);
+  }
+}
+template <class DataFunc>
+Status RunOnPoolSkipInit(ThreadPool* pool, const uint32_t begin,
+                         const uint32_t end, const DataFunc& data_func,
+                         const char* caller) {
```
veluca93, line 114: I believe now we can just have this call RunOnPool without duplicating code :)

(although perhaps at this point I would remove this function entirely...)

tools/hdr/tone_map.cc
```diff
@@ -1,89 +1,89 @@
...
-    c_out.tf.SetTransferFunction(jxl::TransferFunction::kPQ);
-  } else {
-    c_out.tf.SetTransferFunction(jxl::TransferFunction::k709);
-  }
-  JXL_CHECK(c_out.CreateICC());
-  JXL_CHECK(image.TransformTo(c_out, jxl::GetJxlCms(), &pool));
-  image.metadata.m.color_encoding = c_out;
-  JXL_CHECK(jxl::EncodeToFile(image, output_filename, &pool));
-}
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
```
veluca93, line 1: What's up with this file?

jonsneyers: The line endings were CR LF for all files in this directory, so I did a `dos2unix` on all of them because my editor shows the CR visibly as `^M` which annoyed me :)

lib/jxl/base/data_parallel.h
jonsneyers: Right. Removed the function.

tools/hdr/tone_map.cc
veluca93: Yeeeeah... Good idea.

veluca93 (approved):

Issue #871, opened 2021-11-19
New tool request: jxltran


jonsneyers:
It would be nice to have a `jxltran` tool similar to `jpegtran` that can perform some operations on existing jxl images that don't involve decoding / encoding, but things like changing some header fields, stripping or adding metadata, etc.

Perhaps some of that functionality is already or should be exposed through the API, and can be implemented that way, e.g. anything related to box handling. For changing codestream header fields, it might be trickier to do that via the API (we would need new concepts in the API for that), but it would still be useful to have a way to do that.

Examples of useful codestream level functionality:
- change the orientation
- change the intrinsic dimensions
- do a reversible crop (i.e. change the image dimensions while keeping frames intact, only updating their crop size and offset)
- add, remove, or change the photon noise
- sharpen or blur the image by adjusting its gaborish weights
- change the nominal color space / bit depth (in case of XYB encoded images)
- extract the ICC profile
- extract keyframes from an animation
- get or set the names of frames and extra channels
- change the FPS and/or duration of individual frames

Examples of useful file format level functionality:
- get, set or strip exif/xmp metadata
- strip the jbrd box
- compress or decompress exif/xmp to/from brob boxes
- convert between naked codestream and container

Most of the codestream level functionality is probably somewhat tricky to implement and is likely something we don't really want to expose in the API since it involves too much of the "guts" of jxl. However, the things that involve only the _image_ header (not the frame headers) seem like they shouldn't be too hard to do, and it might be worthwhile to have API functions to update those fields (maybe not for 1.0 though). I'm thinking in particular about orientation, intrinsic dimensions, and animation speed (FPS). In particular, for feature parity with the existing approach for orientation (doing it via Exif, with many 'viewers' actually implementing a way for users to change orientation too), I think it would be good to have a way in the API to do that.

jonsneyers:
I forgot I already opened a similar issue: https://github.com/libjxl/libjxl/issues/477

Some more ideas are there. Closing that one though, one issue is enough.

ziemek99:
I wrote about a bunch of ideas in that issue and I'll just link to my own [comment](https://github.com/libjxl/libjxl/issues/477#issuecomment-910965583) in order not to repeat myself 😉

schmorp:
[...]

PR #1006, opened 2021-12-14
add support for encoding cropped frames

jonsneyers:
Adds support in the encode API to encode frames with have_crop.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -369,9 +369,10 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
       to.alpha_channel = from.alpha;
     }
 
-    // TODO(lode): also handle have_crop and the cropping dimensions, this
-    // requires getting the pixel data from the user as a smaller image.
-
+    if (input_frame->option_values.header.layer_info.have_crop) {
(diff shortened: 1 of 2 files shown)


mo271 (approved):

ziemek99:
Any chance to add it as a `cjxl` option?

jonsneyers:
> Any chance to add it as a `cjxl` option?

This is mostly intended for multi-frame encoding where frames can be smaller than the full canvas. Currently cjxl doesn't use the api but it already does cropped frame encoding when it gets apng, gif or psd input that has such frames.

Perhaps in a future api-using cjxl we could also add a way to explicitly encode multiple inputs, either as overlays or as an animation. Then it would be useful to have some syntax to give the offsets and durations of each frame. But for now I wouldn't add more stuff to cjxl, let's first migrate it to use the api...

ziemek99:
> But for now I wouldn't add more stuff to cjxl, let's first migrate it to use the api...

Ok, I don't want to interfere with that since `cjxl` really deserves it 😉

> This is mostly intended for multi-frame encoding where frames can be smaller than the full canvas.

I've thought more about single-frame simple `jxl_from_tree`-like cropping, an equivalent of its `FramePos` command. Could be extracted to `jxltran` in the future (#871).

PR #1014, opened 2021-12-15
fix MSYS test and speedup some more tests

jonsneyers:
Fix the test failing on MSYS2.
Also makes some tests faster by not doing unnecessary iterations or by doing sampling of the whole range instead of exhaustively doing (part of) it.

### DIFF
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -863,7 +863,7 @@ TEST(JxlTest, RoundtripAlphaResamplingOnlyAlpha) {
   CodecInOut io2;
   EXPECT_TRUE(DecodeFile(dparams, compressed, &io2, pool));
 
-  EXPECT_LE(compressed.size(), 34000u);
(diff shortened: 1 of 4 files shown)


lib/jxl/jxl_test.cc
```diff
@@ -863,7 +863,7 @@ TEST(JxlTest, RoundtripAlphaResamplingOnlyAlpha) {
   CodecInOut io2;
   EXPECT_TRUE(DecodeFile(dparams, compressed, &io2, pool));
 
-  EXPECT_LE(compressed.size(), 34000u);
+  EXPECT_LE(compressed.size(), 34020u);
```
veluca93, line 866: Can we be a bit less tight? :D

jonsneyers: Made it a bit less tight, but I would prefer to keep the margin small so we know when density regressions are happening.

veluca93 (approved):

PR #1016, opened 2021-12-15
ensure at least 2 mantissa bits

jonsneyers:
Spec says the number of mantissa bits has to be in the range [2, 23] — so including the sign bit, there have to be at least 3 bits besides the exponent bits.


mo271 (approved):
Nice to keep in sync with the spec! 
Good to have this check only in one place now..

Issue #1017, opened 2021-12-15
jxl_threads: There is not point in exposing the underlying c++ symbols


malaterre:
**Describe the bug**

jxl library properly uses a jxl.version script to hide some of the c++ symbols. This is not the case for jxl_threads. Please apply similar logic.

**To Reproduce**

When building on amd64, the c++ symbols being exported depends on the libstdc++ version:

* https://buildd.debian.org/status/fetch.php?pkg=jpeg-xl&arch=amd64&ver=0.6.1%2Bds-4&stamp=1639576462&raw=0

```
dpkg-gensymbols: error: some symbols or patterns disappeared in the symbols file: see diff output below
dpkg-gensymbols: warning: debian/libjxl0.6/DEBIAN/symbols doesn't match completely debian/libjxl0.6.symbols
--- debian/libjxl0.6.symbols (libjxl0.6_0.6.1+ds-4_amd64)
+++ dpkg-gensymbolsHmRRaw	2021-12-15 13:54:15.234900884 +0000
@@ -79,7 +79,7 @@
  JxlThreadParallelRunnerDefaultNumWorkerThreads@Base 0.6.1
  JxlThreadParallelRunnerDestroy@Base 0.6.1
  _ZNSt6vectorISt6threadSaIS0_EE17_M_default_appendEm@Base 0.6.1
- _ZNSt6vectorISt6threadSaIS0_EE17_M_realloc_insertIJRFvPN6jpegxl20ThreadParallelRunnerEiES6_RjEEEvN9__gnu_cxx17__normal_iteratorIPS0_S2_EEDpOT_@Base 0.6.1
+#MISSING: 0.6.1+ds-4# _ZNSt6vectorISt6threadSaIS0_EE17_M_realloc_insertIJRFvPN6jpegxl20ThreadParallelRunnerEiES6_RjEEEvN9__gnu_cxx17__normal_iteratorIPS0_S2_EEDpOT_@Base 0.6.1
  _ZTINSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPN6jpegxl20ThreadParallelRunnerEiES5_jEEEEEE@Base 0.6.1
  _ZTSNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPN6jpegxl20ThreadParallelRunnerEiES5_jEEEEEE@Base 0.6.1
[...]

PR #1018, opened 2021-12-15
jxl_threads: Hide all the std namespace symbols.

malaterre:
Closes issue (#1017)

### DIFF
--- a/lib/jxl_threads.cmake
+++ b/lib/jxl_threads.cmake
@@ -74,6 +74,18 @@ set_target_properties(jxl_threads PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
+  set_target_properties(jxl_threads PROPERTIES
+      LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/jxl/jxl.version)
+  if(APPLE)
+  set_property(TARGET ${target} APPEND_STRING PROPERTY
+      LINK_FLAGS "-Wl,-exported_symbols_list,${CMAKE_CURRENT_SOURCE_DIR}/jxl/jxl_osx.syms")
+  elseif(WIN32)
+    # Nothing needed here, we use __declspec(dllexport) (jxl_threads_export.h)
+  else()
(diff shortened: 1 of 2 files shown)


lib/jxl_threads.cmake
```diff
@@ -74,6 +74,16 @@ set_target_properties(jxl_threads PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
+  set_target_properties(jxl_threads PROPERTIES
+      LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/jxl/jxl_threads.version)
+  if(APPLE)
```
veluca93, line 79: I understand little about this, but the non-threads version has something for OSX here - why is an equivalent line not here?

veluca93 (dismissed):

lib/jxl_threads.cmake
malaterre: > I understand little about this, but the non-threads version has something for OSX here - why is an equivalent line not here?

Pay attention that I only tested the UNIX portion of the patch, no OSX system here.

malaterre: Also someone *must* review my patch. This is not entirely clear what the long term is with regards to `jxl_threads` vs simply `jxl`. Someone may suggest this is fine to use a single copy of this file (ie. re-use `jxl.version` against `jxl_threads`).

veluca93: Yup, I added the CI:full label which should trigger a run on OSX too.

As for reusing jxl.version - can you see any reason not to? In principle I'd be in favour of reusing it if possible.

malaterre: > As for reusing jxl.version - can you see any reason not to? In principle I'd be in favour of reusing it if possible.

I was influenced by these for the decision:

```
% find . | grep _export
./obj-x86_64-linux-gnu/lib/include/jxl/jxl_threads_export.h
./obj-x86_64-linux-gnu/lib/include/jxl/jxl_export.h
```

which explains why I picked the original solution...anyway I've updated my patch.

veluca93 (approved):

veluca93 (approved):

PR #1019, opened 2021-12-15
Wire up modular mode for outputting to the render pipeline.

veluca93:
Also wire up DC upscaling.

### DIFF
--- a/lib/jxl/dec_modular.cc
+++ b/lib/jxl/dec_modular.cc
@@ -50,36 +50,26 @@ void MultiplySum(const size_t xsize,
 
 void RgbFromSingle(const size_t xsize,
                    const pixel_type* const JXL_RESTRICT row_in,
-                   const float factor, Image3F* decoded, size_t /*c*/, size_t y,
-                   Rect& rect) {
-  JXL_DASSERT(xsize <= rect.xsize());
+                   const float factor, float* out_r, float* out_g,
+                   float* out_b) {
   const HWY_FULL(float) df;
   const Rebind<pixel_type, HWY_FULL(float)> di;  // assumes pixel_type <= float
-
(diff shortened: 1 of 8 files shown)


jonsneyers:
Does this work also for alpha/extra channels, or is this step just making it work for the no-VarDCT RGB case?

veluca93:
> Does this work also for alpha/extra channels, or is this step just making it work for the no-VarDCT RGB case?

Extra channels coming up soon :)

lib/jxl/dec_modular.cc
```diff
@@ -622,8 +636,34 @@ Status ModularFrameDecoder::FinalizeDecoding(PassesDecoderState* dec_state,
 
   auto& decoded = dec_state->decoded;
 
-  JXL_RETURN_IF_ERROR(
-      ModularImageToDecodedRect(gi, dec_state, pool, output, Rect(decoded)));
+  if (dec_state->render_pipeline) {
+    std::atomic<size_t> num_errors{0};
```
sboukortt, line 640: [In the past](https://github.com/libjxl/libjxl/pull/860/files#diff-c7ffee7b87f9e62e559021b02af4ddbeae7d920d0ac7bde1a3b8322e4a9319adR45), I have used a bool for this. Is a size_t preferable? It does maybe seem a bit lighter than the bool notation-wise.

sboukortt (dismissed):

jonsneyers (approved):

lib/jxl/dec_modular.cc
veluca93: Done.

veluca93:
Had to modify the code to deal with GetInputBuffers() multiple times, PTAL.

jonsneyers (approved):

PR #1020, opened 2021-12-15
don't ignore JxlPixelFormat.align when encoding

jonsneyers:
Row alignment (stride) was ignored when passing an image buffer in JxlEncoderAddImageFrame. This should fix that.

Also check if the buffer size has the right size: currently it was only checking if it is large enough, but if it's too large then it's better to return an error than to just silently ignore the additional bytes — most likely this is API usage error.

Fix one of the tests where we had such an API usage error (interpreting a 16-bit input buffer as 8-bit), which didn't really matter for that particular test but still...

### DIFF
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -198,8 +198,8 @@ PaddedBytes CreateTestJXLCodestream(
   io.metadata.m.color_encoding = color_encoding;
   EXPECT_TRUE(ConvertFromExternal(
       pixels, xsize, ysize, color_encoding, /*has_alpha=*/include_alpha,
-      /*alpha_is_premultiplied=*/false, bitdepth, JXL_BIG_ENDIAN,
-      /*flipped_y=*/false, &pool, &io.Main(), /*float_in=*/false));
+      /*alpha_is_premultiplied=*/false, /*bits_per_sample=*/16, JXL_BIG_ENDIAN,
+      /*flipped_y=*/false, &pool, &io.Main(), /*float_in=*/false, /*align=*/0));
   jxl::PaddedBytes jpeg_data;
   if (jpeg_codestream != nullptr) {
 #if JPEGXL_ENABLE_JPEG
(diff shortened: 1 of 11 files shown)


lib/jxl/decode_test.cc
```diff
@@ -203,8 +203,8 @@ PaddedBytes CreateTestJXLCodestream(
   io.metadata.m.color_encoding = color_encoding;
   EXPECT_TRUE(ConvertFromExternal(
       pixels, xsize, ysize, color_encoding, /*has_alpha=*/include_alpha,
-      /*alpha_is_premultiplied=*/false, bitdepth, JXL_BIG_ENDIAN,
-      /*flipped_y=*/false, &pool, &io.Main(), /*float_in=*/false));
+      /*alpha_is_premultiplied=*/false, 16, JXL_BIG_ENDIAN,
+      /*flipped_y=*/false, &pool, &io.Main(), /*float_in=*/false, 0));
```
veluca93, line 207: Can you add /*comments=*/ about numerical arguments?

lib/jxl/enc_external_image.cc
```diff
@@ -124,10 +125,19 @@ Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
...
-  if (ysize && bytes.size() / ysize < row_size) {
+  const size_t last_row_size = xsize * bytes_per_pixel;
+  const size_t row_size =
+      (align > 1 ? jxl::DivCeil(last_row_size, align) * align : last_row_size);
+  const size_t bytes_to_read = row_size * (ysize - 1) + last_row_size;
+  if (xsize == 0 || ysize == 0) return JXL_FAILURE("Empty image");
+  if (bytes.size() < bytes_to_read) {
     return JXL_FAILURE("Buffer size is too small");
   }
+  // Too large buffer is likely an application bug, so also fail for that.
```
veluca93, line 136: I'm not sure I like this idea - I could imagine someone using a larger buffer which is virtually cropped and things getting more complicated. At the very least, I think this should happen in a separate PR.

jonsneyers: If you want to use a larger buffer and virtually crop it, you can always pass a smaller size than its actual size. This only works anyway if you want to crop away rows at the bottom — any other kind of crop requires you to do the math and compute the new pointer (to the crop origin), xsize /ysize and alignment (i.e. the original row_size).

Making it slightly easier to crop rows at the bottom (a rather specific thing) at the expense of losing an opportunity to catch API usage bugs doesn't seem to be worth it to me, but I agree that this is arguable so I'll make a separate PR for that.

lib/jxl/decode_test.cc
jonsneyers: done

lib/jxl/enc_external_image.cc
jonsneyers: split this part out

mo271:
Needs rebase after https://github.com/libjxl/libjxl/pull/1022 is merged.

jonsneyers:
> Needs rebase after #1022 is merged.

Actually it's easier if this one merges first and then #1022 gets rebased, I think.

veluca93 (approved):

PR #1022, opened 2021-12-15
enc buffer api: check if buffer size matches

jonsneyers:
Follow-up on https://github.com/libjxl/libjxl/pull/1020: check that the buffer size matches what is expected, don't just check if it's large enough. If the size doesn't match, it's likely an API usage error (e.g. interpreting an RGBA buffer as RGB) so it's more useful to return a meaningful error in this case than to cause head scratching for the application developer.

Also updating documentation to clarify that the size has to match.

The only degree of freedom this check is leaving, is the padding at the end of the last row in case `align` is being used. You can choose whether to pass a buffer with or without that padding.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -531,7 +531,8 @@ JxlEncoderAddJPEGFrame(const JxlEncoderFrameSettings* frame_settings,
  * contents are copied internally.
  * @param buffer buffer type to input the pixel data from. Owned by the caller
  * and its contents are copied internally.
- * @param size size of buffer in bytes.
+ * @param size size of buffer in bytes. This size should match what is implied
+ * by the frame dimensions and the pixel format.
  * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error
  */
 JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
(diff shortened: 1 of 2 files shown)


jonsneyers:
Opinions on this one please. I think it's a somewhat classic "detect bugs" vs "allow you to shoot yourself in the foot" thing.

mo271 (approved):
I do like the possibility to allow the align argument. Do I understand correctly, that because of a check comparing with the stride, the align argument cannot be used to circumvent the check for matching buffer sizes? I'm not sure if there is a use case for having non-matching buffer sizes, and the case helping out users, who accidently pass and RGBA buffer instead of RGB or similar problems seems realistic.

mo271:
linking @veluca93's comment: https://github.com/libjxl/libjxl/pull/1020#discussion_r769787329

jonsneyers:
Well if you'd accidentally pass an RGBA buffer (with a size equal to the full buffer size) and you say it is RGB, then currently that will result in a garbled images consisting of the top 75% of the buffer misinterpreted as RGB. This check would instead make it an API error.

If you accidentally pass an RGBA buffer and you also set the pixelformat to RGB but with an align that corresponds to an RGBA row, then this check will not complain and it will result in a garbled image of the left 75% of the buffer misinterpreted as RGB. But that's maybe an programming error that is less likely, since it means putting inconsistent things in the pixelformat. The error of passing a buffer that corresponds to the basicinfo instead of the pixelformat is more likely to happen in the wild, imo.

mo271 (approved):

PR #1023, opened 2021-12-15
Add support for extra channels in the rendering pipeline.

veluca93:


lib/jxl/dec_frame.cc
```diff
@@ -714,21 +721,53 @@ void FrameDecoder::PreparePipeline() {
...
+
   if (frame_header_.upsampling != 1) {
     for (size_t c = 0; c < 3; c++) {
       builder.AddStage(GetUpsamplingStage(
           frame_header_.nonserialized_metadata->transform_data, c,
           CeilLog2Nonzero(frame_header_.upsampling)));
     }
   }
+
+  if (late_ec_upsample) {
```
jonsneyers, line 760: Could also integrate it in the loop above, e.g. by having a variable `late_upsample_nb_chans` instead of `late_ec_upsample`, which is equal to 3 in the early ec upsampling case and 3+nb_extra_chans in the other case. In the after-patches upsampling, the upsampling factor is the same anyway...

lib/jxl/dec_modular.cc
```diff
@@ -589,19 +590,33 @@ Status ModularFrameDecoder::ModularImageToDecodedRect(
     const size_t ec_ysize = DivCeil(frame_dim.ysize_upsampled, ecups);
     JXL_ASSERT(c < gi.channel.size());
     Channel& ch_in = gi.channel[c];
-    // For x0, y0 there's no need to do a DivCeil().
-    JXL_DASSERT(rect.x0() % (1ul << ch_in.hshift) == 0);
-    JXL_DASSERT(rect.y0() % (1ul << ch_in.vshift) == 0);
+    if (!render_pipeline_input) {
```
jonsneyers, line 593: Why make the DASSERTs conditional?

jonsneyers (dismissed):

lib/jxl/dec_modular.cc
veluca93: Because `rect` is not used and may be invalid in the render_pipeline_input case :)

lib/jxl/dec_frame.cc
veluca93: How about this?

jonsneyers (approved):

PR #1028, opened 2021-12-17
avoid unnecessary copy from modular full image to modular group

jonsneyers:
Generalizes the `Channel` class in modular so it can be either a self-owned Plane, or a Rect inside an externally-owned Plane. Obviously in the second case you're not working on a copy so you have to be careful when modifying the pixel values.

In the encoder, this saves an unnecessary copy from the full modular image to the group images: instead of a copy, the group images just become rects of the full image, which is fine since the full image isn't needed anymore anyway (so it's OK if the group images still get modified afterwards, e.g. by doing a local Palette or a local RCT transform).

Some speed results for `cjxl ClassA_8bit_p06.ppm.png -m --num_reps=30 -e [effort]`

| effort | before https://github.com/libjxl/libjxl/pull/946 | current | after this PR |
|-------|--------|------|-------|
| 1 (lightning) | 25.91 MP/s | 36.79 MP/s | 41.67 MP/s
| 2 (thunder) | 21.44 MP/s | 26.10 MP/s | 31.88 MP/s
| 3 (falcon) | 14.55 MP/s | 17.47 MP/s | 19.82 MP/s


veluca93:
This seems quite bug-prone - can we make it so that we don't have the two options, and always use externally-owned buffers?

jonsneyers:
> This seems quite bug-prone - can we make it so that we don't have the two options, and always use externally-owned buffers?

That could be done, but it involves a lot more code changes. Having the actual Plane private should avoid most bugs; in a follow-up MR maybe we can get rid of the GetPlane() so we don't need to worry that there are code paths left where its assert would actually fail.

veluca93:
I'm worried about mixing owning semantics and borrowing semantics in the same class - that sounds like it will lead to segmentation faults very easily.

jonsneyers:
> I'm worried about mixing owning semantics and borrowing semantics in the same class - that sounds like it will lead to segmentation faults very easily.

Agreed, it's not really great design. But making it always-borrowing at this point means we'll have to find some other place that owns it, and I don't see a good/convenient place for that without massive code changes. Letting modular `Image` own the planes doesn't really help, that just moves the problem.

eustas:
If we plan to revive this PR in the future, it could be split in 2 parts:
 - adding and using Channel::Xyz that proxies channel.plane.Xyz; that would be logical no-op and picks substantial part of this PR
 - everything else; this could be done later when we have time to solve ownership puzzle

PR #1030, opened 2021-12-20
disable -Werror

mo271:
When building with msan, we disable `-Werror`, in order to fix tests for #999.


mo271:
With a system absl (libabsl-dev), soon to be added to #999, it should be possible to have `Werror`, so it might be good to keep some targets with that on in the pipeline.

veluca93 (approved):

PR #1032, opened 2021-12-20
Standalone fast-lossless JXL encoder

veluca93:
Personal experiment to see how fast we can make a JPEG XL lossless encoder that still has good density.


experimental/fast_lossless/fast_lossless.cc
```diff
@@ -0,0 +1,665 @@
...
+         ((static_cast<uint16_t>(~value) >> 15) - 1);
+}
+
+struct ChannelRowEncoder {
+  inline void ProcessChunk(const int16_t* row, const int16_t* row_left,
+                           const int16_t* row_top, const int16_t* row_topleft,
+                           size_t chunk_size, BitWriter& output) {
+    if (chunk_size == 0) return;
+    bool continue_rle = true;
+    alignas(32) uint16_t residuals[16] = {};
```
jonsneyers, line 516: Is that initialization useful/needed?

experimental/fast_lossless/fast_lossless.cc
```diff
@@ -0,0 +1,665 @@
...
+    for (size_t ix = 0; ix < chunk_size; ix++) {
+      int16_t px = row[ix];
+      int16_t left = row_left[ix];
```
jonsneyers, line 519: Might be better to use row[ix-1] here, no?
Same with topleft, which should always be the same as top[ic -1].
Maybe the compiler generates more efficient code if it knows that than if the values could come from 4 different places...

experimental/fast_lossless/fast_lossless.cc
veluca93: It doesn't change anything for the 16 case (the compiler doesn't even emit it), and it helps avoiding weird memory in the other cases (which are anyway extremely rare), so I'd keep it

experimental/fast_lossless/fast_lossless.cc
veluca93: Ah, but it's not, for row 0 ;)
Either way the emitted code is pretty nice as is so I wouldn't bother.

jonsneyers: Ah right, for row 0 topleft is a bit weird.
Left is still always current-1 though.
But probably doesn't matter.

PR #1037, opened 2021-12-21
Use bundled png for WASM

eustas:


veluca93 (approved):

eustas:
Finally, it works. There is a problem with LTO that sweeps-out setjmp; disabled it.
PTAL

veluca93 (approved):

PR #1039, opened 2021-12-22
add gflags

mo271:


mo271:
The failing CiFuzz should be fixed by adding the corresponding dependency in https://github.com/google/oss-fuzz/blob/master/projects/libjxl/Dockerfile. Will do as soon as this is merged

mo271:
I proposed using this instead of abseil-cpp, so to close #999.

third_party/CMakeLists.txt
```diff
@@ -153,6 +153,28 @@ if (JPEGXL_ENABLE_VIEWERS OR NOT JPEGXL_ENABLE_SKCMS)
...
+  add_subdirectory(gflags)
+  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/gflags/COPYING.txt"
+                 ${PROJECT_BINARY_DIR}/LICENSE.gflags COPYONLY)
+else()
+  # Use sytem libgflags-dev
+  find_package(gflags)
+  if (NOT gflags_FOUND)
+    message(FATAL_ERROR
+      "Gflags not found, install libgflags-dev or download gflags source to"
+      "third_partygflags from https://github.com/gflags/gflags. You can use"
```
veluca93, line 169: Typo

veluca93 (approved):

third_party/CMakeLists.txt
mo271: fixed

veluca93 (approved):

mo271:
working on fixing the CIFuzz check: https://github.com/google/oss-fuzz/pull/7063

.github/workflows/release.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
           doxygen \
           libbrotli-dev \
           libgdk-pixbuf2.0-dev \
+          libgflag-dev \
```
eustas, line 38: Should it be "libgflags-dev"?

mo271: yes.
Thanks for fixing it in https://github.com/libjxl/libjxl/commit/a46cbcc65ca1f22820c25343330a441cf9f658e3 !

PR #1043, opened 2021-12-22
Fix cjxl for non-JPEG input

sboukortt:


sboukortt:
Not sure how I didn’t catch this…

veluca93 (approved):

veluca93:
> Not sure how I didn’t catch this…

The more interesting question is, why did our CI not catch it? Do we have no tests at all for the tools?

PR #1045, opened 2021-12-23
more targets in debug_ci

mo271:
We allow underscores in the name of between `ci-` and `-debug`, so that
it is possible to also run `./ci.sh msan_install` and other targets that
include an underscore.

### DIFF
--- a/.github/workflows/debug_ci.yml
+++ b/.github/workflows/debug_ci.yml
@@ -48,7 +48,7 @@ jobs:
         fetch-depth: 2
     - name: Build
       run: |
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

mo271:
running `ci-msan-debug` takes a long time to before one can login via ssh and running `ci--debug` fails because of the `-DJPEGXL_FORCE_SYSTEM_BROTLI=ON` argument, so for faster msan build debugging naming the branch `ci-msan_install-debug` would be desirable.

PR #1046, opened 2021-12-23
make decoding of fjxl-produced images a bit faster

jonsneyers:
Special modular decode case for RLE-only huffman decoding with gradient predictor (i.e. what fjxl currently produces).

before:
1200 x 1400, geomean: 22.59 MP/s [14.46, 25.23], 100 reps, 0 threads.
after:
1200 x 1400, geomean: 34.35 MP/s [19.44, 37.91], 100 reps, 0 threads.

There's still a lot of room for further speed improvements by specializing for 16-bit modular buffers, avoiding copying (should be better with render-pipeline), SIMD'ing YCoCg, and avoiding conversion to float.

### DIFF
--- a/lib/jxl/dec_ans.h
+++ b/lib/jxl/dec_ans.h
@@ -275,6 +275,33 @@ class ANSSymbolReader {
     return static_cast<uint32_t>(ret);
   }
 
+  // Takes a *clustered* idx. Can only use if HuffRleOnly() is true.
+  void ReadHybridUintClusteredHuffRleOnly(size_t ctx,
+                                          BitReader* JXL_RESTRICT br,
+                                          uint32_t* value, uint32_t* run) {
+    JXL_DASSERT(HuffRleOnly());
+    br->Refill();  // covers ReadSymbolWithoutRefill + PeekBits
+    size_t token = ReadSymbolHuffWithoutRefill(ctx, br);
+    if (JXL_UNLIKELY(token >= lz77_threshold_)) {
+      *run =
+          ReadHybridUintConfig(lz77_length_uint_, token - lz77_threshold_, br) +
+          lz77_min_length_ - 1;
+      return;
+    }
+    *value = ReadHybridUintConfig(configs[ctx], token, br);
+    return;
+  }
+  bool HuffRleOnly() {
+    if (lz77_window_ == nullptr) return false;
+    if (!use_prefix_code_) return false;
+    for (size_t i = 0; i < kHuffmanTableBits; i++) {
+      if (huffman_data_[lz77_ctx_].table_[i].bits) return false;
+      if (huffman_data_[lz77_ctx_].table_[i].value != 1) return false;
+    }
+    if (configs[lz77_ctx_].split_token > 1) return false;
+    return true;
+  }
+
   // Takes a *clustered* idx.
   size_t ReadHybridUintClustered(size_t ctx, BitReader* JXL_RESTRICT br) {
     if (JXL_UNLIKELY(num_to_copy_ > 0)) {
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -197,6 +197,39 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
         }
       }
     } else if (predictor == Predictor::Gradient && offset == 0 &&
+               multiplier == 1 && reader->HuffRleOnly()) {
+      JXL_DEBUG_V(8, "Gradient RLE (fjxl) very fast track.");
+      uint32_t run = 0;
+      uint32_t v = 0;
+      pixel_type_w sv = 0;
+      for (size_t y = 0; y < channel.h; y++) {
+        pixel_type *JXL_RESTRICT r = channel.Row(y);
+        const pixel_type *JXL_RESTRICT rtop = (y ? channel.Row(y - 1) : r - 1);
+        const pixel_type *JXL_RESTRICT rtopleft =
+            (y ? channel.Row(y - 1) - 1 : r - 1);
+        pixel_type_w guess = (y ? rtop[0] : 0);
+        if (run == 0) {
+          reader->ReadHybridUintClusteredHuffRleOnly(ctx_id, br, &v, &run);
+          sv = UnpackSigned(v);
+        } else {
+          run--;
+        }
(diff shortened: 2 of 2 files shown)


lib/jxl/dec_ans.h
```diff
@@ -275,6 +275,37 @@ class ANSSymbolReader {
...
+  // Takes a *clustered* idx. Can only use if RleOnly() is true.
+  void ReadHybridUintClusteredRleOnly(size_t ctx, BitReader* JXL_RESTRICT br,
+                                      uint32_t* value, uint32_t* run) {
+    br->Refill();  // covers ReadSymbolWithoutRefill + PeekBits
+    size_t token = ReadSymbolHuffWithoutRefill(ctx, br);
+    if (JXL_UNLIKELY(token >= lz77_threshold_)) {
+      num_to_copy_ =
+          ReadHybridUintConfig(lz77_length_uint_, token - lz77_threshold_, br) +
+          lz77_min_length_;
+      // br->Refill();  // covers ReadSymbolWithoutRefill + PeekBits
```
veluca93, line 287: Can we remove those?

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -196,6 +196,61 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
+               multiplier == 1 && reader->RleOnly()) {
+      JXL_DEBUG_V(8, "Gradient RLE (fjxl) very fast track.");
+      const intptr_t onerow = channel.plane.PixelsPerRow();
+      uint32_t run = 0;
+      uint32_t v = 0;
+      pixel_type sv = 0;
+      for (size_t y = 0; y < channel.h; y++) {
+        pixel_type *JXL_RESTRICT r = channel.Row(y);
+
+#if true
```
veluca93, line 209: I don't think we should have this :)

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -196,6 +196,61 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
+        const pixel_type *JXL_RESTRICT rtop = (y ? channel.Row(y - 1) : r - 1);
+        const pixel_type *JXL_RESTRICT rtopleft =
+            (y ? channel.Row(y - 1) - 1 : r - 1);
+        {
+          size_t x = 0;
```
veluca93, line 214: Maybe clearer to remove the conditionals here

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -196,6 +196,61 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
+          pixel_type left = (x ? r[x - 1] : y ? *(r + x - onerow) : 0);
+          pixel_type top = (y ? *(r + x - onerow) : left);
+          pixel_type topleft = (x && y ? *(r + x - 1 - onerow) : left);
+          pixel_type guess = ClampedGradient(top, left, topleft);
+          if (run == 0) {
+            reader->ReadHybridUintClusteredRleOnly(ctx_id, br, &v, &run);
+            sv = UnpackSigned(v);
+          } else
```
veluca93, line 222: braces :)

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -196,6 +196,61 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
+          pixel_type left = r[x - 1];
+          pixel_type top = rtop[x];
+          pixel_type topleft = rtopleft[x];
+          pixel_type guess = ClampedGradient(top, left, topleft);
+          if (!run) {
+            reader->ReadHybridUintClusteredRleOnly(ctx_id, br, &v, &run);
+            sv = UnpackSigned(v);
+          } else
+            run--;
+          r[x] = sv + guess;
```
veluca93, line 236: I think this may risk overflowing, which (a) UBSAN would complain about (b) is UB...

jonsneyers: I think only (a) is a problem. IIRC, spec says that modular values not fitting in int32_t is not a valid bitstream...

lib/jxl/dec_ans.h
jonsneyers: Yes :)

lib/jxl/modular/encoding/encoding.cc
jonsneyers: Changed it to use pixel_type_w

lib/jxl/modular/encoding/encoding.cc
jonsneyers: yeah it's not very useful, removing...

lib/jxl/modular/encoding/encoding.cc
jonsneyers: Yes, simplified it - for x=0, clampedgradient boils down to just top :)

lib/jxl/modular/encoding/encoding.cc
veluca93: Is it significantly slower? Can we just use make_pixel?

jonsneyers: It's a bit faster, actually. Make_pixel is a bit slower because it does more UnpackSigned in case of runs...

veluca93: Ah but with sv a pixel_type_w this works out without overflows. NVM then, great ;)

lib/jxl/dec_ans.h
```diff
@@ -275,6 +275,31 @@ class ANSSymbolReader {
     return static_cast<uint32_t>(ret);
   }
 
+  // Takes a *clustered* idx. Can only use if RleOnly() is true.
+  void ReadHybridUintClusteredRleOnly(size_t ctx, BitReader* JXL_RESTRICT br,
+                                      uint32_t* value, uint32_t* run) {
+    br->Refill();  // covers ReadSymbolWithoutRefill + PeekBits
```
veluca93, line 281: Can we add a DASSERT?

veluca93 (approved):

lib/jxl/dec_ans.h
jonsneyers: OK like this?

Issue #1041, opened 2021-12-22
Heap-buffer-overflow in StoreRGBA


pzread:
We found a crash with heap-buffer-overflow when fuzzing djxl_fuzzer target. The crash can only be reproduced after 7c5dfcd35fc6a91ae9d776bda7e23ecfaa8bfb38, so I suspect it is related to coalescing mode. The release versions don't seem to be affected.

**To Reproduce**
It is fuzzed with a modified oss-fuzz, therefore it can be reproduced with the oss-fuzz setup.

1. Follow [https://github.com/libjxl/libjxl/blob/main/doc/fuzzing.md](https://github.com/libjxl/libjxl/blob/main/doc/fuzzing.md) to setup the environment.
2. Build binaries with ASAN `BUILD_DIR=build-fuzzasan ./ci.sh ossfuzz_asan`.
3. Run `build-fuzzasan/tools/djxl_fuzzer ./poc1` to reproduce.

`poc1` can be downloaded here: [poc1](https://drive.google.com/file/d/1k_YsaFfTI3P7jrfNnoxJzkshGLsGWZ8P/view?usp=sharing)

**ASAN log**
```
=================================================================
==1239039==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61a000000504 at pc 0x55ff987f4bca bp 0x7fb9149f66d0 sp 0x7fb9149f66c8
WRITE of size 1 at 0x61a000000504 thread T2
    #0 0x55ff987f4bc9 in StoreRGBA<hwy::N_SCALAR::Simd<unsigned char, 1>, hwy::N_SCALAR::Vec1<unsigned char> > /src/libjxl/lib/jxl/dec_reconstruct.cc:199:10
    #1 0x55ff987f4bc9 in jxl::N_SCALAR::FloatToRGBA8(jxl::Image3<float> const&, jxl::Rect const&, bool, jxl::Plane<float> const*, jxl::Rect const&, jxl::Rect const&, unsigned char*, unsigned long) /src/libjxl/lib/jxl/dec_reconstruct.cc:294:9
[...]

PR #1047, opened 2021-12-23
fix non-coalesced frame dimension in case of upsampling

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/1041


veluca93:
Can we add a test?

eustas:
There is `OrientedCroppedFrameTest`. Perhaps we have to extend it with "upsampling" dimension.

jonsneyers:
#1059 also does this, so closing this one

PR #1051, opened 2021-12-27
Use upsampled dimensions in GetCurrentDimensions

eustas:
djxl_fuzzer expects that frame header layer_info contains dimensions
that correspond to what is passed to callback.


eustas:
Not sure, though, it we should use just upsampled, or upsampled padded.

eustas:
Seems to be a dupe of #1047

PR #1052, opened 2021-12-27
Add a "save DC frame" and a "Save as reference" stage

veluca93:


lib/jxl/dec_frame.cc
```diff
@@ -1298,12 +1319,20 @@ Status FrameDecoder::FinalizeFrame() {
...
+      !dec_state_->render_pipeline) {
     dec_state_->shared_storage
         .dc_frames[dec_state_->shared->frame_header.dc_level - 1] =
         std::move(*decoded_->color());
     decoded_->RemoveColor();
   }
+  if (dec_state_->render_pipeline && frame_header_.CanBeReferenced()) {
+    auto& info = dec_state_->shared_storage
+                     .reference_frames[frame_header_.save_as_reference];
+    info.storage = std::move(dec_state_->frame_storage_for_referencing);
```
eustas, line 1332: What will remain in frame_storage_for_referencing? (what will happen if we accidentally use it)

eustas (approved):

lib/jxl/dec_frame.cc
veluca93: IIRC it'll just become an empty image, (0,0) size, so it should be fine.

PR #1054, opened 2021-12-27
Add a patches stage

veluca93:
Depends on #1052


lib/jxl/blending.cc
```diff
@@ -361,30 +361,32 @@ Status PerformBlending(
     } else if (ec_blending[i].mode == PatchBlendMode::kBlendAbove) {
       size_t alpha = ec_blending[i].alpha_channel;
```
eustas, line 362: why  x0 is not  used  in line 359 (tmp.Row(3 + i)[x] = bg[3 + i][x] + fg[3 + i][x];)

veluca93: Because that's a mistake ;)

eustas (approved):

PR #1055, opened 2021-12-28
Deal with division by zero in spline rasterization

eustas:
FP precision loss

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -272,12 +272,16 @@ Vector operator-(const Spline::Point& a, const Spline::Point& b) {
   return {a.x - b.x, a.y - b.y};
 }
 
-std::vector<Spline::Point> DrawCentripetalCatmullRomSpline(
-    std::vector<Spline::Point> points) {
-  if (points.size() <= 1) return points;
+// TODO(eustas): avoid making a copy of "points".
+void DrawCentripetalCatmullRomSpline(std::vector<Spline::Point> points,
+                                     std::vector<Spline::Point>& result) {
+  if (points.empty()) return;
+  if (points.size() == 1) {
+    result.push_back(points[0]);
+    return;
+  }
   // Number of points to compute between each control point.
   static constexpr int kNumPoints = 16;
-  std::vector<Spline::Point> result;
   result.reserve((points.size() - 1) * kNumPoints + 1);
   points.insert(points.begin(), points[0] + (points[0] - points[1]));
   points.push_back(points[points.size() - 1] +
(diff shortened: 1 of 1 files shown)


eustas:
PTAL

veluca93:
I'm not sure why and how these new checks can be triggered without triggering the decode-time checks.

eustas:
Ahem,... there are no new checks in new version. 
Unfortunately, commit message does not transfer to PR message.
tldr: sometimes `a - (a - b) == 0`, even if `b != 0`...

lib/jxl/splines.cc
```diff
@@ -567,13 +571,15 @@ Status Splines::InitializeDrawCache(const size_t image_xsize,
...
                         : kHardPixelLimit;
+  std::vector<Spline::Point> intermediate_points;
   for (size_t i = 0; i < splines_.size(); ++i) {
     JXL_RETURN_IF_ERROR(
         splines_[i].Dequantize(starting_points_[i], quantization_adjustment_,
                                cmap.YtoXRatio(0), cmap.YtoBRatio(0), spline));
     if (std::adjacent_find(spline.control_points.begin(),
                            spline.control_points.end()) !=
         spline.control_points.end()) {
+      // TODO(eustas): spec does not explicitly say this is a problem...
```
veluca93, line 582: It does - it states that whenever a division by 0 would be required, the number you divide by is non0 in a well-formed codestream.

Couldn't we instead check that the difference between consecutive points is "large enough"?
@sboukortt for opinions

eustas: I mean, it says that we should render Cutmull-Rom spline, and gives some example code. However, mathematically, division by zero is not required if we carefully interprete what happens in the situation of repeated control points.
I see no reason to forbid cusps (spinodes) - those could be useful, e.g. for text rendering.

veluca93: IIRC the "example code" is not an example, but it is supposed to be exactly what needs to happen...

jonsneyers: I'm fine with allowing repeated control points as long as nothing changes in the rendering of cases without repeated control points. We can update the spec in that direction as far as I'm concerned. But it's indeed not just "example code", it is the specification of how the decoded image should look like, and any other implementation needs to produce the same pixels.

eustas: On the second thought - we allow control points to be fairly close to each other, so that visual cusps are achievable. So, let's leave spec intact and just deal with FP precision problems.

sboukortt: We use the centripetal variant of Catmull-Rom splines so cusps should not happen: http://www.cemyuksel.com/research/catmullrom_param/ (α=0.5, blue curves)

lib/jxl/splines.cc
```diff
@@ -272,12 +272,13 @@ Vector operator-(const Spline::Point& a, const Spline::Point& b) {
...
 }
 
-std::vector<Spline::Point> DrawCentripetalCatmullRomSpline(
-    std::vector<Spline::Point> points) {
-  if (points.size() <= 1) return points;
+// TODO(eustas): avoid making a copy of "points".
+void DrawCentripetalCatmullRomSpline(std::vector<Spline::Point> points,
+                                     std::vector<Spline::Point>& result) {
+  // TODO(eustas): in spec "point" splines are allowed
+  if (points.size() <= 1) return;
```
veluca93, line 279: Why don't we push_back() points[0] in result?

lib/jxl/splines.cc
veluca93: Ah, I see what's going on and this change is indeed a noop with infinite precision, so that seems fine - except for the case of a single point...

lib/jxl/splines.cc
eustas: Exactly! That is why I have left the comment.

eustas: Fixed. PTAL

veluca93 (approved):
Have you verified that this change doesn't affect rendering in "usual" cases?

eustas:
Nope. Perhaps we should add "conformance" test into the CI,...

PR #1056, opened 2021-12-28
Do not build testing if not going to run tests

eustas:


eustas:
I believe, not building (A|M)SAN tests on push won't allow big failures. On land those will be built / executed; as well on CI:full.

eustas:
MSAN/ASAN build time for checks is ~12/13 minutes now; could be pushed further down, but for me this time is comfortable.
Plus we will face worker starvation mush less frequently...

veluca93 (approved):

PR #1059, opened 2021-12-29
Use upsampled dimensions in GetCurrentDimensions

eustas:
djxl_fuzzer expects that frame header layer_info contains dimensions
that correspond to what is passed to callback.

### DIFF
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -2981,135 +2981,147 @@ TEST(DecodeTest, SkipFrameWithAlphaBlendingTest) {
 }
 
 TEST(DecodeTest, OrientedCroppedFrameTest) {
-  size_t xsize = 90, ysize = 120;
-  JxlPixelFormat format = {4, JXL_TYPE_UINT16, JXL_BIG_ENDIAN, 0};
+  const auto test = [](bool keep_orientation, uint32_t orientation,
+                       uint32_t resampling) {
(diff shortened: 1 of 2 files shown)


lib/jxl/decode_test.cc
```diff
@@ -2986,135 +2986,144 @@ TEST(DecodeTest, SkipFrameWithAlphaBlendingTest) {
...
+      EXPECT_EQ(JXL_DEC_SUCCESS, JxlDecoderProcessInput(dec));
+      JxlThreadParallelRunnerDestroy(runner);
+      JxlDecoderDestroy(dec);
+    }
+
+    EXPECT_EQ(0u, jxl::test::ComparePixels(frames[0].data(), frames[4].data(),
+                                           oxsize, oysize, format, format));
+  };
+
+  for (bool keep_orientation : {true, false}) {
```
veluca93, line 3115: Not necessary for now, but could we start converting some of those nested for loops into parameterized tests?

veluca93 (approved):

jonsneyers:
This is the same fix as https://github.com/libjxl/libjxl/pull/1047, right?

lib/jxl/decode_test.cc
eustas: Ack. Sure, make sense.

eustas:
Yup, the same, plus some testing. At least it fails without the fix.

veluca93 (approved):

Issue #1057, opened 2021-12-28
Delta palette for bit_depth=32


eustas:
Some `kDeltaPalette` entries are UB to use when `bit_depth==32`, namely:
```{128, 128, 128}, {144, 144, 144}, {0, 128, 128}, {-128, 0, 0}, {128, 128, 0},   {0, 0, -128}```
What decoder should do, when it is asked about such entries?

jonsneyers:
Delta palette was not designed with non-integer samples in mind, the definition does not make sense for the float case (and bitdepth 32 implies float).

I suggest we add in the second edition of the spec that implicit delta palette is not used in the float case. I suggest our decoder just fails if it happens that negative values occur in the fp case. Alternatively it can just do something arbitrary.

veluca93:
How about disallowing Delta palette entirely for float / > 24bpp samples?

jonsneyers:
The thing is, there is no real distinction between normal palette and delta palette. Normal palette (and even delta palette with explicit deltas) _could_ be useful for float — in particular, channel palette gives a way to efficiently encode float images that originate from int images (or that otherwise have only relatively few values).

I don't mind just disallowing Palette completely in case of bitdepth > 24, but it does come at some cost in expressivity of the bitstream...

veluca93:
There *is* distinction though - normal palette has a predictor of Zero,
Delta palette doesn't.
It's true that you can do non-delta palette with a nonzero predictor, but
there's no reason to that I can think of.

jonsneyers:
Ah right. Well you can also do delta palette without ever using the implicit deltas, which could be useful in principle. But I think it's better if we can have something that can fail early rather than something that depends on the actual data (which implies that avoiding UB likely comes at a speed cost).

So prohibiting Palette with nonzero predictor when bits_per_sample > 24 sounds like a good idea to me — shouldn't be a significant expressivity loss, while it makes it easy to avoid UB trouble.

jonsneyers:
This is fixed on the libjxl end. Spec update still needs to be done.

PR #1060, opened 2021-12-29
Prohibit Palette with nonzero predictor when bits_per_sample > 24

eustas:
See https://github.com/libjxl/libjxl/issues/1057


jonsneyers (approved):

jonsneyers:
We should remember to add this condition to the spec. Adding a label for changes that have (potential) spec impact.

jonsneyers:
Actually this is not enough to guarantee that the UB does not happen: you can also have the zero predictor and still use kDeltaPalette entries (it would be pointless, but the bitstream doesn't disallow it).

I also wonder if the default colors (the color cubes when `index >= nb_colours`) don't also cause trouble when bits_per_sample is too high.

I think it might be best not to allow Palette at all when bits_per_sample > 24.

eustas:
Indeed. Original finding (in libvips) still repro.
Will prepare PR disallowing palette for `bits_per_sample > 31`.

PR #1067, opened 2021-12-31
Fix APNG decoding.

### LINKED ISSUE #1065: Some animated pngs with non-full frames with transparency visually differ from source after transcoding with cjxl
**Describe the bug**
I tested the addition in pull request https://github.com/libjxl/libjxl/pull/1058.
There seems to be some trouble with blending some(?) non-full frames with transparency. Instead of showing the previous background it shows a fully transparent background for some "subframes" in some images, creating transparent rectangles.
I created an animation with just three frames to show this effect and attached it to this issue.

**To Reproduce**
- Encode the input.png with cjxl : cjxl -d 0.0 -e 7 in.png out.jxl
- decode out.jxl back to animated png: djxl out.jxl out.png
- compare the animations in an imageviewer or browser (I compared with firefox) and observe the difference.
  (note: I believe the wrong transparency happens during encoding, so the out.jxl file is already "broken". Not 100% sure though...)

**Expected behavior**
Animation before and after conversion should look the same, or cjxl should throw an error if this is not possible.

**Environment**
- OS: Ubuntu 21.10
- Compiler version: gcc version 11.2.0
- CPU type: x86_64
- cjxl/djxl version string: v0.7.0 844b1ec [AVX2,SSE4,SSSE3,Scalar]

Was also able to reproduce on Windows 10 with a build of cjxl from https://artifacts.lucaversari.it/libjxl/libjxl/latest/

**(Correct) Input png**
animated

**(Broken) Animated jxl decoded to apng**
djxl

veluca93:
Fixes #1065.

### DIFF
--- a/lib/extras/dec/apng.cc
+++ b/lib/extras/dec/apng.cc
@@ -388,7 +388,6 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
   PaddedBytes chunkIHDR;
   std::vector<PaddedBytes> chunksInfo;
   bool isAnimated = false;
-  bool skipFirst = false;
   bool hasInfo = false;
   APNGFrame frameRaw = {};
   uint32_t num_channels;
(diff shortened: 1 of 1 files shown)


jonsneyers (approved):

jonsneyers:
Always nice when a bug can be fixed by just removing old lines of code :)

PR #1071, opened 2022-01-03
Fix integer overflow in quant_adjustment calculation

eustas:
`-(min_int)` is not allowed.

Drive-by: fix decorellation - zeores were used instead of decoded values
Drive-by: make calculation a little bit more precise.
Drive-by: get rid of overwordly `square_root`.


mo271 (approved):

eustas:
Test is failing. Going to check if it is because of "better precision" or not...

mo271 (approved):

PR #1073, opened 2022-01-04
Encoder using using C api

mo271:
This is the start for an encoder that uses the C api and eventually
should replace cxjl.


tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+DEFINE_int32(photon_noise, 0,
+             // TODO(tfish): Discuss docstring change with team.
+             // Also: This now is an int, no longer a float.
+             "Adds noise to the image emulating photographic film noise. "
+             "The higher the given number, the grainier the image will be. "
+             "As an example, a value of 100 gives low noise whereas a value "
+             "of 3200 gives a lot of noise. The default value is 0.");
+
+DEFINE_string(
+    distance, "1.0",  // TODO(tfish): wire this up.
```
veluca93, line 191: Why is this DEFINE_string and not DEFINE_double?

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+    // TODO(tfish): Clarify discrepancy with team:
+    // Documentation says default==squirrel(7) here:
+    // https://libjxl.readthedocs.io/en/latest/api_encoder.html#_CPPv424JxlEncoderFrameSettingId
+    // but enc_params.h has kFalcon=7.
+    "Encoder effort setting. Range: 1 .. 9.\n"
+    "    Default: 7. Higher number is more effort (slower).");
+
+namespace {
+
+// RAII-wraps the C-API encoder.
```
veluca93, line 225: This should already exist in encode_cxx.h.

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+                         JxlEncoderFrameSettingId encoder_option) {
+  gflags::CommandLineFlagInfo flag_info =
+      gflags::GetCommandLineFlagInfoOrDie(flag_name);
+  if (!flag_info.is_default) {
+    JxlEncoderFrameSettingsSetOption(frame_settings, encoder_option,
+                                     static_cast<int32_t>(flag_value));
+  }
+  return true;
+}
+// TODO(tfish): Clean this up once we can properly break the codec_in_out.h
```
veluca93, line 265: I think this TODO is not relevant now, right?

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+    return status;
+  }
+  if (image_data.size() < kMinBytes) return JXL_FAILURE("Input too small.");
+  jxl::Span<const uint8_t> encoded(image_data);
+
+  // Default values when not set by decoders.
+  ppf.info.uses_original_profile = true;
+  ppf.info.orientation = JXL_ORIENT_IDENTITY;
+  jxl::extras::ColorHints color_hints;
+  // TODO(tfish): Fix size_constraints1 hack once SizeConstraints moved out of
```
veluca93, line 287: Also this

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+}
+
+}  // namespace
+// tristate flag not necessary, because we can use
+// gflags::GetCommandLineFlagInfoOrDie(const char* name).is_default
+int main(int argc, char** argv) {
+  std::cerr << "Warning: This is work in progress, consider using cjxl "
+               "instead!\n";
+  gflags::SetUsageMessage(
+      std::string("JPEG XL-encodes an image.\n") +
```
veluca93, line 335: Why so many std::string?

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+#endif
+#if JPEGXL_ENABLE_JPEG
+      std::string("JPEG, ") +
+#endif
+      std::string("PPM, PFM, PGX.\n") + std::string("  Sample usage:\n") +
+      std::string(argv[0]) +
+      std::string(" <source_image_filename> <target_image_filename>"));
+  uint32_t version = JxlEncoderVersion();
+  std::stringstream version_string;
+  version_string << version / 1000000 << "." << (version / 1000) % 1000 << "."
```
veluca93, line 351: std::to_string() instead of stringstream?

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+                        JXL_ENC_FRAME_SETTING_PATCHES);
+    ProcessTristateFlag("gaborish", FLAGS_gaborish, jxl_encoder_frame_settings,
+                        JXL_ENC_FRAME_SETTING_GABORISH);
+    ProcessTristateFlag("group_order", FLAGS_group_order,
+                        jxl_encoder_frame_settings,
+                        JXL_ENC_FRAME_SETTING_GROUP_ORDER);
+
+    const int32_t flag_effort = FLAGS_effort;
+    if (!(1 <= flag_effort && flag_effort <= 9)) {
+      // Strictly speaking, custom absl flags-parsing would integrate
```
veluca93, line 437: out of date

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,604 @@
...
+      return EXIT_FAILURE;
+    }
+    if (ppf.frames.size() < 1) {
+      std::cerr << "No frames on input file.\n";
+      return EXIT_FAILURE;
+    }
+    const jxl::extras::PackedFrame& pframe = ppf.frames[0];
+    const jxl::extras::PackedImage& pimage = pframe.color;
+    JxlPixelFormat ppixelformat = pimage.format;
+    std::cerr << "DDD JxlEncoderAddImageFrame() - input frames="
```
veluca93, line 547: debugging leftover?

tools/fetch_encoded.c
```diff
@@ -0,0 +1,106 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "jxl/encode.h"
+
+#define DEBUG_BUFFER_REALLOCATION 0
+
+JXL_BOOL fetch_jxl_encoded_image(JxlEncoder *jxl_encoder,
```
veluca93, line 22: The naming convention is different from the rest of the codebase.

Why can't we use C++ here?

If we do use C, why not `realloc`?

tools/fetch_encoded.c
```diff
@@ -0,0 +1,106 @@
...
+  if (!file) {
+    fprintf(stderr, "Could not open file for writing: %s\n", filename);
+    return JXL_FALSE;
+  }
+  size_t bytes_written = fwrite(bytes, sizeof(uint8_t), size, file);
+  if (bytes_written != size) {
+    fprintf(stderr,
+            "Failure: Wanted to write %d bytes, "
+            "but did write %d bytes to file: %s\n",
+            (int)size, (int)bytes_written, filename);
```
veluca93, line 98: We probably should also print strerror(errno).

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,593 @@
...
+    "(JPEG/GIF).\n"
+    "    1.0 = visually lossless. Default for other input.\n"
+    "    Recommended range: 0.5 .. 3.0.");
+
+DEFINE_int64(target_size, 0,  // TODO(tfish): wire this up.
+             "Aim at file size of N bytes.\n"
+             "    Compresses to 1 % of the target size in ideal conditions.\n"
+             "    Runs the same algorithm as --target_bpp");
+
+DEFINE_string(target_bpp, "0",  // TODO(tfish): wire this up.
```
veluca93, line 203: other non-double here and below

tools/fetch_encoded.c
fischbacher: 1. Why not use C++: this is generic stuff that one has to also do when e.g. writing a FFI for JXL (say, for Python, Perl, Racket, nodejs, etc.). It's more useful to have this in C and allow carrying forward this code rather than forcing whoever wanted to use this for a FFI to also introduce a dependency on a C++ compiler.

2. realloc() is definitely more appropriate here - also since malloc() may have done pre-allocation. I think I just overlooked that because I was too occupied with keeping all this "how to wire up things" state in my head when this was all new to me.

tools/fetch_encoded.c
veluca93: Seems about right.

tools/fetch_encoded.c
veluca93: I don't think tools/ is the correct place for basically a wrapper over the API that makes it so that output is no longer streaming (if I understand correctly). I'd argue that someone writing a FFI layer should still expose the streaming output mode of the C API, so I'm not sure there's a point to make this not c/djxl-private.

tools/cjxl_ng_main.cc
mo271: Done. When wiring this up, need to make sure that `-inf` is correctly parsed for `-quality`.

tools/fetch_encoded.c
mo271: Just as a proposal, I just submitted a version that does not use `fetch_jxl_encoded_image`, `write_jxl_file` nor the `ManagedJxlEncoder` (also to address the first unresolved comment). This now basically follows how it is done in `encode_oneshot.cc`. 
Not sure how this plays with a potential FFI, but perhaps this can be done at another place.

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,597 @@
...
+    "    Default: 7. Higher number is more effort (slower).");
+
+namespace {
+/**
+ * Writes bytes to file.
+ */
+bool WriteFile(const std::vector<uint8_t>& bytes, const char* filename) {
+  FILE* file = fopen(filename, "wb");
+  if (!file) {
+    fprintf(stderr, "Could not open %s for writing\n", filename);
```
veluca93, line 233: This function also could use some `strerror`s

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,597 @@
...
+  // TODO(tfish): Migrate this:
+  // if (!skip_ppf_conversion) {
+  //   JXL_RETURN_IF_ERROR(ConvertPackedPixelFileToCodecInOut(ppf, pool, io));
+  // }
+  return true;
+}
+
+}  // namespace
+// tristate flag not necessary, because we can use
+// gflags::GetCommandLineFlagInfoOrDie(const char* name).is_default
```
veluca93, line 318: Not sure what this comment refers to

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,597 @@
...
+                        JXL_ENC_FRAME_SETTING_DOTS);
+    ProcessTristateFlag("patches", FLAGS_patches, jxl_encoder_frame_settings,
+                        JXL_ENC_FRAME_SETTING_PATCHES);
+    ProcessTristateFlag("gaborish", FLAGS_gaborish, jxl_encoder_frame_settings,
+                        JXL_ENC_FRAME_SETTING_GABORISH);
+    ProcessTristateFlag("group_order", FLAGS_group_order,
+                        jxl_encoder_frame_settings,
+                        JXL_ENC_FRAME_SETTING_GROUP_ORDER);
+
+    const int32_t flag_effort = FLAGS_effort;
```
veluca93, line 421: Maybe a validator with a (template?) parameter for the list of valid values would overall reduce boilerplate, given the number of such checks we have.

tools/cjxl_ng_main.cc
mo271: added 3.

tools/cjxl_ng_main.cc
mo271: Was an outdated comment, removed...

tools/cjxl_ng_main.cc
mo271: I think it would be a good idea: added TODO for that!

tools/cjxl_ng_main.cc
mo271: decided to use `encode_cxx.h` and `thread_parallel_runner_cxx.h`

tools/fetch_encoded.c
mo271: decided to remove `fetch_encoded.*` and instead go with a custom `WriteFile`

tools/cjxl_ng_main.cc
```diff
@@ -0,0 +1,606 @@
...
+  const char* filename_out = argv[2];
+
+  size_t num_worker_threads = JxlThreadParallelRunnerDefaultNumWorkerThreads();
+  {
+    int64_t flag_num_worker_threads = FLAGS_num_threads;
+    if (flag_num_worker_threads != 0) {
+      num_worker_threads = flag_num_worker_threads;
+    }
+  }
+  // ManagedJxlEncoder managed_jxl_encoder =
```
veluca93, line 373: Stale comment

mo271: removed

veluca93 (approved):
Seems at least a reasonable starting point.

PR #1074, opened 2022-01-05
New options to enable plugins individually

novomesk:
Package maintainers (at least from Gentoo project) need that specific features need to be allowed one-by-one, so every feature should be behind a switch.

Before this PR, the 3 plugins were enabled/disabled together.

After this PR, maintainers will be able to control which plug-in to enable or to disable.


mo271 (approved):
Thanks for your contribution!

PR #1078, opened 2022-01-07
ignore tools/conformance/__pycache__

mo271:


.gitignore
```diff
@@ -14,3 +14,4 @@ third_party/hdrvdp-2.2.2.zip.tmp
 # Output plots
 tools/benchmark/metrics/plots
 tools/benchmark/metrics/results.csv
+tools/conformance/__pycache__
```
sboukortt, line 17: This is relatively minor but would it be possible to maintain the final `\n`? It makes `cat` better-behaved, for example.

mo271: sure, makes sense, I didn't intend to no to have the newline..

sboukortt (approved):

PR #1082, opened 2022-01-10
Prevent heap buffer overflow

eustas:


lib/jxl/dec_group.cc
```diff
@@ -259,6 +259,8 @@ Status DecodeGroupImpl(GetBlock* JXL_RESTRICT get_block,
     r[i] =
         Rect(block_rect.x0() >> hshift[i], block_rect.y0() >> vshift[i],
              block_rect.xsize() >> hshift[i], block_rect.ysize() >> vshift[i]);
+    const auto& dc = dec_state->shared->dc->Plane(i);
+    JXL_CHECK(r[i].IsInside({0, 0, dc.xsize(), dc.ysize()}));
```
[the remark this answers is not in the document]
veluca93: (also the variable should probably be behind a `#if JXL_ENABLE_ASSERT`)

lib/jxl/passes_state.cc
```diff
@@ -42,10 +42,12 @@ Status InitializePassesSharedState(const FrameHeader& frame_header,
...
   shared->quant_dc = ImageB(frame_dim.xsize_blocks, frame_dim.ysize_blocks);
-  if (!(frame_header.flags & FrameHeader::kUseDcFrame) || encoder) {
-    shared->dc_storage =
-        Image3F(frame_dim.xsize_blocks, frame_dim.ysize_blocks);
-  } else {
+
+  shared->dc_storage = Image3F(frame_dim.xsize_blocks, frame_dim.ysize_blocks);
+  shared->dc = &shared->dc_storage;
+
+  bool use_dc = !!(frame_header.flags & FrameHeader::kUseDcFrame);
```
veluca93, line 49: maybe `use_dc_frame`?

veluca93: (which reminds me: shouldn't we allocate dc_storage only if not using a DC frame?)

lib/jxl/dec_group.cc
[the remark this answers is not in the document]
eustas: Done

lib/jxl/passes_state.cc
eustas: Done.

(Not unlikely)

veluca93: Can we restore that logic and see if the buffer overflow is still resolved?

eustas: Seems to work.
Also reallocated `dc_storage` to 0x0 image, so that problem would manifest itself earlier next time.
PTAL

veluca93 (approved):

PR #1083, opened 2022-01-10
Queue PR workflows...

eustas:
And (hopefully) cancel the older ones.

This should help when PR gets updated while previous runs are not finished.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -14,6 +14,10 @@ on:
   pull_request:
     types: [opened, reopened, labeled, synchronize]
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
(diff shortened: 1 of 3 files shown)


veluca93:
Should we also add `github.repository_owner` to the key? I'm thinking of two people opening PRs from branches with the same name simultaneously...

Also, I think cancel_pending should only happen if !github.ref_protected (i.e. not main)

eustas:
Going to dump ref and see what is looks like. I believe for PR it will reference to PR number.
Protection is a different dimension, lets filter by event type - e.g. push to PR.

veluca93:
> Going to dump ref and see what is looks like. I believe for PR it will reference to PR number. Protection is a different dimension, lets filter by event type - e.g. push to PR.

Sounds good - the doc made me believe it's the branch/tag name but it may be different on PRs.
Event type is better, yes ;)

eustas:
PTAL

veluca93 (approved):

EwoutH:
Haven't seen this used in this way before, very neat!

PR #1084, opened 2022-01-11
API proposal for frame index box

jyrkialakuijala:


lib/include/jxl/encode.h
```diff
@@ -272,6 +272,15 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_JPEG_RECON_CFL = 30,
 
+  /** Prepare the frame for indexing in the frame index box.
+   * 0 = ignore this frame (same as not setting a value),
+   * 1 = index this frame within the Frame Index Box.
+   * If any frames are indexed, the first frame needs to
+   * be indexed, too. If the first frame is not indexed, and
+   * a later frame is attempted to be indexed, JXL_ENC_ERROR will occur.
```
lvandeve, line 280: It may also make sense to cause an error (or alternatively ignore the setting) if the frame is not a keyframe (since it's required for the indexed frames to be key frames)

jyrkialakuijala: Good catch. I added that to the comment.

lvandeve (approved):

PR #1087, opened 2022-01-12
upscaling demo

mo271:
This adds a python script to demo the upscaling with a factor of 2, 4
and 8.


tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+    -0.025194061504750526]], [[-0.0004507227386051219, -0.005422291349995998,
+    -0.01155272937910007, -0.01081748428801323, -0.00447631814814665],
+    [-0.005422291349995998, -0.03680917952171095, -0.022489385288542607,
+    -0.024475522375569655, -0.03324558280295302], [-0.011552729379100074,
+    -0.02248938528854261, 0.4272005024152728, 0.3575770812164143,
+    -0.0378381168526885], [-0.010817484288013228, -0.02447552237556967,
+    0.3575770812164143, 0.29895328454745285, -0.03706352644207268],
+    [-0.00447631814814665, -0.03324558280295302, -0.0378381168526885,
+    -0.03706352644207269, -0.029286133281073243]]]]
+}
```
fischbacher, line 733: OK for a demo, but overall, there might be better alternatives to cluttering the source file with a 700-lines constant. Would it work (for example) to use a .npz instead?

>>> numpy.savez_compressed('/tmp/data.npz', k2=numpy.zeros([4, 4]), k4=numpy.zeros([4, 4]), k8=numpy.zeros([4, 4]))
>>> d = numpy.load('/tmp/data.npz')
>>> type(d)
<class 'numpy.lib.npyio.NpzFile'>
>>> d
<numpy.lib.npyio.NpzFile object at 0x7f0328f523a0>
>>> list(d)
['k2', 'k4', 'k8']
>>> d['k2'].shape
(4, 4)
>>> d['k2'][:2, :2]
array([[0., 0.],
       [0., 0.]])

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+
+
+def convolution(pixels, coeffs, kernel_size):
+  pad_width = kernel_size//2
+  is_multi_channel = len(pixels.shape) > 2
+  padded_pixels = np.pad(
+      pixels, 2*[2*[pad_width]] + int(is_multi_channel)*[[0, 0]], mode='edge')
+  x, y,  = pixels.shape[:2]
```
fischbacher, line 741: x, y, *_ = pixels.shape

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+  convoluted = np.block([[np.einsum('rc...,RCrc->...RC',
+                                    padded_pixels[i-pad_width: i + pad_width + 1,
+                                                  j - pad_width: j + pad_width + 1],
+                                    coeffs)
+                          for j in range(pad_width, pad_width + y)]
+                         for i in range(pad_width, pad_width + x)])
+  if is_multi_channel:
+    return np.moveaxis(convoluted, 0, -1)
+  else:
+    return convoluted
```
fischbacher, line 751: More concise:

return np.moveaxis(convoluted, 0, -1) if is_multi_channel else convoluted

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+    -0.024475522375569655, -0.03324558280295302], [-0.011552729379100074,
+    -0.02248938528854261, 0.4272005024152728, 0.3575770812164143,
+    -0.0378381168526885], [-0.010817484288013228, -0.02447552237556967,
+    0.3575770812164143, 0.29895328454745285, -0.03706352644207268],
+    [-0.00447631814814665, -0.03324558280295302, -0.0378381168526885,
+    -0.03706352644207269, -0.029286133281073243]]]]
+}
+
+
+def convolution(pixels, coeffs, kernel_size):
```
fischbacher, line 736: It would be good to have a docstring for this.

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+    parser.add_argument(
+        "output_filename",
+        type=str,
+        help="where the upscaled image is written as PNG."
+    )
+
+    args = parser.parse_args()
+    upscaling_factor = args.upscaling_factor[0]
+    kernel_size = 5
+    assert upscaling_factor in [2, 4, 8]
```
fischbacher, line 779: This will raise an AssertionError on user-provided-wrong-value, which looks weird to the user. Better:

if upscaling_factor not in (2, 4, 8):
  raise ValueError(...)

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,793 @@
...
+    kernel = np.array(scaling_kernels[upscaling_factor])
+    assert kernel.shape == (
+        upscaling_factor, upscaling_factor, kernel_size, kernel_size)
+    orig = Image.open(args.input_filename)
```
fischbacher, line 783: If we call this orig_raw, and do

orig = orig_raw.convert(...) if orig.mode == 'P' else orig_raw

we can eliminate the reassignment.

tools/upscaling_coefficients/upscaler_demo.py
mo271: Done, also removed the unnecessary "kernel_size" argument (since we can get it from coeffs). 
Also: renamed `coeffs` to `kernel`

tools/upscaling_coefficients/upscaler_demo.py
mo271: Done

tools/upscaling_coefficients/upscaler_demo.py
mo271: Done

tools/upscaling_coefficients/upscaler_demo.py
mo271: Done

tools/upscaling_coefficients/upscaler_demo.py
mo271: reassignment eliminated

mo271:
Other than the changes suggested by the review comment, I also changed the rounding to nearest int, instead of truncating.

tools/upscaling_coefficients/upscaler_demo.py
fischbacher: Another alternative would be to move these to a separate upscaler_demo_constants.py.

A third alternative - and this is actually sort-of violating the letter of some readability guidelines here, but in this case helps quite a bit with readability, is to use this technique:

===
#!/usr/bin/env python3

FOO_CONSTANTS = _get_constants()

... code ...

def main():
  ...

if __name__ == '__main__':
  main()

def _get_constants():
  # all the distracting stuff goes here, at the very end
===

This sort-of emulates what Perl's __DATA__ achieves, see e.g.: https://perldoc.perl.org/SelfLoader#The-__DATA__-token

I would say there is some value in not having this big wall of numbers at the start of the file.

tools/upscaling_coefficients/upscaler_demo.py
```diff
@@ -0,0 +1,809 @@
...
+    kernel: A [upscaling_factor, upscaling_factor, kernel_size,
+     kernel_size]-array used for the convolution.
+
+  Returns:
+    A [heigth, width]- or [height, width, num_channels]-array representing the
+    convoluted image.
+  """
+  kernel_size = kernel.shape[2]
+  pad_width = kernel_size//2
+  is_multi_channel = len(pixels.shape) > 2
```
fischbacher, line 756: Would it perhaps make sense to simply normalize the input as follows (such tweaks are OKish-to-do right as the first thing that happens to an input arg in the body)...:

pixels = pixels.reshape(pixels.shape[:2] + (1,))

This eliminates special-casing. From there on, we are guaranteed that this has three indices, and can use more uniform logic for processing.

Note:

>>> z = numpy.zeros([40, 20])
>>> z1 = numpy.zeros([40, 20])
>>> z2 = numpy.zeros([40, 20, 3])
>>> z1r = z1.reshape(z1.shape[:2] + (-1,))
>>> z2r = z2.reshape(z2.shape[:2] + (-1,))
>>> z1r.shape
(40, 20, 1)
>>> z2r.shape
(40, 20, 3)

mo271: This could be done, but then we would need to do another reshape in the end, because the intention is that the output has the same shape as the input.

Do you still think that this would be a good idea considering this?

fischbacher: I would say that a final

return output.reshape(input.shape)

is also rather uniform and nicely conveys the idea that "the output should be the same shape as the input". In that case, the structure would roughly be...

def f(input, ...):
  shaped_input = input.reshape(...)
  ...
  return output.reshape(input)

mo271: modified the code accordingly; also clarifiedthat the output shape is actually the upscaled shape in the first two dimensions.

tools/upscaling_coefficients/upscaler_demo.py
mo271: Makes sense, moved the constants to the bottom, in a way similar to what you suggested.

fischbacher (approved):
This looks good now.

PR #1094, opened 2022-01-13
wip: Decoder using C api

mo271:
This is the start of a decoder that uses the C api and eventually
should replace djxl.


tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+#include "jxl/thread_parallel_runner.h"
+#include "jxl/thread_parallel_runner_cxx.h"
+#include "lib/extras/dec/decode.h"
+#include "lib/jxl/base/printf_macros.h"
+
+DECLARE_bool(help);
+DECLARE_bool(helpshort);
+
+DEFINE_int64(num_reps, 1,  // TODO(firsching): Clarify meaning of this
+                           // docstring. Is this simply for benchmarking?
```
veluca93, line 27: yes

tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+             "How many times to decompress.");
+
+DEFINE_int64(num_threads, 0,
+             // TODO(firsching): Sync with team about changed meaning of 0 -
+             // was: No multithreaded workers. Is: use default number.
+             "Number of worker threads (0 == use machine default).");
+
+// TODO(firsching): wire this up.
+DEFINE_bool(print_profile, false, "Print timing information before exiting.");
```
veluca93, line 36: I wouldn't complain if we removed it

tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+DEFINE_string(display_nits, "0.f-255.",
+              "luminance range of the display to which to "
+              "tone-map; the lower bound can be omitted");
+
+// TODO(firsching): wire this up.
+DEFINE_double(preserve_saturation, 0.1,
+              "with --tone_map, how much to favor saturation over luminance");
+
+// TODO(firsching): wire this up.
+DEFINE_string(color_space, "RGB_D65_SRG_Rel_Lin",
```
veluca93, line 57: I think the default ought to be the empty string -- and that the docstring is slightly unclear

tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+DEFINE_bool(allow_more_progressive_steps, false,
+            "allow decoding more progressive steps in truncated "
+            "files. No effect without --allow_partial_files");
+
+#if JPEGXL_ENABLE_JPEG
+// TODO(firsching): wire this up.
+DEFINE_bool(
+    pixels_to_jpeg, false,
+    "By default, if the input JPEG XL contains a recompressed JPEG file, "
+    "djxl "
```
veluca93, line 78: extra newline?

tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+DEFINE_bool(use_sjpeg, false, "use sjpeg instead of libjpeg for JPEG output");
+#endif
+
+// TODO(firsching): wire this up.
+DEFINE_bool(print_read_bytes, false, "print total number of decoded bytes");
+
+// TODO(firsching): wire this up.
+DEFINE_bool(quiet, false, "silence output (except for errors)");
+
+bool WriteFile(const char* filename, const std::vector<uint8_t> bytes) {
```
veluca93, line 100: I think the vector wants to be a const reference

tools/djxl_ng_main.cc
```diff
@@ -0,0 +1,299 @@
...
+  bool can_reconstruct_jpeg = false;
+  std::vector<uint8_t> jpeg_data_chunk(kOutputChunkSize);
+  jpeg_bytes.resize(0);
+  for (;;) {
+    JxlDecoderStatus status = JxlDecoderProcessInput(dec.get());
+    if (status == JXL_DEC_ERROR) {
+      fprintf(stderr, "Failed to decode image\n");
+      break;
+    } else if (status == JXL_DEC_NEED_MORE_INPUT) {
+      size_t remaining = JxlDecoderReleaseInput(dec.get());
```
veluca93, line 169: All of the logic here seems rather unnecessarily complicated, why don't we just read the whole file into memory?

tools/djxl_ng_main.cc
mo271: removed comment

tools/djxl_ng_main.cc
mo271: is removed

tools/djxl_ng_main.cc
mo271: Added remark to consider it, when wiring up the flag.

tools/djxl_ng_main.cc
mo271: made better line breaks

tools/djxl_ng_main.cc
mo271: yes, of course

veluca93 (approved):

tools/djxl_ng_main.cc
mo271: restructured to read the whole file into memory
It does make the logic easier.
I used the read file function from the `decode_oneshot.cc`. 
Right now djxl_ng_main.cc has some overlap of reading and writing files with cjxl_ng_main.cc. Not sure if it would make sense to have file_io in some central place for better reuse-ability.

veluca93: Makes sense, but we can do it in a followup

fischbacher (approved):
Nice work!

Issue #1038, opened 2021-12-22
cjxl, djxl: Add option to preserve file modification timestamp


jendalinda:
When transcoding old image files for archiving, it would be nice to have an option to let the transcoded files keep the modification date and time of the original files.

jonsneyers:
I think this might be tricky to do in a portable way...

jendalinda:
As a workaround, this cmd script does the trick on Windows:
```
@echo off

set jpg_file=my_image.jpg
set jxl_file=my_image.jxl

cjxl.exe %jpg_file% %jxl_file% -e 9

powershell -command "(Get-ChildItem '%jxl_file%').LastWriteTime = (Get-ChildItem '%jpg_file%').LastWriteTime"
```

cagelight:
> I think this might be tricky to do in a portable way...

As of C++17 it's actually pretty easy, you can use `std::filesystem::last_write_time` to fetch and set the time in a cross platform way.

avidseeker:
Adding my vote to this. A part of "lossless" conversion is also keeping the metadata of the older file.

As a workaround, I'm using a wrapper script for cjxl with
```
cjxl "$old_img" "$img".jxl
touch -r "$old_img" "$img".jxl
```

sergeevabc:
Windows 7 x64 user here, with NTFS.

I'd like to convert thousands of images to JXL. It is crucial to transfer their timestamps. Unfortunately, aforementioned workarounds are insufficient: @jendalinda's one and @avidseeker's one do not transfer the creation time. Often that time differs from the modification time, which leads to loss of information about the life path of an image.

``` console
$ exiftool.exe original.jpg | rg.exe time
File Modification Date/Time     : 2023:09:26 13:17:12+03:00
File Access Date/Time           : 2023:09:26 00:00:00+03:00
File Creation Date/Time         : 2023:09:26 12:06:20+03:00   !!!

$ exiftool.exe converted.jxl | rg.exe time
File Modification Date/Time     : 2023:09:26 13:17:12+03:00
File Access Date/Time           : 2023:09:26 00:00:00+03:00
File Creation Date/Time         : 2024:01:04 22:04:42+03:00   !!!
```

Now, what helped me is [NirCmd](https://www.nirsoft.net/utils/nircmd.html) (binaries are available in the footer) as follows
``` console
$ nircmdc clonefiletime original.jpg converted.jxl

$ exiftool original.jpg | rg creation
[...]

PR #1100, opened 2022-01-18
Elevate C++ standard to 17

cagelight:
I would like to propose this pull request to allow more modern C++ to be used in the code base. I feel it has been long enough since 2017 that this should not pose any issues, and it would allow issues like #1038 to be easily fixed.


malaterre:
In case this issue is ever considered. I'd like to also suggest the following approach as was done in related highway project:

* https://github.com/google/highway/issues/486

So the patch would be more like this:

* https://github.com/google/highway/commit/3ddf82d0b25b045ec5790ac9447c67dd93b98ab9

Thanks for consideration.

malaterre:
One last comment, the following would also need to be addressed for consistency. Something like:

```
Index: libjxl/lib/CMakeLists.txt
===================================================================
--- libjxl.orig/lib/CMakeLists.txt
+++ libjxl/lib/CMakeLists.txt
@@ -39,7 +39,7 @@ set(JPEGXL_INTERNAL_FLAGS
   # WARN_FLAGS
   -Wall
   -Wextra
-  -Wc++11-compat
+  -Wc++17-compat
   -Warray-bounds
   -Wformat-security
   -Wimplicit-fallthrough
```

hohMiyazawa:
The default ``gcc`` version in Ubuntu 18.04 long term support is ``v7.4.0``. Full support for c++17 is from ``gcc 8`` and on.

EwoutH:
I support updating to C++17.

> The default `gcc` version in Ubuntu 18.04 long term support is `v7.4.0`. Full support for c++17 is from `gcc 8` and on.

There will always be cases that aren't covered when updating a requirement. Because Ubuntu 20.04 LTS is available and GCC 8 to 11 can be installed easily on Ubuntu 18.04, I don't think this is a blocking isssue.
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt install gcc-11 g++-11
```

eustas:
Shadowed by #3590

PR #1101, opened 2022-01-18
fixing modular animated upsampling

mo271:


lib/jxl/dec_reconstruct.cc
```diff
@@ -1297,7 +1297,8 @@ Status FinalizeFrameDecoding(ImageBundle* decoded,
     std::vector<Rect> rects_to_process;
     for (size_t y = 0; y < frame_dim.ysize; y += kGroupDim) {
```
veluca93, line 1298: also here

mo271: done

veluca93 (approved):

PR #1103, opened 2022-01-18
GFLAGS is supposed to be used by tools only

Vargol:
Current CMakeFiles.txt looks for gflags unless it is bundled, as gflags is only required by the tools this breaks builds where the tools are disabled.

This change only does a search for gflags if tools is set to enabled

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,6 +20,7 @@ Alex Xu (Hello71) <alex_y_xu@yahoo.ca>
 Alexander Sago <cagelight@gmail.com>
 Andrius Lukas Narbutas <andrius4669@gmail.com>
 Artem Selishchev
+David Burnett <vargolsoft@gmail.com>
 Dirk Lemstra <dirk@lemstra.org>
 Don Olmstead <don.j.olmstead@gmail.com>
 Jon Sneyers <jon@cloudinary.com>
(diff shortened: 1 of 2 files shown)


Vargol:
I've now signed a CLA, both unsigned contributors are me, an you please try a re-scan

veluca93:
> I've now signed a CLA, both unsigned contributors are me, an you please try a re-scan

I think you need to change the author of the commit to use your email, rather than the ".local" placeholder.

`git config --local user.email ...` + `git commit --amend --reset-author` ought to do the trick.

Vargol:
I've done the git config change and added myself to to authors file, which was also failing.

veluca93:
> I've done the git config change and added myself to to authors file, which was also failing.

I think you still have one of the (3?) commits with the wrong author.

Vargol:
I've tried a few things but I don't see to be able to change the author of that first commit, I just don't know git that well. Would it be easier to drop this pull request and start again ?

mo271:
> I've tried a few things but I don't see to be able to change the author of that first commit, I just don't know git that well. Would it be easier to drop this pull request and start again ?

If you don't mind, I will give it a try, rebasing and squashing, keeping you as the author.

Vargol:
No I don't mind fell free to do what you can

mo271:
I put the Authors again in alphabetical order.. 
So the Authors file and the cla/google test look green now.
Thanks for you contribution!

Vargol:
Thats get news, If I make another contribution hopefully it won't be as painful.

mo271 (approved):

PR #1110, opened 2022-01-20
Adding a general-purpose bisector tool

fischbacher:
This allows us to do generic 'bisect to
target-size/target-bits-per-pixel' searches without having bisection
in the JXL encoder. Picking target quality settings via bisection is
relevant for performance comparison of codecs (when encoding the same
image to a given size with different encoders), but it makes more
sense to not have support for this directly in the encoder due to
potential for confusion: It in general does not make much sense to
encode images of very different visual complexity in a corpus to
identical bits-per-pixel, since this would give very unhomogeneous
quality.

### DIFF
--- a/tools/bisector
+++ b/tools/bisector
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 r"""General-purpose bisector
 
+Prints a space-separated list of values to stdout:
+1_if_success_0_otherwise left_x left_f(x) right_x right_f(x)
+
 Usage examples:
 
 # Finding the square root of 200 via bisection:
@@ -55,10 +58,10 @@ def _expandvars(vardef, env,
   """os.path.expandvars() variant using parameter env rather than os.environ."""
   current_expanded = vardef
   for num_recursions in range(max_recursion):
-    if verbosity >= 4:
+    if verbosity >= 3:
       print(f'_expandvars(): num_recursions={num_recursions}, '
             f'len={len(current_expanded)}' +
-            (', current: ' + current_expanded if verbosity >= 5 else ''))
+            (', current: ' + current_expanded if verbosity >= 4 else ''))
     if len > max_length:
         break
     current_expanded, num_replacements = re.subn(
@@ -157,7 +160,7 @@ def _bisect(*,
       delta_arg = bisect_right - bisect_left
       # Also check whether the argument is "within tolerance".
       # Here, we have to be careful if bisect_left and bisect_right
-      # have different sign: Then, their absolute magnitude
+      # have different signs: Then, their absolute magnitude
       # "sets the relevant scale".
       if abs(delta_arg) <= atol_arg + (
               rtol_arg * 0.5 * (abs(bisect_left) + abs(bisect_right))):
--- a/tools/bisector
+++ b/tools/bisector
@@ -4,37 +4,37 @@ r"""General-purpose bisector
 Usage examples:
 
 # Finding the square root of 200 via bisection:
-bisect.py --var=BB --range=0.0,100.0 --target=200 --maxiter=100 \
-          --atol_val=1e-12 --rtol_val=0 --cmd='echo "$BB * $BB" | bc'
+bisector --var=BB --range=0.0,100.0 --target=200 --maxiter=100 \
+         --atol_val=1e-12 --rtol_val=0 --cmd='echo "$BB * $BB" | bc'
 # => 1 14.142135623730923 199.99999999999923 14.142135623731633 200.0000000000193
 
 # Finding an integer approximation to sqrt(200) via bisection:
-bisect.py --var=BB --range=0,100 --target=200 --maxiter=100 \
-          --atol_arg=1 --cmd='echo "$BB * $BB" | bc'
+bisector --var=BB --range=0,100 --target=200 --maxiter=100 \
+         --atol_arg=1 --cmd='echo "$BB * $BB" | bc'
 # => 1 14 196.0 15 225.0
 
 # Finding a change-id that broke something via bisection:
-bisect.py --var=BB --range=0,1000000 --target=0.5 --maxiter=100 \
-          --atol_arg=1 \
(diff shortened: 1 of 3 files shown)


jonsneyers:
I agree with the description but perhaps it would be useful to have a more user-friendly way to use this to produce a jxl file of a given bpp, e.g. a script that calls this script so you can do something like `cjxl_bpp 0.9 input.png output.jxl [options passed to cjxl]` instead of having to type the monstrous command line that is in the example comment :)

I agree that it's good not to have this in cjxl itself since this is useful only for evaluation, not for practical use. But that doesn't mean we have to make it excessively tricky to do evaluation :)

tools/bisector
```diff
@@ -0,0 +1,278 @@
...
+              f'left: {value_left} at {bisect_left}, '
+              f'right: {value_right} at {bisect_right}, '
+              f'mid: {value_mid} at {bisect_mid}')
+      delta_val = target - value_mid
+      if abs(delta_val) <= atol_val + rtol_val * abs(target):
+        return 1, bisect_left, value_left, bisect_right, value_right
+      delta_arg = bisect_right - bisect_left
+      # Also check whether the argument is "within tolerance".
+      # Here, we have to be careful if bisect_left and bisect_right
+      # have different sign: Then, their absolute magnitude
```
mo271, line 160: super nit: plural "signs"

tools/bisector
```diff
@@ -0,0 +1,278 @@
...
+        bisect_right=high_val,
+        rtol_val=parsed.rtol_val,
+        atol_val=parsed.atol_val,
+        rtol_arg=parsed.rtol_arg,
+        atol_arg=parsed.atol_arg,
+        maxiter=parsed.maxiter,
+        extra_env_defs=extra_env_defs,
+        verbosity=parsed.verbosity,
+    )
+    print(' '.join(map(str, ret)))
```
mo271, line 272: Is it explained somewhere what the first return value (`1` or `0`) means? Presumably success or not. Consider adding an explanation in the doc string of `_bisect`, in the help or elsewhere.

fischbacher:
> I agree with the description but perhaps it would be useful to have a more user-friendly way to use this to produce a jxl file of a given bpp, e.g. a script that calls this script so you can do something like `cjxl_bpp 0.9 input.png output.jxl [options passed to cjxl]` instead of having to type the monstrous command line that is in the example comment :)
> 
> I agree that it's good not to have this in cjxl itself since this is useful only for evaluation, not for practical use. But that doesn't mean we have to make it excessively tricky to do evaluation :)

Makes perfect sense to add ~two simple #!/bin/sh scripts that shoot for a given target-bpp / target-size simply by calling the generic swiss army knife bisector-tool (while passing through further arguments to cjxl). Will see if I can get round to adding this to this pull request tomorrow.

mo271 (dismissed):
Unless I have overlooked something, there is no distinction between 2 and 3 for verbosity, so perhaps just merge them into 2 and change 4 to 3 and 5 to 4? 
Also in the `help` of `--verbosity` the relevant levels could be mentioned. 
In the functions where the verbosity is passed down, it's default value is always 0, but in the arg-parsing when nothing is provided the default it is `1`.  Might be a little less confusing if it was always default 1 in the definition of those function and not just then they are called.

Is the name going to be `tools/bisector` or `tools/bisect.py`?

veluca93:
I'm not sure a general-purpose bisector is the best approach here: the optimization in cjxl makes use of the information that `bpp*target_distance ~ constant` on a given image, and a general purpose bisector would be significantly slower.

fischbacher:
On: Verbosity levels. This was partly influenced by 80's line-number-infilling thinking, i.e. we might want to add other verbosity levels in-between later without changing the meaning of the existing ones. In hindsight, it is however unlikely that this tool will see relevant evolution, so compacting the range makes more sense. Adjusted levels as proposed.

tools/bisector
fischbacher: Done. (Even though grammatically, both would work. But it aligns the sentence structure with numpy's corresponding ValueError.)

tools/bisector
fischbacher: Explaining this at module level now.

fischbacher:
Ad more specialized tools for jxl-bisecting: We now have these.

fischbacher:
Ad: "a general purpose bisector would be significantly slower".

(a) We zoom in on the value logarithmically, gaining one bit per step. We typically do not need more than 20 bits.
(b) This functionality is only required in rather uncommon circumstances - when collecting data for codec performance evaluation.

In that sense: Is this actually relevant?

If we do consider this to be relevant, I still think that "separation of concerns", i.e. having a separate thing that is not jxl-specific, makes more sense here. What might be the right thing to do here is to make the bisector a bit smarter so that it optionally can use interpolation to guide the (currently always even) split, or perhaps a caller-provided model. For the time being, this however looks like a rather minor thing to fix to me - is there a reason why this would be more important?

fischbacher:
Ad: "a general purpose bisector would be significantly slower".

(a) We zoom in on the value logarithmically, gaining one bit per step. We typically do not need more than 20 bits.
(b) This functionality is only required in rather uncommon circumstances - when collecting data for codec performance evaluation.

In that sense: Is this actually relevant?

If we do consider this to be relevant, I still think that "separation of concerns", i.e. having a separate thing that is not jxl-specific, makes more sense here. What might be the right thing to do here is to make the bisector a bit smarter so that it optionally can use interpolation to guide the (currently always even) split, or perhaps a caller-provided model. For the time being, this however looks like a rather minor thing to fix to me - is there a reason why this would be more important?

mo271 (approved):
LGTM
One thing to consider might be to move the three new file to a dedicated directory inside `tools` in order to avoid cluttering `tools` too much. Perhaps a sub-directory `bisection` might be a good name?

fischbacher:
Did think about introducing a dedicated directory. Overall, this change is introducing two shell scripts and one python script. Once we have 5 or so, having a scripts/ directory might make sense.

PR #1111, opened 2022-01-21
adds .pam input support

jonsneyers:
We didn't have a way to pass uncompressed images with alpha to cjxl: .ppm and .pfm cannot do alpha. This adds .pam input. Still to be added is .pam output, but that's for another pull request.


mo271:
Do we plan on doing .pam output as well at some point?

jonsneyers:
> Do we plan on doing .pam output as well at some point?

We probably should. Atm we have ppm for uncompressed output, but that cannot handle alpha so atm you're forced to use slow png output when decoding images with alpha.

For output, I recommend we use the extension `.pnm` for "pick whatever is needed", which would use .pam if there's alpha, .ppm if there's no alpha, and .pgm if it's grayscale. That's the most convenient way to do it — and if you want specific output as .pam, .ppm or .pgm, you can use those extensions (in that case you'd get all-opaque alpha or grayscale-as-RGB if you use a format that can do "more than needed", and an error if you use a format that can do "less than needed").

veluca93 (approved):

PR #1115, opened 2022-01-24
cjxl_ng: Wiring up more flags

fischbacher:


tools/cjxl_ng_main.cc
```diff
@@ -46,35 +56,19 @@ DEFINE_bool(responsive, false, "[modular encoding] do Squeeze transform");
...
             "pixels instead of doing lossless transcode).");
 
-DEFINE_bool(jpeg_transcode_disable_cfl, false,  // TODO(tfish): Wire this up.
-            "Disable CFL for lossless JPEG recompression");
-
 DEFINE_bool(premultiply, false,  // TODO(tfish): Wire this up.
             "Force premultiplied (associated) alpha.");
 
-DEFINE_bool(centerfirst, false,  // TODO(tfish): Wire this up.
-            "Put center groups first in the compressed file.");
```
mo271, line 72: this is removed, because it is the same as the "group_order" flag, correct?

mo271 (approved):

mo271:
I rebased this because there was an error for the AUTHORS check from the merge in github

tools/cjxl_ng_main.cc
fischbacher: Yes, I think this functionality is now available via --group_order.

PR #1117, opened 2022-01-25
convert CodecInOut to PackedPixelFile

mo271:
Adding a new function fo convert a CodecInOut to a PackedPixelFile. The
idea is to use this for djxl when changing the encoder in `extras/enc/`
to use PackedPixelFile instead of CodecInOut.

### DIFF
--- a/lib/extras/packed_image_convert.cc
+++ b/lib/extras/packed_image_convert.cc
@@ -160,8 +160,7 @@ Status ConvertCodecInOutToPackedPixelFile(const CodecInOut& io,
                io.metadata.m.bit_depth.exponent_bits_per_sample);
     alpha_premultiplied = alpha_channel->alpha_associated;
   }
-  const bool is_gray = io.metadata.m.color_encoding.IsGray();
-  (void)is_gray;
+
   // Convert the image metadata
   ppf->info.xsize = io.metadata.size.xsize();
   ppf->info.ysize = io.metadata.size.ysize();
@@ -185,10 +184,12 @@ Status ConvertCodecInOutToPackedPixelFile(const CodecInOut& io,
   ppf->info.animation.num_loops = io.metadata.m.animation.num_loops;
 
   // Convert the color encoding
-  // TODO(firsching): do specific ICC profiles need special treatment here?
-  ConvertInternalToExternalColorEncoding(io.metadata.m.color_encoding,
-                                         &ppf->color_encoding);
-  io.metadata.m.color_encoding.IsSRGB();
+  ppf->icc.assign(io.metadata.m.color_encoding.ICC().begin(),
+                  io.metadata.m.color_encoding.ICC().end());
+  if (ppf->icc.empty()) {
+    ConvertInternalToExternalColorEncoding(io.metadata.m.color_encoding,
+                                           &ppf->color_encoding);
+  }
 
   // Convert the extra blobs
   ppf->metadata.exif.assign(io.blobs.exif.begin(), io.blobs.exif.end());
--- a/lib/extras/packed_image_convert.cc
+++ b/lib/extras/packed_image_convert.cc
@@ -5,6 +5,13 @@
 
 #include "lib/extras/packed_image_convert.h"
 
+#include <cstdint>
+
+#include "jxl/color_encoding.h"
+#include "jxl/types.h"
+#include "lib/jxl/base/status.h"
+#include "lib/jxl/color_encoding_internal.h"
+#include "lib/jxl/dec_external_image.h"
 #include "lib/jxl/enc_external_image.h"
 
 namespace jxl {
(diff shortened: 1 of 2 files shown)


lib/extras/packed_image_convert.cc
```diff
@@ -156,5 +163,106 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
   return true;
 }
 
+Status ConvertCodecInOutToPackedPixelFile(const CodecInOut& io,
```
jyrkialakuijala, line 166: add a single line 'purpose' comment here

jyrkialakuijala (approved):

lib/extras/packed_image_convert.cc
mo271: Done

lib/extras/packed_image_convert.cc
```diff
@@ -156,5 +163,90 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
...
+  if (has_alpha) {
+    JXL_ASSERT(io.metadata.m.GetAlphaBits() ==
+               io.metadata.m.bit_depth.bits_per_sample);
+    const auto* alpha_channel = io.metadata.m.Find(ExtraChannel::kAlpha);
+    JXL_ASSERT(alpha_channel->bit_depth.exponent_bits_per_sample ==
+               io.metadata.m.bit_depth.exponent_bits_per_sample);
+    alpha_premultiplied = alpha_channel->alpha_associated;
+  }
+  const bool is_gray = io.metadata.m.color_encoding.IsGray();
+  (void)is_gray;
```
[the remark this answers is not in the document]
jonsneyers: I would recommend to make it work (eventually) to produce grayscale ppf when possible. It is quite annoying for use cases that deal with large grayscale images to always get things as RGB. It could also be a problem when grayscale ICC profiles get attached to RGB png files, for example.

lib/extras/packed_image_convert.cc
```diff
@@ -156,5 +163,90 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
...
+    JXL_RETURN_IF_ERROR(ConvertToExternal(
+        frame.Copy(), frame_bits_per_sample, float_out, format.num_channels,
+        format.endianness,
+        /* stride_out=*/packed_frame.color.stride, pool,
+        packed_frame.color.pixels(), packed_frame.color.pixels_size,
+        /*out_callback=*/nullptr, /*out_opaque=*/nullptr,
+        frame.metadata()->GetOrientation()));
+
+    // TODO(firsching): Convert the extra channels. FIXME!
+    JXL_CHECK(frame.extra_channels().empty());
```
jonsneyers, line 245: I think at least alpha should be made to work. Other extra channels can be ignored for now.

mo271: If I understand correctly, currently the (png, jpg, etc) decoder, that decode images with also as having 4 color channels and 0 extra channels, hence they are using those two fields in `JxlBasicInfo` with a different meaning. 
I agree that alpha should work, and for now, it should work when it comes as interleaved. 
I plan to get some more clarity of what is going on exactly here, and take action in a follow up pull request.

mo271 (commented):
Thanks for the review @sboukortt  and @jonsneyers!

lib/extras/packed_image_convert.cc
jonsneyers: `frame.color().kNumPlanes` is always going to be 3, if I'm not mistaken. Alpha is never included in `color()`, and I think it's always going to be RGB, even in the case of grayscale (not sure though).

jonsneyers: Things are confusing here.

`frame` here is a `CodecInOut` frame, which at least at the moment has the convention that `color()` is always Image3F (so planar float32 RGB) and anything else (including alpha) is in the extra channels.

A `PackedPixelFile` uses a representation similar to what we use in the API (interleaved, arbitrary pixel types, 1 to 4 channels (Y, YA, RGB, RGBA), any other extra channels are separate).

But in this direction, from CodecInOut to PackedPixelFile, the input alpha will always be separate, and it's probably best to return a PackedPixelFile with the alpha interleaved with the RGB because that's what the PNG encoder (and PAM if we add that at some point) will need.

mo271: There is also `frame.alpha`, right?



> But in this direction, from CodecInOut to PackedPixelFile, the input alpha will always be separate, and it's probably best to return a PackedPixelFile with the alpha interleaved with the RGB because that's what the PNG encoder (and PAM if we add that at some point) will need.
I agree and this is exactly what I'm trying to do. I also think that this works already in the code, because `ConvertToExtrenal` should do the interleaving if alpha is present in the image bundle and the num_channels is provided correctly. 

I'll change the check 
```frame.extra_channels().empty()```, which should have been 
```frame.HasExtraChannels()``` anyhow to check that it has any other extra channels, besides a potential alpha channel. Also I make sure that num_channels is correct.

File: lib/extras/packed_image_convert.cc [hunk 4]
```diff
@@ -136,5 +146,114 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
...
+
+  // Convert animation metadata
+  JXL_ASSERT(io.frames.size() == 1 || io.metadata.m.have_animation);
+  ppf->info.have_animation = io.metadata.m.have_animation;
+  ppf->info.animation.tps_numerator = io.metadata.m.animation.tps_numerator;
+  ppf->info.animation.tps_denominator = io.metadata.m.animation.tps_denominator;
+  ppf->info.animation.num_loops = io.metadata.m.animation.num_loops;
+
+  // Convert the color encoding
+  ppf->icc.assign(io.metadata.m.color_encoding.ICC().begin(),
```
jonsneyers, line 193: Shouldn't this be the c_desired color profile? Otherwise the ppf gets the original icc profile with pixels in the desired converted space...

mo271: changed to using `c_desired`

jonsneyers (approved):

PR #1118, opened 2022-01-25
Remove duplicate function converting JxlDataType

mo271:
There are two identical functions converting JxlDataType to the number
of bits per channel. We remove one of them.


mo271:
Not entirely sure that yet a better place for that function would be the top of [packed_image.h](main/lib/extras/packed_image.h) outside the `PackedImage` class.

jyrkialakuijala (approved):

PR #1121, opened 2022-01-26
decode to PackedPixelFile

mo271:
We add a function fo decode compressed jxl data to a PackeDPixelFile.

### DIFF
--- a/tools/djxl_ng_main.cc
+++ b/tools/djxl_ng_main.cc
@@ -234,7 +234,7 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
 }
 
 int DecompressJxlToPackedPixelFile(const std::vector<uint8_t>& compressed,
-                                   jxl::extras::PackedPixelFile& pixels,
+                                   jxl::extras::PackedPixelFile& ppf,
                                    JxlPixelFormat& format, JxlDecoderPtr dec,
                                    JxlThreadParallelRunnerPtr runner) {
   if (JXL_DEC_SUCCESS != JxlDecoderSetParallelRunner(dec.get(),
@@ -267,9 +267,9 @@ int DecompressJxlToPackedPixelFile(const std::vector<uint8_t>& compressed,
       fprintf(stderr, "Error, already provided all input\n");
       return EXIT_FAILURE;
     } else if (status == JXL_DEC_BASIC_INFO) {
-      if (JXL_DEC_SUCCESS != JxlDecoderGetBasicInfo(dec.get(), &pixels.info)) {
+      if (JXL_DEC_SUCCESS != JxlDecoderGetBasicInfo(dec.get(), &ppf.info)) {
         fprintf(stderr, "JxlDecoderGetBasicInfo failed\n");
-        break;
+        return EXIT_FAILURE;
       }
       // TODO(firsching): handle extra channels
     } else if (status == JXL_DEC_COLOR_ENCODING) {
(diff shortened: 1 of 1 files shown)


tools/djxl_ng_main.cc
```diff
@@ -233,6 +235,107 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
   return EXIT_SUCCESS;
 }
 
+int DecompressJxlToPackedPixelFile(const std::vector<uint8_t>& compressed,
+                                   jxl::extras::PackedPixelFile& pixels,
```
fischbacher, line 239: The name `pixels` looks a bit misleading here. How about renaming this to `ppf`?

tools/djxl_ng_main.cc
```diff
@@ -233,6 +235,107 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
...
+  if (JXL_DEC_SUCCESS != JxlDecoderSetParallelRunner(dec.get(),
+                                                     JxlThreadParallelRunner,
+                                                     runner.get())) {
+    fprintf(stderr, "JxlEncoderSetParallelRunner failed\n");
+    return EXIT_FAILURE;
+  }
+  if (JXL_DEC_SUCCESS !=
+      JxlDecoderSubscribeEvents(dec.get(),
+                                JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING |
+                                    JXL_DEC_FRAME | JXL_DEC_FULL_IMAGE)) {
```
fischbacher, line 251: (indenting is a bit weird here. Might need a `clang-format -i code.cc`.)

tools/djxl_ng_main.cc
```diff
@@ -233,6 +235,107 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
...
+    if (status == JXL_DEC_ERROR) {
+      fprintf(stderr, "Failed to decode image\n");
+      return EXIT_FAILURE;
+    } else if (status == JXL_DEC_NEED_MORE_INPUT) {
+      fprintf(stderr, "Error, already provided all input\n");
+      return EXIT_FAILURE;
+    } else if (status == JXL_DEC_BASIC_INFO) {
+      if (JXL_DEC_SUCCESS != JxlDecoderGetBasicInfo(dec.get(), &pixels.info)) {
+        fprintf(stderr, "JxlDecoderGetBasicInfo failed\n");
+        break;
```
fischbacher, line 274: What's the principle that decides that we should break here rather than returning EXIT_FAILURE?

tools/djxl_ng_main.cc
```diff
@@ -233,6 +235,107 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
...
+      if (JXL_DEC_SUCCESS !=
+          JxlDecoderGetICCProfileSize(dec.get(), &format, target, &icc_size)) {
+        fprintf(stderr, "JxlDecoderGetICCProfileSize failed\n");
+      }
+      if (icc_size != 0) {
+        pixels.icc.resize(icc_size);
+        if (JXL_DEC_SUCCESS !=
+            JxlDecoderGetColorAsICCProfile(dec.get(), &format, target,
+                                           pixels.icc.data(), icc_size)) {
+          fprintf(stderr, "JxlDecoderGetColorAsICCProfile failed\n");
```
fischbacher, line 290: If icc_size != 0, but we cannot GetColorAsICCProfile, would we then not also want to early-exit?

A `break` will make us run into a `return EXIT_SUCCESS` in l.336, no?

tools/djxl_ng_main.cc
```diff
@@ -233,6 +235,107 @@ int DecompressJxlReconstructJPEG(const std::vector<uint8_t>& compressed,
...
+        if (JXL_DEC_SUCCESS !=
+            JxlDecoderGetColorAsEncodedProfile(dec.get(), &format, target,
+                                               &pixels.color_encoding)) {
+          fprintf(stderr, "JxlDecoderGetColorAsEncodedProfile failed\n");
+        }
+      }
+    } else if (status == JXL_DEC_FRAME) {
+      jxl::extras::PackedFrame frame(pixels.info.xsize, pixels.info.ysize,
+                                     format);
+      pixels.frames.push_back(std::move(frame));
```
fischbacher, line 302: Are we using C++17? Could we do .emplace_back() here?

fischbacher (approved):
This is work-in-progress towards getting the C API based decoder wired up.

tools/djxl_ng_main.cc
veluca93: We aren't, but why couldn't we emplace_back regardless?

mo271: changed to emplace_back

mo271 (commented):
Thanks for the review!

sboukortt (approved):

PR #1127, opened 2022-01-27
cjxl_ng: Adding animated png/gif support and more flags.

fischbacher:


tools/cjxl_ng_main.cc
```diff
@@ -695,49 +795,83 @@ int main(int argc, char** argv) {
...
+      basic_info.animation.tps_numerator = ppf_info.animation.tps_numerator;
+      basic_info.animation.tps_denominator = ppf_info.animation.tps_denominator;
+      basic_info.animation.num_loops = ppf_info.animation.num_loops;
+
+      basic_info.num_extra_channels =
+          (ppf.frames[0].color.format.num_channels == 2 ||
+           ppf.frames[0].color.format.num_channels == 4)
+              ? 1
+              : 0;
+      basic_info.uses_original_profile = JXL_TRUE;
```
mo271, line 817: doing it that way does not work on animations without alpha, I think. Did you test that?
I suggest to use the `ppf.info` directly without making another `basic_info`, basically like here:
https://github.com/mo271/libjxl/commit/18703a0312b6af8e4832d13a68ae2abb4f751eae

Hardcoding 'uses_original_profile` will lead to file size increase.

mo271 (commented):
Might it make sense to make "adding the animations" and "adding more flags" in two seperate commits? (No problem if not)

tools/cjxl_ng_main.cc
mo271: Also currently this does not work with animation containing frame with frame dimensions less than the dimension of the image. I think we should either catch that (rather common) case and give a meaningful error or fix it

fischbacher: This needs tests anyhow. Overall, we are still in "let's get this feature-complete" phase. There will have to be a cleanup afterwards.

tools/cjxl_ng_main.cc
```diff
@@ -309,8 +337,9 @@ jxl::Status LoadInput(const char* filename_in,
   if (image_data.size() < kMinBytes) return JXL_FAILURE("Input too small.");
   jxl::Span<const uint8_t> encoded(image_data);
 
-  // Default values when not set by decoders.
-  ppf.info.uses_original_profile = true;
+  // Manually fix intensity target 0 problem, decoder should set a
+  // reasonable default, but at least apng doesn't:
+  ppf.info.intensity_target = 255.f;
```
mo271, line 342: This should be taken care of soon, when we initialize ppf properly.

mo271: see #1129

tools/cjxl_ng_main.cc
```diff
@@ -611,6 +671,18 @@ int main(int argc, char** argv) {
           jxl_encoder_frame_settings, JXL_ENC_FRAME_SETTING_MODULAR_COLOR_SPACE,
           FLAGS_modular_colorspace);
     }
+    if (modular_ma_tree_learning_percent_set) {
+      if (!(-1 <= FLAGS_modular_ma_tree_learning_percent &&
+            FLAGS_modular_ma_tree_learning_percent <= 100)) {
```
jonsneyers, line 676: Values up to 1000 actually have different encoder behavior (it's the percent of samples used in MA tree learning, but also the promille of samples used to determine how to do property quantization when doing sample collection, iirc).

jonsneyers (approved):

mo271:
Is this superseded by https://github.com/libjxl/libjxl/pull/1179 ? If yes, perhaps we should close this pull request here...

fischbacher:
Obsolete now

PR #1130, opened 2022-01-28
Add unified safeguard for truncated group decoding.

eustas:
Before planes were zeroified only if the very late stage failed.
Now the scope guard is placed after meta-apply, so that there is no
escape path to leave planes uninitialized (if truncated groups are allowed).

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -11,6 +11,7 @@
 #include <queue>
 
 #include "lib/jxl/base/printf_macros.h"
+#include "lib/jxl/base/scope_guard.h"
 #include "lib/jxl/modular/encoding/context_predict.h"
 #include "lib/jxl/modular/options.h"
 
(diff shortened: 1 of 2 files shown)


eustas:
Used ScopeGuard. PTAL

lib/jxl/modular/encoding/encoding.cc
```diff
@@ -403,6 +404,13 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   }
 
   size_t nb_channels = image.channel.size();
+  size_t next_channel = 0;
+  auto scope_guard = MakeScopeGuard([&]() {
```
jonsneyers, line 408: In the case where `num_chans == 0` in the code below (which I think might happen e.g. when decoding modular DC groups in a large image that doesn't do squeeze, so the DC groups are empty), starting the scope guard here will cause all channels to get zero-initialized while that is not really necessary in such a case.

Maybe start the scope guard after the line with `if (num_chans == 0) return true;`?

jonsneyers:
Could you check if this code gets called in normal non-truncated decodes? Because we're trying to avoid unnecessary zero initialization (for speed reasons), and I'm a bit worried that the way it is written now, the initialization might trigger too easily. E.g. check if the output of `cjxl -m -e 1` and `cjxl -m -R 1 -e 1` (on images of various sizes) can be decoded without triggering the ZeroFill...

lib/jxl/modular/encoding/encoding.cc
eustas: Sure! Thanks.

eustas:
Tried on few examples, no ZF.
To guarantee no ZF happens added "disarm" for guard and used it before the only non-error return.

jonsneyers (approved):

PR #1132, opened 2022-01-28
Make sure PNG memory is released in any case.

eustas:


lib/extras/dec/apng.cc
```diff
@@ -405,6 +405,14 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
 
   std::vector<FrameInfo> frames;
 
+  auto before_return = [&](void*) {
+    // It is safe to invoke this method repeatedely; so make sure memory is
+    // released in any case.
+    png_destroy_read_struct(&png_ptr, &info_ptr, 0);
+  };
+  std::unique_ptr<void, decltype(before_return)> scope_guard((void*)1,
+                                                             before_return);
```
veluca93, line 414: I'd rather

struct Guard {
 ~Guard() { ... }
} guard;

eustas: Sure. It seems that there are 2 guards already -> time to extract template for that.

eustas:
PTAL

lib/jxl/base/scope_guard.h
```diff
@@ -0,0 +1,46 @@
...
+
+template <typename Callback>
+class ScopeGuard {
+ public:
+  // Discourage unnecessary moves / copies.
+  ScopeGuard(const ScopeGuard &) = delete;
+  ScopeGuard &operator=(const ScopeGuard &) = delete;
+  ScopeGuard &operator=(ScopeGuard &&) = delete;
+
+  // Some compilers have different RVO rules -> require move constructor.
```
veluca93, line 21: More precisely, before C++17 RVO is not guaranteed to happen...

veluca93 (approved):

PR #1136, opened 2022-01-30
Macro Based Definitions For Public Facing API

KoromaruKoruko:
Example code of using Macros to define Structures and Enumerables.
And then an Example of how it could be beneficial to have this.

The real reason this would be good to implement, would be dynamically generating binding data.
For a real world use case, I am working on getting libjxl to work nicely with client wasm.
Allowing for use of the libjxl library within the client browser.
I need to generate proper bindings for the JavaScript to nicely play with the native c/c++.
Having all the used enums and structs, defined like this, means I can write a few preprocessors include a bunch of files, and have all the bindings for those types be dealt with, in a future proofed manor.
Otherwise to do something similar I would need to wrap all the structures and inject the required code, then I would manually have to write each individual struct   member,   enum and value.
Which would then also mean that any future edits would have to be carried out across multiple files to preserve functionality.


KoromaruKoruko:
Note this is not the full implementation,
this is mearly a PR to gather everyone's opinion on implementation.
as well as a vote weather or not to implement at all.

eustas:
We have used similar approach in brotli... And it seems that no-one has actually used those enums.

I believe, we should rather look into the neutral ground, e.g. WebIDL. We could define all those structs, enums, methods in that language and generate header files from them. On CI we could check that headers correspond to definition.
If any binding is required, this could be also generated from idl file.

KoromaruKoruko:
I have looked into WebIDL and it seems, to be less feature rich then just using embind (as I have been doing for the current wasm interface) lacking a key feature such as unique pointers, which is used in construction of the Decoder and Encoder.

further more it would require new means of documentation and still require us to maintain 2 versions of the interface. one in c++ and one is idl. webidl specifically does not seem to like functional non-bound functions (always requiring and interface or typename). thus meaning a separate API structure would need to be maintained for the library. which probably isn't that drastic for the end-user (given they are more likely used to OOP Environments) but would mean a bit more thinking on where a global method should be placed.

eustas:
That was one thing I have been worrying about. Perhaps it will be better to write some AST transformer to convert c++ header to a different representation....

KoromaruKoruko:
> That was one thing I have been worrying about. Perhaps it will be better to write some AST transformer to convert c++ header to a different representation....

I have been debating this as well, never the less I would incorporate it using this basic structure. As abstracting the complexities of c/c++ type definitions into an easier to parse layout, would make implementing such a transformer/transpiler much much simpler.

KoromaruKoruko:
@eustas I have done the main conversion.
However I am unable to properly build for anything other then wasm...
But the docs do get generated correctly (haven't tested it with Sphinx or Dot installed yet)

and I have gone over the files twice to check for any coding mistakes, as simply said converting these files was mind numbing.
especially decode.h and encode.h

mo271:
This seems like it is very much out-of-date, because of various restructures of our build system.

PR #1142, opened 2022-02-01
Pipeline stage for blending

sboukortt:
The stage only handles the part of the image corresponding to the current crop, and therefore the pipeline does not yet work for animations where not all frames have the same size.


veluca93 (approved):

sboukortt:
Sorry, forgot about build_cleaner.py…

veluca93 (approved):

PR #1148, opened 2022-02-01
SIMDify inverse YCoCg

jonsneyers:
Small speedup for decoding e.g. jxl files produced with fjxl or cjxl -m -e 1-3.

### DIFF
--- a/lib/jxl/modular/transform/rct.cc
+++ b/lib/jxl/modular/transform/rct.cc
@@ -4,8 +4,15 @@
 // license that can be found in the LICENSE file.
 
 #include "lib/jxl/modular/transform/rct.h"
-
+#undef HWY_TARGET_INCLUDE
+#define HWY_TARGET_INCLUDE "lib/jxl/modular/transform/rct.cc"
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/transform/rct.cc
```diff
@@ -15,7 +22,25 @@ void InvRCTRow(const pixel_type* in0, const pixel_type* in1,
...
+      auto Cg = Load(d, in2 + x);
+      Y -= ShiftRight<1>(Cg);
+      auto G = Cg + Y;
+      Y -= ShiftRight<1>(Co);
+      auto R = Y + Co;
+      Store(R, d, out0 + x);
+      Store(G, d, out1 + x);
+      Store(Y, d, out2 + x);
+    }
+  }
```
veluca93, line 42: While we're at it, can we also SIMDfy the other case? it shouldn't be hard.

jonsneyers: Done.

veluca93 (approved):

PR #1149, opened 2022-02-02
More fjxl decode speedups

jonsneyers:
Some more tweaks to the specialized decode path for the case of fjxl:
(building on top of #1148)

Before:
3072 x 2048, geomean: 32.41 MP/s [25.23, 34.52], 30 reps, 0 threads.
3072 x 2048, geomean: 58.47 MP/s [41.42, 61.41], 30 reps, 2 threads.
3072 x 2048, geomean: 94.37 MP/s [53.14, 100.48], 30 reps, 4 threads.

After:
3072 x 2048, geomean: 46.57 MP/s [33.95, 48.11], 30 reps, 0 threads.
3072 x 2048, geomean: 79.85 MP/s [51.41, 85.18], 30 reps, 2 threads.
3072 x 2048, geomean: 122.51 MP/s [62.37, 131.39], 30 reps, 4 threads.


lib/jxl/modular/encoding/encoding.cc
```diff
@@ -210,19 +210,19 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
...
         if (run == 0) {
-          reader->ReadHybridUintClusteredHuffRleOnly(ctx_id, br, &v, &run);
+          reader->ReadHybridUintClusteredHuffRleOnly(br, &v, &run);
           sv = UnpackSigned(v);
         } else {
           run--;
         }
         r[0] = sv + guess;
-        for (size_t x = 1; x < channel.w; x++) {
+        for (size_t x = 1; x != channel.w; x++) {
```
veluca93, line 219: Why this change?

lib/jxl/modular/encoding/context_predict.h
```diff
@@ -334,21 +334,20 @@ constexpr size_t kGradientProp = 9;
...
-  // grad = M + m - l = n + w - l
-  const int32_t grad =
-      static_cast<int32_t>(static_cast<uint32_t>(n) + static_cast<uint32_t>(w) -
-                           static_cast<uint32_t>(l));
-  // We use two sets of ternary operators to force the evaluation of them in
-  // any case, allowing the compiler to avoid branches and use cmovl/cmovg in
-  // x86.
-  const int32_t grad_clamp_M = (l < m) ? M : grad;
-  return (l > M) ? m : grad_clamp_M;
+  const int32_t ab =
```
veluca93, line 337: A bunch of comments here would be very appreciated :)

lib/jxl/dec_ans.h
```diff
@@ -275,25 +275,36 @@ class ANSSymbolReader {
     return static_cast<uint32_t>(ret);
   }
 
-  // Takes a *clustered* idx. Can only use if HuffRleOnly() is true.
-  void ReadHybridUintClusteredHuffRleOnly(size_t ctx,
-                                          BitReader* JXL_RESTRICT br,
-                                          uint32_t* value, uint32_t* run) {
+  // Takes a *clustered* idx, implicit ctx==1. Can only use if HuffRleOnly() is
+  // true.
+  void JXL_INLINE ReadHybridUintClusteredHuffRleOnly(BitReader* JXL_RESTRICT br,
```
veluca93, line 280: Probably also should mention it's 000/100 only...

lib/jxl/modular/encoding/encoding.cc
jonsneyers: Ah, no particular reason, I tried (unsuccessfully) to write that loop in some other way and that's a leftover from that attempt.

lib/jxl/dec_ans.h
veluca93: At this point one might replace "HuffRleOnly" with something like FJXLFastPath...

lib/jxl/modular/encoding/context_predict.h
jonsneyers: Added a bunch of them to help explain / prove why this actually works.

veluca93: Maybe we could also add a test that checks a bunch of cases vs the naive impl?

jonsneyers: When I was trying to come up with this somewhat less-branchy reformulation, I always did it by just brute-force checking, which is much easier and less error-prone than proving correctness :)
Something like this:
```C
for (int A = -300; A < 300; A++)
for (int B = -300; B < 300; B++)
for (int C = -300; C < 300; C++) {
    int a = ClampedGradient(A,B,C);
    int b = OtherClampedGradient(A,B,C);
    if (a != b) printf("NOT OK: %i %i %i -> %i != %i\n",A,B,C,a,b);
}
```

Something like that could be added as a unit test too but I'm not sure if that is very useful.

veluca93: I think it would be - especially for randomly weird systems...

jonsneyers: Fair enough, added a test that compares random combinations to the result of a naive implementation.

With values that cause int32 overflows in the intermediate results, there is a difference, so the test only tests values in the range [-2^29.. 2^29) — i.e. int30. Since libjxl cannot deal with integer bit depths > 24 anyway (since things have to be representable in binary32 floats), that's probably fine.

Lossless binary32 float encoding (reinterpreted as int32) still works, but the input does have to be in the [0.f,1.f) range (which maps to a subset of [0, 2^30) when reinterpreting as integer) or else weird overflows will happen. But that was already the case with the previous implementation too, I suppose.

lib/jxl/modular_test.cc
```diff
@@ -322,5 +324,22 @@ TEST(ModularTest, RoundtripLosslessCustomFloat) {
...
 
+TEST(ModularTest, ClampedGradientTest) {
+  Rng rng(0);
+  constexpr int max_bits = 29;
+  for (size_t i = 0; i < 100000; i++) {
+    int32_t a = rng.UniformI(-(1 << max_bits), 1 << max_bits);
+    int32_t b = rng.UniformI(-(1 << max_bits), 1 << max_bits);
+    int32_t c = rng.UniformI(-(1 << max_bits), 1 << max_bits);
+    int32_t cg = ClampedGradient(a, b, c);
+    int64_t g = a + b - c;
```
veluca93, line 335: This line does not do what you think it does in the presence of int32 overflow :) -- in fact, this "simple" implementation is in fact incorrect

jonsneyers: Sigh, it does the arithmetic in int32 with overflow and then stores it in an int64, right?
Great, I managed to make the "obviously correct" version incorrect :)

Is this an OK solution?
```
int64_t la = a, lb = b, lc = c;
int64_t g = la + lb - lc;
```

KoromaruKoruko: This should do what you want it to, so long as the compiler doesn't 'optimize' it by adding b and c first.
```c
int64_t g = static_cast<int64_t>(a) + b + c;
```

veluca93: that, or just int64_t{a}+b-c should work too.

KoromaruKoruko: actually looking at the underlying differenced it should be safe, and slightly faster to use a 'reinterpret_cast' as we're going from a smaller to a larger buffer, and this should just cause the compiler to push the int32_t into a int64_t compatible register.
without any checks.

lib/jxl/dec_ans.h
jonsneyers: Makes sense, done.

veluca93 (approved):

veluca93 (approved):

PR #1157, opened 2022-02-04
Faster unsqueeze / lossy modular

jonsneyers:
Some more specialization for the slower modular decode paths.

~~Vectorizeable unsqueeze.
Alas, my clang (12.0.1) requires a `pragma` to be convinced to vectorize the inner loop. Not sure how to proceed with that...~~

SIMDified unsqueeze.

Good decode speedup for images where this applies (lossy modular, lossless progressive modular, also should help a little for DC frames and VarDCT mode alpha channels).

Decoding a lossy modular image (produced with cjxl -m -q 80 -e 2), with AVX2:

Before:
2749 x 4049, geomean: 9.90 MP/s [8.67, 10.66], 10 reps, 0 threads.
2749 x 4049, geomean: 18.35 MP/s [15.42, 18.81], 10 reps, 2 threads.
2749 x 4049, geomean: 29.92 MP/s [23.30, 31.61], 10 reps, 4 threads.

After:
~~2749 x 4049, geomean: 14.94 MP/s [12.36, 15.50], 10 reps, 0 threads.
2749 x 4049, geomean: 26.91 MP/s [20.13, 28.25], 10 reps, 2 threads.
2749 x 4049, geomean: 39.48 MP/s [28.23, 40.84], 10 reps, 4 threads.~~

2749 x 4049, geomean: 17.56 MP/s [14.44, 17.95], 10 reps, 0 threads.
2749 x 4049, geomean: 31.24 MP/s [26.89, 32.48], 10 reps, 2 threads.
2749 x 4049, geomean: 46.14 MP/s [38.19, 50.40], 10 reps, 4 threads.

### DIFF
--- a/lib/jxl/modular/encoding/context_predict.h
+++ b/lib/jxl/modular/encoding/context_predict.h
@@ -304,7 +304,7 @@ class MATreeLookup {
     int64_t offset;
     int32_t multiplier;
   };
-  LookupResult Lookup(const Properties &properties) const {
+  JXL_INLINE LookupResult Lookup(const Properties &properties) const {
     uint32_t pos = 0;
     while (true) {
       const FlatDecisionNode &node = nodes_[pos];
@@ -416,6 +416,7 @@ enum PredictorMode {
   kUseWP = 2,
   kForceComputeProperties = 4,
   kAllPredictions = 8,
+  kNoEdgeCases = 16
 };
 
 JXL_INLINE pixel_type_w PredictOne(Predictor p, pixel_type_w left,
@@ -461,7 +462,7 @@ JXL_INLINE pixel_type_w PredictOne(Predictor p, pixel_type_w left,
 }
 
 template <int mode>
-inline PredictionResult Predict(
+JXL_INLINE PredictionResult Predict(
     Properties *p, size_t w, const pixel_type *JXL_RESTRICT pp,
     const intptr_t onerow, const size_t x, const size_t y, Predictor predictor,
     const MATreeLookup *lookup, const Channel *references,
@@ -470,13 +471,15 @@ inline PredictionResult Predict(
   size_t offset = 3;
   constexpr bool compute_properties =
       mode & kUseTree || mode & kForceComputeProperties;
-  pixel_type_w left = (x ? pp[-1] : (y ? pp[-onerow] : 0));
-  pixel_type_w top = (y ? pp[-onerow] : left);
-  pixel_type_w topleft = (x && y ? pp[-1 - onerow] : left);
-  pixel_type_w topright = (x + 1 < w && y ? pp[1 - onerow] : top);
-  pixel_type_w leftleft = (x > 1 ? pp[-2] : left);
-  pixel_type_w toptop = (y > 1 ? pp[-onerow - onerow] : top);
-  pixel_type_w toprightright = (x + 2 < w && y ? pp[2 - onerow] : topright);
+  constexpr bool nec = mode & kNoEdgeCases;
+  pixel_type_w left = (nec || x ? pp[-1] : (y ? pp[-onerow] : 0));
+  pixel_type_w top = (nec || y ? pp[-onerow] : left);
+  pixel_type_w topleft = (nec || (x && y) ? pp[-1 - onerow] : left);
+  pixel_type_w topright = (nec || (x + 1 < w && y) ? pp[1 - onerow] : top);
+  pixel_type_w leftleft = (nec || x > 1 ? pp[-2] : left);
+  pixel_type_w toptop = (nec || y > 1 ? pp[-onerow - onerow] : top);
+  pixel_type_w toprightright =
+      (nec || (x + 2 < w && y) ? pp[2 - onerow] : topright);
 
   if (compute_properties) {
     // location
@@ -506,7 +509,7 @@ inline PredictionResult Predict(
     wp_pred = wp_state->Predict<compute_properties>(
         x, y, w, top, left, topright, topleft, toptop, p, offset);
   }
-  if (compute_properties) {
+  if (!nec && compute_properties) {
     offset += weighted::kNumProperties;
     // Extra properties.
     const pixel_type *JXL_RESTRICT rp = references->Row(x);
@@ -562,6 +565,15 @@ inline PredictionResult PredictTreeNoWP(Properties *p, size_t w,
                                         const MATreeLookup &tree_lookup,
                                         const Channel &references) {
   return detail::Predict<detail::kUseTree>(
+      p, w, pp, onerow, x, y, Predictor::Zero, &tree_lookup, &references,
+      /*wp_state=*/nullptr, /*predictions=*/nullptr);
+}
+// Only use for y > 1, x > 1, x < w-2, and empty references
+JXL_INLINE PredictionResult
+PredictTreeNoWPNEC(Properties *p, size_t w, const pixel_type *JXL_RESTRICT pp,
+                   const intptr_t onerow, const int x, const int y,
+                   const MATreeLookup &tree_lookup, const Channel &references) {
+  return detail::Predict<detail::kUseTree | detail::kNoEdgeCases>(
       p, w, pp, onerow, x, y, Predictor::Zero, &tree_lookup, &references,
       /*wp_state=*/nullptr, /*predictions=*/nullptr);
 }
--- a/lib/jxl/modular/transform/squeeze.cc
+++ b/lib/jxl/modular/transform/squeeze.cc
@@ -12,13 +12,79 @@
 #include "lib/jxl/common.h"
 #include "lib/jxl/modular/modular_image.h"
 #include "lib/jxl/modular/transform/transform.h"
+#undef HWY_TARGET_INCLUDE
+#define HWY_TARGET_INCLUDE "lib/jxl/modular/transform/squeeze.cc"
+#include <hwy/foreach_target.h>
+#include <hwy/highway.h>
 
+#include "lib/jxl/simd_util-inl.h"
+
+HWY_BEFORE_NAMESPACE();
 namespace jxl {
+namespace HWY_NAMESPACE {
+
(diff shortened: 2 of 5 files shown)


lib/jxl/modular/encoding/context_predict.h
```diff
@@ -484,13 +485,15 @@ inline PredictionResult Predict(
...
   constexpr bool compute_properties =
       mode & kUseTree || mode & kForceComputeProperties;
-  pixel_type_w left = (x ? pp[-1] : (y ? pp[-onerow] : 0));
-  pixel_type_w top = (y ? pp[-onerow] : left);
-  pixel_type_w topleft = (x && y ? pp[-1 - onerow] : left);
-  pixel_type_w topright = (x + 1 < w && y ? pp[1 - onerow] : top);
-  pixel_type_w leftleft = (x > 1 ? pp[-2] : left);
-  pixel_type_w toptop = (y > 1 ? pp[-onerow - onerow] : top);
-  pixel_type_w toprightright = (x + 2 < w && y ? pp[2 - onerow] : topright);
+  constexpr bool nec = mode & kNoEdgeCases;
```
veluca93, line 488: I think it would be a better idea to allocate a slightly larger buffer, and prepare row pointers accordingly... (similarly to how fjxl does it)

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -43,36 +96,109 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
     return true;
   }
 
+#if HWY_TARGET != HWY_SCALAR
+
+  // somewhat complicated trickery just to be able to SIMD this
```
veluca93, line 101: Can you expand on what is that trickery? :)

lib/jxl/modular/encoding/context_predict.h
veluca93: I'm not sure it is in the long term...

jonsneyers: Agreed, we should do this at some point. It might be a bit trickier than in fjxl though, because decode-side you cannot do all padding in advance.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Added a comment that attempts to explain it.

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -12,13 +12,155 @@
...
 namespace jxl {
+namespace HWY_NAMESPACE {
+
+using hwy::HWY_NAMESPACE::ShiftLeft;
+using hwy::HWY_NAMESPACE::ShiftRight;
+
+#if HWY_TARGET != HWY_SCALAR
+
+template <size_t count>
+JXL_INLINE void fast_unsqueeze(const pixel_type *JXL_RESTRICT p_residual,
```
veluca93, line 29: style: FastUnsqueeze

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -12,13 +12,155 @@
...
+    auto next_avg = Load(d, p_navg + x);
+    auto top = Load(d, p_pout + x);
+    // Equivalent to SmoothTendency(top,avg,next_avg), but without branches
+    auto Ba = top - avg;
+    auto an = avg - next_avg;
+    auto nonmono = Ba ^ an;
+    auto absBa = Abs(Ba);
+    auto absan = Abs(an);
+    auto absBn = Abs(top - next_avg);
+    auto a3e = BitCast(d, ShiftRight<32>(MulEven(absBa, onethird)));
```
veluca93, line 49: A comment here would be _very_ helpful (or factoring out to a DivideBy3)

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -12,13 +12,155 @@
...
+    auto diff = diff_minus_tendency + tendency;
+    auto out = ShiftRight<1>(
+        ShiftLeft<1>(avg) + diff +
+        IfThenElse(diff < Zero(d), (diff & Set(d, 1)), Neg(diff & Set(d, 1))));
+    Store(out, d, p_out + x);
+    Store(out - diff, d, p_nout + x);
+  }
+}
+#if HWY_CAP_GE256
+JXL_INLINE void Transpose8x8Block(const pixel_type *JXL_RESTRICT from,
```
veluca93, line 79: Can we have those in transpose-inl.h, or maybe simd_utils.h?

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -43,36 +185,105 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
...
+            for (size_t y = 0; y < kRowsPerThread; y++) {
+              b_p_avg[8 * 8 + y] = p_avg[x + 8 + onerow_in * y];
+            }
+            for (size_t i = 0; i < 8; i++) {
+              fast_unsqueeze<kRowsPerThread>(
+                  b_p_residual + 8 * i, b_p_avg + 8 * i, b_p_avg + 8 * (i + 1),
+                  (x + i ? b_p_out_odd : b_p_avg + 8 * i), b_p_out_even,
+                  b_p_out_odd);
+              for (size_t y = 0; y < kRowsPerThread; y++) {
+                p_out[((x + i) << 1) + onerow_out * y] = b_p_out_even[y];
```
veluca93, line 224: Probably want to use StoreInterleaved from simd_util (and probably you want to do this directly in FastUnsqueeze, i.e. templatizing over a functor that takes the vector to be written, and transposes them and then writes them here or writes them directly in vertical squeeze)

Another option is to have FastUnsqueeze return a std::array<xxx, 8> (and perhaps making kRowsPerThread a global constant).

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -43,36 +185,105 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
...
+                p_avg[chin.w - 1 + onerow_in * y];
+        }
+      },
+      "InvHorizontalSqueeze"));
+  size_t firstrow = chin.h / kRowsPerThread * kRowsPerThread;
+#else
+  size_t firstrow = 0;
+#endif
+
+  JXL_RETURN_IF_ERROR(RunOnPool(
```
veluca93, line 257: I would either extract the one row scalar code to a function or get rid of this second case (it doesn't make that much of a difference, and you can fall back to scalar code if you don't have enough rows in the SIMD code).

Or both ;)

lib/jxl/modular/transform/squeeze.cc
jonsneyers: It looks like at least with my clang, autovec seems to already figure it out.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Added a comment.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Done.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Extracted it to a one row scalar unsqueeze function. It does have to be called in two spots because there's both the `height % 8` rows at the bottom and the `width % something` columns on the right that have to be done without simd.

Possibly at some point in the future if we have row padding in modular planes, the columns on the right will no longer require special handling.

lib/jxl/modular/transform/squeeze.cc
veluca93: I'd still prefer to see it explicitly :)

lib/jxl/modular/transform/squeeze.cc
veluca93: A solution to that would be having num_simd = 0 for last rows, (width/8*8) otherwise :)

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -43,36 +185,105 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
     return true;
   }
 
+#if HWY_TARGET != HWY_SCALAR
```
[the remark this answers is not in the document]
veluca93: I'm mostly concerned about overflows in $weird situations in Squeeze tendency :)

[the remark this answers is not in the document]
jonsneyers: Fair enough. I'll add a test using high bit depth + ycocg + squeeze. I expect the breaking point to be a bit earlier with the int32_t simd code than with the scalar SmoothTendency, but I assume it will be past the 24-bit limit that we have as an implementation limit anyway for uint image types. But you're right that it's probably a good idea to check that with a test.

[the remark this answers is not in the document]
jonsneyers: I added the test in a separate PR: https://github.com/libjxl/libjxl/pull/1267

Once that one lands, I'll rebase this one so we can check that the test still succeeds.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Did something like that. The second RunOnPool() is now gone.

lib/jxl/modular/transform/squeeze.cc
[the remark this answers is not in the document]
jonsneyers: Rebased it, so we'll soon see if the test breaks or not.

lib/jxl/modular/transform/squeeze.cc
jonsneyers: Done. Now the un-transpose and interleaved storing is made explicit.

veluca93 (dismissed):

veluca93 (approved):

PR #1160, opened 2022-02-07
Replace usage of __MACH__ with __APPLE__

malaterre:
It turns out that __MACH__ macro refer to the Mach kernel, which is also
used for GNU/Hurd. Prefer usage of macro __APPLE__ in this case.

This should solve the following compilation error:

lib/extras/time.cc:25:10: fatal error: mach/mach_time.h: No such file or directory

For reference:

* https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW13

### DIFF
--- a/lib/jxl/base/os_macros.h
+++ b/lib/jxl/base/os_macros.h
@@ -20,7 +20,7 @@
 #define JXL_OS_LINUX 0
 #endif
 
-#ifdef __MACH__
+#ifdef __APPLE__
 #define JXL_OS_MAC 1
 #else
 #define JXL_OS_MAC 0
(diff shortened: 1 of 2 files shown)


malaterre:
Just FYI, this is strictly identical to:

* https://github.com/google/highway/commit/711a7a5a3b3cfe28f3f369afab07a41c7021468a

cc: @jan-wassenberg

sboukortt (approved):

sboukortt:
Thanks!

PR #1163, opened 2022-02-07
Allow the pipeline to handle multiple frames of different sizes

sboukortt:


lib/jxl/render_pipeline/simple_render_pipeline.cc
```diff
@@ -188,6 +189,59 @@ void SimpleRenderPipeline::ProcessBuffers(size_t group_id, size_t thread_id) {
...
+      for (size_t c = 0; c < old_channels.size(); c++) {
+        channel_data_.emplace_back(2 * kRenderPipelineXOffset + image_xsize,
+                                   2 * kRenderPipelineXOffset + image_ysize);
+      }
+      for (size_t y = 0; y < image_ysize; ++y) {
+        for (size_t c = 0; c < channel_data_.size(); c++) {
+          output_rows[c].resize(1);
+          output_rows[c][0] = channel_data_[c].Row(kRenderPipelineXOffset + y);
+        }
+        // TODO(sboukortt): consider doing this only on the parts of the
```
veluca93, line 210: If this is just a question of optimization, then the slow pipeline doesn't need to care.
If this is required by the stage, then yes :D

lib/jxl/render_pipeline/render_pipeline_test.cc
```diff
@@ -421,5 +421,38 @@ JXL_GTEST_INSTANTIATE_TEST_SUITE_P(RenderPipelineTest, RenderPipelineTestParam,
...
+  DecompressParams dparams;
+  CodecInOut io_default;
+  ASSERT_TRUE(DecodeFile(dparams, compressed, &io_default, &pool));
+  CodecInOut io_slow_pipeline;
+  dparams.use_slow_render_pipeline = true;
+  ASSERT_TRUE(DecodeFile(dparams, compressed, &io_slow_pipeline, &pool));
+
+  ASSERT_EQ(io_default.frames.size(), io_slow_pipeline.frames.size());
+  for (size_t i = 0; i < io_default.frames.size(); i++) {
+#if JXL_HIGH_PRECISION
```
veluca93, line 440: If the files uses modular, then probably we ought to skip this #if

veluca93 (approved):

lib/jxl/render_pipeline/simple_render_pipeline.cc
veluca93: Then no need to keep the todo around :)

jonsneyers:

PR #1168, opened 2022-02-08
Jpeg serializer: deal with corner-case (negation of min-int)

eustas:


veluca93 (approved):
Isn't -min_int already

PR #1169, opened 2022-02-09
Gif decoder: Provide optional alpha channel as an optional extra_channel, rather than using rgba pixels.

fischbacher:


sboukortt (dismissed):
Looks good, thanks! Let’s maybe just squash the commits before merging.

sboukortt (dismissed):

sboukortt (approved):

PR #1174, opened 2022-02-14
Add support for extra channels to PPF -> codecio conversion.

veluca93:


lib/extras/packed_image_convert.cc
```diff
@@ -126,8 +146,15 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
...
-    JXL_CHECK(frame.extra_channels.empty());
+    for (size_t i = 0; i < frame.extra_channels.size(); i++) {
+      bundle.extra_channels().emplace_back(frame.extra_channels[i].xsize,
+                                           frame.extra_channels[i].ysize);
+      JXL_CHECK(BufferToImageF(
+          frame.extra_channels[i].format, frame.extra_channels[i].xsize,
+          frame.extra_channels[i].ysize, frame.extra_channels[i].pixels(),
+          frame.extra_channels[i].pixels_size, pool,
+          &bundle.extra_channels().back()));
+    }
```
sboukortt, line 157: Seems like this could be a range-based for loop, right?

sboukortt (dismissed):

PR #1179, opened 2022-02-16
cjxl_ng: Fix encoding (some) animated gifs.

fischbacher:
cjxl_ng should now support animated gifs with and without alpha channel.
There still can be issues with variable-size frames.

This also completes missing modular-mode support flags.


mo271 (approved):
Cool, let's fix the gifs with cropped frames next..

PR #1184, opened 2022-02-17
Implement the low-memory rendering pipeline.

veluca93:
Includes some fixes to the various stages.

The low-memory rendering pipeline is enabled by default. This does not
change the decoding output, and has minor effects on speed (that can
likely be compensated for by better optimizing the various stages and
removing remaining non-rp code).

Two TODOs are still left before being able to fully remove the non-rp
code: DC upsampling and encoder-side roundtrip still use code that
belongs to the non-rendering-pipeline path.

Speed measurements when decoding to uint8 (the slowdown is slightly more
significant when decoding to full float buffers, but that's not a common usecase):
```
x86 baseline: 54.481 MP/s
ARM baseline: geomean: 9.14383 MP/s [9.08262, 9.23499]

x86 after: 53.598 MP/s                                         
ARM after: geomean: 9.25677 MP/s [9.21585, 9.31553]
```

Also includes various changes all over the place to make the rendering pipeline work properly.


sboukortt (dismissed):

jonsneyers:

PR #1185, opened 2022-02-18
callback for decoding to ppf

mo271:


mo271:
Based on #1180, in order to be able to test it.

veluca93 (approved):

PR #1189, opened 2022-02-20
Remove set-but-never-used variables.

veluca93:


tools/conformance/djxl_conformance.cc
```diff
@@ -401,7 +400,6 @@ bool DecodeJXL(const DecodeOptions& opts) {
         std::vector<char> name(extra_channels[i].name_length + 1);
         EXPECT_SUCCESS(JxlDecoderGetExtraChannelName(dec.get(), i, name.data(),
                                                      name.size()));
```
sboukortt, line 402: Doesn’t it mean that this is also not used? (Should it be?)

veluca93: I do suspect it *should* be, yes... I'm not really sure though

sboukortt (approved):

PR #1194, opened 2022-02-23
add brotli_effort option

jonsneyers:
Currently, when doing JPEG recompression, the `jbrd` is brotli-compressed with effort 11 (hardcoded), and when `brob` boxes are created, they are brotli-compressed with effort 9 (also hardcoded). This adds an encoder option to set a custom brotli effort for the uses of brotli in the JPEG XL file format.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -283,6 +283,13 @@ typedef enum {
    */
   JXL_ENC_FRAME_INDEX_BOX = 31,
 
+  /** Sets brotli encode effort for use in JPEG recompression and compressed
+   * metadata boxes (brob). Can be -1 (default) or 0 (fastest) to 11 (slowest).
+   * Default is based on the general encode effort in case of JPEG
+   * recompression, and 4 for brob boxes.
+   */
+  JXL_ENC_FRAME_SETTING_BROTLI_EFFORT = 32,
+
   /** Enum value not to be used as an option. This value is added to force the
    * C compiler to have the enum to take a known size.
    */
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -211,7 +211,7 @@ PaddedBytes CreateTestJXLCodestream(
                             jpeg_bytes.data() + jpeg_bytes.size());
     EXPECT_TRUE(jxl::jpeg::DecodeImageJPG(
         jxl::Span<const uint8_t>(jpeg_bytes.data(), jpeg_bytes.size()), &io));
-    EXPECT_TRUE(EncodeJPEGData(*io.Main().jpeg_data, &jpeg_data));
+    EXPECT_TRUE(EncodeJPEGData(*io.Main().jpeg_data, &jpeg_data, cparams));
     io.metadata.m.xyb_encoded = false;
 #else   // JPEGXL_ENABLE_JPEG
     JXL_ABORT(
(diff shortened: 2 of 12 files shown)


lib/include/jxl/encode.h
```diff
@@ -283,6 +283,12 @@ typedef enum {
    */
   JXL_ENC_FRAME_INDEX_BOX = 31,
 
+  /** Sets brotli encode effort for use in JPEG recompression and compressed
+   * metadata boxes (brob). Ranges from 0 (fastest) to 11 (slowest).
```
eustas, line 287: LGTM, but it would be better to have looser coupling with brotli encoder.
Imagine, one day we have brotli 2.0, which have, e.g. quality 12.
For JXL users 11 meant max compression and it will no longer be true.
So, would be cool to have some coarse grained quality setting here (e.g. 1..5) and choose brotli quality inside codec.

eustas (dismissed):

lib/include/jxl/encode.h
jonsneyers: I see what you mean, but if and when that happens, can't we just remap 0-11 to whatever the new range would be?

I don't see why it would help if the effort setting is more coarse-grained. If anything, I would make it more fine-grained, say a setting from 0 to 100 that maps to `setting/9`, leaving room for future brotli changes.

I like the 0..11 scale for effort though, and I hope brotli will keep it, and libjxl will eventually expand its current 1..9 effort scale to 0..11 too :)

ziemek99:
The default Brotli effort is still hardcoded in cjxl. What about making it dependent on cjxl's effort (of course with possibility to override manually)?

jonsneyers:
> The default Brotli effort is still hardcoded in cjxl. What about making it dependent on cjxl's effort (of course with possibility to override manually)?

This PR does add an option `--brotli_effort` to cjxl.

ziemek99:
> 
> > The default Brotli effort is still hardcoded in cjxl. What about making it dependent on cjxl's effort (of course with possibility to override manually)?
> 
> This PR does add an option `--brotli_effort` to cjxl.
> 

I might be wrong, but from what I've seen `--brotli-effort` is an independent option which defaults to hardcoded 9. What I propose is to couple it with cjxl's `--effort` option (just like some encoding options are disabled/enabled by default based on `--effort`).

jonsneyers:
Ah, that makes sense, yes.

lib/include/jxl/encode.h
```diff
@@ -283,6 +283,12 @@ typedef enum {
...
+   * Default: 9.
```
veluca93, line 288: I think the default should be -1, and chosen based on the encoder effort :)

jonsneyers: Makes sense. At least for jbrd we can do that. For the brob boxes, we don't really have an encoder effort, since that is part of the frame settings, so there is no encoder effort defined for these.

For brob boxes perhaps a sensible default is something lower than 9. I wanted to add this option because 9 was too slow for psd reconstruction boxes, which could be quite large in some cases so the brobbing was then actually a big bottleneck in the recompression process...

jonsneyers:
Updated to have a default of -1, which means it will use a default value of 2+effort for `jbrd` and 4 for `brob`.

veluca93 (dismissed):

veluca93 (approved):

ziemek99:
> Updated to have a default of -1, which means it will use a default value of 2+effort for `jbrd` and 4 for `brob`.

I smell some trouble with `jbrd` when maximum `--effort` will be increased above 9...

jonsneyers:
> > Updated to have a default of -1, which means it will use a default value of 2+effort for `jbrd` and 4 for `brob`.
> 
> I smell some trouble with `jbrd` when maximum `--effort` will be increased above 9...

If and when we change the range of effort, we'll anyway have to go through all places where it is used, to make sure things still make sense. Good point though.

PR #1199, opened 2022-02-28
Speedup EPF stages

veluca93:


sboukortt (approved):

jonsneyers:
This caused a conformance fail, is that expected or not?

veluca93:
> This caused a conformance fail, is that expected or not?

Yup, the whole chain of PRs requires an update of conformance tests (it contains 2 changes to EPF, one to bring it in line with the current spec and the other to fix discontinuities)

jonsneyers:

PR #1202, opened 2022-02-28
Use the rendering pipeline in the encoder too.

veluca93:
Also removes two tests that have long been just a source of annoyances
rather than a way of finding bugs.


lib/jxl/enc_adaptive_quantization.cc
```diff
@@ -740,6 +740,27 @@ void FindBestQuantization(const ImageBundle& linear, const Image3F& opsin,
...
+    }
+    TemporaryShrink(const TemporaryShrink&) = delete;
+    TemporaryShrink(TemporaryShrink&&) = delete;
+
+    ~TemporaryShrink() { bundle.ShrinkTo(orig_xsize, orig_ysize); }
+
+    ImageBundle& bundle;
+    size_t orig_xsize;
+    size_t orig_ysize;
+  } t(const_cast<ImageBundle&>(linear),
```
sboukortt, line 760: 😂

tools/fuzzer_corpus.cc
```diff
@@ -357,6 +357,7 @@ int main(int argc, const char** argv) {
   }
 
   struct stat st;
+  memset(&st, 0, sizeof(st));
```
sboukortt, line 360: Out of curiosity, why is this necessary? I don’t see any suggestion in man 2 stat that stat expects the target of the pointer to have already been initialized.

sboukortt (approved):

jonsneyers:

PR #1208, opened 2022-03-03
Hazard elimination: C++ enum class defs must parallel C API.

fischbacher:


veluca93:
Couldn't we remove the C++ enum entirely?

fischbacher:
Removing the C++ enums: Your call - it's more your code than mine, after all.

There might be some value in having proper a C++ enum class.

veluca93:
> Removing the C++ enums: Your call - it's more your code than mine, after all.
> 
> There might be some value in having proper a C++ enum class.

@sboukortt do you know if the field visitors should still work with C enums? If so, I'd think it better to only stick with the C enums.

sboukortt:
> > Removing the C++ enums: Your call - it's more your code than mine, after all.
> > There might be some value in having proper a C++ enum class.
> 
> @sboukortt do you know if the field visitors should still work with C enums? If so, I'd think it better to only stick with the C enums.

From what I can tell, they should, but the C++ side will still contain some duplication through the `EnumBits` functions that it needs to overload.

veluca93 (approved):

PR #1221, opened 2022-03-09
cjxl: add --strip_jpeg_metadata option

Hello71:
it suppresses generating "jbrd" chunk in output. it is no-op if not in
jpeg transcode mode.

dunno if this is really that useful. i tested it and it seems to work, and seems like minimal added maintenance burden. i don't mind if you don't merge it though because cjxl_ng.


jonsneyers:
Makes sense to have an option like this, in case you don't want to strip Exif/XMP metadata (the --strip option will strip everything) but you don't care about jpeg bitstream reconstruction.

Could you just run `clang-format --style=file -i cjxl.cc` so the code formatting check doesn't complain?

PR #1223, opened 2022-03-10
fixing animations with cropped frames

mo271:


jonsneyers:
Does the old cjxl still work after this?

mo271:
> Does the old cjxl still work after this?

For gif it seems to work. For apng, I'm having a file where I see new artifacts after the change. Will look into fixing this...

mo271:
> Does the old cjxl still work after this?

Now cjxl gives byte-identical files for both gifs and apngs before and after this change.

veluca93 (approved):

sboukortt:
I seem to be getting 0-byte JXL files from single-frame PNG files using `cjxl_ng` starting with this change, do you also observe this behavior?

mo271:
> I seem to be getting 0-byte JXL files from single-frame PNG files using `cjxl_ng` starting with this change, do you also observe this behavior?

This was not intended. Will look into it...

PR #1227, opened 2022-03-11
use actual num_channels instead of guessing from color encoding

### LINKED ISSUE #1198: encode api: grayscale input sent as RGB(A) buffer is broken
BufferToImageBundle / ConvertFromExternal in enc_external_image.cc does not look at pixel_format to determine the number of channels given in the input buffer, it just assumes that if the color encoding is grayscale, there is 1 color channel, and otherwise there are 3.

This causes bigtime trouble if you try encoding an image that happens to be grayscale by giving the encoder an RGB(A) buffer (where R=G=B). It's quite common for applications to do everything with RGBA buffers regardless of whether the image actually has alpha or is grayscale or not, to reduce the number of cases / code paths. If you pass an RGBA buffer (marked as such in the pixel_format), the current code will treat it as if it's a Gray+Alpha buffer if the color encoding is grayscale, i.e. it will interpret R and B as gray samples and G and A as alpha samples, which is of course not what it should do.
Possible solutions to correctly deal with this case:

- check that R=G=B if the color encoding is grayscale (otherwise return error since "you lied") but an RGB(A) buffer gets passed
- assume that R=G=B in this case, and e.g. just ignore R and B and put G in the gray channel
- don't assume anything, and convert to grayscale in some way (e.g. 0.299*R + 0.587*G + 0.114*B) 

It probably also causes trouble if you try to pass a grayscale buffer when the image is RGB (though that's less likely to actually happen in practice).

jonsneyers:
Fixes #1198 

When an RGB(A) buffer is passed while the image is grayscale, use R and A instead of misinterpreting the buffer as Y(A) and ignoring the bottom part. G and B just get ignored in this case and G and B get copied from R (we could instead check that R==G==B in the input, but that comes at a small overhead just to catch potential wrong API usage; I think it will be sufficiently easy to debug for application developers if they see their color image gets incorrectly grayscaled when they accidentally incorrectly set the color encoding to grayscale).

When a Y(A) buffer is passed while the image is color, return a relevant error instead of just saying that the buffer is too small.

### DIFF
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -197,7 +197,7 @@ PaddedBytes CreateTestJXLCodestream(
   // image match.
   io.metadata.m.color_encoding = color_encoding;
   EXPECT_TRUE(ConvertFromExternal(
-      pixels, xsize, ysize, color_encoding, /*has_alpha=*/include_alpha,
+      pixels, xsize, ysize, color_encoding, num_channels,
       /*alpha_is_premultiplied=*/false, /*bits_per_sample=*/16, JXL_BIG_ENDIAN,
       /*flipped_y=*/false, &pool, &io.Main(), /*float_in=*/false, /*align=*/0));
   jxl::PaddedBytes jpeg_data;
@@ -1223,7 +1223,8 @@ TEST_P(DecodeTestParam, PixelTest) {
     io.SetSize(config.xsize, config.ysize);
 
     EXPECT_TRUE(ConvertFromExternal(
-        bytes, config.xsize, config.ysize, color_encoding, config.include_alpha,
+        bytes, config.xsize, config.ysize, color_encoding,
+        config.output_channels,
         /*alpha_is_premultiplied=*/false, 16, JXL_BIG_ENDIAN,
         /*flipped_y=*/false, nullptr, &io.Main(), /*float_in=*/false,
         /*align=*/0));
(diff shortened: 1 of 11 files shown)


lib/jxl/decode_test.cc
```diff
@@ -1542,20 +1543,19 @@ TEST(DecodeTest, PixelTestWithICCProfileLossy) {
   jxl::Span<const uint8_t> span0(pixels.data(), pixels.size());
   jxl::CodecInOut io0;
   io0.SetSize(xsize, ysize);
-  EXPECT_TRUE(ConvertFromExternal(
-      span0, xsize, ysize, color_encoding0,
-      /*has_alpha=*/false, false, 16, format_orig.endianness,
-      /*flipped_y=*/false, /*pool=*/nullptr, &io0.Main(), /*float_in=*/false,
-      /*align=*/0));
+  EXPECT_TRUE(ConvertFromExternal(span0, xsize, ysize, color_encoding0, 3,
```
veluca93, line 1546: /*num_channels=*/3

lib/jxl/decode_test.cc
```diff
@@ -1542,20 +1543,19 @@ TEST(DecodeTest, PixelTestWithICCProfileLossy) {
...
+                                  false, 16, format_orig.endianness,
```
veluca93, line 1547: (this was already the case, but also for false and 16)

lib/jxl/decode_test.cc
```diff
@@ -1600,32 +1600,20 @@ TEST(DecodeTest, PixelTestOpaqueSrgbLossy) {
...
+                                    false, 16, format_orig.endianness,
+                                    /*flipped_y=*/false, /*pool=*/nullptr,
+                                    &io0.Main(), /*float_in=*/false,
+                                    /*align=*/0));
 
     jxl::ColorEncoding color_encoding1 = jxl::ColorEncoding::SRGB(false);
     jxl::Span<const uint8_t> span1(pixels2.data(), pixels2.size());
     jxl::CodecInOut io1;
-    if (channels == 4) {
-      io1.metadata.m.SetAlphaBits(8);
```
veluca93, line 1613: Why is this code gone?

jonsneyers: I think it dates back to a time when ConvertFromExternal would not work when the buffer is RGBA while the image does not have alpha, so this probably was a workaround to fix that. But ConvertFromExternal just ignores the A if the image does not have alpha, so this workaround is not needed anymore.

lib/jxl/decode_test.cc
jonsneyers: Updated

veluca93 (approved):

PR #1228, opened 2022-03-11
fix default resampling in encode api

### LINKED ISSUE #1210: Setting JXL_ENC_FRAME_SETTING_RESAMPLING with its documented default value breaks encoding
**Describe the bug**
https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/lib/include/jxl/encode.h#L100-L106

This matches its setter:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/lib/jxl/encode.cc#L768

But according to the encoder itself, the default value is not -1, it's **0** according to enc_file.cc:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/lib/jxl/enc_file.cc#L204-L208

or **1** if we check enc_params.h:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/lib/jxl/enc_params.h#L248-L253

This means that if I try to set it, e.g. when extracting the value from an UI dialog, this happens:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/lib/jxl/enc_frame.cc#L322-L324

**To Reproduce**

Just append this line here:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/examples/encode_oneshot.cc#L186-L190

```cpp
JxlEncoderFrameSettingsSetOption(frameSettings,
                                         JXL_ENC_FRAME_SETTING_RESAMPLING,
                                         cfg->getInt("resampling", -1));
```

**Expected behavior**

Encoding succeeds.

**Screenshots**

imagen

**Environment**
 - OS: [e.g. Windows] Windows 10 21H2
 - Compiler version: [e.g. clang 11.0.1] MSVC 14.29.30133
 - CPU type: [e.g. x86_64] x86_64
 - cjxl/djxl version string: [e.g. cjxl [v0.3.7 | SIMD supported: SSE4,Scalar]] JPEG XL encoder v0.7.0 0.6.1 [AVX2,SSE4,SSSE3,Scalar]

**Additional context**

The only place where `JXL_ENC_FRAME_SETTING_RESAMPLING` is declared to be `-1` by default is here:

https://github.com/libjxl/libjxl/blob/ec7adc703541daad0b6b1ec8bde5cbe3a3bfcf7c/tools/cjxl_ng_main.cc#L526-L536

jonsneyers:
Fixes #1210 so it does the correct thing when setting resampling to -1 in the API.

### DIFF
--- a/lib/jxl/enc_params.h
+++ b/lib/jxl/enc_params.h
@@ -246,14 +246,30 @@ struct CompressParams {
   bool use_new_heuristics = false;
 
   // Down/upsample the image before encoding / after decoding by this factor.
-  // The resampling value can also be set to 0 to automatically choose based
+  // The resampling value can also be set to <= 0 to automatically choose based
   // on distance, however EncodeFrame doesn't support this, so it is
-  // required to process the CompressParams to set a valid positive resampling
+  // required to call PostInit() to set a valid positive resampling
   // value and altered butteraugli score if this is used.
-  size_t resampling = 1;
-  size_t ec_resampling = 1;
+  int resampling = -1;
+  int ec_resampling = -1;
   // Skip the downsampling before encoding if this is true.
   bool already_downsampled = false;
+
+  void PostInit() {
+    if (resampling <= 0) {
+      resampling = 1;
+      // For very low bit rates, using 2x2 resampling gives better results on
+      // most photographic images, with an adjusted butteraugli score chosen to
+      // give roughly the same amount of bits per pixel.
+      if (!already_downsampled && butteraugli_distance >= 20) {
+        resampling = 2;
+        butteraugli_distance = 6 + ((butteraugli_distance - 20) * 0.25);
+      }
+    }
+    if (ec_resampling <= 0) {
+      ec_resampling = resampling;
+    }
+  }
 };
 
 static constexpr float kMinButteraugliForDynamicAR = 0.5f;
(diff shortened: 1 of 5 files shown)


lib/jxl/enc_params.h
```diff
@@ -254,6 +254,19 @@ struct CompressParams {
   size_t ec_resampling = 1;
   // Skip the downsampling before encoding if this is true.
   bool already_downsampled = false;
+
+  void PostInit() {
+    if (resampling == 0) {
```
veluca93, line 259: why not "== 0 || == -1" and the C API just sets the value?

Can we add a test?

veluca93: We can make it signed too :)

As for tests: mostly that setting it to -1 works and doesn't crash/...

jonsneyers: Made it signed, and also made it -1 by default, because we claim in the documentation that -1 is the default but actually 1 was the default.

I don't think a separate test is needed now, because all the existing tests that don't explicitly change it will now have it set to -1.

veluca93 (approved):

Issue #1116, opened 2022-01-24
UInt32 JxlDecoderImageOutBufferSize != JXL_DEC_SUCCESS


KoromaruKoruko:
**Describe the bug**
While decoding an image using UInt32 Buffers, i get a non- JXL_DEC_SUCCESS result from JxlDecoderImageOutBufferSize.
decoding with UInt8, UInt16 and Float32 works as expected, only UInt32 doesn't

**To Reproduce**
compile wasm32 with a emscripten executable wrapper then try and decode an image using JXL_TYPE_UINT32 pixel type.

**Expected behavior**
it returns a successful decoded pixel buffer.

**Screenshots**
![image](https://user-images.githubusercontent.com/17232928/150842049-a0373515-e136-4c87-a4ad-2d1acb1a13f4.png)

**Environment**
 - OS: -
 - Compiler version: clang-13 (emscripten)
 - CPU type: wasm_x32 [no SIMD]
 - browser: known firefox & chrome

**Additional context**
CPPCode => https://github.com/KoromaruKoruko/libjxl/blob/main/lib/wasm_extras/dec_extras.cc
 JSCODE (Ran in console, after loading wasm wrapper)
![image](https://user-images.githubusercontent.com/17232928/150842438-50a43c9d-6729-4a5b-9067-1d427f81951d.png)

PR #1231, opened 2022-03-11
allow decoding to u32

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/1116

For some reason, the decode api didn't allow decoding to UINT32, while it actually works fine as far as I can tell (besides the fact that only the 24 msb will be useful, given that everything goes through float32 internally). Only the test_utils.h code for testing UINT32 decoding had some bugs, fixed those.


jonsneyers:
An alternative solution is to deprecate UINT32 pixelformats, since the use case for it is not very clear, also considering that jxl does not actually support lossless uint32 encoding.

jonsneyers:
Opinions please: the options are "implement uint32 in the api" or "deprecate uint32 in the api".

I don't have a strong preference myself: keeping it could be nice for applications that don't want to deal with float, or for a future where we no longer necessarily convert modular to float before returning it, so we could support up to 31-bit integer (which fits in uint32 but not in float32). Deprecating it is currently no big loss since float32 is available as an alternative and is what is used internally anyway, and it is confusing to have a uint32 pixel_format when the bitstream itself doesn't actually support uint32 samples.

jonsneyers:
(pinging @jyrkialakuijala @veluca93  @sboukortt @eustas @mo271 @fischbacher for opinions on this)

malaterre:
Just for reference, here is what openjpeg states for 32bits input:

```
% opj_decompress -i input32.j2k -o output.pgx

[INFO] Start to read j2k main header (0).
[ERROR] Invalid values for comp = 0 : prec=32 (should be between 1 and 38 according to the JPEG2000 norm. OpenJpeg only supports up to 31)
[ERROR] Marker handler function failed to read the marker segment
ERROR -> opj_decompress: failed to read the header
```

jonsneyers:
Closing in favor of https://github.com/libjxl/libjxl/pull/1337

PR #1233, opened 2022-03-12
add Noise and XYBFactors options to jxl_from_tree

jonsneyers:
Adds three new options to `jxl_from_tree` to manually specify:

- Noise lookup table: `Noise [8 floats in [0.0,1.0) range]`
- custom XYB quantization factors: `XYBFactors [3 floats]`
- Delta Palette: `DeltaPalette` (signals a default zero-color delta palette transform)

This allows setting XYB factors to their defaults `4096 512 256`, which makes the LfChannelDequantization bundle 1 bit instead of 49 bits, and using different factors than the default ones for modular to get different 'bit depths' in XYB too. In particular, this enables producing the currently known smallest jxl file as follows:
```
XYB
XYBFactors 4096 512 256
Gaborish
EPF 2
Width 512
Height 256
- Set 0
```

For noise this helps to produce test bitstreams to debug noise seeding.

Delta Palette is just to explore jxl art :)

### DIFF
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -213,7 +213,8 @@ uint64_t FrameFlagsFromParams(const CompressParams& cparams) {
   // noise is stored within the compressed image and adding noise makes things
   // worse.
   if (ApplyOverride(cparams.noise, dist >= kMinButteraugliForNoise) ||
-      cparams.photon_noise_iso > 0) {
+      cparams.photon_noise_iso > 0 ||
+      cparams.manual_noise.size() == NoiseParams::kNumNoisePoints) {
     flags |= FrameHeader::kNoise;
   }
 
@@ -303,7 +304,7 @@ Status MakeFrameHeader(const CompressParams& cparams,
   frame_header->flags = FrameFlagsFromParams(cparams);
   // Non-photon noise is not supported in the Modular encoder for now.
   if (frame_header->encoding != FrameEncoding::kVarDCT &&
-      cparams.photon_noise_iso == 0) {
+      cparams.photon_noise_iso == 0 && cparams.manual_noise.size() == 0) {
     frame_header->UpdateFlag(false, FrameHeader::Flags::kNoise);
   }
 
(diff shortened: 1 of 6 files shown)


veluca93 (approved):

lib/jxl/enc_frame.cc
```diff
@@ -1291,6 +1291,12 @@ Status EncodeFrame(const CompressParams& cparams_orig,
     lossy_frame_encoder.State()->shared.image_features.noise_params =
         SimulatePhotonNoise(ib.xsize(), ib.ysize(), cparams.photon_noise_iso);
   }
+  if (cparams.manual_noise.size() > 0) {
```
veluca93, line 1294: Couldn't this segfault with wrong lut size?

veluca93 (changes requested):

lib/jxl/enc_frame.cc
jonsneyers: Yes. Not anymore.

lib/jxl/enc_frame.cc
```diff
@@ -213,7 +213,7 @@ uint64_t FrameFlagsFromParams(const CompressParams& cparams) {
   // noise is stored within the compressed image and adding noise makes things
   // worse.
   if (ApplyOverride(cparams.noise, dist >= kMinButteraugliForNoise) ||
-      cparams.photon_noise_iso > 0) {
+      cparams.photon_noise_iso > 0 || cparams.manual_noise.size() > 0) {
```
veluca93, line 216: Also here (and maybe, for both options, a warning/error if their size is incorrect)

veluca93 (changes requested):

lib/jxl/enc_frame.cc
jonsneyers: Modified the cparams PostInit to check this and return error if it's incorrect.

jonsneyers:
Also added a `DeltaPalette` keyword, PTAL.
The stuff to only do MetaApply instead of doing the fwd transform is an optimization that is not really needed but it's nice since the fwd delta palette transform is pretty slow, and of course it's completely unnecessary to actually do the fwd transform (actually all jxl_from_tree transforms could be optimized this way).

veluca93 (approved):

PR #1236, opened 2022-03-15
Resolve most clang-tidy reports

eustas:


veluca93 (approved):

sboukortt:
Was the rename of `ssimulacra.cpp` to `ssimulacra.txt` intentional?

tools/ssimulacra.cc
```diff
@@ -3,7 +3,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Re-implementation of //third_party/ssimulacra/ssimulacra.cpp using jxl's
+// Re-implementation of //tools/ssimulacra.tct using jxl's
```
jonsneyers, line 6: It was renamed to .txt, but actually why don't we just remove that file? It doesn't really serve any purpose. If someone needs the old opencv-based version, they can find it at https://github.com/cloudinary/ssimulacra, no need to store that old code on this repo too.

PR #1239, opened 2022-03-16
Deal with unteminated c-strings in JXL messages

eustas:
Context: b/224883274

### DIFF
--- a/lib/extras/dec/pnm.cc
+++ b/lib/extras/dec/pnm.cc
@@ -215,7 +215,12 @@ class Parser {
           return JXL_FAILURE("PAM: unknown TUPLTYPE");
         }
       } else {
-        return JXL_FAILURE("PAM: unknown header keyword: %s", pos_);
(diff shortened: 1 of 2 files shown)


lib/extras/dec/pnm.cc
```diff
@@ -215,7 +215,12 @@ class Parser {
           return JXL_FAILURE("PAM: unknown TUPLTYPE");
         }
       } else {
-        return JXL_FAILURE("PAM: unknown header keyword: %s", pos_);
+        constexpr size_t kMaxHeaderLength = 20;
+        char unknown_header[kMaxHeaderLength + 1];
+        size_t len = std::min<size_t>(kMaxHeaderLength, end_ - pos_);
+        strncpy(unknown_header, reinterpret_cast<const char*>(pos_), len);
```
veluca93, line 221: Why is this reinterpret_cast necessary?

eustas: Because `pos_` is `uint8_t*`

Issue #1088, opened 2022-01-12
SOF9 Arithmetic Extended Sequential DCT => Corrupt or CMYK JPEG


malaterre:
**Describe the bug**

`cjxl` reports a confusing error message for JPEG using arithmetic coding (instead of default huffman).

```
% cjxl black.jpg black.jxl
JPEG XL encoder v0.6.1 086a674 [AVX2,SSE4,Scalar]
Corrupt or CMYK JPEG.
Failed to read image black.jpg.
```
**To Reproduce**

```
$ convert -size 512x512 -depth 8 xc:black black.pgm
$ cjpeg -arithmetic -quality 100 -grayscale -outfile black.jpg black.pgm
$ cjxl black.jpg black.jxl
```

**Expected behavior**

I've reported this issue under "bug" section. This is not clear why arithmetic coding is not supported in cjxl code path (no remaining legal restrictions on arithmetic coding AFAIK).

So either:

1. Improve the error message in this case, or
2. Implement proper support for SOF9 marker

**Environment**
 - OS: Linux Debian/sid
 - cjxl/djxl version string: Default Debian package (from experimental section) on amd64

malaterre:
Just for reference, I see that arithmetic coding perform a bit better than JPEG-XL in this special case (both lossless):

```
% ls -al black.j*
-rw-r--r-- 1 mathieu mathieu 128 Jan 12 11:03 black.jpg
-rw-r--r-- 1 mathieu mathieu 231 Jan 12 11:16 black.jxl
```

jonsneyers:
Arithmetic coding is not supported by `jbrd` (jpeg bitstream reconstruction data). We could possibly add support it as an input format and recompress the coefficients losslessly, but we cannot do bitstream-exact reconstruction.

malaterre:
@jonsneyers I did not realize by-design JPEG-XL could only handle huffman JPEG and not arithmetic. I've updated the bug report so that suggestion (1) is implemented. It would be confusing for users if `cjxl input.jpg output.jxl` would be lossy for JPEG with arthmetic coding.

jonsneyers:
The current error it gives on that image is this: (in a debug build)
```
JPEG XL encoder v0.7.0 09e08f89 [AVX2,SSE4,SSSE3,Scalar]
./lib/jxl/jpeg/enc_jpeg_data_reader.cc:171: Invalid comps_in_scan: 1
[...]

PR #1252, opened 2022-03-21
Clarify requirement for 8bits precision.

malaterre:
Thanks to Jon Sneyers for clarification:
https://github.com/libjxl/libjxl/issues/1088#issuecomment-1073577246

### DIFF
--- a/lib/jxl/jpeg/enc_jpeg_data_reader.cc
+++ b/lib/jxl/jpeg/enc_jpeg_data_reader.cc
@@ -95,6 +95,7 @@ bool ProcessSOF(const uint8_t* data, const size_t len, JpegReadMode mode,
   int height = ReadUint16(data, pos);
   int width = ReadUint16(data, pos);
   int num_components = ReadUint8(data, pos);
+  // 'jbrd' is hardcoded for 8bits:
   JXL_JPEG_VERIFY_INPUT(precision, 8, 8, PRECISION);
   JXL_JPEG_VERIFY_INPUT(height, 1, kMaxDimPixels, HEIGHT);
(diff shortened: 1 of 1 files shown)


jonsneyers:
We _could_ in principle change this. Representing 12-bit DCT coeffs can be done losslessly in jxl. Doing it in a way that also allows jpeg bitstream reconstruction is a bit trickier since it requires changing that code to not be hardcoded for only 8-bit, but it's not impossible.

Question is if it is worth the trouble though, since such jpeg files are "exotic", i.e. they don't exist "in the wild" and cannot be decoded by most existing jpeg decoders. So it is somewhat questionable if there really is a use case for this (as opposed to just doing a full-fledged lossless or vardct jxl encoding from the 12-bit source image).

malaterre:
@jonsneyers There is currently a Supplement under review by DICOM commitee to add support for JPEG-XL. The wording at this time is rather elusive, so we can decide which way to go:

> If a JPEG encoded image is recoded as JPEG XL it shall use the JPEG compatible mode in JPEG XL.

But technically within the DICOM world, we tend to have access to (so-called) `JPEG Extended (Process 2 & 4): Default Transfer Syntax for Lossy JPEG 12 Bit Image Compression (Process 4 only)` (aka 1.2.840.10008.1.2.4.51) (*). I do not know how widespread they are, but decision to support 12bits, will definitely impact the wording of the Supplement.

cc: @wayfarer3130

(*) https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_8.7.3.html#table_8.7.3-2

wayfarer3130:
Thank you - will update that to mention that it is JPEG Baseline only that
shall use the JPEG compatible mode in JPEG XL.

jonsneyers (approved):

PR #1254, opened 2022-03-21
Deviate from upstream ppm specification and assume color space is sRGB

malaterre:
[...]
In the raster, the sample values are "nonlinear." They are proportional
to the intensity of the ITU-R Recommendation BT.709 red, green, and blue
in the pixel, adjusted by the BT.709 gamma transfer function. (That
transfer function specifies a gamma number of 2.2 and has a linear
section for small intensities). A value of Maxval for all three samples
represents CIE D65 white and the most intense color in the color
universe of which the image is part (the color universe is all the
colors in all images to which this image might be compared).
[...]

* http://netpbm.sourceforge.net/doc/ppm.html

### DIFF
--- a/lib/extras/dec/pnm.cc
+++ b/lib/extras/dec/pnm.cc
@@ -312,6 +312,9 @@ Status DecodeImagePNM(const Span<const uint8_t> bytes,
     return JXL_FAILURE("PNM: bits_per_sample invalid");
   }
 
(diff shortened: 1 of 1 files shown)


jonsneyers:
Heh, I didn't know that was in the ppm spec. Back then, probably sRGB wasn't defined yet (?). Today I think sRGB is a more reasonable assumption than 709. It's good to clarify this though, thanks!

jonsneyers (approved):

PR #1262, opened 2022-03-22
Fix access to non-existing EC plane

eustas:
Currently we do not populate image bundle when callback used.
This means that there is no EC planes to copy data from.

Does not seem to be a (popular) use-case.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -1065,6 +1065,7 @@ static JxlDecoderStatus ConvertImageInternal(
 
   jxl::Status status(true);
   if (want_extra_channel) {
+    JXL_ASSERT(extra_channel_index < frame.extra_channels().size());
     status = jxl::ConvertToExternal(
         frame.extra_channels()[extra_channel_index],
         BitsPerChannel(format.data_type), float_format, format.endianness,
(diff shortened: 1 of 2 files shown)


lib/jxl/decode.cc
```diff
@@ -1563,14 +1564,20 @@ JxlDecoderStatus JxlDecoderProcessCodestream(JxlDecoder* dec, const uint8_t* in,
...
           dec->image_out_buffer_set = false;
 
+          bool has_ec = !dec->ib->extra_channels().empty();
           for (size_t i = 0; i < dec->extra_channel_output.size(); ++i) {
             void* buffer = dec->extra_channel_output[i].buffer;
             // buffer nullptr indicates this extra channel is not requested
             if (!buffer) continue;
+            if (!has_ec) {
+              JXL_WARNING(
+                  "Extra channels are not supported when callback is used");
```
jonsneyers, line 1574: Doesn't chrome use the callback? Does that mean that alpha no longer works? Or is this only about extra channels beyond alpha?

eustas: rgba works via callback. it only will fail if (float) callback is set *and* extra channels are requested.

PR #1267, opened 2022-03-23
fix accuracy of float2int / int2float and add test

jonsneyers:
This makes the conversion from float to int or back use double precision multiplication when the bitdepth is 23 or more. Otherwise things don't roundtrip losslessly. When the bitdepth is at most 22, the SIMD conversion can still be used, so this is only slowing things down when the bitdepth is > 22.

For integer bitdepths of 25 and more, we can't roundtrip losslessly anyway because single-precision floats only have 24 mantissa bits, but we can roundtrip integer values representable in floats losslessly, up to 30-bit (or 29-bit when Squeeze is used).

Adds a test to see if an image (with artificial noise added in the lsb) roundtrips losslessly at bitdepths 1 to 30 without squeeze and 1 to 29 with squeeze.

### DIFF
--- a/lib/jxl/modular_test.cc
+++ b/lib/jxl/modular_test.cc
@@ -25,6 +25,7 @@
 #include "lib/jxl/dec_file.h"
 #include "lib/jxl/dec_params.h"
 #include "lib/jxl/enc_butteraugli_comparator.h"
+#include "lib/jxl/enc_butteraugli_pnorm.h"
 #include "lib/jxl/enc_cache.h"
 #include "lib/jxl/enc_color_management.h"
 #include "lib/jxl/enc_file.h"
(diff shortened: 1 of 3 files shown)


lib/jxl/modular_test.cc
```diff
@@ -281,6 +282,75 @@ TEST(ModularTest, RoundtripLosslessCustomSqueeze) {
                                      /*distmap=*/nullptr, pool));
 }
 
+TEST(ModularTest, RoundtripLosslessSqueezeHighBitdepth) {
```
veluca93, line 285: I understand that we already do this in other tests, but I'd prefer if we made the test parametric instead of having an internal loop that makes debugging harder if a test does fail.

lib/jxl/modular_test.cc
```diff
@@ -281,6 +282,75 @@ TEST(ModularTest, RoundtripLosslessCustomSqueeze) {
...
+  const PaddedBytes orig =
+      ReadTestData("wesaturate/500px/u76c0g_bliznaca_srgb8.png");
+  CodecInOut io1;
+  ASSERT_TRUE(SetFromBytes(Span<const uint8_t>(orig), &io1, pool));
+
+  for (int responsive = 0; responsive <= 1; responsive++) {
+    for (int bitdepth = 1; bitdepth < 31; bitdepth++) {
+      if (responsive && bitdepth > 29) continue;
+
+      size_t xsize = 423 + bitdepth;
```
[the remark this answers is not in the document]
jonsneyers: Just to have some variety of dimensions, in case there's a bug that only happens when dimensions are even or odd or something like that.

lib/jxl/modular_test.cc
jonsneyers: Good point, I'll just have to check how to do that again.

jonsneyers:
Updated: now also roundtrips for 31-bit without squeeze and 30-bit with squeeze.
This failed before because 1) the input values didn't get clamped to 0..1 range, causing the effective bitdepth to be one more, and 2) enc_modular was still using an intermediate float value to store the multiplier, causing trouble since `(int)((float)((1<<31)-1)) == -2147483648`.

veluca93 (approved):

PR #1269, opened 2022-03-24
WIP candidate replacement for cache_aligned.cc

veluca93:
- [ ] depends on: #1100

### DIFF
--- a/lib/jxl/base/cache_aligned.cc
+++ b/lib/jxl/base/cache_aligned.cc
@@ -22,23 +22,9 @@
 
 #include "lib/jxl/base/printf_macros.h"
 #include "lib/jxl/base/status.h"
+#include "lib/jxl/sanitizers.h"
 
 namespace jxl {
-namespace {
-
-#pragma pack(push, 1)
-struct AllocationHeader {
-  void* allocated;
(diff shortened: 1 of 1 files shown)


mo271:
blocked by #1100

eustas:
Hi. Why it depends on c++17?
https://en.cppreference.com/w/c/memory/aligned_alloc says `since C11`.

veluca93:
> Hi. Why it depends on c++17? https://en.cppreference.com/w/c/memory/aligned_alloc says `since C11`.

That's C11, not C++11, right?

eustas:
True: https://en.cppreference.com/w/cpp/memory/c/aligned_alloc

What about `std::aligned_storage`? Does `new` initialize it? (and is it possible to `delete` it after we "forget" its true type).

veluca93:
I'm honestly not sure... I don't really remember what I was going for with this PR either.

eustas:
Then I'll just add a TODO to `memory_manager` that we could switch to aligned_alloc after c++17 is a mainline.

eustas:
(nope, doesn't fit external memory manager API)

PR #1271, opened 2022-03-24
remove psd input

jonsneyers:
Removes PSD as an input codec, for a number of reasons:

- It's not in the scope of cjxl/djxl to support all kinds of formats, and this was certainly one of the more exotic ones;
- PSD output wasn't implemented;
- PSD input was only implemented partially (not all kinds of PSD files would work), and that's to some extent always going to be a problem, because
- PSD is a proprietary format that doesn't have a full spec, it is basically "whatever Photoshop happens to produce".
- I'm planning to release as a separate project a tool that uses libjxl to do lossless recompression of PSD files. This will be a better way to convert PSD to jxl, but it's outside the scope of the libjxl project itself.
- It would be tricky anyway to keep supporting PSD input in cjxl_ng, for the same reasons as why it couldn't be adapted to use the PackedPixelFile struct like all other codecs.


sboukortt (approved):

jonsneyers:
Flyby fix: two tests in encode_test were broken after the combination of https://github.com/libjxl/libjxl/pull/1263 and https://github.com/libjxl/libjxl/pull/1247, fixed them.

veluca93 (approved):

ziemek99:
How is it possible to make multi-layered JXL images since PSD is now removed?

PR #1277, opened 2022-03-25
Take upsampling into account for border

eustas:


veluca93 (approved):

lib/jxl/render_pipeline/low_memory_render_pipeline.cc
```diff
@@ -373,9 +373,10 @@ void LowMemoryRenderPipeline::PrepareForThreadsInternal(size_t num,
   }
   // TODO(veluca): avoid reallocating buffers if not needed.
   stage_data_.resize(num);
+  size_t upsampling = 1u << base_color_shift_;
   size_t stage_buffer_xsize =
-      (frame_dimensions_.group_dim << base_color_shift_) +
-      2 * group_data_x_border_ +   // maximum size of a rect
+      frame_dimensions_.group_dim * upsampling +
+      2 * group_data_x_border_ * upsampling +   // maximum size of a rect
```
veluca93, line 379: I *think* `group_data_x_border * (1+upsampling)` rather than `group_data_x_border * 2 * upsampling` is sufficient, but I'm not certain.

eustas: Hmm. Sounds reasonable. But here is what I see: `GroupDone` uses 0-th and 3-rd values in `xpos` array, i.e. it has `xpad` from each side.
So the total delta-x passed to upsampler is `160` (`16+128+16`).

veluca93: Nevermind, yes, you are right :)

veluca93 (approved):

PR #1279, opened 2022-03-25
fix ppm input where bitdepth is not 8 or 16

jonsneyers:
Currently, e.g. 10-bit ppm files are loaded incorrectly: while the range of values in the ppm is 0..1023, they get interpreted as UINT16 so basically everything becomes very dark.

This solves that.

### DIFF
--- a/lib/extras/packed_image.h
+++ b/lib/extras/packed_image.h
@@ -64,6 +64,12 @@ class PackedImage {
   // Whether the y coordinate is flipped (y=0 is the last row).
   bool flipped_y = false;
 
+  // Whether the range is determined by format or by JxlBasicInfo
+  // e.g. if format is UINT16 and JxlBasicInfo bits_per_sample is 10,
+  // then if bitdepth_from_format == true, the range is 0..65535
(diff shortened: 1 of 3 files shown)


mo271 (commented):
great!

mo271 (approved):

lib/extras/packed_image.h
```diff
@@ -64,6 +64,12 @@ class PackedImage {
   // Whether the y coordinate is flipped (y=0 is the last row).
   bool flipped_y = false;
 
+  // Whether the range is determined by format or by JxlBasicInfo
+  // e.g. if format is UINT16 and JxlBasicInfo bits_per_sample is 10,
+  // then if bitdepth_from_format == true, the range is 0..65535
+  // while if bitdepth_from_format == true, the range is 0..1023.
```
ziemek99, line 70: Shouldn't that be:
`// while if bitdepth_from_format == false, the range is 0..1023.`?

jonsneyers: Ah yes indeed. Good catch!

Issue #1283, opened 2022-03-26
build problem on riscv64: undefined reference to `__atomic_fetch_or_1


novomesk:
Hello,
On Gentoo Linux on riscv64 architecture, we have problems to build libjxl.a
```
/usr/lib/gcc/riscv64-unknown-linux-gnu/11.2.1/../../../../riscv64-unknown-linux-gnu/bin/ld: lib/libjxl.a(dec_group_border.cc.o): in function `.L0 ':
dec_group_border.cc:(.text+0x158): undefined reference to `__atomic_fetch_and_1'
```

Complete build log: 
[build.log.txt](https://github.com/libjxl/libjxl/files/8355338/build.log.txt)

jonsneyers:
Closing, should be resolved. Please reopen if needed.

PR #1166, opened 2022-02-08
Pthread dependencies

malaterre:

PR #1128, opened 2022-01-27
libjxl implementation rely on c11 atomics (cache_aligned.cc)

malaterre:
It turns out that GCC requires explicitly linking to a library `atomic`
in order to support c11/atomics. gcc spec file will not handle it
directly at least not on the following Debian archs: armel, mipsel,
powerpc and riscv64.

Introduce a new cmake module to detect usage of gcc/atomic and add
missing library to the link step.

For reference:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104248

PR #1284, opened 2022-03-26
fix missing atomic issue (#1283)

dlan17:
Due to there is no 1, 2byte atomic instruction in 64bit RISC-V hardware,
the software layer have to emulate relavant function in atomic library

Let's explicitly pass -pthread here to work around pthread builtin since glibc version 2.34
as the "-pthread" option will pull in libatomic for machines like RISC-V

the command of "gcc dumpspecs | grep pthread" will show accordingly in RISC-V:
   pthread:--push-state --as-needed -latomic --pop-state

Signed-off-by: Yixun Lan <dlan@gentoo.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,9 @@ endif()  # JPEGXL_STATIC
 # Threads
 set(THREADS_PREFER_PTHREAD_FLAG YES)
 find_package(Threads REQUIRED)
+if(CMAKE_USE_PTHREADS_INIT)
+	target_link_libraries(Threads::Threads INTERFACE -pthread)
+endif()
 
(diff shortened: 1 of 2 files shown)


Hello71:
it seems that `set(THREADS_PREFER_PTHREAD_FLAG True); target_link_libraries(... Threads::Threads)` is the standard way to do this. `target_link_libraries(... -pthread)` seems like it should break windows builds.

malaterre:
duplicate of https://github.com/libjxl/libjxl/pull/1128 ?

dlan17:
> duplicate of #1166 ?

yes, let's mark this one as duplicated issue, "automatical detection + link atomic library" would be a better way to go, thanks

PR #1286, opened 2022-03-28
noise seeding: match frame index of jxli box

jonsneyers:
In the `jxli` box as defined in 18181-2, we don't include zero-duration non-last frames in the frame count. If we want to be able to correctly do the noise seeding when doing `jxli` seeking, we have to match that.


veluca93 (commented):
I think we have a IsDisplayed method somewhere for this...

jonsneyers:
> I think we have a IsDisplayed method somewhere for this...

I thought so too but I couldn't immediately find a suitable existing method...

veluca93 (approved):

veluca93:
I guess we could make one :)

PR #945, opened 2021-12-03
remove shadow variable

maddin200:

PR #1287, opened 2022-03-28
add author

maddin200:


jonsneyers:
Could you add this to the existing https://github.com/libjxl/libjxl/pull/945 pull request? That one will have to be rebased anyway to have the updated AUTHORS list before it can merge, so can just as well just update that pull request to also do the AUTHORS update.

Issue #384, opened 2021-07-30
Incomplete manpage for cjxl


Artoria2e5:
**Describe the bug**
`doc/man/cjxl.txt` seems to be missing several of the options described in `tools/cjxl.cc`'s `AddCommandLineOptions` function. I noticed this problem when looking for how to use `--resampling`.

I am used to thinking a manpage should show all options and provide more explanation on them than `--help`.

**To Reproduce**
Run `cjxl -v --help`. I suppose the verbosity-gate is what caused the original man-writer to omit these options. I guess the gate can be reflected in the man by putting them in an "Advanced Options" section.

Wait, `cjxl -v -v --help` reveals more? Time for a "Professional options" section.   

**Environment**
It's not useful here, but I'll do it.

 - OS: macOS
 - Compiler version: Apple clang version 12.0.5 (clang-1205.0.22.11)
 - CPU type: x86_64
 - cjxl/djxl version string: cjxl [v0.3.7 | SIMD supported: AVX2,SSE4]

**Additional context**
None for now.

Artoria2e5:
I will try and handle this myself. (Copy-and-pasting documentation should be relaxing, right?) If you don't hear from me in two weeks, feel free to take over.

fabiorug:
are photon noise and how to use/force 8, 10 bits written in the manpage?

Artoria2e5:
Not currently, no. `--override_bitdepth=8` is a thing, so it will be added. From what I can tell it is per-channel

fabiorug:
and whatever iulia comsa added to delta, i don't think there are instructions how to use l 1 p 0 s 1    that short of thing, which parameter are needed, which you can avoid and what are the best to combine.

jonsneyers:
One of the reasons we didn't put many of the encode options in the manpage is that they are still experimental and subject to change. For example for the delta palette approach to near lossless: it is still evolving, including possibly the command line syntax for it. There might also be options that will disappear once we make cjxl libjxl-based (internally using the encode api, instead of directly calling encode functions), since some experimental things will not be worth exposing to the api if we have good enough defaults/heuristics for them.

Artoria2e5:
@jonsneyers, is it okay to put the "verbose" `-v` thing in only instead? That set feels generally less experimental and more useful. Or we can wait till PR time to figure that out.

jonsneyers:
Yes, makes sense to start with the options in the default help screen and the -v one, and wait with the -v -v ones...

[...]

PR #1288, opened 2022-03-28
Tools manpages

malaterre:


jonsneyers (approved):

malaterre:
> We kind of intentionally didn't include all options in the manpages

I figured something like this. 

<debian hat one>Debian will enter freeze next year, so I wanted to have something `brushed up` somewhere around summer 2022 for libjxl package.</>

veluca93:
> > We kind of intentionally didn't include all options in the manpages
> 
> I figured something like this.
> 
> /debian hat on/Debian will enter freeze next year, so I wanted to have something _brushed up_ somewhere around summer 2022 for libjxl package./

FWIW 1.0 ought to be comeing soon and c/djxl_ng will come with that, so perhaps it's worth waiting a little...

mo271:
closes #384

PR #1289, opened 2022-03-28
cjxl_ng: fixing single-frame pngs

mo271:
Fixing an issue introduced in #1223. Drive-by: don't assume cropping by default.

### DIFF
--- a/lib/extras/dec/apng.cc
+++ b/lib/extras/dec/apng.cc
@@ -745,7 +745,9 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
     pframe.frame_info.duration = frame.duration;
     pframe.frame_info.layer_info.blend_info.blendmode =
         should_blend ? JXL_BLEND_BLEND : JXL_BLEND_REPLACE;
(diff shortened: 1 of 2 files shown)


jonsneyers (approved):

lib/extras/dec/apng.cc
```diff
@@ -745,7 +745,6 @@ Status DecodeImageAPNG(const Span<const uint8_t> bytes,
     pframe.frame_info.duration = frame.duration;
     pframe.frame_info.layer_info.blend_info.blendmode =
         should_blend ? JXL_BLEND_BLEND : JXL_BLEND_REPLACE;
-    pframe.frame_info.layer_info.have_crop = 1;
```
jonsneyers, line 748: Shouldn't this be true in case x0/y0 are nonzero or xsize/ysize do not match the image dimensions?

mo271: Yes, added that.

jonsneyers (approved):

mo271:
rebased because of the failing arm cross compile check

jonsneyers (approved):

PR #1296, opened 2022-03-29
Remove a simple warning on 32bits arch where SIZE_MAX < ULLONG_MAX

malaterre:
This commit removes the following warning:

libjxl/lib/jxl/headers.cc:64:15: warning: comparison is always false due to limited range of data type [-Wtype-limits]
   64 |   if (xsize64 > 0xFFFFFFFFull || ysize64 > 0xFFFFFFFFull) {
      |       ~~~~~~~~^~~~~~~~~~~~~~~


veluca93 (commented):
@eustas can you check that this function works on 32-bit systems?

eustas:
Will take a look soon.

eustas:
Exact values are not important here, it is just a safeguard.
So if we just replace `> 0xFFFF'FFFF` with `>= 0xFFFF'FFFF` then 32-bit compiler will become happy, and de don't need `#if`.

malaterre:
> replace `> 0xFFFF'FFFF` with `>= 0xFFFF'FFFF`

Very smart indeed, but this change the behavior, which I could not do for a simple warning.

veluca93:
> > replace `> 0xFFFF'FFFF` with `>= 0xFFFF'FFFF`
> 
> Very smart indeed, but this change the behavior, which I could not do for a simple warning.

I think changing the behaviour is fine :)

jonsneyers:
Probably we could change that constant by the largest size allowed in level 10...

eustas (approved):

eustas:
True. Is it `2^30 + 1`?
BTW, we need to fix `FieldsTest.TestOutOfRange` as well. It seems to test oversized extensions, but uses extremal w/h values in size header as well.

jonsneyers:
> True. Is it `2^30 + 1`? BTW, we need to fix `FieldsTest.TestOutOfRange` as well. It seems to test oversized extensions, but uses extremal w/h values in size header as well.

Largest xsize or ysize allowed in level 10 is  `1 << 30`.

It doesn't mean we cannot support larger sizes (it's OK if we also decode codestreams that don't conform to level 10), but that's the largest size the encode api will allow.

The bitstream limit (largest possible dimensions that can be signaled) is `1 << 30` for the height (using 1 + u(30)) and `1 << 31` for the width (using `ratio == 7` for a 2:1 aspect ratio with the largest height).

Level 10 also specifies that width times height is at most `1 << 40`, so that further restricts things.

lib/jxl/headers.cc
```diff
@@ -58,7 +58,7 @@ size_t SizeHeader::xsize() const {
 }
 
 Status SizeHeader::Set(size_t xsize64, size_t ysize64) {
-  if (xsize64 > 0xFFFFFFFFull || ysize64 > 0xFFFFFFFFull) {
+  if (xsize64 >= 0xFFFFFFFFull || ysize64 >= 0xFFFFFFFFull) {
```
jonsneyers, line 61: You can safely change these conditions to
```suggestion
  if (xsize64 > 0x40000000ull || ysize64 > 0x40000000ull) {
```
since there is no valid way in which the dimensions can be larger than that.

jonsneyers:
There's a test in `fields_test.cc` that should be updated since it tries to set a too-large dimension and expects no error. Maybe make it expect error for that value and no error for smaller, valid numbers.

PR #1297, opened 2022-03-29
Fix crash in set_from_bytes fuzzer

eustas:
AdoptWhitePoint error is silenced in ConvertExternalToInternalColorEncoding ->
-> ConvertPackedPixelFileToCodecInOut did not notice ICC was not set ->
-> CHECK triggered.

### DIFF
--- a/lib/extras/packed_image_convert.cc
+++ b/lib/extras/packed_image_convert.cc
@@ -69,6 +69,9 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
   } else {
     JXL_RETURN_IF_ERROR(ConvertExternalToInternalColorEncoding(
         ppf.color_encoding, &io->metadata.m.color_encoding));
+    if (io->metadata.m.color_encoding.ICC().empty()) {
+      return JXL_FAILURE("Failed to serialize ICC");
+    }
   }
 
   // Convert the extra blobs


lib/extras/packed_image_convert.cc
```diff
@@ -59,16 +59,22 @@ Status ConvertPackedPixelFileToCodecInOut(const PackedPixelFile& ppf,
...
     JXL_RETURN_IF_ERROR(ConvertExternalToInternalColorEncoding(
         ppf.color_encoding, &io->metadata.m.color_encoding));
+    // Successful ConvertExternalToInternalColorEncoding does not mean there
+    // were no errors (and ICC was actually set).
+    want_fallback_icc = io->metadata.m.color_encoding.ICC().empty();
+  }
+
+  if (want_fallback_icc) {
+    fprintf(stderr, "Warning: error setting ICC profile, assuming SRGB");
+    io->metadata.m.color_encoding = ColorEncoding::SRGB(is_gray);
```
veluca93, line 77: Couldn't this condition be false because the color encoding *was* set, but not as an ICC? In that case this condition would do something wrong...
I believe we have a function to set the ICC from the internal colorencoding, perhaps we should use that?

eustas: At the end of `ConvertExternalToInternalColorEncoding` makes an attempt to create ICC (`ColorEncoding::CreateICC` -> `ColorEncoding::MaybeCreateProfile`). But it fails (`AdoptWhitePoint` recognises invalid input).

So, if there is no ICC, it means it can not be created. And we need it for `CheckMetadata`.

eustas: Fine for me. One sec, will send an update.

eustas:
PTAL

veluca93 (approved):

PR #1275, opened 2022-03-25
remove modular quality

jonsneyers:
Get rid of the separate cparams.quality_pair for modular quality, instead use cparams.butteraugli_distance also in modular mode as the main setting that controls lossy quality.

Also taking the opportunity to better calibrate modular and vardct quality and to tweak the quality settings chosen for alpha, dc frames, and patch frames.

## Effect on dc frames

Test corpus: jyrki31.

Before:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:d0.5        13270  4225280    2.5471813   3.049  22.687   1.12016439   0.37185960  0.947193828701      0
jxl:d1          13270  2680279    1.6157880   3.333  24.225   1.68695998   0.60068375  0.970577588663      0
jxl:d2          13270  1685403    1.0160337   3.511  25.091   2.98979163   0.94922260  0.964442190920      0
jxl:d4          13270  1006130    0.6065386   3.296  17.630   4.82071018   1.50086778  0.910334293335      0
jxl:d5          13270   825027    0.4973619   3.258  16.790   5.96885347   1.76033233  0.875522269601      0
jxl:d6          13270   718021    0.4328541   3.274  16.965   7.37481737   1.97648186  0.855528239885      0
jxl:d8          13270   576361    0.3474553   3.231  16.312   8.18236828   2.35794399  0.819280165669      0
jxl:d10         13270   478064    0.2881976   3.291  16.836   9.00066471   2.74497689  0.791095844090      0
jxl:d15         13270   333269    0.2009090   3.306  15.588  12.30572796   3.70734670  0.744839140403      0
Aggregate:      13270  1004866    0.6057768   3.281  18.822   4.68458678   1.43966693  0.872116784553      0
```

After:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------
jxl:d0.5        13270  4225280    2.5471813   2.944  21.976   1.12016439   0.37185960  0.947193828701      0
jxl:d1          13270  2680279    1.6157880   3.214  24.537   1.68695998   0.60068375  0.970577588663      0
jxl:d2          13270  1685403    1.0160337   3.412  23.954   2.98979163   0.94922260  0.964442190920      0
[...]

PR #1303, opened 2022-03-29
Adjust alpha quality expectations.

eustas:
Alpha quality got better, especially for scalar build.
Readjusted expectationss to unbreak tests.

Related: #1275

### DIFF
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -786,7 +786,7 @@ TEST(JxlTest, RoundtripAlpha) {
 
   EXPECT_THAT(ButteraugliDistance(io, io2, cparams.ba_params, GetJxlCms(),
                                   /*distmap=*/nullptr, pool),
-              IsSlightlyBelow(1.4));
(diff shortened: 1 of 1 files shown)


jonsneyers:
Is it different for scalar vs simd? Because that is somewhat strange...

eustas:
Yes. E.g. scalar butteraugli score is ~0.95, whereas regular is ~1.15 in those tests.

veluca93 (approved):

veluca93:
> Yes. E.g. scalar butteraugli score is ~0.95, whereas regular is ~1.15 in those tests.

That _is_ rather suspicious...

eustas:
I can check if it was the case before #1275.
Likely it was such a way for a long time, and we have just shifted a baseline for both.

jonsneyers:
That's a larger difference than what I would expect just from float arithmetic rounding differences. The alpha itself should be identical since it uses only integer arithmetic.

It smells like a bug, maybe in butteraugli itself when dealing with alpha?

PR #1305, opened 2022-03-30
Refactor padding calculation to improve readability.

eustas:
Drive-by: generalise Rect to use with ssize_t (negative origin).

### DIFF
--- a/lib/jxl/image.h
+++ b/lib/jxl/image.h
@@ -194,87 +194,94 @@ class Image3;
 // shifting the pointer by x0/y0 allows this to apply to multiple images with
 // different resolutions (e.g. color transform and quantization field).
 // Can compare using SameSize(rect1, rect2).
-class Rect {
+template <typename T>
+class RectT {
  public:
   // Most windows are xsize_max * ysize_max, except those on the borders where
   // begin + size_max > end.
-  constexpr Rect(size_t xbegin, size_t ybegin, size_t xsize_max,
-                 size_t ysize_max, size_t xend, size_t yend)
+  constexpr RectT(T xbegin, T ybegin, size_t xsize_max, size_t ysize_max,
+                  T xend, T yend)
       : x0_(xbegin),
         y0_(ybegin),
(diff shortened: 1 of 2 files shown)


lib/jxl/image.h
```diff
@@ -194,21 +194,65 @@ class Image3;
...
+
+  JXL_MUST_USE_RESULT RectT Translate(int64_t x_offset,
+                                      int64_t y_offset) const {
+    return RectT(x0_ + x_offset, y0_ + y_offset, xsize_, ysize_);
+  }
+
+  T x0() const { return x0_; }
+  T y0() const { return y0_; }
+  size_t xsize() const { return xsize_; }
+  size_t ysize() const { return ysize_; }
```
veluca93, line 230: Can we do x1, y1 while we're at it?

lib/jxl/image.h
```diff
@@ -194,21 +194,65 @@ class Image3;
...
+  T x0_;
+  T y0_;
+
+  size_t xsize_;
+  size_t ysize_;
+};
+
+class Rect : public RectT<size_t> {
+ public:
+  constexpr Rect(size_t xbegin, size_t ybegin, size_t xsize_max,
```
veluca93, line 249: Could do using RectT::RectT, no?

veluca93 (approved):

lib/jxl/image.h
eustas: Sure.

lib/jxl/image.h
eustas: Merged back Rect and RectT. Inheritance of templatized producer without CRTP is a hell, and with it - it looks disgusting.

veluca93: Just thinking out loud, but couldn't we make all rects use ssize_t?

veluca93 (approved):

lib/jxl/image.h
eustas: We could try. One thing I foresee is problem with large images on 32-bit system... But it is likely, we will get OOM first.

PR #1309, opened 2022-03-31
Fix q100

jonsneyers:
The changes in https://github.com/libjxl/libjxl/pull/1275 accidentally caused `-q 100` to correspond to d0.1, which was not the intention. This makes `cjxl -q 100` lossless again. Another changed behavior is that `cjxl -m` now defaults to the same distance as non-modular encoding (instead of defaulting to lossless), but I think that's OK to keep like that.


veluca93:
There's an extra commit :)

jonsneyers:
Oops, fixed.

veluca93 (approved):

PR #1310, opened 2022-03-31
cjxl_ng: Extensive clean-up

fischbacher:
- Greatly streamline flag-handling
 - Add support for --num_reps
 - Add --quality in addition to --depth
 - Many detail improvements


tools/cjxl_ng_main.cc
```diff
@@ -427,467 +413,437 @@ int main(int argc, char** argv) {
...
-    }
-    if (modular_predictor_set) {
-      if (!(0 <= FLAGS_modular_predictor && FLAGS_modular_predictor <= 3)) {
-        std::cerr << "Invalid --modular_predictor: " << FLAGS_modular_predictor
-                  << std::endl;
-        return EXIT_FAILURE;
+
+      process_flag("center_x", FLAGS_center_x,
+                   JXL_ENC_FRAME_SETTING_GROUP_ORDER_CENTER_X,
+                   [](int32_t x) -> std::string { return ""; });
```
mo271, line 605: I guess we could check if `-1 <= x` and return "valid values are -1 and [0, xsize)" or something similar if this is not the case.

mo271 (approved):

tools/cjxl_ng_main.cc
```diff
@@ -35,9 +46,9 @@ DEFINE_bool(encoder_version, false,
             "Print encoder library version number and exit.");
 
 DEFINE_bool(add_jpeg_frame, false,
-            // This supersedes --jpeg_transcode
-            "Use JxlEncoderAddJPEGFrame to add a JPEG frame, "
-            "rather than JxlEncoderAddImageFrame.");
+            "Use JxlEncoderAddJPEGFrame to add a JPEG frame "
+            "(i.e. transcoding JPEG), rather than using "
+            "JxlEncoderAddImageFrame.");
```
jonsneyers, line 51: Current cjxl does this by default (including jbrd) when the input is a jpeg. It would be nice if we can keep that behavior.

tools/cjxl_ng_main.cc
```diff
@@ -65,26 +76,17 @@ DEFINE_bool(qprogressive_ac, false, "Use progressive mode for AC.");
...
 DEFINE_bool(modular_lossy_palette, false, "Use delta-palette.");
 
-DEFINE_bool(jpeg_transcode, false,  // TODO(tfish): Wire this up.
-            "Do lossy transcode of input JPEG file (decode to "
-            "pixels instead of doing lossless transcode).");
-
-DEFINE_bool(premultiply, false,  // TODO(tfish): Wire this up.
+DEFINE_bool(premultiply, false,
+            // TODO(tfish): Discuss - should `false` be treated differently from
+            // `unset` w.r.t. what default to use?
```
jonsneyers, line 81: I think we only have input formats that use non-premultiplied alpha, so I think it's a sensible default to have false here. If and when we would add an input format that (can) use premultiplied alpha, I think it would make sense to have a default that is "do the same as the input format", with an override to change it (which is a lossy operation in both directions).

tools/cjxl_ng_main.cc
```diff
@@ -149,49 +150,34 @@ DEFINE_int32(
...
-             // upper limit here.
-             "other values 0..(ysize-1) set this to a particular coordinate.");
+             "other values [0..ysize) set this to a particular coordinate.");
 
 DEFINE_int64(num_threads, 0,
-             // TODO(tfish): Sync with team about changed meaning of 0 -
-             // was: No multithreaded workers. Is: use default number.
-             "Number of worker threads (0 == use machine default).");
+             "Number of worker threads (0 == use machine default, "
+             "-1 == do not use multithreading).");
```
jonsneyers, line 164: To me it makes more sense to have -1 = default, 0 = everything in main thread. We use -1 as the default in just about every other setting, so why not here?

tools/cjxl_ng_main.cc
```diff
@@ -235,37 +216,23 @@ DEFINE_int32(photon_noise, 0,
 DEFINE_int32(codestream_level, 5, "The codestream level. Either `5` or `10`.");
```
jonsneyers, line 216: It might make sense to have a default of -1 here, meaning "let the encoder set the codestream level to the minimum needed to represent the input". Otherwise you have to manually set the level to 10 when you're e.g. trying to losslessly encode a 16-bit png or a 32-bit pfm, or an image that is more than 268 megapixels, since level 5 encoding will fail for those.

tools/cjxl_ng_main.cc
```diff
@@ -235,37 +216,23 @@ DEFINE_int32(photon_noise, 0,
...
 
 DEFINE_double(
-    distance, 1.0,  // TODO(tfish): wire this up.
-    "Max. butteraugli distance, lower = higher quality. Range: 0 .. 25.\n"
+    distance, 1.0,
+    "Max. butteraugli distance, lower = higher quality.\n"
     "    0.0 = mathematically lossless. Default for already-lossy input "
     "(JPEG/GIF).\n"
```
jonsneyers, line 222: I think we still have to make sure that this actually happens: doing lossless when the input is JPEG or GIF. I think we currently just load the input and use d1 regardless of what the input format was.

tools/cjxl_ng_main.cc
```diff
@@ -277,60 +244,76 @@ DEFINE_string(frame_indexing, "",
...
+  }
+  if (distance_set) {
+    if (JXL_ENC_SUCCESS != JxlEncoderSetFrameDistance(
+                                                      jxl_encoder_frame_settings, FLAGS_distance)) {
+      std::cerr << "Setting --distance parameter failed." << std::endl;
+      exit(EXIT_FAILURE);
+    }
+  }
+  if (quality_set) {
+    double distance = FLAGS_quality >= 30? 0.1 + (100 - FLAGS_quality) * 0.09 : 6.4 + pow(2.5, (30 - FLAGS_quality) / 5.0) / 6.25;
```
jonsneyers, line 304: Please apply the fix from https://github.com/libjxl/libjxl/pull/1309 also here: quality 100 should map to distance 0.

tools/cjxl_ng_main.cc
```diff
@@ -277,60 +244,76 @@ DEFINE_string(frame_indexing, "",
               "the frame index box.");
 
 DEFINE_string(
-    // TODO(tfish): Clarify with team whether changing from int-param to string
-    // is OK here.
     colortransform, "",
     "The color transform to use. Valid values are: '' (= \"use default\"), "
     "'RGB', 'XYB', 'YCbCr'.");
```
jonsneyers, line 249: It's probably a good idea to mention here that `YCbCr` actually means that the input will be interpreted as if it is in YCbCr already, i.e. we're not doing the forward color transform (RGB to YCbCr) but just encoding the pixel values as-is and signaling that they are YCbCr. This is surprising and is only useful for lossless compression of already-YCbCr data (like video frames in .y4m) that also happens to be 4:4:4, since we don't currently have a way to pass chroma subsampled data using the encode api.

tools/cjxl_ng_main.cc
```diff
@@ -35,9 +46,9 @@ DEFINE_bool(encoder_version, false,
...
 
 DEFINE_bool(add_jpeg_frame, false,
-            // This supersedes --jpeg_transcode
-            "Use JxlEncoderAddJPEGFrame to add a JPEG frame, "
-            "rather than JxlEncoderAddImageFrame.");
+            "Use JxlEncoderAddJPEGFrame to add a JPEG frame "
+            "(i.e. transcoding JPEG), rather than using "
+            "JxlEncoderAddImageFrame.");
 
 DEFINE_bool(jpeg_store_metadata, false,
```
jonsneyers, line 53: Wouldn't it make more sense to let this default to `true` so recompressed jpegs can actually be reconstructed? That's the current behavior of `cjxl`.

jonsneyers:
I left a bunch of comments, of course not all of them necessarily need to be addressed by this pull request but please keep track of them (maybe with a TODO).

tools/cjxl_ng_main.cc
jonsneyers: Ah right. So in that case I recommend having a default "do whatever the input format does".

fischbacher: Turned into an int-flag to allow three options.

tools/cjxl_ng_main.cc
fischbacher: Switched -1 and 0.

tools/cjxl_ng_main.cc
fischbacher: Do we actually know what the "minimum needed" is before we have seen every frame?

tools/cjxl_ng_main.cc
fischbacher: Done.

tools/cjxl_ng_main.cc
fischbacher: Done.

tools/cjxl_ng_main.cc
fischbacher: Changed docstring. Does this address the issue?

tools/cjxl_ng_main.cc
fischbacher: ? Duplicate comment?

fischbacher:
Duh... as I thought, the "multiprocessing hanging" thing *was* something seriously stupid.

tools/cjxl_ng_main.cc
fischbacher: Changed behavior to: if input is JPEG, do lossless transcoding by default, but say on stderr that cjxl_ng is relying on that default, asking the user to make a choice in order to silence that message.

tools/cjxl_ng_main.cc
```diff
@@ -277,60 +244,76 @@ DEFINE_string(frame_indexing, "",
               "the frame index box.");
```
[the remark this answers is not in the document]
fischbacher: For now, added a TODO here.

tools/cjxl_ng_main.cc
jonsneyers: No, we don't, at least not in the case of animations, since some restrictions are things like the minimum nonzero frame duration and the maximum number of pixels in consecutive zero-duration frames.

But most of the level restrictions apply to things that can be known early, like bitdepth, number of extra channels, and image dimensions.

tools/cjxl_ng_main.cc
jonsneyers: SGTM.

tools/cjxl_ng_main.cc
```diff
@@ -235,58 +216,40 @@ DEFINE_int32(photon_noise, 0,
...
-    colortransform, "",
-    "The color transform to use. Valid values are: '' (= \"use default\"), "
-    "'RGB', 'XYB', 'YCbCr'.");
-
-DEFINE_string(
-    mquality, "",  // TODO(tfish): Wire this up.
-    "[modular encoding] lossy 'quality', in the form luma_q[,chroma_q] "
-    "(100=lossless, lower is more lossy)");
+DEFINE_string(colortransform, "",
+              "The color transform used by the input. Valid values are: '' (= "
```
jonsneyers, line 250: It's a bit more complicated than that:
- `RGB` means: don't do any color transform (this is the default when doing lossless)
- `XYB` means: encoder will convert the RGB input to XYB (this is the default when doing lossy)
- `YCbCr` means: the input is already in YCbCr, the encoder will not do any color transform but it will signal in the frame header(s) that the data is YCbCr so it will get decoded correctly (this is the default when doing lossless JPEG recompression, and it would also be the thing to do when doing lossless compression from y4m input, if we want to support that).

To be honest, I don't think this parameter for cjxl is very useful, since doing anything different than the default will either lead to incorrect results, or to poor results. So I wouldn't object to just removing this parameter.

tools/cjxl_ng_main.cc
```diff
@@ -295,61 +258,99 @@ namespace {
...
+      std::cerr << "Setting --quality parameter failed." << std::endl;
+      exit(EXIT_FAILURE);
+    }
+    return;
+  }
+  // No flag set, but input is JPG or GIF: Use distance 0 default.
+  if (codec == jxl::extras::Codec::kJPG || codec == jxl::extras::Codec::kGIF) {
+    if (JXL_ENC_SUCCESS !=
+        JxlEncoderSetFrameDistance(jxl_encoder_frame_settings, 0.0)) {
+      std::cerr << "Setting '100% quality' default for GIF or JPEG input."
```
jonsneyers, line 325: This should probably be in the `else` branch of that test. Also I don't like `100% quality`, it's better to say lossless.

tools/cjxl_ng_main.cc
fischbacher: So... the idea here is that if we see anything in the parameters that force level 10, we set level 10 straightaway, and if it can be determined only while processing frames, and the user did not set level 10, they get a failure message?

Sounds like a reasonable for-convenience idea, overall, but the API documentation should have a list of things that automatically trigger level 10. Not being too familiar with the internals here, I would struggle to come up with a complete list of criteria, and in the interest of having behavior documented, cjxl_ng users should be able to reason about when this automatic switch to level 10 happens.

tools/cjxl_ng_main.cc
fischbacher: Ah! We are talking about jpeg_store_metadata. Agree that this should be default-true.

tools/cjxl_ng_main.cc
fischbacher: Agree. Removed the colortransform flag related code.

tools/cjxl_ng_main.cc
fischbacher: Thanks for spotting. Fixed.

tools/cjxl_ng_main.cc
jonsneyers: For JPEG recompression, since we only do 8-bit JPEGs without CMYK, level 5 should usually be enough. The exception is if it's a huge JPEG with more than 1 << 28 pixels or one with a very large ICC profile.

In terms of API, I think we need to clarify how we want the API to work, and then make sure it works like that and properly document it. I see three options:

- Application is responsible for setting the codestream level _before_ setting basic info, and setting basic info fails when trying to set things that are not compatible with the codestream level. In this case, GetRequiredCodestreamLevel is not useful and should get deprecated. This is quite inconvenient because it means applications have to know what level to set.
- Application is responsible for setting the codestream level _after_ setting basic info (but before encoding the first frame), and can use GetRequiredCodestreamLevel to set it based on what is implied by basic info. This is somewhat error-prone since applications will likely forget to call GetRequiredCodestreamLevel and set the level accordingly, which will usually be fine (since many images can be done in the default level 5), but will cause unexpected failure when e.g. the dimensions are too large or when trying to encode a 16-bit image.
- Application can optionally manually set the codestream level, but if it doesn't, the encoder automatically picks the lowest level that is compatible with what is known when it has to decide what to write as the codestream level. So if it wasn't manually set, it chooses the lowest level possible, and if it was manually set, it uses that level and returns error if the level turns out to be too low. In this case, GetRequiredCodestreamLevel is not needed. This is probably the most convenient way to do it. It also has the advantage that AddJPEGFrame can still set the level to 10 in case of a very large JPEG.

novomesk: > * but will cause unexpected failure when e.g. the dimensions are too large or when trying to encode a 16-bit image.

Is level 10 really necessary for 16bit images?

According https://github.com/libjxl/libjxl/blob/eb152f5259df9af806c5b47b192072444b9dc41b/lib/include/jxl/encode.h#L943 level 5 has supports _maximum 16-bit color channel depth_.

jonsneyers: Level 10 is necessary to do _lossless_ 16-bit.

Technically, what makes the difference is if `int16_t` buffers can be used for the modular parts or not. When doing lossy VarDCT, this will be fine, since modular is then only used for quantized DC and control fields, which are all things that fit in `int16_t`. But when doing lossless Modular with uint16 input, `int16_t` is not enough. The current encoder will signal that `int16_t` is not enough whenever trying to do lossless with more than 12-bit input (it puts the boundary there so it still has spare bits for doing RCT and/or Squeeze), and that means level 5 is not enough.

Note that this is unrelated to how buffers are passed in the API, e.g. you can still get/set buffers using UINT16 or FLOAT32 sample types regardless of the codestream level.

novomesk: This is important information.

16bit lossless always worked in 0.6.1

But now in git master I have to set codestream level to 10.

jonsneyers: In 0.6.1, the encoder is producing a codestream that is not properly marked as being level 10. The current decoder doesn't care about it, but it's kind of an encoder bug. But I do think that it would be a good idea to implement the third option I proposed, i.e. the encoder automatically sets the level to 10 when needed.

sboukortt (approved):

jonsneyers:
I think this needs a rebase.

sboukortt (approved):

sboukortt (approved):

PR #1311, opened 2022-03-31
Fix OOM caused by large negative origin

eustas:


lib/jxl/render_pipeline/low_memory_render_pipeline.cc
```diff
@@ -397,15 +398,16 @@ void LowMemoryRenderPipeline::PrepareForThreadsInternal(size_t num,
   }
   if (first_image_dim_stage_ != stages_.size()) {
     out_of_frame_data_.resize(num);
+    size_t left_padding = std::max<ssize_t>(0, frame_origin_.x0);
+    size_t middle_padding = group_dim;
```
veluca93, line 402: What is middle_padding?

eustas: Actually it is top-bottom padding. But x-wise it is "middle" - for non-first non-last gx.

veluca93 (approved):

PR #1312, opened 2022-04-01
Pnm max val log2

malaterre:


veluca93 (approved):

veluca93 (dismissed):

malaterre:
@jonsneyers This is the easy portion of the patch. I get confused by this:

```
% convert -size 512x512 -depth 1 xc:white white.1.pgm
 % tools/cjxl white.1.pgm l.jxl
[...]
./lib/jxl/enc_external_image.cc:231: JXL_FAILURE: packed 1-bit per sample is not yet supported
```

where:

```
Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
[...]
  // TODO(deymo): Implement 1-bit per sample as 8 samples per byte. In
  // any other case we use DivCeil(bits_per_sample, 8) bytes per pixel per
  // channel.
  if (bits_per_sample == 1) {
    return JXL_FAILURE("packed 1-bit per sample is not yet supported");
  }
```

lib/extras/dec/pnm.cc
```diff
@@ -231,7 +231,8 @@ class Parser {
     if (max_val == 0 || max_val >= 65536) {
       return JXL_FAILURE("PAM: bad MAXVAL");
     }
-    header->bits_per_sample = CeilLog2Nonzero(max_val);
+    // e.g When `max_val` is 4096, we want 13 bits:
```
jonsneyers, line 234: Maybe we should return error if maxval is not of the form 2^n - 1. I'm not sure if it is obligatory to do that in pnm, but I know that's what basically all applications will expect, including this one.

jonsneyers:
> @jonsneyers This is the easy portion of the patch. I get confused by this (pgm storage, not pbm):
> 
> ```
> % convert -size 512x512 -depth 1 xc:white white.1.pgm
> % head -3 white.1.pgm
> P5
> 512 512
> 1
> % tools/cjxl white.1.pgm l.jxl
> [...]
> ./lib/jxl/enc_external_image.cc:231: JXL_FAILURE: packed 1-bit per sample is not yet supported
> ```
> 
> where:
> 
> ```
> Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
> [...]
>   // TODO(deymo): Implement 1-bit per sample as 8 samples per byte. In
>   // any other case we use DivCeil(bits_per_sample, 8) bytes per pixel per
>   // channel.
>   if (bits_per_sample == 1) {
>     return JXL_FAILURE("packed 1-bit per sample is not yet supported");
>   }
> ```
> 
> My confusion is that information from `bitdepth_from_format` (commit [d272638](https://github.com/libjxl/libjxl/commit/d2726387710c22b3e0350d20d843111987c157a7)) is lost at this point.

Right, I think the original confusion was that bits_per_sample == 1 implies things are packed (8 pixels per byte) while it's of course also perfectly possible to use 1 byte per pixel for 1-bit images (which is what pgm does).

Since we don't support packed 1-bit anyway, and atm supporting it wouldn't help much to save memory because we'll internally store it as 3x32-bit floats anyway, I'm inclined to say that we should just assume things are _not_ packed in ConvertFromExternal, i.e. treat 1-bit just like how we treat 2-bit.

veluca93:
> > @jonsneyers This is the easy portion of the patch. I get confused by this (pgm storage, not pbm):
> > ```
> > % convert -size 512x512 -depth 1 xc:white white.1.pgm
> > % head -3 white.1.pgm
> > P5
> > 512 512
> > 1
> > % tools/cjxl white.1.pgm l.jxl
> > [...]
> > ./lib/jxl/enc_external_image.cc:231: JXL_FAILURE: packed 1-bit per sample is not yet supported
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where:
> > ```
> > Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
> > [...]
> >   // TODO(deymo): Implement 1-bit per sample as 8 samples per byte. In
> >   // any other case we use DivCeil(bits_per_sample, 8) bytes per pixel per
> >   // channel.
> >   if (bits_per_sample == 1) {
> >     return JXL_FAILURE("packed 1-bit per sample is not yet supported");
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My confusion is that information from `bitdepth_from_format` (commit [d272638](https://github.com/libjxl/libjxl/commit/d2726387710c22b3e0350d20d843111987c157a7)) is lost at this point.
> 
> Right, I think the original confusion was that bits_per_sample == 1 implies things are packed (8 pixels per byte) while it's of course also perfectly possible to use 1 byte per pixel for 1-bit images (which is what pgm does).
> 
> Since we don't support packed 1-bit anyway, and atm supporting it wouldn't help much to save memory because we'll internally store it as 3x32-bit floats anyway, I'm inclined to say that we should just assume things are _not_ packed in ConvertFromExternal, i.e. treat 1-bit just like how we treat 2-bit.

Note that IIRC that packed bits is how libpng will give you 1-bit images :)

lib/extras/dec/pnm.cc
malaterre: `CeilLog2Nonzero(1u)` is legal

malaterre:
> Right, I think the original confusion was that bits_per_sample == 1 implies things are packed (8 pixels per byte) while it's of course also perfectly possible to use 1 byte per pixel for 1-bit images (which is what pgm does).

Ok, I've pushed a commit with your comment. Let me know if that is acceptable now.

lib/extras/dec/pnm.cc
malaterre: Ok I added the missing `JXL_RETURN_IF_ERROR` stuff

jonsneyers:
> > > @jonsneyers This is the easy portion of the patch. I get confused by this (pgm storage, not pbm):
> > > ```
> > > % convert -size 512x512 -depth 1 xc:white white.1.pgm
> > > % head -3 white.1.pgm
> > > P5
> > > 512 512
> > > 1
> > > % tools/cjxl white.1.pgm l.jxl
> > > [...]
> > > ./lib/jxl/enc_external_image.cc:231: JXL_FAILURE: packed 1-bit per sample is not yet supported
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > where:
> > > ```
> > > Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
> > > [...]
> > >   // TODO(deymo): Implement 1-bit per sample as 8 samples per byte. In
> > >   // any other case we use DivCeil(bits_per_sample, 8) bytes per pixel per
> > >   // channel.
> > >   if (bits_per_sample == 1) {
> > >     return JXL_FAILURE("packed 1-bit per sample is not yet supported");
> > >   }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > My confusion is that information from `bitdepth_from_format` (commit [d272638](https://github.com/libjxl/libjxl/commit/d2726387710c22b3e0350d20d843111987c157a7)) is lost at this point.
> > 
> > 
> > Right, I think the original confusion was that bits_per_sample == 1 implies things are packed (8 pixels per byte) while it's of course also perfectly possible to use 1 byte per pixel for 1-bit images (which is what pgm does).
> > Since we don't support packed 1-bit anyway, and atm supporting it wouldn't help much to save memory because we'll internally store it as 3x32-bit floats anyway, I'm inclined to say that we should just assume things are _not_ packed in ConvertFromExternal, i.e. treat 1-bit just like how we treat 2-bit.
> 
> Note that IIRC that packed bits is how libpng will give you 1-bit images :)

Libpng has an option to expand 1, 2 and 4 bit images to 8 bit. We use that option.

malaterre:
> Note that IIRC that packed bits is how libpng will give you 1-bit images :)

Hence my comment about `pbm` format ;)

jonsneyers:
Could you make it return an error on `pbm` input? It will error anyway because it will expect non-bitpacked data so it will not get enough bytes, but that will be a confusing error message and it would be better to immediately do a `JXL_FAILURE("pbm is not supported")` when encountering a pbm header.

lib/extras/dec/pnm.cc
```diff
@@ -231,7 +231,9 @@ class Parser {
     if (max_val == 0 || max_val >= 65536) {
       return JXL_FAILURE("PAM: bad MAXVAL");
     }
-    header->bits_per_sample = CeilLog2Nonzero(max_val);
+    // e.g When `max_val` is 1 , we want 1 bit:
+    header->bits_per_sample = FloorLog2Nonzero(max_val) + 1;
+    JXL_RETURN_IF_ERROR((1u << header->bits_per_sample) - 1 == max_val);
```
jonsneyers, line 236:
```suggestion
    if ((1u << header->bits_per_sample) - 1 != max_val) return JXL_FAILURE("PNM: unsupported MaxVal (expected 2^n - 1)");
```

jonsneyers (approved):

PR #1314, opened 2022-04-01
Add missing install rules for jar and jni targets

malaterre:


malaterre:
Could somone confirm the naming convention `jxl_jni_wrapper.jar`, I would have expected `org.jpeg.jpegxl.jar`, right ?

eustas:
Lets switch to `org.jpeg.jpegxl.jar`.

eustas (approved):

eustas (approved):

PR #1315, opened 2022-04-02
Build `jxl_extras_dec` only as a shared library if necessary

kleisauke:
This should fix libvips' CIFuzz and OSS-Fuzz integration.
https://github.com/libvips/libvips/runs/5799006301
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libvips

Alternatively, we could avoid building `jxl_extras_dec` as a shared
library, since it's not being linked against.

(The logic is derived from `jxl_threads.cmake`)


sboukortt (approved):

sboukortt:
Thanks!

PR #1320, opened 2022-04-06
Try to fix MSAN build

eustas:


mo271 (approved):
seems like it worked!

PR #1322, opened 2022-04-06
Allow ImageOutCallbacks to receive more information about threading

sboukortt:


lib/extras/enc/apng.cc
```diff
@@ -169,8 +169,11 @@ Status EncodeImageAPNG(const CodecInOut* io, const ColorEncoding& c_desired,
     JXL_RETURN_IF_ERROR(ConvertToExternal(
         *transformed, bits_per_sample, /*float_out=*/false,
         c_desired.Channels() + (ib.HasAlpha() ? 1 : 0), JXL_BIG_ENDIAN, stride,
-        pool, raw_bytes.data(), raw_bytes.size(), /*out_callback=*/nullptr,
-        /*out_opaque=*/nullptr, metadata.GetOrientation()));
+        pool, raw_bytes.data(), raw_bytes.size(),
+        /*out_init_callback=*/nullptr,
```
veluca93, line 173: Should we make those into a struct perhaps?

lib/include/jxl/decode.h
```diff
@@ -917,6 +917,55 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetImageOutBuffer(
...
+ * Worker callback for @c JxlDecoderSetMultithreadedImageOutCallback.
+ *
+ * @see JxlDecoderSetMultithreadedImageOutCallback
+ *
+ * @param run_opaque user data returned by the @c init callback.
+ * @param thread_id number in `[0, num_threads)` identifying the thread of the
+ * current invocation of the callback.
+ * @param x horizontal position of the first (leftmost) pixel of the pixel data.
+ * @param y vertical position of the pixel data.
+ * @param num_pixels number of pixels in the pixel data. May be less than the
```
veluca93, line 947: and is certainly less than num_pixels_per_thread?

lib/include/jxl/decode.h
```diff
@@ -963,6 +1012,30 @@ JXL_EXPORT JxlDecoderStatus
 JxlDecoderSetImageOutCallback(JxlDecoder* dec, const JxlPixelFormat* format,
                               JxlImageOutCallback callback, void* opaque);
 
+/** Similar to @c JxlDecoderSetImageOutCallback except that the callback is
+ * allowed an initialization phase during which it is informed of how many
+ * threads will call it concurrently, and those calls are further informed of
+ * which thread they are occurring in.
+ *
+ * @param dec decoder object
+ * @param format format of the pixels. Object owned by user and its contents
```
veluca93, line 1021: s/and its/whose/

lib/jxl/dec_external_image.cc
```diff
@@ -293,13 +293,17 @@ Status ConvertChannelsToExternal(const ImageF* channels[], size_t num_channels,
   const size_t bytes_per_pixel = num_channels * bytes_per_channel;
 
   std::vector<std::vector<uint8_t>> row_out_callback;
-  auto InitOutCallback = [&](size_t num_threads) {
-    if (out_callback) {
+  void* out_run_opaque = nullptr;
+  auto InitOutCallback = [&](size_t num_threads) -> Status {
+    if (out_init_callback && out_run_callback) {
```
veluca93, line 298: We very likely should assert that they are all null, or all non-null.

Passing a struct might help for that. (or a class with constructor + getters, but that gets more red tape-y)

lib/jxl/dec_external_image.cc
```diff
@@ -466,12 +465,15 @@ Status ConvertChannelsToExternal(const ImageF* channels[], size_t num_channels,
...
-          if (out_callback) {
-            (*out_callback)(out_opaque, 0, y, xsize, row_out);
+          if (out_run_callback) {
+            out_run_callback(out_run_opaque, thread, 0, y, xsize, row_out);
           }
         },
         "ConvertUint"));
   }
+  if (out_destroy_callback) {
+    out_destroy_callback(out_run_opaque);
```
veluca93, line 475: probably better to have this happen with RAII

lib/jxl/decode.cc
```diff
@@ -2646,19 +2667,49 @@ JxlDecoderStatus JxlDecoderSetImageOutCallback(JxlDecoder* dec,
                                                const JxlPixelFormat* format,
                                                JxlImageOutCallback callback,
                                                void* opaque) {
+  dec->simple_image_out_callback.callback = callback;
+  dec->simple_image_out_callback.opaque = opaque;
+  const auto init_callback =
+      +[](void* init_opaque, size_t num_threads, size_t num_pixels_per_thread) {
```
veluca93, line 2673: Why are we forcing a conversion to fn pointer?

lib/jxl/decode.cc
```diff
@@ -2646,19 +2667,49 @@ JxlDecoderStatus JxlDecoderSetImageOutCallback(JxlDecoder* dec,
...
+  const auto run_callback =
+      +[](void* run_opaque, size_t thread_id, size_t x, size_t y,
+          size_t num_pixels, const void* pixels) {
+        const auto* const simple_callback =
+            static_cast<const JxlDecoder::SimpleImageOutCallback*>(run_opaque);
+        simple_callback->callback(simple_callback->opaque, x, y, num_pixels,
+                                  pixels);
+      };
+  return JxlDecoderSetMultithreadedImageOutCallback(
+      dec, format, init_callback, run_callback, /*destroy_callback=*/nullptr,
```
veluca93, line 2686: I'd probably disallow making either of the callbacks a nullptr -- that sounds like something that could give us problems down the line.

lib/jxl/render_pipeline/stage_write.cc
```diff
@@ -269,24 +269,32 @@ class WriteToImage3FStage : public RenderPipelineStage {
...
         width_(width),
         height_(height),
         rgba_(rgba),
         has_alpha_(has_alpha),
         alpha_c_(alpha_c),
-        opaque_alpha_(kMaxPixelsPerCall, 1.0f) {
-    settings_.temp_buffer_size = kMaxPixelsPerCall * (rgba_ ? 4 : 3);
+        opaque_alpha_(kMaxPixelsPerCall, 1.0f) {}
+
+  ~WriteToPixelCallbackStage() override {
```
veluca93, line 289: We should disallow copy + assign then, and implement / disallow move

veluca93 (approved):

mo271 (commented):
Maybe already add a bullet point to the [CHANGELOG.md](https://github.com/libjxl/libjxl/blob/main/CHANGELOG.md), so that we don't forget about it...

lib/extras/enc/apng.cc
sboukortt: Sure. Only on the C++ side, right?

veluca93: Not in the API, if that's what you're asking :)

lib/jxl/decode.cc
sboukortt: Not sure, I have gotten into the habit of doing that when it is the end-goal, perhaps to get a compilation error as soon as possible in case of accidental capture. For what it’s worth, there are a few other places where we do this.

lib/include/jxl/decode.h
sboukortt: Not sure about this one, we might get some opposition: https://en.wikipedia.org/wiki/Inanimate_whose

veluca93: I was more objecting to "and its" rather than suggesting "whose" :P

sboukortt: Fair enough :p How is this? I also changed the function from which I copied it in the first place but there are a few more still.

veluca93 (dismissed):

veluca93 (approved):

Issue #1295, opened 2022-03-29
32 bits: pgx.cc:122: JXL_FAILURE: PGX: >16 bits not yet supported


malaterre:
**Describe the bug**

I'd like to use JPEG-XL to compress unsigned int (32bits). I tried using PGX inputs but it fails.

**To Reproduce**

```
% convert -depth 32 -size 512x512 xc:black black32.pgx
% ./tools/cjxl --num_threads=0 -d 0 black32.pgx black32.jxl
JPEG XL encoder v0.7.0 ff032c3 [AVX2,SSE4,SSSE3,Scalar]
./lib/extras/dec/pgx.cc:122: JXL_FAILURE: PGX: >16 bits not yet supported
./lib/extras/dec/exr.cc:93: JXL_FAILURE: OpenEXR failed to parse input
./lib/extras/dec/decode.cc:127: JXL_FAILURE: Codecs failed to decode
./lib/extras/codec.cc:49: JXL_FAILURE: Codecs failed to decode
Failed to read image black32.pgx.
```

Applying the following patch:

```
% git diff lib/extras/dec/pgx.cc
diff --git a/lib/extras/dec/pgx.cc b/lib/extras/dec/pgx.cc
index 7b79eaf..e319a36 100644
--- a/lib/extras/dec/pgx.cc
+++ b/lib/extras/dec/pgx.cc
@@ -117,9 +117,11 @@ class Parser {
     // 0xa, or 0xd 0xa.
     JXL_RETURN_IF_ERROR(SkipLineBreak());

+#if 0
     if (header->bits_per_sample > 16) {
       return JXL_FAILURE("PGX: >16 bits not yet supported");
     }
+#endif
     // TODO(lode): support signed integers. This may require changing the way
     // external_image works.
     if (header->is_signed) {
```

lead to:

```
JPEG XL encoder v0.7.0 ff032c3 [AVX2,SSE4,SSSE3,Scalar]
./lib/extras/dec/color_hints.cc:56: No color_space/icc_pathname given, assuming sRGB
Read 512x512 image, 107.9 MP/s
Encoding [Modular, lossless, squirrel], 0 threads.
./lib/jxl/image_metadata.cc:56: JXL_FAILURE: Invalid bits_per_sample: 32
./lib/jxl/fields.cc:595: JXL_ABORT: AllDefault should never fail
zsh: illegal hardware instruction  ./tools/cjxl --num_threads=0 -d 0 black32.pgx black32.jxl
```

with:

[...]

PR #1328, opened 2022-04-08
Remove strict requirement for 16bits since we support 24bits losslessly

malaterre:
Closes issue (#1295)


malaterre:
@mo271 sorry for being picky

jonsneyers (approved):

PR #1332, opened 2022-04-11
Provide a mechanism to override default testdata location

malaterre:
This will allow user to configure libjxl using an existing testdata
location:

```
% cmake -DJPEGXL_TEST_DATA_PATH:PATH=/usr/share/libjxl-testdata .
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -316,6 +316,10 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/LICENSE"
 # Enable tests regardless of where they are defined.
 enable_testing()
 include(CTest)
+# Specify default location of `testdata`:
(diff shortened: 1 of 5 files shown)


jonsneyers:
I don't object, but I wonder what's the use case for this. When distributing libjxl, surely you're not going to also distribute the testdata and/or the tests?

malaterre:
@jonsneyers I can now run the full test suite on Debian archs (including big-endian) using simply:

* https://salsa.debian.org/debian-phototools-team/libjxl/-/blob/master/debian/rules#L57

Before that I had to include the testdata (19M) within each release (even minor) of libjxl. See recent work from @mo271.

jonsneyers (approved):

PR #1333, opened 2022-04-11
Change position bytes to codestream bytes

jyrkialakuijala:
Frame index box needs the count of bytes in codestream box(es),
not the absolute byte position in the resulting file.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -252,6 +252,12 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
     // TODO(lode): preview should be added here if a preview image is added
 
     writer.ZeroPadToByte();
+
+    // Not actually the end of frame, but the end of metadata/ICC, but helps
+    // the next frame to start here for indexing purposes.
+    codestream_bytes_written_end_of_frame +=
+        jxl::DivCeil(writer.BitsWritten(), 8);
+
     bytes = std::move(writer).TakeBytes();
 
     if (MustUseContainer()) {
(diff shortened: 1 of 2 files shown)


lib/jxl/encode_internal.h
```diff
@@ -117,12 +117,15 @@ struct JxlEncoderStruct {
...
   std::vector<jxl::JxlEncoderQueuedInput> input_queue;
   std::deque<uint8_t> output_byte_queue;
-  size_t output_bytes_flushed;
 
-  // Get the current write position in the stream (for indexing use).
-  size_t BytePosition() const {
-    return output_bytes_flushed + output_byte_queue.size();
-  }
+  // How many codestream bytes have been written, i.e.,
+  // content of jxlc and jxlj boxes. Frame index box jxli
```
lvandeve, line 122: I think jxlj should be jxlp (also below), the codestream is jxlc and jxlp boxes

lib/jxl/encode.cc
```diff
@@ -387,17 +387,20 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
...
+    JXL_ASSERT(writer.BitsWritten() == 0);
     if (!jxl::EncodeFrame(input_frame->option_values.cparams, frame_info,
                           &metadata, input_frame->frame, &enc_state, cms,
                           thread_pool.get(), &writer,
                           /*aux_out=*/nullptr)) {
       return JXL_API_ERROR("Failed to encode frame");
     }
+    codestream_bytes_written_beginning_of_frame =
+        codestream_bytes_written_end_of_frame;
+    codestream_bytes_written_end_of_frame += (writer.BitsWritten() + 7) / 8;
```
lvandeve, line 399: nit: could use jxl::DivCeil(writer.BitsWritten(), 8) here since it's the convention in the rest of the code

lib/jxl/encode.cc
```diff
@@ -387,17 +387,20 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
...
       ib.origin.y0 = input_frame->option_values.header.layer_info.crop_y0;
     }
+    JXL_ASSERT(writer.BitsWritten() == 0);
     if (!jxl::EncodeFrame(input_frame->option_values.cparams, frame_info,
                           &metadata, input_frame->frame, &enc_state, cms,
                           thread_pool.get(), &writer,
                           /*aux_out=*/nullptr)) {
       return JXL_API_ERROR("Failed to encode frame");
     }
+    codestream_bytes_written_beginning_of_frame =
```
lvandeve, line 397: As far as I can tell, it's now only counting the bytes produced by EncodeFrame, but not including the ones produced by WriteHeaders as well as WriteICC higher up, so updating the codestream bytes pos is also needed there.

I think that should be all (anything that outputs jxlc or jxlp), but it may be worth double checking. A preview frame also goes to the codestream should also be counted, but preview image is not supported yet by the encoder so that can't be done yet.

lib/jxl/encode_internal.h
jyrkialakuijala: fixed

lib/jxl/encode.cc
jyrkialakuijala: fixed

lib/jxl/encode.cc
jyrkialakuijala: added

PR #1337, opened 2022-04-12
deprecate JXL_TYPE_BOOLEAN and JXL_TYPE_UINT32

jonsneyers:
These two sample types are in the enum but were never fully implemented so they couldn't actually be used by applications. It probably is not a great idea to implement them either, for the following reasons:

- Supporting UINT32 as a sample type at the API boundary can give the misleading impression that JPEG XL can losslessly store uint32 samples, which is not true. The spec limits integer bitdepths to what fits in int32 (so 31-bit uint), and the current libjxl implementation uses internal float representations so it is in practice limited to 24-bit uint.
- Saving memory for 1-bit images by having a bit-packed representation at the API boundary is a somewhat futile exercise when we're still internally using 32-bit floats to represent each bit. It's also somewhat arbitrary to support bit packing specifically for the case of 1-bit, but not for e.g. 2-bit or 4-bit, or for e.g. fitting 10-bit RGB in 4 bytes. No bit packing at all keeps things simple. Also the code was assuming that 1-bit is by definition bit-packed, while e.g. for the `pgm`/`ppm` formats, you need non-bitpacked 1-bit (i.e. bytes that are 0 or 1).

This ~~deprecates~~ removes these two sample types and removes the partial implementations.

(original plan was to keep them around in the enum and just deprecate them, but the documentation generator could not handle that, see https://github.com/sphinx-doc/sphinx/issues/10341, so just removing them is the easiest solution — note that no application could actually be really using these, since it would lead to runtime errors in both encoder and decoder)

### DIFF
--- a/lib/extras/dec/pgx.cc
+++ b/lib/extras/dec/pgx.cc
@@ -117,12 +117,10 @@ class Parser {
     // 0xa, or 0xd 0xa.
     JXL_RETURN_IF_ERROR(SkipLineBreak());
 
-    // libjxl currently does use float buffers internally for most parts of the
-    // codec so the actual precision limit for lossless integer encoding is
-    // 24-bit (higher than that starts losing least significant bits in the
-    // conversion to/from float).
-    if (header->bits_per_sample > 24) {
-      return JXL_FAILURE("PGX: >24 bits not yet supported");
+    // TODO(jon): could do up to 24-bit by converting the values to
+    // JXL_TYPE_FLOAT.
+    if (header->bits_per_sample > 16) {
+      return JXL_FAILURE("PGX: >16 bits not yet supported");
     }
     // TODO(lode): support signed integers. This may require changing the way
     // external_image works.
@@ -131,9 +129,7 @@ class Parser {
     }
 
     size_t numpixels = header->xsize * header->ysize;
-    size_t bytes_per_pixel = header->bits_per_sample <= 8    ? 1
-                             : header->bits_per_sample <= 16 ? 2
-                                                             : 4;
+    size_t bytes_per_pixel = header->bits_per_sample <= 8 ? 1 : 2;
     if (pos_ + numpixels * bytes_per_pixel > end_) {
       return JXL_FAILURE("PGX: data too small");
     }
@@ -178,9 +174,7 @@ Status DecodeImagePGX(const Span<const uint8_t> bytes,
   ppf->info.orientation = JXL_ORIENT_IDENTITY;
 
   JxlDataType data_type;
-  if (header.bits_per_sample > 16) {
-    data_type = JXL_TYPE_UINT32;
-  } else if (header.bits_per_sample > 8) {
+  if (header.bits_per_sample > 8) {
     data_type = JXL_TYPE_UINT16;
   } else {
     data_type = JXL_TYPE_UINT8;
--- a/lib/extras/dec/pnm.cc
+++ b/lib/extras/dec/pnm.cc
@@ -19,7 +19,6 @@ namespace {
 struct HeaderPNM {
   size_t xsize;
   size_t ysize;
-  bool is_bit;     // PBM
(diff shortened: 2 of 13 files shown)


malaterre:
```
 % git diff -- lib/extras/enc/pnm.cc
diff --git a/lib/extras/enc/pnm.cc b/lib/extras/enc/pnm.cc
index 035ad54..e723a92 100644
--- a/lib/extras/enc/pnm.cc
+++ b/lib/extras/enc/pnm.cc
@@ -50,15 +50,6 @@ Status EncodeHeader(const PackedPixelFile& ppf, const size_t bits_per_sample,
                  type, oriented_xsize, oriented_ysize, scale);
     JXL_RETURN_IF_ERROR(static_cast<unsigned int>(*chars_written) <
                         kMaxHeaderSize);
-  } else if (bits_per_sample == 1) {  // PBM
-    if (is_gray) {
-      return JXL_FAILURE("Cannot encode color as PBM");
-    }
-    *chars_written =
-        snprintf(header, kMaxHeaderSize, "P4\n%" PRIuS " %" PRIuS "\n",
-                 oriented_xsize, oriented_ysize);
-    JXL_RETURN_IF_ERROR(static_cast<unsigned int>(*chars_written) <
-                        kMaxHeaderSize);
   } else {  // PGM/PPM
     const uint32_t max_val = (1U << bits_per_sample) - 1;
     if (max_val >= 65536) return JXL_FAILURE("PNM cannot have > 16 bits");
```

jonsneyers:
> Use case:
> 
> ```
> % convert -depth 1 -size 512x512 xc:white white.pgm && ./tools/cjxl -d 0 white.pgm white.jxl && ./tools/djxl white.jxl white.jxl.pgm
> ```
> 
> ```
>  % git diff -- lib/extras/enc/pnm.cc
> diff --git a/lib/extras/enc/pnm.cc b/lib/extras/enc/pnm.cc
> index 035ad54..e723a92 100644
> --- a/lib/extras/enc/pnm.cc
> +++ b/lib/extras/enc/pnm.cc
> @@ -50,15 +50,6 @@ Status EncodeHeader(const PackedPixelFile& ppf, const size_t bits_per_sample,
>                   type, oriented_xsize, oriented_ysize, scale);
>      JXL_RETURN_IF_ERROR(static_cast<unsigned int>(*chars_written) <
>                          kMaxHeaderSize);
> -  } else if (bits_per_sample == 1) {  // PBM
> -    if (is_gray) {
> -      return JXL_FAILURE("Cannot encode color as PBM");
> -    }
> -    *chars_written =
> -        snprintf(header, kMaxHeaderSize, "P4\n%" PRIuS " %" PRIuS "\n",
> -                 oriented_xsize, oriented_ysize);
> -    JXL_RETURN_IF_ERROR(static_cast<unsigned int>(*chars_written) <
> -                        kMaxHeaderSize);
>    } else {  // PGM/PPM
>      const uint32_t max_val = (1U << bits_per_sample) - 1;
>      if (max_val >= 65536) return JXL_FAILURE("PNM cannot have > 16 bits");
> ```

Addressing this in https://github.com/libjxl/libjxl/pull/1345.

veluca93:
I'm not sure I like the "remove because the doc doesn't handle it" approach - could we perhaps instead make them global `static const int`s with a deprecated attribute?

jonsneyers:
> I'm not sure I like the "remove because the doc doesn't handle it" approach - could we perhaps instead make them global `static const int`s with a deprecated attribute?

I'll do that. I don't think it makes a big difference in practice though, since the removed types didn't work anyway so if anyone had code that used it, it wouldn't work — it would compile, but not run. But adding the deprecated consts would allow them to keep compiling such broken code and/or get a somewhat more informative compiler error/warning.

veluca93 (dismissed):

veluca93 (approved):

PR #1339, opened 2022-04-12
Fix aarc64 write stage.

eustas:
"extra" should be the number of pixels to group boundary, not image edge.


lib/jxl/render_pipeline/stage_write.cc
```diff
@@ -85,7 +85,9 @@ class WriteToU8Stage : public RenderPipelineStage {
         height_(height),
         rgba_(rgba),
         has_alpha_(has_alpha),
-        alpha_c_(alpha_c) {}
+        alpha_c_(alpha_c) {
+    (void)width_;  // unused
```
veluca93, line 89: Should we remove the field instead then?

eustas: Sure.

veluca93 (approved):

veluca93 (approved):

PR #1341, opened 2022-04-12
relax high bitdepth palette restriction and improve lossless float compression

### LINKED ISSUE #1323: Poor compression ratio for 32bits float
**Describe the bug**

I'd like to achieve good compression ratio using `-e 9` using cjxl command line tool with pfm (rgb) input.

Here is the entropy plot (-d 0 -e 9):

sample_1920×1280 pfm jxl

Compared to a zstd (no option) :

sample_1920×1280 pfm zstd

or gzip (no option):

sample_1920×1280 pfm gz

**To Reproduce**
```
% wget https://telparia.com/fileFormatSamples/image/pfm/sample_1920×1280.pfm
% sha1sum sample_1920×1280.pfm
bd579ee0144d6d3dd4495de3cb95d2283fd6c0e0  sample_1920×1280.pfm
% time cjxl -d 0 -e 9 sample_1920×1280.pfm sample_1920×1280.pfm.9.jxl
JPEG XL encoder v0.7.0 e3a3558 [AVX2,SSE4,SSSE3,Scalar]
./lib/extras/dec/color_hints.cc:56: No color_space/icc_pathname given, assuming sRGB
Read 1920x1280 image, 107.4 MP/s
Encoding [Modular, lossless, tortoise], 8 threads.
Compressed to 13425615 bytes (43.703 bpp).
1920 x 1280, 0.01 MP/s [0.01, 0.01], 1 reps, 8 threads.
cjxl -d 0 -e 9 sample_1920×1280.pfm sample_1920×1280.pfm.9.jxl  197.76s user 0.20s system 102% cpu 3:12.67 total
```
For comparison:
```
% time zstd -o sample_1920×1280.pfm.zstd sample_1920×1280.pfm
sample_1920×1280.pfm : 30.04%   (29491217 => 8858010 bytes, sample_1920×1280.pfm.zstd)
zstd -o sample_1920×1280.pfm.zstd sample_1920×1280.pfm  0.19s user 0.02s system 106% cpu 0.198 total
```
```
% time gzip -c  sample_1920×1280.pfm >  sample_1920×1280.pfm.gz
gzip -c sample_1920×1280.pfm > sample_1920×1280.pfm.gz  1.91s user 0.01s system 99% cpu 1.920 total

```
**Expected behavior**

I'd like to have a flat entropy histogram on pfm inputs (32bits float). Here is a typical plot for uint16 (grayscale pgm original input):

MG jxl

**Screenshots**

See above

**Environment**
 - OS: Linux
 - Compiler version: gcc-10
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.7.0 e3a3558 [AVX2,SSE4,SSSE3,Scalar]

**Additional context**

To generate plot I used:

```
% binwalk -EJ <input file>
```

jonsneyers:
Fixes #1323 

Currently we're not allowing Palette when bitdepth is > 24, but this is a bit overly restrictive. Instead, let's not have such a restriction, but just don't do default delta scaling in this case.

With that restriction relaxed, the encoder can freely use channel palette also in case of float32 images. This PR adds an encode codepath for the case where the lookup table would be too large, doing it the slower way instead.

Also, RCTs are currently not tried in case of float32 (since bitdepth is 32 so there are no spare bits for it), but in case channel palette reduces the effective bitdepth, they can actually be used without issues.

Before, default speed:
Compressed to 16111054 bytes (52.445 bpp).
1920 x 1280, 0.56 MP/s [0.56, 0.56], 1 reps, 4 threads.

After, default speed:
Compressed to 2807063 bytes (9.138 bpp).
1920 x 1280, 0.36 MP/s [0.36, 0.36], 1 reps, 4 threads.

Before, -e 3:
Compressed to 22134261 bytes (72.052 bpp).
1920 x 1280, 9.48 MP/s [9.48, 9.48], 1 reps, 4 threads.

After, -e 3:
Compressed to 3229866 bytes (10.514 bpp).
1920 x 1280, 2.26 MP/s [2.26, 2.26], 1 reps, 4 threads.

Encoding obviously gets slower — transforms are not just skipped, and the implementation of high-bitdepth forward channel palette is quite naive/slow) — but I think the huge difference in compression ratio justifies that.

### DIFF
--- a/lib/jxl/modular/transform/palette.h
+++ b/lib/jxl/modular/transform/palette.h
@@ -84,7 +84,7 @@ static pixel_type GetPaletteValue(const pixel_type *const palette, int index,
     pixel_type result =
         kDeltaPalette[((index + 1) >> 1)][c] * kMultiplier[index & 1];
     if (bit_depth > 8) {
-      result *= static_cast<pixel_type>(1) << (bit_depth - 8);
+      result *= static_cast<pixel_type>(1) << (std::min(bit_depth, 24) - 8);
     }
     return result;
   } else if (palette_size <= index && index < palette_size + kLargeCubeOffset) {
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -675,6 +675,8 @@ Status ModularFrameEncoder::ComputeEncodingData(
        (do_color && metadata.bit_depth.bits_per_sample > 8))) {
     // single channel palette (like FLIF's ChannelCompact)
     size_t nb_channels = gi.channel.size() - gi.nb_meta_channels;
+    int orig_bitdepth = max_bitdepth;
+    max_bitdepth = 0;
     for (size_t i = 0; i < nb_channels; i++) {
       int min, max;
       compute_minmax(gi.channel[gi.nb_meta_channels + i], &min, &max);
@@ -694,7 +696,11 @@ Status ModularFrameEncoder::ComputeEncodingData(
         // effective bit depth is lower, adjust quantization accordingly
         compute_minmax(gi.channel[gi.nb_meta_channels + i], &min, &max);
         if (max < maxval) maxval = max;
-      }
+        int ch_bitdepth =
+            (max > 0 ? CeilLog2Nonzero(static_cast<uint32_t>(max)) : 0);
+        if (ch_bitdepth > max_bitdepth) max_bitdepth = ch_bitdepth;
+      } else
+        max_bitdepth = orig_bitdepth;
     }
   }
 
@@ -738,7 +744,7 @@ Status ModularFrameEncoder::ComputeEncodingData(
   }
 
   // don't do an RCT if we're short on bits
-  if (cparams.color_transform == ColorTransform::kNone && do_color && !fp &&
+  if (cparams.color_transform == ColorTransform::kNone && do_color &&
       gi.channel.size() - gi.nb_meta_channels >= 3 &&
       max_bitdepth + 1 < level_max_bitdepth) {
     if (cparams.colorspace == 1 ||
--- a/lib/jxl/modular/transform/enc_palette.cc
+++ b/lib/jxl/modular/transform/enc_palette.cc
@@ -179,7 +179,43 @@ Status FwdPaletteIteration(Image &input, uint32_t begin_c, uint32_t end_c,
     size_t lookup_table_size =
         static_cast<int64_t>(maxval) - static_cast<int64_t>(minval) + 1;
     if (lookup_table_size > palette_internal::kMaxPaletteLookupTableSize) {
-      return false;  // too large lookup table
+      // a lookup table would use too much memory, instead use a slower approach
+      // with std::set
+      std::set<pixel_type> chpalette;
+      pixel_type idx = 0;
+      for (size_t y = 0; y < h; y++) {
+        const pixel_type *p = input.channel[begin_c].Row(y);
+        for (size_t x = 0; x < w; x++) {
+          const bool new_color = chpalette.insert(p[x]).second;
+          if (new_color) {
+            idx++;
+            if (idx > (int)nb_colors) return false;
+          }
+        }
+      }
+      JXL_DEBUG_V(6, "Channel %i uses only %i colors.", begin_c, idx);
+      Channel pch(idx, 1);
+      pch.hshift = -1;
+      nb_colors = idx;
+      idx = 0;
+      pixel_type *JXL_RESTRICT p_palette = pch.Row(0);
+      for (pixel_type p : chpalette) {
+        p_palette[idx++] = p;
+      }
+      for (size_t y = 0; y < h; y++) {
+        pixel_type *p = input.channel[begin_c].Row(y);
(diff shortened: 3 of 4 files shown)


malaterre:
Just for later reference, this patch will break backward compatibility. 2cts.

```
% djxl sample_1920×1280.pfm.9.jxl sample_1920×1280.pfm.9.jxl.pfm
JPEG XL decoder v0.7.0 e3a3558 [AVX2,SSE4,SSSE3,Scalar]
Read 2807063 compressed bytes.
./lib/jxl/modular/transform/transform.cc:72: JXL_FAILURE: Palette is not allowed for bitdepth > 24
./lib/jxl/modular/encoding/encoding.cc:441: JXL_RETURN_IF_ERROR code=1: transform.MetaApply(image)
./lib/jxl/modular/encoding/encoding.cc:563: JXL_RETURN_IF_ERROR code=1: dec_status
./lib/jxl/dec_modular.cc:229: JXL_RETURN_IF_ERROR code=1: dec_status
./lib/jxl/dec_frame.cc:191: JXL_RETURN_IF_ERROR code=1: frame_decoder.ProcessSections( section_info.data(), section_info.size(), section_status.data())
./lib/jxl/dec_file.cc:140: JXL_RETURN_IF_ERROR code=1: dec_ok
Failed to decompress to pixels.
```

jonsneyers:
> Just for later reference, this patch will break backward compatibility. 2cts.
> 
> ```
> % djxl sample_1920×1280.pfm.9.jxl sample_1920×1280.pfm.9.jxl.pfm
> JPEG XL decoder v0.7.0 e3a3558 [AVX2,SSE4,SSSE3,Scalar]
> Read 2807063 compressed bytes.
> ./lib/jxl/modular/transform/transform.cc:72: JXL_FAILURE: Palette is not allowed for bitdepth > 24
> ./lib/jxl/modular/encoding/encoding.cc:441: JXL_RETURN_IF_ERROR code=1: transform.MetaApply(image)
> ./lib/jxl/modular/encoding/encoding.cc:563: JXL_RETURN_IF_ERROR code=1: dec_status
> ./lib/jxl/dec_modular.cc:229: JXL_RETURN_IF_ERROR code=1: dec_status
> ./lib/jxl/dec_frame.cc:191: JXL_RETURN_IF_ERROR code=1: frame_decoder.ProcessSections( section_info.data(), section_info.size(), section_status.data())
> ./lib/jxl/dec_file.cc:140: JXL_RETURN_IF_ERROR code=1: dec_ok
> Failed to decompress to pixels.
> ```

It indeed will not decode with versions between this commit: https://github.com/libjxl/libjxl/pull/1155 and the current one which removes that check. The previously released version (0.6.1) should decode it though.

lib/jxl/modular/transform/palette.h
```diff
@@ -83,7 +83,7 @@ static pixel_type GetPaletteValue(const pixel_type *const palette, int index,
     static constexpr int kMultiplier[] = {-1, 1};
     pixel_type result =
         kDeltaPalette[((index + 1) >> 1)][c] * kMultiplier[index & 1];
-    if (bit_depth > 8) {
+    if (bit_depth > 8 && bit_depth <= 24) {
```
veluca93, line 86: What if instead we replace the bit_depth value with std::max(bit_depth, 24)?

jonsneyers: Yes, that would also work (std::min though).

Another, maybe more radical option would be to just remove the bitdepth scaling altogether in the default deltas. I don't really see why it is useful in the first place. It basically causes high bit depth delta palette to just boil down to 8-bit delta palette — unless of course you add custom deltas — but that doesn't really seem to be the most desirable thing to do. If you want to use delta palette on a 10-bit image, then maybe you actually want to have default deltas with smaller steps, so slow gradients can be done more accurately.

For the default absolute colors (the 4x4x4 cube and the 5x5x5 cube), bitdepth scaling does of course make sense. But for the default deltas, I think it actually doesn't really make sense.

@jyrkialakuijala what's your take on this? So what I'm proposing is to just remove the bitdepth scaling completely from the default delta definition, i.e. the current line `if (bitdepth > 8) value <<= bitdepth - 8;` in the spec. By removing that line, you get default deltas that get increasingly more fine-grained as the bitdepth increases, and at some point you'll probably need to add non-default deltas to make effective use of delta palette, but I think that makes more sense than scaling the defaults so it's always as if things are 8-bit.

jonsneyers: OK, in that case, shall we change it into this?
```
if (bitdepth > 8) value <<= std::min(bitdepth,24) - 8;
```

So keep the scaling as it is now for 9..24-bit, and for higher bitdepth, clamp the scaling to what is done for 24-bit (as opposed to not allowing the (delta) palette transform at all).

jonsneyers: The above is how it currently is in the CD of 18181-1-Ed2.

jyrkialakuijala: Perfect

jyrkialakuijala (approved):

PR #1343, opened 2022-04-13
Prepare for EMU128 target: avoid x86-specific code

jan-wassenberg:


lib/jxl/convolve-inl.h
```diff
@@ -44,7 +44,7 @@ class Neighbors {
     return c;  // Same (the first mirrored value is the last valid one)
 #else  // 128 bit
     // c = LKJI
-#if HWY_ARCH_X86
+#if HWY_TARGET <= (1 << HWY_HIGHEST_TARGET_BIT_X86)
```
eustas, line 47: Just an idea: perhaps we should have macro for expression like that (i.e. the one that includes comparison and the second operand).

eustas (approved):

lib/jxl/convolve-inl.h
jan-wassenberg: Thanks for suggesting it. hm, probably best if we don't encourage this stye of x86-dependent code. (Using such a macro would also require a Highway release.)

eustas (approved):

PR #1346, opened 2022-04-14
promote jxlinfo to installed tool

jonsneyers:
Currently `jxlinfo` is in `examples`, i.e. it is "only an example application". It is however a useful tool, so it makes sense to include it in `tools` and install it by default. Other image formats also have tools like this (e.g. `pnginfo`, `webpinfo`).


tools/CMakeLists.txt
```diff
@@ -164,32 +165,45 @@ if(JPEGXL_ENABLE_TOOLS)
...
     djxl.cc
     speed_stats.cc
   )
   target_link_libraries(djxltool
-      box
-      jxl-static
-      jxl_extras-static
-      jxl_threads-static
+    box
+    jxl-static
```
veluca93, line 185: Uh, that doesn't seem right.

jonsneyers: I didn't change anything except the indentation...

veluca93 (approved):

Issue #1330, opened 2022-04-10
cjpeg: "APNG chunk size is too big" when trying to compress very large (142.1Mpix) png image


RarogCmex:
**Describe the bug**
It seems like built'in decoder can't compress some type of large files
**To Reproduce**
```# cjxl sample.png compressed.jxl```
_gives:_
```JPEG XL encoder v0.7.0 [SSE4]
./lib/extras/dec/apng.cc:298: APNG chunk size is too big
./lib/extras/dec/color_hints.cc:56: No color_space/icc_pathname given, assuming sRGB
./lib/extras/dec/decode.cc:127: JXL_FAILURE: Codecs failed to decode
./lib/extras/codec.cc:49: JXL_FAILURE: Codecs failed to decode
Failed to read image comment_1620718615PMoN9rgOE6Az4AXDSibg8e.png.
```
**Expected behavior**
It should be compressed as usual.

**Environment**
 - OS: Gentoo Linux
 - Compiler version: gcc 11.2.0
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.7.0 [SSE4] / version 0.7.0_pre20220329

**Additional context**
Sample image too big to attach it directly, I'll upload it to Yandex Disk (can contain NSFW parts):
https://disk.yandex.ru/d/Aj7gv1NKTWvTYw
```sample_cjxl_bug.png: PNG image data, 10032 x 14164, 8-bit/color RGB, non-interlaced```
Please say if you got downloading issues.

**Comment by TheDecryptor** [comment]
If I run the image through imagemagick first (just `convert sample_cjxl_bug.png sample_cjxl_bug.2.png`) then cjxl encodes it without issue, it's just something about the exact source file that's causing a problem

**Comment by RarogCmex** [comment]
Yes. It seems like built'in cjxl decoder can't compress some type of large files

**Comment by RarogCmex** [comment]
Also with cjxl_ng:
```
cjxl_ng sample_cjxl_bug.png output.jxl
Warning: This is work in progress, consider using cjxl instead!
./lib/extras/dec/apng.cc:298: APNG chunk size is too big
./lib/extras/dec/color_hints.cc:56: No color_space/icc_pathname given, assuming sRGB
./tools/cjxl_ng_main.cc:379: JXL_FAILURE: Codecs failed to decode input.
Loading input file failed.

```

jonsneyers:
[...]

PR #1349, opened 2022-04-15
allow bigger png chunks

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/1330

Bumps up the maximum png chunk size from 100 MB to 1 GB.


eustas (approved):
See no problem with bigger input (chunk) size.

Issue #949, opened 2021-12-05
Reference API generates corrupt JXL files that cannot be decoded to JPEG anymore.


schmorp:
**Describe the bug**

While cjxl, which uses an undocumented API, can create jxl files that allow jpeg reconstruction, the official (C) API currently (commit ea2612d6df99e9878b51e315935f9d6201f5fe47) generates corrupt files that can be viewed, but fail to reconstruct to the original JPEG files.

**To Reproduce**

I have an encoder (not written in C) that basically does this:

JxlEncoderCreate
JxlEncoderUseContainer
JxlEncoderStoreJPEGMetadata
JxlEncoderFrameSettingsCreate
JxlEncoderSetFrameLossless
JxlEncoderAddJPEGFrame
JxlEncoderCloseInput

That is, it doesn't set any options and relies on AddJPEGFrame to set defaults. The expectation is that it is possibe to reconstruct the original JPEG from the encoded file. The resulting jxl file is 283kb large and has the following structure:

box: type: "JXL " size: 12
box: type: "ftyp" size: 20
box: type: "jxlp" size: 22
box: type: "jbrd" size: 284
box: type: "jxlp" size: 282464

The image displays fine, i.e. it can be (pixel-)decoded without problems both by djxl as well as by my own image viewer that uses the reference API.

Trying to reconstruct the original JPEG with the C api fails with status 2, while trying to reconstruct using djxl fails with:

Read 282802 compressed bytes.
Unknown compressed image format

Debugging the problem, it seems the Exif information is missing. Indeed, if I use cjxl, I get a different, larger file (316k), with the following structure, which reconstructs fine:

box: type: "JXL " size: 12
box: type: "ftyp" size: 20
box: type: "Exif" size: 8719
box: type: "jbrd" size: 284
box: type: "jxlc" size: 282470

The original JPEG has the following structure:

 SOI*(start_of_image) @0x00000000
 APP0 @0x00000002
 APP1 @0x00000014
 APPE @0x00000148
 DQT(quantization_table) @0x00000158
 SOF0(huff_baseline) @0x0000019d
 DRI(restart_interval) @0x000001aa
 DHT(huff_table) @0x000001b0
 SOS(start_of_scan) @0x00000284
 EOI*(end_of_image) @0x00054398

**Expected behavior**

1. JPEG reconstruction should not fail for files encoded by the reference implementation.
[...]

PR #1350, opened 2022-04-15
jpeg recompression: preserve exif/xmp/jumbf

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/949.

Currently, AddJPEGFrame in the encode api silently strips Exif/XMP/JUMBF blobs that may exist in the JPEG bitstream. That causes JPEG reconstruction to fail, and is also just a bad thing to do even if JPEG reconstruction is not desired.

This PR makes it instead always preserve such metadata. It currently stores the metadata using `brob`, which seems to work fine with the decode api (e.g. `djxl_ng`), though it doesn't work when decoding with `djxl` (which has its own partial implementation of the container format, which does not include handling `brob`). It's probably not worth it to fix `djxl` at this point.

We could later add an option to strip the data instead (only useful when `jbrd` is stripped too), and/or to store it uncompressed.

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -1167,7 +1167,27 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
     // Can't XYB encode a lossless JPEG.
     return JXL_ENC_ERROR;
   }
-
+  if (!io.blobs.exif.empty()) {
+    size_t exif_size = io.blobs.exif.size();
+    // Exif data in JPEG is limited to 64k
+    if (exif_size > 0xFFFF) return JXL_ENC_ERROR;
+    exif_size += 4;  // prefix 4 zero bytes for tiff offset
+    std::vector<uint8_t> exif(exif_size);
+    memcpy(exif.data() + 4, io.blobs.exif.data(), io.blobs.exif.size());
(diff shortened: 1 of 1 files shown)


lib/jxl/encode.cc
```diff
@@ -1167,7 +1167,27 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  if (!io.blobs.exif.empty()) {
+    size_t exif_size = io.blobs.exif.size();
+    // Exif data in JPEG is limited to 64k
+    if (exif_size > 0xFFFF) return JXL_ENC_ERROR;
+    exif_size += 4;  // prefix 4 zero bytes for tiff offset
+    std::vector<uint8_t> exif(exif_size);
+    memcpy(exif.data() + 4, io.blobs.exif.data(), io.blobs.exif.size());
+    JxlEncoderUseBoxes(frame_settings->enc);
+    JxlEncoderAddBox(frame_settings->enc, "Exif", exif.data(), exif_size,
+                     JXL_TRUE);
```
veluca93, line 1179: /*brotli=*/ or whatever the parameter name is :)

jonsneyers: updated.

veluca93 (approved):

PR #1357, opened 2022-04-19
Do not render padding if there is nothing to pad

eustas:


eustas:
This is not the first time. Going to check if we can stick to some "stable" distro.

PR #1358, opened 2022-04-20
Avoid (temporary?) integer overflow

eustas:
Drive-by: simplify division by 2:
https://godbolt.org/z/sq91EMKT8

### DIFF
--- a/lib/jxl/modular/transform/squeeze.cc
+++ b/lib/jxl/modular/transform/squeeze.cc
@@ -119,8 +119,7 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
       pixel_type left = (x ? p_out[(x << 1) - 1] : avg);
       pixel_type tendency = SmoothTendency(left, avg, next_avg);
       pixel_type diff = diff_minus_tendency + tendency;
-      pixel_type A =
-          ((avg * 2) + diff + (diff > 0 ? -(diff & 1) : (diff & 1))) >> 1;
+      pixel_type A = avg + (diff / 2);
       p_out[(x << 1)] = A;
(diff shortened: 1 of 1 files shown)


veluca93:
@jonsneyers to double-check if this simplification can be applied to the SIMD code too

lib/jxl/modular/transform/squeeze.cc
```diff
@@ -119,8 +119,7 @@ Status InvHSqueeze(Image &input, uint32_t c, uint32_t rc, ThreadPool *pool) {
       pixel_type left = (x ? p_out[(x << 1) - 1] : avg);
       pixel_type tendency = SmoothTendency(left, avg, next_avg);
       pixel_type diff = diff_minus_tendency + tendency;
-      pixel_type A =
-          ((avg * 2) + diff + (diff > 0 ? -(diff & 1) : (diff & 1))) >> 1;
+      pixel_type A = avg + (diff / 2);
```
jonsneyers, line 122: There's a similar line in InvVSqueeze, can do the same simplification there.

eustas: Sure. One moment.

jonsneyers:
Wow, good find! How did I ever end up with that complicated expression if it just boils down to `avg + diff/2`?

eustas:
I did that in a thought experiment. Both addends (`avg * 2` and "fixed" diff) have the rightmost bit set to 0 -> their sum have the rightmost bit set to 0 as well -> likely no surprises. But going to double-check soon.

jonsneyers:
We should also clean this up in the spec, and probably the SIMD code can be made slightly faster by doing the simplification also there.

Issue #541, opened 2021-09-04
Awfully Bad Rounding Errors when Decoding Lossy JXL (Dithering Needed!)


Matt-Gore:
The decoder of libjxl produces ugly rounding errors when decoding a lossy 8 bpc JXL, which results in visible banding.

We will take those two images as examples:
![preview.webp](https://res.cloudinary.com/mattgore/preview.webp)
(full resolution versions found [here](https://res.cloudinary.com/mattgore/063.webp) and [there](https://res.cloudinary.com/mattgore/064.webp))

This is how (a part) of the image looks after encoding it "visually lossless" (-d 1) and decoding it back to 8 bpc pixels:
![064_crop_bad.webp](https://res.cloudinary.com/mattgore/064_crop_bad.webp)

Notice how wrong it looks? Here's an exaggerated view of the situation (that could happen on a cheap display or to people sitting in a dark room):
![064_crop_bad_extra.webp](https://res.cloudinary.com/mattgore/064_crop_bad_extra.webp)

You might ask yourself how this visual bug could happen, because if you look at the [lossless original picture](https://res.cloudinary.com/mattgore/064.webp) (which is also just 8 bpc), everything seems fine. But don't worry, we don't need to throw away JXL since it all can be fixed!
So, JPEG XL uses DCTs for lossy images (by default) and very small details like the patterns of noise are not considered worthy or important to store when using "visually lossless" quality. It thus leads to simpler DCTs, however, they can still represent a smooth gradient. The actual issue happens at a late stage of decoding. JPEG XL uses 32 Bit floating point precision to operate as accurate as possible but when outputting to 8 bpc pixels again, it messes up by just dropping essential information and doing very inaccurate and unnecessarily simple rounding. Instead, quantization of the float values should utilize a brand new, still rather unknown discovery in digital signal processing called DITHERING! Jokes aside, but a quick and easy improvement with a huge benefit would be doing something like applying 4x4 ordered dither. (This would be better than just having to add [photon] noise.)

That's how great it would look with dither:
![064_crop_good.webp](https://res.cloudinary.com/mattgore/064_crop_good.webp)

Which is orders of magnitudes better than what is currently been done. And causes no noticeable banding:
![064_crop_good_extra.webp](https://res.cloudinary.com/mattgore/064_crop_good_extra.webp)

Here is the other example (same picture order):
![063_crop_bad.webp](https://res.cloudinary.com/mattgore/063_crop_bad.webp)
[...]

PR #1362, opened 2022-04-20
apply 8x8 ordered dither to uint decode output

jonsneyers:
When converting the internal float sample values to uint ones (when JXL_TYPE_UINT8 or JXL_TYPE_UINT16 pixel formats are requested), currently this is done simply by picking the nearest integer value.  This can be a problem, especially with 8-bit output on sufficiently bright displays, as noted by @Matt-Gore in https://github.com/libjxl/libjxl/issues/541. It can lead to quite noticeable banding in slow dark gradients.

This adds 8x8 ordered dithering when converting to uint output.  Even just 2x2 or 4x4 dithering already alleviates most of the banding issues, but 8x8 is still very cheap. I couldn't measure any speed difference compared to non-dithered.

### DIFF
--- a/lib/jxl/dec_cache.h
+++ b/lib/jxl/dec_cache.h
@@ -241,6 +241,28 @@ struct GroupDecCache {
   size_t max_block_area_ = 0;
 };
 
+// 8x8 ordered dithering pattern from
+// https://en.wikipedia.org/wiki/Ordered_dithering
+const int8_t kIntegerDither[64] = {
+    0,  32, 8,  40, 2,  34, 10, 42, 48, 16, 56, 24, 50, 18, 58, 26,
+    12, 44, 4,  36, 14, 46, 6,  38, 60, 28, 52, 20, 62, 30, 54, 22,
+    3,  35, 11, 43, 1,  33, 9,  41, 51, 19, 59, 27, 49, 17, 57, 25,
+    15, 47, 7,  39, 13, 45, 5,  37, 63, 31, 55, 23, 61, 29, 53, 21};
+
+struct Ordered8x8Dithering {
+  float dither[64];
+  Ordered8x8Dithering() {
+    for (size_t i = 0; i < 64; i++) {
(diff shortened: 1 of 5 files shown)


jonsneyers:
(btw this was inspired by @veluca93's code from a year ago, see https://gitlab.com/wg1/jpeg-xlm/-/merge_requests/3457)

jonsneyers:
> > (btw this was inspired by @veluca93's code from a year ago, see https://gitlab.com/wg1/jpeg-xlm/-/merge_requests/3457)
> 
> Unless we are very much in a hurry, I'd prefer to first refactor the Write stages to handle all the cases that are currently handled in decode.cc.
> 
> Also, I guess we should decide whether dithering should be mandatory, the default, or explicitly requested when writing to U8.
> 
> Finally, and it's a bit of an aside, we should explore how to allow people to use U8 output more for this to be really useful, because i.e. you don't want to dither before color conversions.

Sure, no urgent need to merge this, I just wanted to put it here so people can try it if they want, and so we don't forget about it.

Probably we need a color management stage before conversion to u8 for u8 output to be more useful. That's true regardless of dithering, but indeed it makes more sense to do dithering after color conversion instead of before.

Matt-Gore:
Dithering should always be used when decreasing the bit depth because this is way more accurate than dropping the least significant bits. If the original was lossless 8 bit, then dithering also won't harm since the pixels will round back to the same values.

And yes, color management should not be done at 8 bit precision if possible. So it should be unrelated to dither and happen earlier on the float values.

Matt-Gore:
> Does this need to be included in 1.0?

Yes, it's essential for (lossy) 8 bit output and color management. Dithering was also part of Pik and so it should definitely be included in 1.0.

cfeck:
I suggest to use a blue noise matrix instead of an ordered matrix. Precomputed CC0 blue noise textures are available via http://momentsingraphics.de/BlueNoise.html

veluca93:
Superceded by #3090

PR #1366, opened 2022-04-21
API only (proposal): set CMS and desired output color encoding in decoder

jonsneyers:
No implementation yet, just an API proposal.

The aim is to make uint8 output more useful by allowing applications to let libjxl do the color conversion before conversion to uint8. That way, the uint8 output can be guaranteed to be in display space, and it's no longer necessary to use the float pixel callback and manually do the color management and conversion to uint8.

Open question: in the encode API there's a default cms available, but I guess that here it also has to work in the lightweight decode-only `libjxl_dec` which doesn't include a default cms. We could perhaps allow calling JxlDecoderSetColorEncoding or JxlDecoderSetICCProfile when no explicit cms was set, but only in `libjxl` builds where we can have a default cms (i.e. if you do that when linking with `libjxl_dec`, it will return error).

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -842,7 +842,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
  *     instead.
  */
 // TODO(firsching): deprecate this function once JxlDecoderSetOutputColorProfile
-// is implemted.
+// is implemented.
 JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
     JxlDecoder* dec, const JxlColorEncoding* color_encoding);
 
@@ -897,7 +897,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  * num_color_channels from the basic info is 1, RGB if num_color_channels from
  * the basic info is 3.
  *
- * This function must not be called after JxlDecoderSetCms.
+ * This function must not be called before JxlDecoderSetCms.
  *
  * @param dec decoder orbject
  * @param color_encoding the output color encoding
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -883,9 +883,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  *
  * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
  * before any other event occurred, and should be used before getting
- * JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding or ICC profile must be
- * grayscale if num_color_channels from the basic info is 1, RGB if
- * num_color_channels from the basic info is 3.
+ * JXL_COLOR_PROFILE_TARGET_DATA.
  *
  * This function must not be called before JxlDecoderSetCms.
  *
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -877,10 +877,10 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  * If no CMS has been set with @ref JxlDecoderSetCms, there are two cases:
  *
  * (1) Calling this function with a color encoding will convert XYB images to
- * the desired color encoding. Non-XYB images will not be converted to the
- * desired color space. In this case, if the requested color encoding has a
+ * the desired color encoding. In this case, if the requested color encoding has a
  * narrower gamut, or the white points differ, then the resulting image can have
- * significant color distortion.
+ * significant color distortion. Non-XYB images will not be converted to the
+ * desired color space.
  *
  * (2) Calling this function with an ICC profile will result in an error.
  *
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -830,15 +830,13 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
 /** Sets the desired output color profile of the decoded image by calling
  * @ref JxlDecoderSetOutputColorProfile, passing on @c color_encoding and
  * setting @c icc_data to NULL. See @ref JxlDecoderSetOutputColorProfile for
- * details. See @ref JxlDecoderSetOutputColorProfile for details.
+ * details.
  *
  * @param dec decoder object
  * @param color_encoding the default color encoding to set
  * @return @ref JXL_DEC_SUCCESS if the preference was set successfully, @ref
  *     JXL_DEC_ERROR otherwise.
  */
-// TODO(firsching): deprecate this function once JxlDecoderSetOutputColorProfile
-// is implemented.
 JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
     JxlDecoder* dec, const JxlColorEncoding* color_encoding);
 
@@ -881,11 +879,11 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  * (2) Calling this function with an ICC profile will result in an error.
  *
  * If called with an ICC profile (after a call to @ref JxlDecoderSetCms), the
- * ICC profile has to be a valid RGB or grayscale color profile
+ * ICC profile has to be a valid RGB or grayscale color profile.
  *
  * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
  * before any other event occurred, and should be used before getting
- * JXL_COLOR_PROFILE_TARGET_DATA The color_encoding or ICC profile must be
+ * JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding or ICC profile must be
  * grayscale if num_color_channels from the basic info is 1, RGB if
  * num_color_channels from the basic info is 3.
  *
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -829,7 +829,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
     const JxlDecoder* dec, const JxlPixelFormat* unused_format,
     JxlColorProfileTarget target, uint8_t* icc_profile, size_t size);
 
-/** Sets the desired output color profile either of the decoded image by calling
+/** Sets the desired output color profile of the decoded image by calling
  * @ref JxlDecoderSetOutputColorProfile, passing on @c color_encoding and
  * setting @c icc_data to NULL. See @ref JxlDecoderSetOutputColorProfile for
  * details.
@@ -860,18 +860,18 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
     JxlDecoder* dec, float desired_intensity_target);
 
 /**
- * Sets the desired output color profile either of the decoded
- * image either from color a color encoding or an ICC profile. Valid calls of
- * this function have either @c color_encoding or @c icc_data set to NULL and
- * @c icc_size must be 0 if and only if @c icc_data is NULL.
+ * Sets the desired output color profile of the decoded image either from a
+ * color encoding or an ICC profile. Valid calls of this function have either @c
+ * color_encoding or @c icc_data set to NULL and @c icc_size must be 0 if and
+ * only if @c icc_data is NULL.
  *
  * Depending of whether a color management system (CMS) has been set the
  * behavior is as follows:
  *
- * If an color management system (CMS) has been set with @ref JxlDecoderSetCms,
- * and the CMS suppports output to the desired color encoding or icc profile,
- * the it will provide the output in that color encoding or icc profile. If the
- * desired color encoding or icc is not supported, then an error will be
+ * If a color management system (CMS) has been set with @ref JxlDecoderSetCms,
+ * and the CMS suppports output to the desired color encoding or ICC profile,
+ * the it will provide the output in that color encoding or ICC profile. If the
+ * desired color encoding or the ICC is not supported, then an error will be
  * returned.
  *
  * If no CMS has been set with @ref JxlDecoderSetCms, there are two cases:
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -876,6 +876,69 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
     JxlDecoder* dec, float desired_intensity_target);
 
 /**
+ * Sets the color management system (CMS) that will be used for color conversion
+ * (if applicable) during decoding. May only be set before starting decoding. If
+ * left unset, the decoder will not do arbitrary color conversions:
+ * JxlDecoderSetPreferredColorProfile can be used, but there is no guarantee
+ * that the decoded image will be returned in the preferred color encoding. If
+ * however a CMS is set using this function, then the function
+ * JxlDecoderSetColorEncoding or JxlDecoderSetICCProfile can be used and the
+ * decoded image will then be guaranteed to be in the requested color encoding.
+ *
+ * Only one of JxlDecoderSetPreferredColorProfile, JxlDecoderSetColorEncoding
+ * and JxlDecoderSetICCProfile should be called. The CMS will not be used in
+ * case JxlDecoderSetPreferredColorProfile is used; in the other two cases
+ * the CMS will be used if the desired conversion cannot be performed by the
+ * decoder. In any case, the color encoding of the decoded image data can be
+ * obtained through JxlDecoderGetColorAsICCProfile with
+ * target=JXL_COLOR_PROFILE_TARGET_DATA. If a desired color encoding was set
+ * with JxlDecoderSetColorEncoding or JxlDecoderSetICCProfile, the resulting ICC
+ * profile will always be equivalent to the desired encoding; if a preferred
+ * color encoding was set with JxlDecoderSetPreferredColorProfile, this will not
+ * necessarily be the case (it will only be the case if the decoder can do the
+ * conversion without using the CMS).
+ *
+ * @param dec decoder object.
+ * @param cms structure representing a CMS implementation. See JxlCmsInterface
+ * for more details.
+ */
+JXL_EXPORT void JxlDecoderSetCms(JxlDecoder* dec, JxlCmsInterface cms);
+
+/**
+ * Sets the desired output color encoding of the decoded image. This function
+ * can only be used if JxlDecoderSetCms was previously set. This is an
+ * alternative to JxlDecoderSetICCProfile and only one of these two must be
+ * used. This one sets the color encoding as a @ref JxlColorEncoding, while the
+ * other sets it as ICC binary data.
+ *
+ * @param dec decoder object.
+ * @param color color encoding. Object owned by the caller and its contents are
+ * copied internally.
+ * @return JXL_DEC_SUCCESS if the operation was successful, JXL_DEC_ERROR
+ * otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetColorEncoding(JxlDecoder* dec, const JxlColorEncoding* color);
+
+/**
+ * Sets the desired output color encoding of the decoded image, as an ICC color
+ * profile. This function can only be used if JxlDecoderSetCms was previously
+ * set. The ICC profile has to be a valid RGB color profile (not grayscale or
+ * CMYK). This is an alternative to JxlDecoderSetColorEncoding and only one of
+ * these two must be used. This one sets the color encoding as ICC binary data,
+ * while the other defines it as a @ref JxlColorEncoding.
+ *
+ * @param dec decoder object.
+ * @param icc_profile bytes of the ICC profile
+ * @param size size of the icc_profile buffer in bytes
+ * @return JXL_DEC_SUCCESS if the operation was successful, JXL_DEC_ERROR
+ * otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetICCProfile(JxlDecoder* dec,
+                                                    const uint8_t* icc_profile,
+                                                    size_t size);
+
+/**
  * Returns the minimum size in bytes of the preview image output pixel buffer
  * for the given format. This is the buffer for @ref
  * JxlDecoderSetPreviewOutBuffer. Requires the preview header information is
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -16,8 +16,6 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#include <cstdint>
-
 #include "jxl/cms_interface.h"
 #include "jxl/codestream_header.h"
 #include "jxl/color_encoding.h"
@@ -843,7 +841,9 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
  * JxlDecoderSetOutputColorProfile (possibly with @c icc_data set to NULL)
  *     instead.
  */
-JXL_DEPRECATED JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
+// TODO(firsching): deprecate this function once JxlDecoderSetOutputColorProfile
+// is implemted.
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
     JxlDecoder* dec, const JxlColorEncoding* color_encoding);
 
 /** Requests that the decoder perform tone mapping to the peak display luminance
@@ -885,7 +885,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  * (2) Calling this function with an ICC profile will result in an error.
  *
  * In case @ref JxlDecoderSetOutputColorProfile is used with an ICC profile,
- * (after a in this case required call to @ref JxlDecocderSetCms), the ICC
+ * (after a in this case required call to JxlDecocderSetCms), the ICC
  * profile has to be a valid RGB or grayscale color profile.
  *
  * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
@@ -897,7 +897,7 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
  * num_color_channels from the basic info is 1, RGB if num_color_channels from
  * the basic info is 3.
  *
- * This function must not be called after @ref JxlDecoderSetCms.
+ * This function must not be called after JxlDecoderSetCms.
  *
  * @param dec decoder orbject
  * @param color_encoding the output color encoding
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -16,6 +16,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include "jxl/cms_interface.h"
 #include "jxl/codestream_header.h"
 #include "jxl/color_encoding.h"
 #include "jxl/jxl_export.h"
@@ -833,10 +834,19 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
  * No matter what, the @ref JXL_COLOR_PROFILE_TARGET_DATA must still be queried
  * to know the actual data format of the decoded pixels after decoding.
  *
- * The JXL decoder has no color management system built in, but can convert XYB
- * color to any of the ones supported by JxlColorEncoding. Note that if the
- * requested color encoding has a narrower gamut, or the white points differ,
- * then the resulting image can have significant color distortion.
+ * The JXL decoder has no color management system (CMS) built in, but can
+ * convert XYB color to any of the ones supported by JxlColorEncoding. Note that
+ * if the requested color encoding has a narrower gamut, or the white points
+ * differ, then the resulting image can have significant color distortion.
+ *
+ * Can be used with a CMS by setting it via JxlDecoderSetCms, in which case it
+ * not only can convert to color profile supported by JxlColorEncoding, but to
+ * those supported by the CMS. JxlDecoderSetCms must not be called after
+ * JxlDecoderSetPreferredColorProfile, but it can be optionally called before.
+ * When JxlDecoderSetCms has been called, JxlDecoderSetPreferredColorProfile is
+ * an alternative to JxlDecoderSetICCProfile and only one of these two must be
+ * used. This one sets the color encoding as a @ref JxlColorEncoding, while the
+ * other sets it as ICC binary data.
  *
  * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
  * before any other event occurred, and can affect the result of @ref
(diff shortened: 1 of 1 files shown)


mo271:
If we are going to do that, we should do it before 1.0

lib/include/jxl/decode.h
```diff
@@ -746,6 +746,56 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
+ * other sets it as ICC binary data.
+ *
+ * @param dec decoder object.
+ * @param color color encoding. Object owned by the caller and its contents are
+ * copied internally.
+ * @return JXL_DEC_SUCCESS if the operation was successful, JXL_DEC_ERROR
+ * otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetColorEncoding(JxlDecoder* dec, const JxlColorEncoding* color);
```
veluca93, line 779: How does this interact with SetPreferredColorEncoding?

jonsneyers: I'd say SetPreferred only does a conversion if it can do that without calling cms (even when a cms has been set), so lossless arbitrary colorspace remains untouched (including in particular CMYK, where conversion to RGB is particularly lossy).

veluca93: What happens if you call both?

In general, I think we should better document when to use one and when to use the other.

@sboukortt should we also mention the desired intensity target, perhaps?

lib/include/jxl/decode.h
```diff
@@ -746,6 +746,56 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
+
+/**
+ * Sets the desired output color encoding of the decoded image, as an ICC color
+ * profile. This function can only be used if JxlDecoderSetCms was previously
+ * set. The ICC profile has to be a valid RGB color profile (not grayscale or
```
novomesk, line 784: What is the problem with grayscale profiles so they cannot be allowed here?

jonsneyers: It's not strictly a problem but I think it will be convenient if only the code paths to RGB have to be implemented: gray to RGB, RGB to RGB, CMYK to RGB. If we allow gray output encodings, it doubles the number of cases...

I think in most general-purpose applications you want to be able to decode both color and grayscale images, and the target display space is RGB, so I think you'll want to set this to whatever the display space is (probably sRGB, Display P3, or Rec2100 PQ in most cases).

For special-purpose applications, I think they can do the appropriate color management themselves.

lib/include/jxl/decode.h
jonsneyers: I added a paragraph of documentation to JxlDecoderSetCms to try to clarify things.

Desired intensity target can currently not be set, you only get the intensity target of the image from the basicinfo and you'll get output with the same intensity target, but probably it would be useful to also be able to set a desired intensity target for the output so applications don't have to do re-scaling themselves (in particular if they want to decode HDR images to SDR uint).

jonsneyers:
It has been three weeks now, any further comments/opinions on this?

thebombzen:
I think this is overall a good proposal, particularly `JxlDecoderSetICCProfile` because if you know the ICC profile of the target device or otherwise desired space, then without this API you need to request the pixel data in some containing gamut and provide the CMS yourself, despite libjxl having a CMS built-in.

lib/include/jxl/decode.h
thebombzen: I think it would make sense to allow Gray -> Gray at least, so you could get the pixel data in the TARGET_ORIGINAL space in the case of grayscale.

jonsneyers: Agreed, Gray to Gray does make sense. So let's say the following are allowed:

- RGB to RGB
- CMYK to RGB
- Gray to RGB
- Gray to Gray

I don't think we want to allow conversions _to_ CMYK; perhaps it is also useful to allow conversions from RGB/CMYK to Gray but I think that's probably not a big use case and I want to avoid having to implement more code paths than needed.

In case the original space is a CMYK space, I propose we only allow you to get the pixels converted to an RGB space, or else to get them as they are encoded (which is the original space in the non-XYB case). Probably we should not allow the combination of XYB and CMYK — I think the spec currently allows that, but it's ambiguous how to interpret such data.

lib/include/jxl/decode.h
sboukortt: Desired intensity target sounds like a bit of a pain (it means we would probably have to do some tone mapping, and likely different for HLG content), but perhaps it could be good to have indeed.

mo271:
Can we get this merged soon?
To summarize, the two remaining questions seem to be:
  (1) Desired intensity target: probably yes, if I interpret @sboukortt answer correctly
  (2) Allow gray-to-gray: yes, please

mo271:
@jonsneyers 
I'm thinking about how to implement this internally:

1. is the plan to add a `JxlCmsInterface cms` field to the `PassesDecoderState` struct? (and perhaps a bool `cms_is_set`?)
2. When using `JxlDecoderSetColorEncoding` or `JxlDecoderSetICCProfile`, how to detect if the cms is actually needed? Should we adapt `OutputEncodingInfo::Set` to signal if we can ouput without cms in the desired color encoding?
3. Should we init the cms already when calling `JxlDecoderSetColorEncoding` or `JxlDecoderSetICCProfile` (and it turns out that we actually need it) or sometimes later?

mo271:
Some unordered thoughts from an offline discussion about this with @veluca93 and @sboukortt:

 - `JxlDecoderSetPreferredColorProfile` should only fail in those cases, where it is is really impossible to decode without cms to the desired `JxlColorEncoding`.
  - Not all cms can handle decoding to all ICC profiles, so `JxlDecoderSetICCProfile` perhaps somehow signal if the combination requested was successful, and/or set good fallbacks if it was not.
  - Having 3 functions to set the desired color encoding seems a bit much. Perhaps we could only use JxlDecoderSetPreferredColorProfile and JxlDecoderSetICCProfile and instead of using `JxlDecoderSetColorEncoding` for a `JxlColorEncoding`, we convert that to an ICC and then use `JxlDecoderSetICCProfile`?!
  - For some HDR content, it might be beneficial to know more than just the intensity target, and we don't really want to rely on reading out if the original profile was PQ or HLG in order to do the correct color transform (Perhaps this is not stricly related to the design of the api functions discussed here, but we should keep it in mind)

jonsneyers:
I kind of wonder how useful JxlDecoderSetPreferredColorProfile will still be if we have a function that can guarantee to give you output in the preferred space (which would only work when providing a cms, but standard libjxl builds could have a default cms since they have it for encode anyway).

I think there are three main things an application might want to do:

- Get data in the original space, which involves no conversion in case of lossless/no-xyb, but in case of xyb it does require conversion;
- Get data in the display space, which does require conversion unless the display space happens to be the original space.
- Get the data in sRGB, since that's basically the colorspace of everything that doesn't support color profiles.

There's also the case of CMYK, where I think most applications just want to avoid having to deal with that and want to get the data in any space as long as it's an RGB one.

I think it's fine if we don't support conversion _to_ LUT spaces, but I think we should support conversion _from_ LUT spaces (which is needed for lossless images with exotic RGB icc profiles and I guess for all CMYK images). Or maybe we can just make it something that can return failure if the cms cannot handle it, and then the behavior you get depends on what cms you use.

I expect conversion from Rec2100 PQ to smaller gamut SDR spaces (sRGB / Display P3) to be something that will be frequently required in practice, since SDR devices will remain a thing and environments that can only deal with uint8 too.

sboukortt:
> I kind of wonder how useful JxlDecoderSetPreferredColorProfile will still be if we have a function that can guarantee to give you output in the preferred space (which would only work when providing a cms, but standard libjxl builds could have a default cms since they have it for encode anyway).

We were thinking that perhaps we could extend `JxlDecoderSetPreferredColorProfile` to provide that functionality, by relaxing its current constraints.

> I think it's fine if we don't support conversion _to_ LUT spaces, but I think we should support conversion _from_ LUT spaces (which is needed for lossless images with exotic RGB icc profiles and I guess for all CMYK images). Or maybe we can just make it something that can return failure if the cms cannot handle it, and then the behavior you get depends on what cms you use.
> 
> I expect conversion from Rec2100 PQ to smaller gamut SDR spaces (sRGB / Display P3) to be something that will be frequently required in practice, since SDR devices will remain a thing and environments that can only deal with uint8 too.

True. Given this, it probably does make sense to still require a CMS for some functionality.

> This is a potentially important use case, because it means that a lossy encode->decode round trip is possible while using XYB and comparing the final pixel data to the input pixel data, even with an attached ICC profile that describes that space.

For what it’s worth, in certain cases it’s not going to be possible anyway. (Some profiles can only be converted from, not to.) Generally speaking, for this sort of comparison, it probably makes sense to convert both the input and the output to a specific colorspace instead of the arbitrary colorspace that the input happened to be in. After all, that is what Butteraugli (XYB) and SSIMULACRA (L\*a\*b\*) do.

thebombzen:
> > This is a potentially important use case, because it means that a lossy encode->decode round trip is possible while using XYB and comparing the final pixel data to the input pixel data, even with an attached ICC profile that describes that space.
> 
> For what it’s worth, in certain cases it’s not going to be possible anyway. (Some profiles can only be converted from, not to.) Generally speaking, for this sort of comparison, it probably makes sense to convert both the input and the output to a specific colorspace instead of the arbitrary colorspace that the input happened to be in. After all, that is what Butteraugli (XYB) and SSIMULACRA (L*a*b*) do.

The big issue here is with a libjxl client that is not intending to be the final consumer of the pixel data, such as conversion utilities like FFmpeg or ImageMagick. There is currently no "sane default" with which to decode an XYB-based JPEG XL file that has an attached ICC profile.

This is the current behavior in FFmpeg:

1. If there is no attached ICC profile, and the Enums encoded in the image header are supported by FFmpeg, then `SetPreferredColorEncoding` is called on those enums and pixel data is requested in the enum space, and tagged appropriately.
2. If there is no attached ICC profile and the enums are not supported by FFmpeg, then pixel data is requested in BT.2020/D65/PQ/Relative, which is the widest gamut that FFmpeg currently supports.
3. If there is an attached ICC profile and `xyb_encoded = false`, then pixel data is requested in the original space and the attached ICC profile is forwarded as-is. (This is common with reconstructed JPEGs.)
4. If there is an attached ICC profile and `xyb_encoded = true`, then pixel data is requested in BT.2020/D65/PQ/Relative, which is the widest gamut that FFmpeg currently supports.

I'm interested in case (4) above, which I expect to be relatively common, and I don't see a sane default for it. If an application like FFmpeg or ImageMagick encounters a JPEG XL file as in (4), there needs to be a sane default which is not requesting ultra-wide gamut pixel data and hoping the next part of the chain can handle it. What if that ICC profile is something narrow-gamut like Adobe RGB, or simply just sRGB? The client application should not have to parse the ICC profile to determine that, since the client may or may not have a CMS or ICC-parsing tools.

While it may not *always* be possible to convert *to* the space described by the embedded ICC profile, it often will be possible especially if it just contains H.273 information or equivalent. In this event, converting to that space provides a *sane default.* If this functionality is not provided, another sane default needs to exist for intermediary tools, because requesting ultra-wide gamut as a fallback (in order to prevent losing any information) is not ideal.

sboukortt:
Case 4 should be relatively rare, because in VarDCT mode, if the ICC profile represents a colorspace that can be expressed with enum values, the encoder currently replaces it with the enum values altogether. (Although I should check that this is indeed a feature of libjxl and not just legacy `cjxl`.)

Also, we plan for at least linear float output to remain unbounded regardless of the nominal gamut of the requested output colorspace, similar to how it works in OpenEXR. Does that sound helpful?

thebombzen:
> Case 4 should be relatively rare, because in VarDCT mode, if the ICC profile represents a colorspace that can be expressed with enum values, the encoder currently replaces it with the enum values altogether. (Although I should check that this is indeed a feature of libjxl and not just legacy `cjxl`.)

If case (4) is indeed going to be rare, then I see much less need for this.

EwoutH:
This sounds like a very interesting proposal! Any update on this efford?

mo271:
Discussing this again gave the following conclusion (in contrast to the post just above):

 - Having `SetPreferredColorEncoding` and `JxlDecoderSetDesiredIntensityTarget` not working in all cases is a problem and make those api functions less useful.
 - When the cms passed to the new api `JxlDecoderSetCms` can handle a certain color space, then `SetPreferredColorEncoding` and `JxlDecoderSetDesiredIntensityTarget` should work for that color space. 
 - Then there is no need for a `JxlDecoderSetColorEncoding` function, so let's not add it. 
 
Is this a good way forward, @jonsneyers (and @sboukortt @veluca93 @szabadka)?

jonsneyers:
> Discussing this again gave the following conclusion (in contrast to the post just above):
> 
> * Having `SetPreferredColorEncoding` and `JxlDecoderSetDesiredIntensityTarget` not working in all cases is a problem and make those api functions less useful.
> * When the cms passed to the new api `JxlDecoderSetCms` can handle a certain color space, then `SetPreferredColorEncoding` and `JxlDecoderSetDesiredIntensityTarget` should work for that color space.
> * Then there is no need for a `JxlDecoderSetColorEncoding` function, so let's not add it.
> 
> Is this a good way forward, @jonsneyers (and @sboukortt @veluca93 @szabadka)?

Sounds good to me.

When no cms is passed with `JxlDecoderSetCms`, can we make sure that it then uses the default cms if one is available? Basically in a normal libjxl build we have a default cms anyway (for encoding), so we might as well make it work for decoding too. I understand that in a libjxl-dec build we may want to avoid forcing a dependency on a cms, so then you'll have to bring your own cms if you want to get output in a specific colorspace, but in the full libjxl I think we should make it "just work" by default.

How does an application know if the (default or custom) cms can handle the conversion or not? What are the cases where we expect it not to work?

lib/include/jxl/decode.h
mo271: as discussed below, settling for not having `JxlDecoderSetColorEncoding`, but letting `JxlDecoderSetPreferredIntensityTarget` take that role whether or not a CMS

lib/include/jxl/decode.h
mo271: What do we do now here, @jonsneyers?

mo271:
I tried to reflect the discussed changes into the pull request, please take a look @jonsneyers (and @veluca93 @sboukortt @szabadka)

veluca93:
Another question to consider: do we need to support decoding to an
arbitrary ICC? I'd argue it is not fundamental for our API to allow it.

lib/include/jxl/decode.h
jonsneyers: I think that what I wrote in the comment above back in May still makes sense. At least I think that's the minimum we should allow/implement. If it's not much trouble to just allow anything the cms can handle (including stuff like decoding an RGB image to CMYK, if the cms can do that), then that's also fine for me.

For any viewing application, they need to be able to convert any jxl to the display colorspace, or at least to an RGB(A) buffer plus an ICC profile that they can then pass to something that knows how to deal with that. So (anything) to RGB is needed.

For grayscale images, it can be annoying to force them to be converted to RGB since that requires a buffer 3 times as large. So gray to gray is useful too imo.

lib/include/jxl/decode.h
```diff
@@ -833,10 +834,19 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
- * The JXL decoder has no color management system built in, but can convert XYB
- * color to any of the ones supported by JxlColorEncoding. Note that if the
- * requested color encoding has a narrower gamut, or the white points differ,
- * then the resulting image can have significant color distortion.
+ * The JXL decoder has no color management system (CMS) built in, but can
+ * convert XYB color to any of the ones supported by JxlColorEncoding. Note that
+ * if the requested color encoding has a narrower gamut, or the white points
+ * differ, then the resulting image can have significant color distortion.
+ *
+ * Can be used with a CMS by setting it via JxlDecoderSetCms, in which case it
```
jonsneyers, line 842: Should we require JxlDecoderSetCms to be called even if libjxl has a default cms available?
In the encode api we don't require that, if I understand correctly.

mo271: Yes, good idea. I tried to make that clear in the latest commit.

lib/include/jxl/decode.h
mo271: Explained it like that in the doc string now.

mo271:
The latest commits have been done together with @veluca93 and contain the following ideas:

 - merge together `JxlDecoderSetPreferredColorProfile` and `JxlDecoderSetICCProfile` as `JxlDecoderSetOutputColorProfile` order avoid having two functions to do very simliar things, having only to document it once.
  - Allow transformation to any valid RGB and grayscale ICC profile (in order not to make it fail conversion from RGB images while it works for grayscale->grayscale), even though this might mean slighly more work in implementing this.
  - Shorten the doc for `JxlDecoderSetCms` and refer to the one of `JxlDecoderSetOutputColorProfile`
  
  From my point of view this looks good now. What do you think, @jonsneyers and @sboukortt?

mo271:
(will fix the failing build of course...)

mo271 (dismissed):

mo271 (dismissed):

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ *
+ * In case @ref JxlDecoderSetOutputColorProfile is used with an ICC profile,
+ * (after a in this case required call to JxlDecocderSetCms), the ICC
+ * profile has to be a valid RGB or grayscale color profile.
+ *
+ * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
+ * before any other event occurred, and can affect the result of @ref
+ * JXL_COLOR_PROFILE_TARGET_DATA (but not of @ref
+ * JXL_COLOR_PROFILE_TARGET_ORIGINAL), so should be used after getting @ref
+ * JXL_COLOR_PROFILE_TARGET_ORIGINAL but before getting @ref
```
jonsneyers, line 895: Just say "should be used before getting JXL_COLOR_PROFILE_TARGET_DATA" — you can still get the original space whenever you want, right?

lib/include/jxl/decode.h
```diff
@@ -826,39 +827,22 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
  * @param color_encoding the default color encoding to set
  * @return @ref JXL_DEC_SUCCESS if the preference was set successfully, @ref
  *     JXL_DEC_ERROR otherwise.
+ *
+ * @deprecated This function will be removed. Use @ref
+ * JxlDecoderSetOutputColorProfile (possibly with @c icc_data set to NULL)
+ *     instead.
  */
+// TODO(firsching): deprecate this function once JxlDecoderSetOutputColorProfile
+// is implemted.
```
jonsneyers, line 845:
```suggestion
// is implemented.
```

Do we really need to deprecate this one? I don't think it really hurts to keep it around, just as syntactic sugar so you don't need the `, NULL, 0` and so applications using the 0.7 API don't break.

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetOutputColorProfile(
+    JxlDecoder* dec, const JxlColorEncoding* color_encoding,
+    const uint8_t* icc_data, size_t icc_size);
+
+/**
+ * Sets the color management system (CMS) that will be used for color
+ * conversion (if applicable) during decoding. May only be set before starting
+ * decoding and must not be called after @ref JxlDecoderSetOutputColorProfile.
+ *
```
jonsneyers, line 917: In the encode api, we also have `If left unset, the default CMS implementation will be used.`. Can we do that here too? Probably only in case `libjxl` is not used and not in case `libjxl_dec` is used.

jonsneyers:
LGTM modulo some typos/nitpicks.

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
 JXL_EXPORT JxlDecoderStatus JxlDecoderSetDesiredIntensityTarget(
     JxlDecoder* dec, float desired_intensity_target);
 
+/**
+ * Sets the desired output color profile of the decoded image either from a
+ * color encoding or an ICC profile. Valid calls of this function have either @c
+ * color_encoding or @c icc_data set to NULL and @c icc_size must be 0 if and
+ * only if @c icc_data is NULL.
+ *
+ * Depending of whether a color management system (CMS) has been set the
```
szabadka, line 868: typo: Depending on

veluca93:
There is no default CMS implementation in some cases and I am very much
against introduction behaviour that works in some cases but not others :)

lib/include/jxl/decode.h
szabadka: I agree, if we are clear that it is just a shorthand for the other function, we could keep it.

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ *
+ * (1) Calling this function with a color encoding will convert XYB images to
+ * the desired color encoding. In this case, if the requested color encoding has
+ * a narrower gamut, or the white points differ, then the resulting image can
+ * have significant color distortion. Non-XYB images will not be converted to
+ * the desired color space.
+ *
+ * (2) Calling this function with an ICC profile will result in an error.
+ *
+ * In case @ref JxlDecoderSetOutputColorProfile is used with an ICC profile,
```
szabadka, line 887: I think referencing the function we are documenting is unnecessary here, how about something like
"If called with an ICC profile (after a call to @ref JxlDecoderSetCms), the ICC profile has to be a valid RGB or grayscale color profile".

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ * (after a in this case required call to JxlDecocderSetCms), the ICC
+ * profile has to be a valid RGB or grayscale color profile.
+ *
+ * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
+ * before any other event occurred, and can affect the result of @ref
+ * JXL_COLOR_PROFILE_TARGET_DATA (but not of @ref
+ * JXL_COLOR_PROFILE_TARGET_ORIGINAL), so should be used after getting @ref
+ * JXL_COLOR_PROFILE_TARGET_ORIGINAL but before getting @ref
+ * JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding must be grayscale if
```
szabadka, line 896: Is this restriction only for the color_encoding case? Or should it say "color encoding or ICC profile must be grayscale ..."? (Sorry if I missed the discussion on this). In either case, I think this restriction would fit better with the previous paragraph.

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ * num_color_channels from the basic info is 1, RGB if num_color_channels from
+ * the basic info is 3.
+ *
+ * This function must not be called after JxlDecoderSetCms.
+ *
+ * @param dec decoder orbject
+ * @param color_encoding the output color encoding
+ * @param icc_data bytes of the icc profile
+ * @param icc_size size of the icc profile in bytes
+ * @return @ref JXL_DEC_SUCCESS if the preference was set successfully, @ref
```
szabadka, line 906: preference -> color profile

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ *     JXL_DEC_ERROR otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetOutputColorProfile(
+    JxlDecoder* dec, const JxlColorEncoding* color_encoding,
+    const uint8_t* icc_data, size_t icc_size);
+
+/**
+ * Sets the color management system (CMS) that will be used for color
+ * conversion (if applicable) during decoding. May only be set before starting
+ * decoding and must not be called after @ref JxlDecoderSetOutputColorProfile.
```
szabadka, line 916: The second half of the sentence seems redundant.

lib/include/jxl/decode.h
```diff
@@ -875,6 +859,71 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ * color encoding or an ICC profile. Valid calls of this function have either @c
+ * color_encoding or @c icc_data set to NULL and @c icc_size must be 0 if and
+ * only if @c icc_data is NULL.
+ *
+ * Depending of whether a color management system (CMS) has been set the
+ * behavior is as follows:
+ *
+ * If a color management system (CMS) has been set with @ref JxlDecoderSetCms,
+ * and the CMS suppports output to the desired color encoding or ICC profile,
+ * the it will provide the output in that color encoding or ICC profile. If the
```
szabadka, line 873: typo: the->then

szabadka (commented):
Some last minute comments, mainly typos and style.

jonsneyers:
> There is no default CMS implementation in some cases and I am very much against introduction behaviour that works in some cases but not others :)

What are the cases when there is no default CMS implementation? Are those the same cases as where the entire encode API is not available?

I don't know what's the best way to do this, but from an application point of view it seems strange that to encode an image, things just work out of the box, but to decode it, you first need to manually implement a JxlCmsInterface from scratch and explicitly call `JxlDecoderSetCms`. At the very least, I think we need to expose `JxlGetCms()` (which can return error or return the default cms) so applications can at least use the default cms if there is one. But that basically just moves the "behavior that works in some case not others" from one point to the other.

I think we just have to say that `libjxl` is guaranteed to have a default cms while `libjxl_dec` does not have one, and that this means that if you don't call `JxlDecoderSetCms`, it is initialized to the default cms when using `libjxl` but not when using `libjxl_dec`. I don't see why we have to guarantee that applications that work with `libjxl` will also work with `libjxl_dec` — this is in fact already not at all the case.

lib/include/jxl/decode.h
mo271: ok, let's keep it around

lib/include/jxl/decode.h
mo271: I didn't include using a default CMS, because of the fear that the behavior will be unexpectedly different using different compile time options.
Strong opinions on this, @veluca93?

jonsneyers: As long as there's a convenient way for applications to get color managed decoding, I'm happy. In my view, defining a custom CMS is pretty advanced usage of the API and it's not reasonable to expect all applications to do that.

I don't think there will be a significant difference in behavior: in case there is no default cms, `JxlDecoderSetOutputColorProfile` will fail on more input jxl bitstreams than if there is one, but an application in any case needs to deal with the case where it fails, so I don't think it makes a big difference. It would only make a difference if an application is using libjxl and doesn't check the result of `JxlDecoderSetOutputColorProfile` (which, to be fair, is something I do expect some applications to do, since as long as they pass it a valid output colorspace, it is not expected to fail with the default CMS), **and then** they switch to libjxl_dec instead of libjxl.

The way I see it (but please correct me if I'm wrong, because I may be misunderstanding something), there are two main use cases for `libjxl_dec`: 1) Chrome and possibly other browsers/viewers that already have their own cms, so they don't want to introduce a dependency on another cms; 2) games and other decode-only applications that only need to decode their own images, so they don't need cms since they can just avoid jxl files with icc profiles.
In the case of 1), they will set a custom cms so there is no issue. In the case of 2), they will not need a cms so there is also no issue.

Most applications will want to depend on `libjxl` though, because they either need encoding too or because they don't care about binary size enough for it to be worth the hassle of figuring out how to do custom cms (or because they are on a sane platform where dynamic linking with a system libjxl is possible, so binary size is not an issue). For those applications, having a convenient way to get a decoded image in the colorspace they want is important imo.

The main issue here is that it's quite a bit of hassle to define a custom cms. I don't mind that people have to add a single line to call `JxlDecoderSetCms()`, but they need to do much more than that, they have to actually implement that whole interface themselves. That means they basically need to write most of `enc_color_management.cc`.

How about we just add a function `JxlDecoderSetDefaultCms()` that either can return failure, or maybe even better, that doesn't exist at all in `libjxl_dec`? That way we can ensure that there is no different behavior, but still have a simple way for application to get access to the default cms if there is one?

lib/include/jxl/decode.h
mo271: done

lib/include/jxl/decode.h
mo271: done

lib/include/jxl/decode.h
mo271: I think "starting decoding" and calling JxlDecoderSetOutputColorProfile are two different things, so this does not seem redundant to me, or am I misunderstanding something?

lib/include/jxl/decode.h
mo271: right, done.

lib/include/jxl/decode.h
mo271: add "or ICC profile"

lib/include/jxl/decode.h
mo271: Having a default function `JxlDecoderSetDefaultCms()` sounds good to me. Added a todo about that.

lib/include/jxl/decode.h
```diff
@@ -826,39 +827,18 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
- * If @ref JxlDecoderSetPreferredColorProfile is not used, then for images for
- * which uses_original_profile is false and with ICC color profile, the decoder
- * will choose linear sRGB for color images, linear grayscale for grayscale
- * images. This function only sets a preference, since for other images the
- * decoder has no choice what color profile to use, it is determined by the
- * image.
+/** Sets the desired output color profile of the decoded image by calling
+ * @ref JxlDecoderSetOutputColorProfile, passing on @c color_encoding and
+ * setting @c icc_data to NULL. See @ref JxlDecoderSetOutputColorProfile for
+ * details. See @ref JxlDecoderSetOutputColorProfile for details.
```
szabadka, line 833: Last sentence is repeated.

lib/include/jxl/decode.h
```diff
@@ -826,39 +827,18 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetColorAsICCProfile(
...
  *
  * @param dec decoder object
  * @param color_encoding the default color encoding to set
  * @return @ref JXL_DEC_SUCCESS if the preference was set successfully, @ref
  *     JXL_DEC_ERROR otherwise.
  */
+// TODO(firsching): deprecate this function once JxlDecoderSetOutputColorProfile
```
szabadka, line 840: remove todo?

lib/include/jxl/decode.h
```diff
@@ -875,6 +855,69 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ * (1) Calling this function with a color encoding will convert XYB images to
+ * the desired color encoding. In this case, if the requested color encoding has
+ * a narrower gamut, or the white points differ, then the resulting image can
+ * have significant color distortion. Non-XYB images will not be converted to
+ * the desired color space.
+ *
+ * (2) Calling this function with an ICC profile will result in an error.
+ *
+ * If called with an ICC profile (after a call to @ref JxlDecoderSetCms), the
+ * ICC profile has to be a valid RGB or grayscale color profile
```
szabadka, line 884: full stop at the end.

lib/include/jxl/decode.h
```diff
@@ -875,6 +855,69 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ *
+ * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
+ * before any other event occurred, and should be used before getting
+ * JXL_COLOR_PROFILE_TARGET_DATA The color_encoding or ICC profile must be
```
szabadka, line 888: full stop after JXL_COLOR_PROFILE_TARGET_DATA

lib/include/jxl/decode.h
szabadka: I mean it would be simpler to just say: "May only be called before@ref JxlDecoderSetOutputColorProfile."

lib/include/jxl/decode.h
mo271: done

lib/include/jxl/decode.h
mo271: yes, done

lib/include/jxl/decode.h
mo271: done

lib/include/jxl/decode.h
mo271: done

szabadka (dismissed):

lib/include/jxl/decode.h
```diff
@@ -875,6 +853,69 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
...
+ * (2) Calling this function with an ICC profile will result in an error.
+ *
+ * If called with an ICC profile (after a call to @ref JxlDecoderSetCms), the
+ * ICC profile has to be a valid RGB or grayscale color profile.
+ *
+ * Can only be set after the @ref JXL_DEC_COLOR_ENCODING event occurred and
+ * before any other event occurred, and should be used before getting
+ * JXL_COLOR_PROFILE_TARGET_DATA. The color_encoding or ICC profile must be
+ * grayscale if num_color_channels from the basic info is 1, RGB if
+ * num_color_channels from the basic info is 3.
```
thebombzen, line 888: What if the image is tagged as grayscale in the Basic Info, but the `JxlPixelFormat` requests it as RGB? Would an RGB profile be supported in this instance? I'd imagine this should work for XYB since grayscale is just XYB with X=B=0, right?

mo271: Perhaps we go with allowing transformation to any valid RGB and grayscale ICC profile (in order not to make it fail conversion from RGB images while it works for grayscale->grayscale), even though this might mean slighly more work in implementing this.

I'll remove the restriction from the doc string.

jonsneyers:
LGTM as it is now, I wanted to approve but I cannot do that since I started this PR :)

szabadka (approved):

mo271 (approved):

mo271:
> LGTM as it is now, I wanted to approve but I cannot do that since I started this PR :)

Cool, I approved it, you can merge it...

PR #1367, opened 2022-04-21
simplify simd unsqueeze

jonsneyers:
Same simplification as in https://github.com/libjxl/libjxl/pull/1358, but in the SIMD unsqueeze path.

It's a very tiny little bit faster this way, almost not measurable.

Before:
4064 x 2704, geomean: 17.34 MP/s [15.22, 18.43], 30 reps, 0 threads.

After:
4064 x 2704, geomean: 17.40 MP/s [15.29, 18.57], 30 reps, 0 threads.

### DIFF
--- a/lib/jxl/modular/transform/squeeze.cc
+++ b/lib/jxl/modular/transform/squeeze.cc
@@ -23,6 +23,7 @@ HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
 
+using hwy::HWY_NAMESPACE::RebindToUnsigned;
 using hwy::HWY_NAMESPACE::ShiftLeft;
 using hwy::HWY_NAMESPACE::ShiftRight;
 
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/transform/squeeze.cc
```diff
@@ -35,6 +35,8 @@ JXL_INLINE void FastUnsqueeze(const pixel_type *JXL_RESTRICT p_residual,
                               pixel_type *JXL_RESTRICT p_out,
                               pixel_type *p_nout) {
   const HWY_CAPPED(pixel_type, 8) d;
+  const HWY_CAPPED(uint32_t, 8) du;
+  static_assert(sizeof(pixel_type) == 4, "code assumes pixel_type is int32_t");
```
veluca93, line 39: IIRC there's something like Rebind that would avoid this.

veluca93 (approved):

PR #1371, opened 2022-04-22
deprecate JxlDecoderDefaultPixelFormat

jonsneyers:
This function is not very useful. At the moment it just always returns "little-endian float RGBA", but even if we would change the implementation so it returns an image-dependent "good default pixel format", it would still not be a very useful function.

No application is going to set the pixelformat to something that can be anything, because that means the application will have to deal with all possible pixelformats (which is a big pain, since it could be things like non-native-endian grayscale+alpha float16, for all the application knows). It is way more likely that the application will want to set the pixelformat to something specific that can be used directly by the framework it is using or that corresponds to the convention the application uses elsewhere.

So I think this function is not useful, and now we can still remove it, we probably should.


mo271:
should we rebase this and get it reviewed again?

mo271 (approved):

mo271:
Are you planning to merge this, @jonsneyers?

jonsneyers:
Yes, let's.

PR #1373, opened 2022-04-22
make coverage run all tests, and only build it if it's going to make a report

jonsneyers:
Coverage reports currently don't include jxl_test and slow tests because of some philosophical distinction between "unit tests" and "roundtrip tests" which I don't think really makes sense.

Also, when not pushing to main, no coverage report is produced (which is intended) but the instrumented code still gets built (which is a waste of time if no report is going to be produced).

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -59,8 +59,6 @@ jobs:
             apt_pkgs: gcovr
             # Coverage builds require a bit more RAM.
             env_test_stack_size: 2048
-            # Exclude roundtrip tests from the unittest coverage.
-            ctest_args: -E '^JxlTest'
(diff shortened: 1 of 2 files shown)


jonsneyers:
I'll check on my own fork how long it takes to do the coverage with this change, it'll probably take quite a bit longer but the report should more accurately reflect what code is actually reached by the tests.

jonsneyers:
It takes about 1h to run the coverage tests, about twice as long as without jxl_test and without the slow tests. That's about as long as the debug build. I think that's still acceptable.

jonsneyers:
It takes about 1h to run the coverage tests, about twice as long as without jxl_test and without the slow tests. That's about as long as the debug build. I think that's still acceptable.

veluca93 (approved):

PR #1383, opened 2022-04-25
tools/jxlinfo: improve CLI option parsing

thebombzen:
~~Calling "jxlinfo --help" should provide the default help
message rather than failing with "failed to read file --help."
This fixes that behavior to work as expected.~~

This patch adds --help, --verbose, and other convenience
features, such as the ability to read multiple files with
jxlinfo.

### DIFF
--- a/tools/jxlinfo.c
+++ b/tools/jxlinfo.c
@@ -400,7 +400,7 @@ int PrintBasicInfo(FILE* file, int verbose) {
 static void print_usage(const char* name) {
   fprintf(stderr,
           "Usage: %s [-v] INPUT\n"
-          "  INPUT      input JPEG XL image filename\n"
+          "  INPUT      input JPEG XL image filename(s)\n"
           "  -v         more verbose output\n",
           name);
 }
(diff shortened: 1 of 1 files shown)


jonsneyers:
Maybe allow -h too then?

thebombzen:
`-h` and `-?` could both be added, if we desired

mo271:
> `-h` and `-?` could both be added, if we desired

Personally, I think that `-h` and `--help` would be enough, but if you think `-?` would also be used, then feel free to add that as well!

thebombzen:
Aight, ready for review again, after a bunch of forcepushing

tools/jxlinfo.c
```diff
@@ -397,39 +397,65 @@ int PrintBasicInfo(FILE* file, int verbose) {
   return seen_basic_info;
 }
 
+static void print_usage(const char* name) {
+  fprintf(stderr,
+          "Usage: %s [-v] INPUT\n"
+          "  INPUT      input JPEG XL image filename\n"
```
jonsneyers, line 403:
```suggestion
          "  INPUT      input JPEG XL image filename(s)\n"
```

thebombzen: Ah right, since it now supports multiple inputs.

jonsneyers (approved):

Issue #1353, opened 2022-04-16
No implementation of JxlDecoderSetProgressiveDetail(), defined in decode.h


Imroy:
Back in December, JxlDecoderSetProgressiveDetail() was added to decode.h (4ca53d55b075846124f9783461c01ffbffb431f8).
But there is still no implementation of this function anywhere.

Has this been forgotten about in someone's private repository?

PR #1392, opened 2022-05-05
implement JxlDecoderSetProgressiveDetail

mo271:
closes https://github.com/libjxl/libjxl/issues/1353
Co-authored-by: Luca Versari <veluca@google.com>

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -1279,7 +1279,8 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxSizeRaw(const JxlDecoder* dec,
 
 /**
  * Configures at which progressive steps in frame decoding the @ref
- * JXL_DEC_FRAME_PROGRESSION event occurs.
+ * JXL_DEC_FRAME_PROGRESSION event occurs. The default value for the level of
+ * detail if this function is never called is `kDC`.
  *
  * @param dec decoder object
  * @param detail at which level of detail to trigger JXL_DEC_FRAME_PROGRESSION
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -1279,17 +1279,15 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxSizeRaw(const JxlDecoder* dec,
 
 /**
  * Configures at which progressive steps in frame decoding the @ref
- * JXL_DEC_FRAME_PROGRESSION event occurs. By default, this is 0. The detail
- * values mean: 0 = only trigger for the DC image, the 8x8th lower resolution
- * image. 1 = also trigger when a full pass of groups is ready. Higher values
- * indicate more steps but are not yet implemented. Higher values always include
- * the events of lower values as well.
+ * JXL_DEC_FRAME_PROGRESSION event occurs.
  *
  * @param dec decoder object
  * @param detail at which level of detail to trigger JXL_DEC_FRAME_PROGRESSION
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as an
+ * invalid value for the progressive detail.
  */
-JXL_EXPORT void JxlDecoderSetProgressiveDetail(JxlDecoder* dec,
-                                               uint32_t detail);
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetProgressiveDetail(JxlDecoder* dec, JxlProgressiveDetail detail);
 
 /**
  * Outputs progressive step towards the decoded image so far when only partial
--- a/lib/include/jxl/types.h
+++ b/lib/include/jxl/types.h
@@ -115,6 +115,32 @@ typedef struct {
  */
 typedef char JxlBoxType[4];
 
+/** Types of progressive detail.
+ * Setting a progressive detail with value N implies all progressive details
+ * with smaller or equal value. Currently only the following level of
+ * progressive detail is implemented:
+ *  - kDC (which implies kFrames)
+ *  - kPasses (which implies kLastPasses, kDC and kFrames)
+ */
+typedef enum {
+  // after completed kRegularFrames
+  kFrames = 0,
+  // after completed DC (1:8)
+  kDC = 1,
+  // after completed AC passes that are the last pass for their resolution
+  // target.
+  kLastPasses = 2,
+  // after completed AC passes that are not the last pass for their resolution
+  // target.
+  kPasses = 3,
+  // during DC frame when lower resolution are completed (1:32, 1:16)
+  kDCProgressive = 4,
+  // after completed groups
+  kDCGroups = 5,
+  // after completed groups
+  kGroups = 6,
+} JxlProgressiveDetail;
+
 #if defined(__cplusplus) || defined(c_plusplus)
 }
 #endif
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -15,6 +15,7 @@
 #include <utility>
 #include <vector>
 
+#include "jxl/types.h"
 #include "lib/jxl/ac_context.h"
 #include "lib/jxl/ac_strategy.h"
 #include "lib/jxl/ans_params.h"
@@ -656,7 +657,9 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
   std::vector<std::vector<size_t>> ac_group_sec(
       frame_dim_.num_groups,
       std::vector<size_t>(frame_header_.passes.num_passes, num));
-  std::vector<size_t> num_ac_passes(frame_dim_.num_groups);
+  // This keeps track of the number of ac passes we want to process during this
+  // call of ProcessSections.
+  std::vector<size_t> desired_num_ac_passes(frame_dim_.num_groups);
   bool single_section =
       frame_dim_.num_groups == 1 && frame_header_.passes.num_passes == 1;
   if (single_section) {
@@ -666,7 +669,7 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
       processed_section_[0] = true;
       ac_group_sec[0].resize(1);
       dc_global_sec = ac_global_sec = dc_group_sec[0] = ac_group_sec[0][0] = 0;
-      num_ac_passes[0] = 1;
+      desired_num_ac_passes[0] = 1;
     } else {
       section_status[0] = SectionStatus::kDuplicate;
     }
@@ -706,7 +709,7 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
           break;
         }
       }
-      num_ac_passes[g] = j;
+      desired_num_ac_passes[g] = j;
     }
   }
   if (dc_global_sec != num) {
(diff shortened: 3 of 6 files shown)


lib/jxl/decode.cc
```diff
@@ -451,6 +451,11 @@ struct JxlDecoderStruct {
   bool have_container;
   size_t box_count;
 
+  // The number of progressive steps in frame decoding.
+  // 0 = only after DC image ready
+  // 1 = also after full pass of group is ready
```
jonsneyers, line 456: This assumes single-frame images. What about multi-layer images (with zero-duration frames that get blended/combined)? Is the first step only when the DC of the final layer is available? It could make sense to show intermediate results, i.e. show the current blended image at each kRegularFrame (that would finally introduce an actual difference between kRegularFrame and kSkipProgressive).

Going from coarser to finer grained, I think this is the order:
- Show only fully decoded last/non-zero duration frames — i.e. not progressive at all, e.g. for loading a local file
- Show all fully decoded kRegularFrame frames, also the ones with zero duration (so multi-layer images get updated for every layer) — useful if we want to do things like layered encoding (e.g. starting with a subsampled layer), or something like a screenshot that is encoded with a modular layer for the nonphoto parts and one or more vardct frames for the photo parts
- Show DC images too (1:8) — useful for cheap basic progression
- Show full AC passes that are the last pass for their resolution target (so at most four: one for 1:8, one for 1:4, one for 1:2 and one for 1:1) — this would be useful when you don't need full resolution, so you can stop decoding e.g. when 1:2 is available
- Show all full AC passes — useful for general progressive decode, including saliency progression
- Show DC progressively too (e.g. 1:32, 1:16, only possible in case of dc frame) — useful to replace LQIP placeholders
- Show incomplete passes (trigger progression event whenever a group has been decoded) — useful when using e.g. center-first group ordering

The current value 0 is the third bullet and 1 is the fifth bullet.

I'm not saying we urgently need to implement all of these levels of granularity right now, but it would be good to take into account that more detail levels may be added later, and leave room for them.

veluca93: This reminds me: kSkipProgressive frames should skip emitting progressive events at all.

lib/jxl/dec_frame.h
```diff
@@ -135,13 +144,20 @@ class FrameDecoder {
...
+  };
+
   // If enabled, ProcessSections will stop and return true when the DC
   // sections have been processed, instead of starting the AC sections. This
   // will only occur if supported (that is, flushing will produce a valid
   // 1/8th*1/8th resolution image). The return value of true then does not mean
   // all sections have been processed, use HasDecodedDC and HasDecodedAll
   // to check the true finished state.
-  void SetPauseAtProgressive() { pause_at_progressive_ = true; }
+  void SetPauseAtProgressive(uint32_t prog_detail) {
```
veluca93, line 158: Why doesn't this just take the enum directly?

lib/jxl/dec_frame.h
```diff
@@ -285,7 +301,7 @@ class FrameDecoder {
   // Testing setting: whether or not to use the slow rendering pipeline.
   bool use_slow_rendering_pipeline_;
 
-  bool pause_at_progressive_ = false;
+  int pause_at_progressive_ = 0;
```
veluca93, line 304: ProgressiveDetail, not int

lib/jxl/dec_frame.cc
```diff
@@ -748,29 +748,31 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
     FinalizeDC();
     JXL_RETURN_IF_ERROR(AllocateOutput());
     if (pause_at_progressive_ && !single_section) {
-      bool can_return_dc = true;
+      bool can_do_progressive = true;
```
veluca93, line 751: This logic should go out of the DC handling, no?

lib/jxl/dec_frame.h
```diff
@@ -89,6 +89,29 @@ class FrameDecoder {
...
+    // target.
+    kLastPasses = 4,
+    // after completed AC passes that are not the last pass for their resolution
+    // target.
+    kPasses = 8,
+    // during DC Frame when lower resolution are completed (1:32, 1:16)
+    kDCProgressive = 16,
+    // after completed groups
+    kGroups = 32,
+  };
```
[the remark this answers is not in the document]
jonsneyers: So maybe to make it somewhat easier to use, kLastPasses could imply kDC (because DC can be seen as the 'zero-th' pass for the 1:8 resolution target), kPasses could imply kLastPasses, and kGroups could imply everything else.

lib/jxl/decode.cc
jonsneyers: > I added the 7 different stages of progressiveness you suggested to the enum. However in the api, we still expect only two different values. I don't know how to make this future-proof in the best way. Any suggestions?

Perhaps in the API you should be able to set a bitfield (as described by the ProgressiveDetail enum, but just as a uint32). Even when things are implemented, there will be no guarantee that any progressive event will be produced (e.g. single-group images will probably always just immediately do a full decode), so applications cannot expect that to happen anyway. So we could just allow setting any detail, and some of them will never actually make a difference (for now) but that's fine.

lib/include/jxl/decode.h
```diff
@@ -1287,9 +1287,11 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxSizeRaw(const JxlDecoder* dec,
  *
  * @param dec decoder object
  * @param detail at which level of detail to trigger JXL_DEC_FRAME_PROGRESSION
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as an
+ * invalid value for the progressive detail.
  */
-JXL_EXPORT void JxlDecoderSetProgressiveDetail(JxlDecoder* dec,
-                                               uint32_t detail);
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetProgressiveDetail(JxlDecoder* dec,
+                                                           uint32_t detail);
```
jonsneyers, line 1294: One question is if this should only affect when the progression event gets triggered, or if it should also affect what happens when a Flush is done on partial input. That way you can ignore the progression events but still make sure that e.g. only a limited amount of refreshes are done even if Flush is called multiple times.

veluca93: You don't need a new API function to do Flush only when more input is needed and a progressive event exists: you just need to remember whether you got a progressive event until the next time ProcessInput returns NEED_MORE_INPUT.

jonsneyers: Yes, that avoids doing multiple flushes per arriving bitstream chunk, but it may lead to the flush showing "too much" information, e.g. part of a scan when you only want to show complete scans.

lib/jxl/dec_frame.h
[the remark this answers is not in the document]
mo271: I'm making it linear scale, where higher values imply smaller ones.

lib/jxl/decode.cc
mo271: Now I made an enum outside, (which does not contain `kNone`, since we signal "don't use progression" in another way).
I raise an error saying it is not implemented when it is not `kDC` or `kPasses` for now.

mo271: kSkipProgressive check is added.

mo271:
I'm not entirely sure how to test this best. Should I set `cparams.progressive_mode=1` to get an image with multiple passes that triggers return of `JXL_DEC_FRAME_PROGRESSION` when the appropriate conditions are met?
When I did this, the old (with default settings) test failed since it was returned twice, but it only expected it once...

lib/include/jxl/types.h
```diff
@@ -115,6 +115,30 @@ typedef struct {
  */
 typedef char JxlBoxType[4];
 
+/** Types of progressive detail
+ * Currently only the following level of progressive detail is implemented:
+ *  - kDC (which implies kFrames)
+ *  - kPasses (which implies kLastPasses, kDC and kFrames)
```
jonsneyers, line 121: Maybe just write that every N implies events for 0..N?

mo271: done

jonsneyers (approved):

mo271:
added Test (that actually work).

Ready for review now.

mo271:
.. rebased and added Luca as a co-author..

lib/jxl/dec_frame.cc
```diff
@@ -797,24 +812,25 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
           return PrepareStorage(num_threads,
                                 decoded_passes_per_ac_group_.size());
         },
-        [this, &ac_group_sec, &num_ac_passes, &num, &sections, &section_status,
-         &has_error](size_t g, size_t thread) {
-          if (num_ac_passes[g] == 0) {  // no new AC pass, nothing to do.
+        [this, &ac_group_sec, &desired_num_ac_passes, &num, &sections,
+         &section_status, &has_error](size_t g, size_t thread) {
+          if (desired_num_ac_passes[g] ==
+              0) {  // no new AC pass, nothing to do.
```
veluca93, line 818: Can you move the comment one line up?

lib/jxl/dec_frame.h
```diff
@@ -286,6 +295,7 @@ class FrameDecoder {
   bool use_slow_rendering_pipeline_;
 
   bool pause_at_progressive_ = false;
+  JxlProgressiveDetail progressive_detail_ = kDC;
```
veluca93, line 298: Maybe this and pause_at_progressive_ should become a single field

lib/jxl/dec_frame.cc
```diff
@@ -737,6 +740,31 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
   }
   if (has_error) return JXL_FAILURE("Error in DC group");
 
+  bool do_progressive = false;
+  if (pause_at_progressive_) {
```
veluca93, line 744: Shouldn't we have a check about kRegularFrame here too?

lib/jxl/dec_frame.cc
```diff
@@ -737,6 +740,31 @@ Status FrameDecoder::ProcessSections(const SectionInfo* sections, size_t num,
...
+      // for DC and the entire full resolution image is available at once.
+      do_progressive = false;
+    }
+    if (!decoded_->metadata()->extra_channel_info.empty()) {
+      // If extra channels are encoded with modular without squeeze, they
+      // don't support DC. If the are encoded with squeeze, DC works in theory
+      // but the implementation may not yet correctly support this for Flush.
+      // Therefore, can't correctly pause for a progressive step if there is
+      // an extra channel (including alpha channel)
+      do_progressive = false;
```
veluca93, line 757: I'd add a TODO about checking whether this is still the case (@jonsneyers too)

lib/jxl/dec_frame.cc
mo271: sure

lib/jxl/dec_frame.h
veluca93: Right, don't have std::optional in C++11.. Nevermind then.

lib/jxl/dec_frame.cc
mo271: done

lib/jxl/dec_frame.cc
mo271: sure. I added the check later, but it makes sense to check as early as possible.

lib/include/jxl/decode.h
```diff
@@ -1279,17 +1279,15 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxSizeRaw(const JxlDecoder* dec,
...
  *
  * @param dec decoder object
  * @param detail at which level of detail to trigger JXL_DEC_FRAME_PROGRESSION
+ * @return JXL_DEC_SUCCESS on success, JXL_DEC_ERROR on error, such as an
+ * invalid value for the progressive detail.
  */
-JXL_EXPORT void JxlDecoderSetProgressiveDetail(JxlDecoder* dec,
-                                               uint32_t detail);
+JXL_EXPORT JxlDecoderStatus
+JxlDecoderSetProgressiveDetail(JxlDecoder* dec, JxlProgressiveDetail detail);
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/dec_frame.cc
jonsneyers: I think https://github.com/libjxl/libjxl/pull/607 and https://github.com/libjxl/libjxl/pull/629 should make progressive alpha work (assuming it is squeezed of course, if not then there is only 1:1 group data, no progressive previews). 
Best to test if it actually still works.

The example application from https://github.com/libjxl/libjxl/pull/606 can be useful as a start to make a debug tool that dumps the image at every progression event (for the various settings of progressive detail). It dumps .pam files (not .ppm or .pfm) exactly to make it possible to see what is happening with the alpha channel too.

lib/jxl/dec_frame.cc
mo271: actually the linter didn't allow it, moving it one line below.

veluca93 (approved):

PR #1394, opened 2022-05-06
New API to encode PackedPixelFiles to other formats

sboukortt:


lib/extras/codec_test.cc
```diff
@@ -396,6 +406,78 @@ TEST(CodecTest, TestWideGamut) {
 
 TEST(CodecTest, TestPNM) { TestCodecPNM(); }
 
+TEST(CodecTest, FormatNegociation) {
```
veluca93, line 409: negotiation

lib/extras/codec_test.cc
```diff
@@ -396,6 +406,78 @@ TEST(CodecTest, TestWideGamut) {
...
+  const std::vector<JxlPixelFormat> accepted_formats = {
+      {.num_channels = 4,
```
veluca93, line 411: I don't think this is standard C++11... (IIRC it's either C++17 or a GNU extension)

lib/extras/codec_test.cc
```diff
@@ -396,6 +406,78 @@ TEST(CodecTest, TestWideGamut) {
...
+      "third_party/wesaturate/500px/tmshre_riaphotographs_srgb8.png");
+  PackedPixelFile ppf;
+  ASSERT_TRUE(extras::DecodeBytes(Span<const uint8_t>(original_png),
+                                  ColorHints(), SizeConstraints(), &ppf));
+
+  const JxlPixelFormat& format = ppf.frames.front().color.format;
+  ASSERT_THAT(
+      png_encoder->AcceptedFormats(),
+      Contains(AllOf(
+          Field("num_channels", &JxlPixelFormat::num_channels,
```
veluca93, line 460: Ahhhh, pointers to members... :D

lib/extras/enc/apng.cc
```diff
@@ -61,9 +62,38 @@ namespace extras {
 
 namespace {
 
+class APNGEncoder : public Encoder {
+ public:
+  std::vector<JxlPixelFormat> AcceptedFormats() const override {
+    std::vector<JxlPixelFormat> formats;
+    for (const uint32_t num_channels : {1, 2, 3, 4, 5}) {
```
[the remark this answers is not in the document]
jonsneyers: PNG can only do 4 channels, and only RGBA, not CMYK.

CMYKA is a thing, but you'll need TIFF, MIFF or PSD for that.

lib/extras/codec_test.cc
sboukortt: Thanks, I’m adding this to my list of unnecessary spelling differences between English and French, next to conne**ct**ion/conne**x**ion, distor**t**ion/distor**s**ion, appear**a**nce/appar**e**nce and li**t**era**ll**y/li**tt**éra**l**ement.

lib/extras/codec_test.cc
sboukortt: Oh, right, I keep forgetting. Thanks.

lib/extras/enc/apng.cc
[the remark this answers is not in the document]
sboukortt: I had assumed that it was, but I just checked and actually, it would seem that not even CMYK is, as far as PNG is concerned.

lib/extras/enc/apng.h
```diff
@@ -17,10 +18,12 @@
 namespace jxl {
 namespace extras {
 
+std::unique_ptr<Encoder> GetAPNGEncoder();
+
 // Encodes `io` into `bytes`.
 Status EncodeImageAPNG(const CodecInOut* io, const ColorEncoding& c_desired,
```
veluca93, line 24: I assume eventually this won't be needed anymore, right?

lib/extras/enc/encode.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "lib/extras/enc/jpg.h"
+#endif
+#include "lib/extras/enc/pgx.h"
+#include "lib/extras/enc/pnm.h"
+#include "lib/jxl/base/printf_macros.h"
+
+namespace jxl {
+namespace extras {
+
+Status SelectFormat(const std::vector<JxlPixelFormat>& accepted_formats,
```
veluca93, line 26: This seems like something that could be useful enough to have in C / the C API :)

lib/extras/enc/encode.cc
```diff
@@ -0,0 +1,84 @@
...
+    JXL_WARNING("encoding %" PRIuS "-bit original to %" PRIuS " bits",
+                original_bit_depth, current_bit_depth);
+  }
+  return true;
+}
+
+std::unique_ptr<Encoder> Encoder::FromExtension(std::string extension) {
+  std::transform(
+      extension.begin(), extension.end(), extension.begin(),
+      [](char c) { return std::tolower(c, std::locale::classic()); });
```
veluca93, line 60: What will happen here if the extension is not valid ASCII?

veluca93: (TBC, "doing something weird" is fine with me, as long as it doesn't die)

lib/extras/enc/encode.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+class Encoder {
+ public:
+  static std::unique_ptr<Encoder> FromExtension(std::string extension);
+
+  virtual ~Encoder() = default;
+
+  virtual std::vector<JxlPixelFormat> AcceptedFormats() const = 0;
+  virtual Status Encode(const PackedPixelFile& ppf, EncodedImage* encoded_image,
```
veluca93, line 44: I have to say I don't really like the output parameter, but given that we don't have a StatusOr I don't think we have a choice here...

That said, can we document what happens to any already-existing data in `encoded_image`?

lib/extras/enc/jpg.cc
```diff
@@ -235,5 +233,90 @@ Status EncodeImageJPG(const CodecInOut* io, JpegEncoder encoder, size_t quality,
...
+                              option.second.second.data());
+    }
+  }
+  return false;
+}
+
+class JPEGEncoder : public Encoder {
+  std::vector<JxlPixelFormat> AcceptedFormats() const override {
+    std::vector<JxlPixelFormat> formats;
+    for (const uint32_t num_channels : {1, 3}) {
```
veluca93, line 256: FWIW, this one _does_ support CMYK, although I guess we don't really have a way to express that.

lib/extras/enc/jpg.cc
```diff
@@ -235,5 +233,90 @@ Status EncodeImageJPG(const CodecInOut* io, JpegEncoder encoder, size_t quality,
...
+    return formats;
+  }
+  Status Encode(const PackedPixelFile& ppf, EncodedImage* encoded_image,
+                ThreadPool* pool = nullptr) const override {
+    const auto& options = this->options();
+    int quality = 100;
+    auto it_quality = options.find("q");
+    if (it_quality != options.end()) {
+      std::istringstream is(it_quality->second);
+      JXL_RETURN_IF_ERROR(static_cast<bool>(is >> quality));
```
veluca93, line 273: Why not stoll?

lib/extras/enc/pnm.cc
```diff
@@ -32,6 +32,74 @@ namespace {
...
+    }
+    return true;
+  }
+};
+
+class PPMEncoder : public PNMEncoder {
+ public:
+  std::vector<JxlPixelFormat> AcceptedFormats() const override {
+    std::vector<JxlPixelFormat> formats;
+    for (const uint32_t num_channels : {1, 2, 3, 4}) {
```
veluca93, line 56: IIRC PPM is just 3 channels.

lib/extras/enc/pnm.cc
```diff
@@ -32,6 +32,74 @@ namespace {
...
+    }
+    return formats;
+  }
+};
+
+class PFMEncoder : public PNMEncoder {
+ public:
+  std::vector<JxlPixelFormat> AcceptedFormats() const override {
+    std::vector<JxlPixelFormat> formats;
+    for (const uint32_t num_channels : {1, 2, 3, 4}) {
```
veluca93, line 74: And so is PFM

lib/extras/enc/pnm.cc
```diff
@@ -32,6 +32,74 @@ namespace {
...
+    return formats;
+  }
+};
+
+class PGMEncoder : public PPMEncoder {
+ public:
+  std::vector<JxlPixelFormat> AcceptedFormats() const override {
+    std::vector<JxlPixelFormat> formats = PPMEncoder::AcceptedFormats();
+    for (auto it = formats.begin(); it != formats.end();) {
+      if (it->num_channels > 2) {
```
veluca93, line 93: PGM is only 1 channel too (no Gray+Alpha)

lib/extras/enc/encode.cc
sboukortt: Is it, considering that the encoding API is itself also C++?

lib/extras/enc/encode.cc
sboukortt: I don’t think it should die: https://en.cppreference.com/w/cpp/locale/tolower

> Returns the lowercase form of `ch` if one is listed in the locale, otherwise return `ch` unchanged.

lib/extras/enc/encode.h
sboukortt: Good point, thanks.

lib/extras/enc/jpg.cc
sboukortt: It appears to rely on exceptions for error handling: https://en.cppreference.com/w/cpp/string/basic_string/stol#Exceptions

lib/extras/enc/encode.cc
veluca93: Not for encoding to file, but I could imagine using this to pick what format to use for display, for example...

lib/extras/enc/jpg.cc
veluca93: Well, CMYK JPEGs exist, although how they get created I do not know :)

lib/extras/enc/jpg.cc
veluca93: Fair enough. I guess we could strtoll, but that's not exactly very nice either...

lib/extras/enc/jpg.cc
jonsneyers: We don't currently support reading nor writing of CMYK jpegs in that code, but it could be added, it shouldn't be too hard to add. Just can only decode them to pixels and encode them from pixels, not dct coeffs, since VarDCT has 3 channels.

(theoretically we could represent the CMY losslessly as dct coeffs, and only convert the K to pixels, but that would require something other than PackedPixelFiles and I don't think it's worth the trouble).

sboukortt: I think it makes sense to implement CMYK encoding later and add it to `AcceptedFormats` at that point, but in the meantime, it would seem strange for the encoder to advertise support for it when it is not actually implemented.

lib/extras/enc/encode.cc
veluca93: Sgtm.

veluca93 (approved):

PR #1402, opened 2022-05-09
Refactor the low memory rendering pipeline and fix patch + noise combination.

### LINKED ISSUE #1397: Regression in decoder in case of patches + noise
When both patches and noise are used (e.g. a screenshot encoded with --photon_noise), the decoder does something wrong in the patch rendering: patches disappear or end up shifted a little in position.

Older decoder does it fine, so likely a regression caused by the transition to the render pipeline approach.

veluca93:
This PR contains two related commits, one that fixes #1397:

    Fix the patch stage when combined with noise.
    
    The patch stage declared to affect the additional noise channels (which
    it doesn't). Because of this, the pipeline execution got confused about
    the amount of y axis padding that the stage required (as it assumes that
    it is the same for all channels for which it matters), resulting in
    bogus values for the y position and, thus, in patches being moved by 2
    pixels up.

And one that hopefully improves clarity for the rendering pipeline and aids future debugging:

    Refactor the low memory rendering pipeline.
    
    The code should now hopefully be more clear and undestandable.
    
    As a nice side effect, this remove a potential race condition and
    optimizes rendering of frames partially outside the current image.

### DIFF
--- a/lib/jxl/dec_cache.cc
+++ b/lib/jxl/dec_cache.cc
@@ -89,7 +89,9 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
   }
 
   if ((frame_header.flags & FrameHeader::kPatches) != 0) {
-    builder.AddStage(GetPatchesStage(&shared->image_features.patches));
+    builder.AddStage(
+        GetPatchesStage(&shared->image_features.patches,
+                        3 + shared->metadata->m.num_extra_channels));
   }
   if ((frame_header.flags & FrameHeader::kSplines) != 0) {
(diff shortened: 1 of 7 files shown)


jonsneyers:
Nice! That was a pretty subtle one!

Are there any other stages that might have the same problem?

veluca93:
> Nice! That was a pretty subtle one!
> 
> Are there any other stages that might have the same problem?

I don't think so, as far as I remember all other "processing" stages didn't blanket say "all channels".

jonsneyers (approved):

Issue #1400, opened 2022-05-09
pkg-config files are broken when CMAKE_INSTALL_{INCLUDE,LIB}DIR is absolute


alexshpilkin:
**Describe the bug**

As per title: `lib/jxl/libjxl.pc.in` and `lib/threads/libjxl_threads.pc.in` have

```
prefix=@INSTALLDIR@
exec_prefix=${prefix}
libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```

and so can’t handle absolute paths in `CMAKE_INSTALL_{INCLUDE,LIB}DIR`. This leads to broken .pc files on NixOS in particular.

Similar to https://github.com/open-source-parsers/jsoncpp/pull/1199. See “[Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files)” for a discussion of the problem and a suggested fix (I don’t know CMake myself, sorry).

**Environment**
```
$ uname -a
Linux localhost 5.15.36 #1-NixOS SMP Wed Apr 27 12:39:02 UTC 2022 x86_64 GNU/Linux
```

kmilos:
Yeah, not sure about this either - [usually all other OS-es then have to revert/patch this by converting absolute to relative](https://github.com/msys2/MINGW-packages/blob/fed9c186f9c795843116002112e1c8e2e33be8fa/mingw-w64-cjson/PKGBUILD#L47) so relocation would work...

You want to convert the NixOS absolute case to a relative one somehow (would some CMake equivalent of removing CMAKE_INSTALL_PREFIX from CMAKE_INSTALL_{INCLUDE,LIB}DIR work?), or treat is as an exception and let it be absolute like in the suggested jsoncpp example fix linked above...

For the first option, something like this?

```
if(IS_ABSOLUTE CMAKE_INSTALL_INCLUDEDIR)
  string(REPLACE CMAKE_INSTALL_PREFIX "\${prefix}" includedir_for_pc_file CMAKE_INSTALL_INCLUDEDIR)
endif()
```

mo271:
[...]

PR #1403, opened 2022-05-10
Allow adding prefix for pkg-config

mo271:
This fixes https://github.com/libjxl/libjxl/issues/1400.

### DIFF
--- a/lib/jxl.cmake
+++ b/lib/jxl.cmake
@@ -616,6 +616,20 @@ set(JPEGXL_LIBRARY_REQUIRES
 if(NOT JPEGXL_ENABLE_SKCMS)
   set(JPEGXL_LIBRARY_REQUIRES "${JPEGXL_LIBRARY_REQUIRES} lcms2")
 endif()
+
+# Allow adding prefix if CMAKE_INSTALL_INCLUDEDIR not absolute.
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(PKGCONFIG_TARGET_INCLUDES "${CMAKE_INSTALL_INCLUDEDIR}")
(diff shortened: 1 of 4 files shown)


mo271:
I'm not sure if this is the correct way of fixing this.

mo271:
Could you take a look, @alexshpilkin?!

alexshpilkin:
Yes, I’m sorry @mo271. I’ve been trying to get the existing Nixpkgs recipe to build this version, but it isn’t as straightforward as I’ve hoped because some change since 0.6.1 seems to mess with LD_LIBRARY_PATH and cause the tests to fail to run because they can’t find libgmock. I’ll have another look today.

mo271:
> Yes, I’m sorry @mo271. I’ve been trying to get the existing Nixpkgs recipe to build this version, but it isn’t as straightforward as I’ve hoped because some change since 0.6.1 seems to mess with LD_LIBRARY_PATH and cause the tests to fail to run because they can’t find libgmock. I’ll have another look today.

Cool, thanks! Perhaps dissecting it would help; might have been this change: https://github.com/libjxl/libjxl/pull/575?!

alexshpilkin:
Neither of those is it, sorry, and Git (2.36.0)’s submodule support is so horribly broken, it seems, that it can’t even switch from this PR’s branch to the tag for 0.6.1, let alone bisect between them. I give up, I’ll leave figuring out how to build the eventual version 0.7 to the actual Nixpkgs maintainer.

That said, I did get the build to proceed far enough to generate the .pc files, and those look fine to me, so if this PR works in your more conventional environment it should be OK to merge. Thank you for the quick response, and I apologize for the confusion about who was representing what in the issue discussion.

Issue #1396, opened 2022-05-07
Crash in djxl from a file with JPEG reconstruction data with XMP in the JFIF file but no xml JXL box


rouault:
**Describe the bug**
djxl from a file, that has JPEG reconstruction data with XMP in the JFIF file but no corresponding xml box in the JXL file, to JPEG crashes

**To Reproduce**
djxl out4.jxl out.jpg
with out4.jxl from [out4.jxl.zip](https://github.com/libjxl/libjxl/files/8645667/out4.jxl.zip)

**Expected behavior**
should not crash

**Environment**
 - OS: Linux
 - Compiler version: g++ 9.4.0
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL decoder v0.7.0 2f872a1 [AVX2,SSE4,SSSE3,Scalar]

**Additional context**
```
$ valgrind  ~/libjxl/build/tools/djxl out4.jxl out.jpg
==1311894== Memcheck, a memory error detector
==1311894== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1311894== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==1311894== Command: /home/even/libjxl/build/tools/djxl out4.jxl out.jpg
==1311894== 
JPEG XL decoder v0.7.0 2f872a1 [AVX2,SSE4,SSSE3,Scalar]
Read 757 compressed bytes.
==1311894== Invalid read of size 8
==1311894==    at 0x128D58: jpegxl::tools::DecodeJpegXlToJpeg(jxl::DecompressParams, jpegxl::tools::JpegXlContainer const&, jxl::CodecInOut*, jxl::ThreadPool*) (box.cc:318)
==1311894==    by 0x121FB9: jpegxl::tools::DecompressJxlToJPEG(jpegxl::tools::JpegXlContainer const&, jpegxl::tools::DecompressArgs const&, jxl::ThreadPool*, jxl::PaddedBytes*, jpegxl::tools::SpeedStats*) (djxl.cc:209)
==1311894==    by 0x11D111: jpegxl::tools::(anonymous namespace)::DecompressMain(int, char const**) (djxl_main.cc:116)
==1311894==    by 0x4FFD0B2: (below main) (libc-start.c:308)
==1311894==  Address 0x8 is not stack'd, malloc'd or (recently) free'd
```

mo271:
[...]

PR #1404, opened 2022-05-10
checking if xml empty

mo271:
This prevents crashing for certain jxl files and thereby fixes https://github.com/libjxl/libjxl/issues/1396.

### DIFF
--- a/tools/box/box.cc
+++ b/tools/box/box.cc
@@ -315,13 +315,16 @@ jxl::Status DecodeJpegXlToJpeg(jxl::DecompressParams params,
     if (jpeg_data->app_marker_type[i] == jxl::jpeg::AppMarkerType::kXMP) {
       if (have_xmp)
         return JXL_FAILURE("Unexpected: more than one XMP box required?");
-      if (jpeg_data->app_data[i].size() != container.xml[0].second + 32) {
-        return JXL_FAILURE(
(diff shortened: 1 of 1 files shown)


mo271:
It is not quite clear to me if this then really reconstructs the jpeg faithfully.

veluca93 (approved):

tools/box/box.cc
```diff
@@ -315,13 +315,16 @@ jxl::Status DecodeJpegXlToJpeg(jxl::DecompressParams params,
...
       if (have_xmp)
         return JXL_FAILURE("Unexpected: more than one XMP box required?");
-      if (jpeg_data->app_data[i].size() != container.xml[0].second + 32) {
-        return JXL_FAILURE(
-            "XMP box size does not match JPEG reconstruction data");
+      // Container xml might be empty if xmp data not in xml box.
+      if (!container.xml.empty()) {
+        if (jpeg_data->app_data[i].size() != container.xml[0].second + 32) {
+          return JXL_FAILURE(
+              "XMP box size does not match JPEG reconstruction data");
```
jonsneyers, line 322: This error (or a variant saying there is no XMP box at all) should also be returned in case XMP is empty while jpeg reconstruction expects to have one.

Not that it matters too much, this is legacy code that will become redundant when djxl gets replaced by djxl_ng.

Issue #1401, opened 2022-05-09
`Passes::GetDownsamplingBracket` is wildly different from the ISO specification


lifthrasiir:
I've noticed this during the independent reimplementation efforts. As specified, `minshift` is determined to be `downsample[n]` if there is an index `n` such that `last_pass[n]` is equal to the pass index, otherwise it's set to `maxshift`. In the current implementation though, `n` is the largest index such that `last_pass[n]` is **no less than** the pass index, or **0** if there's no such `n`.

Theoretically, and with my limited understanding of the modular mode, this can result in the following problems:

* Some modular group data will be placed in wrong passes. By itself they will decode just fine (as the decoder and encoder have the exactly same bug), but it degrades the progressive decoding performance.
* Some modular group data will be *omitted* (concrete example: `num_passes` = 2, `num_ds` = 1, `downsample[0]` = 2, `last_pass[0]` = 1). I believe the bitstream still decodes, but the decoded image would no longer match the original.
* Some modular group data will be *duplicated* (concrete example: `num_passes` = 3, `num_ds` = 2, `downsample[0..1]` = [4, 2], `last_pass[0..1]` = [1, 0]; keep in mind that the bitstream has zero guarantee about the sortedness of these arrays). I believe the bitstream still decodes, but the same channel might get painted multiple times so again the decoded image would no longer match the original.

Thankfully the current libjxl only emits five combinations of `Passes`, which greatly limit the issue's extent. The following is a summary:

| Case | A | B | C | D | E |
|---|---|---|---|---|---|
| **`--progressive_ac`** | Yes | Yes | Yes | Yes | No |
| **`--saliency_num_progressive_steps=`** | 1 | 2 | 3 | 4¹ | - |
| | | | | | |
| **`num_passes`** | 1 | 2 | 3 | 4 | 2 |
| **`downsample[0 .. num_ds-1]`** | - | 4 | 4, 2 | 4, 2 | 2 |
| **`last_pass[0 .. num_ds-1]`** | - | 0 | 0, 1 | 0, 1 | 0 |
| | | | | | |
| **`[minshift, maxshift)` (spec)** | [0,3) | [2,3), [0,2) | [2,3), [1,2), [0,1) | [2,3), [1,2), [1,1), [0,1) | [1,3), [0,1) |
[...]

PR #1406, opened 2022-05-10
follow spec regarding shift brackets in multi-pass modular

jonsneyers:
See the discussion in https://github.com/libjxl/libjxl/issues/1401

This makes libjxl follow the spec regarding minshift/maxshift for modular passes, bringing the modular part of a pass in sync with the vardct part.

It also adds checks to require that the downscale sequence is strictly decreasing, so [8, 4, 2, 1] is allowed but not [2, 4] or [4, 4]. Also the last_pass sequence is required to be strictly increasing.

These constraints should ensure that all shifts are always covered exactly once (modulo brainos on my end), and should be added to the spec too.

This should not break any bitstreams encoded with "normal" cjxl options (i.e. the ones available in the non-verbose help screen).  It also does not break any bitstreams encoded with "advanced" cjxl options (the ones available in the help screen with one `-v`). It does break bitstreams encoded with "experimental" cjxl options (the ones you only see when doing `-v -v -h`), in particular when using this combination of options: `--progressive_ac --saliency_num_progressive_steps={3,4} -m` (or without the `-m` if they have an alpha channel). Things break in the sense that after this change, the encoder will produce bitstreams that fail to decode by an older decoder, and the decoder will fail to decode bitstreams that were encoded by an older encoder.

The extra checks should not affect any existing bitstreams, nor should they reduce the expressivity of the bitstream in any meaningful way.

### DIFF
--- a/lib/jxl/frame_header.cc
+++ b/lib/jxl/frame_header.cc
@@ -120,10 +120,16 @@ Status Passes::VisitFields(Visitor* JXL_RESTRICT visitor) {
     for (uint32_t i = 0; i < num_downsample; ++i) {
       JXL_QUIET_RETURN_IF_ERROR(
           visitor->U32(Val(1), Val(2), Val(4), Val(8), 1, &downsample[i]));
+      if (i > 0 && downsample[i] >= downsample[i - 1]) {
+        return JXL_FAILURE("downsample sequence should be decreasing");
+      }
     }
     for (uint32_t i = 0; i < num_downsample; ++i) {
       JXL_QUIET_RETURN_IF_ERROR(
           visitor->U32(Val(0), Val(1), Val(2), Bits(3), 0, &last_pass[i]));
+      if (i > 0 && last_pass[i] <= last_pass[i - 1]) {
+        return JXL_FAILURE("last_pass sequence should be increasing");
+      }
       if (last_pass[i] >= num_passes) {
         return JXL_FAILURE("last_pass %u >= num_passes %u", last_pass[i],
                            num_passes);
--- a/lib/jxl/frame_header.h
+++ b/lib/jxl/frame_header.h
@@ -262,10 +262,10 @@ struct Passes : public Fields {
 
   void GetDownsamplingBracket(size_t pass, int& minShift, int& maxShift) const {
     maxShift = 2;
-    minShift = 0;
+    minShift = 3;
     for (size_t i = 0;; i++) {
       for (uint32_t j = 0; j < num_downsample; ++j) {
-        if (i <= last_pass[j]) {
+        if (i == last_pass[j]) {
           if (downsample[j] == 8) minShift = 3;
           if (downsample[j] == 4) minShift = 2;
           if (downsample[j] == 2) minShift = 1;
@@ -275,7 +275,6 @@ struct Passes : public Fields {
       if (i == num_passes - 1) minShift = 0;
       if (i == pass) return;
       maxShift = minShift - 1;
-      minShift = 0;
     }
   }
 


veluca93:
> > As for downscale: I'd be happier with allowing it to be weakly decreasing, but at most once, i.e. allowing a downscale value to be repeated. This allows segmenting the pass into two parts in VarDCT mode (i.e. foreground and background); for modular mode, we could decide that a repeated downscale value just doesn't encode any channels.
> 
> What would be the advantage of being able to signal that e.g. pass 1 and 2 are both the last pass for 1:2, as opposed to just having those two passes but only signaling one of them to be the last pass for 1:2? Remember, num_passes can be larger than num_ds, so you can have intermediate passes between two resolutions but they're just not "last passes".
> 
> I don't think we lose expressivity by requiring every downsample resolution to have at most one last pass (except resolution 1:1 which always has the last pass as implicit "last pass" but you can optionally signal an earlier pass to be "already good enough for 1:1").
> 
> For example, what can still be done is this: num_passes = 11, num_ds = 4, and have downsample = [8,4,2,1], last_pass = [1,5,7,9] which would mean:
> 
> * somehow, DC and pass 0 isn't considered enough data yet for a "good enough" 1:8 image
> * after pass 1 you finally are ready to call it "good enough for 1:8"
> * passes 2, 3 and 4 do whatever they want to do in vardct
> * (up to this point, no modular data has been sent in passes, only the modular dcgroups are available, i.e. everything with shift >= 3)
> * after pass 5 things are "good enough for 1:4", this is also the pass where shift==2 modular data gets sent
> * pass 6 does whatever it wants in vardct, no modular data sent here
> * after pass 7 things are "good enough for 1:2", this is also the pass where shift==1 modular data gets sent
> * pass 8 does whatever it wants in vardct, no modular data sent here
> * after pass 9 things are "good enough for 1:1", this is the pass where the final (shift==0) modular data gets sent
> * pass 10 can add some more (presumably non-salient) vardct data

This sounds reasonable, although I'd perhaps make it so that modular data is sent with the first pass immediately after the previous shift (i.e. pass 2 for shift >=3). WDYT?

jonsneyers:
> > > As for downscale: I'd be happier with allowing it to be weakly decreasing, but at most once, i.e. allowing a downscale value to be repeated. This allows segmenting the pass into two parts in VarDCT mode (i.e. foreground and background); for modular mode, we could decide that a repeated downscale value just doesn't encode any channels.
> > 
> > 
> > What would be the advantage of being able to signal that e.g. pass 1 and 2 are both the last pass for 1:2, as opposed to just having those two passes but only signaling one of them to be the last pass for 1:2? Remember, num_passes can be larger than num_ds, so you can have intermediate passes between two resolutions but they're just not "last passes".
> > I don't think we lose expressivity by requiring every downsample resolution to have at most one last pass (except resolution 1:1 which always has the last pass as implicit "last pass" but you can optionally signal an earlier pass to be "already good enough for 1:1").
> > For example, what can still be done is this: num_passes = 11, num_ds = 4, and have downsample = [8,4,2,1], last_pass = [1,5,7,9] which would mean:
> > 
> > * somehow, DC and pass 0 isn't considered enough data yet for a "good enough" 1:8 image
> > * after pass 1 you finally are ready to call it "good enough for 1:8"
> > * passes 2, 3 and 4 do whatever they want to do in vardct
> > * (up to this point, no modular data has been sent in passes, only the modular dcgroups are available, i.e. everything with shift >= 3)
> > * after pass 5 things are "good enough for 1:4", this is also the pass where shift==2 modular data gets sent
> > * pass 6 does whatever it wants in vardct, no modular data sent here
> > * after pass 7 things are "good enough for 1:2", this is also the pass where shift==1 modular data gets sent
> > * pass 8 does whatever it wants in vardct, no modular data sent here
> > * after pass 9 things are "good enough for 1:1", this is the pass where the final (shift==0) modular data gets sent
> > * pass 10 can add some more (presumably non-salient) vardct data
> 
> This sounds reasonable, although I'd perhaps make it so that modular data is sent with the first pass immediately after the previous shift (i.e. pass 2 for shift >=3). WDYT?

Well the shift >= 3 modular data is considered part of DcGlobal or DcGroups (what fits in one group goes in global, the rest goes in dcgroups), so passes is really only about shifts 0, 1 and 2. If we want to change that, it's a major spec change and it would break any image using modular (except the non-progressive ones that only have shift==0 channels and are bigger than one group).

The encoder can choose where at which passes it puts the modular data (and whether to send each shift separately or do two or all three shifts together in one pass) by adjusting the downsample/last_pass array, so I think in practice it doesn't make that much of a difference, at least not when assuming a decoder that is showing all passes. For a decoder that shows only the last_passes, it does make a difference in which exact pass what modular data is available. Spec says to send the modular data basically as late as possible to meet the downsample goal (with the assumption that with all shift >= log2(N) channels, you can reconstruct an image "good enough for 1:N"). I don't see a strong reason why it would be better to change the spec so modular data gets sent earlier or later than that.

jonsneyers:
In the case of multiple passes at the same 1:n resolution, you could set the 1:n modular data to be sent together with the first of those passes (by declaring that pass to be the last_pass for 1:n).

veluca93:
> In the case of multiple passes at the same 1:n resolution, you could set the 1:n modular data to be sent together with the first of those passes (by declaring that pass to be the last_pass for 1:n).

Yes, but then the last_pass information becomes misleading at best :)

jonsneyers:
> > In the case of multiple passes at the same 1:n resolution, you could set the 1:n modular data to be sent together with the first of those passes (by declaring that pass to be the last_pass for 1:n).
> 
> Yes, but then the last_pass information becomes misleading at best :)

I guess that depends on how you interpret the semantics of that what it means to be a last_pass. Spec says it's the pass needed to "achieve a downsampling factor of [at] least downsample[i]"  (hey, a spec typo! and shouldn't it be "at most"?)

That's a pretty vague description. If you interpret it to mean that at that point the image is pixel-identical to what the full image would look like when both are downscaled (using what method?) with that downscaling factor, then yes, it would be misleading, since in that case it wouldn't be sufficient to have e.g. just the foreground/salient part at that resolution, it would have to be the full image.
But that interpretation doesn't really make sense when we also allow an explicit last_pass for 1:1,  since the implicit final pair (1, num_passes - 1) can always be the only actual last_pass for 1:1 if the last pass contains any information.

So I think the interpretation of "achieving a downsampling factor" should not be as strict as that. After all, 4:2:0 images exist where even the final image is to some extent not really 1:1. I think it's fine if the interpretation is a somewhat more vague "all the modular data and 'the most important' vardct data is available at that resolution", leaving open the possibility that maybe the unimportant background is still at a lower resolution but the main foreground object is actually available already.

Either way, whatever we decide to do with this, we should probably clarify in both spec and libjxl documentation what the semantics is of a last_pass.

szabadka (approved):

szabadka:
I would like to merge this, because the fuzzers are starting to find bitstreams where modular decoding is incomplete and this leads to msan failures. I had to add some extra failure conditions lately, probably to the wrong place, since this would most likely fix those.

jonsneyers:
I think it's fine to merge it as it is now. Perhaps an issue could be opened regarding the discussion above on when exactly what modular data gets sent in case of multiple passes for the same resolution target. My personal opinion on it is that it probably doesn't matter enough to justify additional behavior/spec change, but I'm OK with still changing it if that doesn't break any existing 'normal' bitstreams.

PR #1419, opened 2022-05-12
Move third_party/testdata to testdata/

eustas:
testdata is not completely "third_party"


mo271:
testdata does have a difference license than libjxl, though.
Do you think therefore it would be good to keep it under thirdparty?

mo271 (approved):

PR #1428, opened 2022-05-18
fix range check for RCT

jonsneyers:
The valid range for the rct type is [0,41].
The spec also needs to be updated to require that range.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -281,7 +281,7 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_COLOR_TRANSFORM = 24,
 
-  /** Reversible color transform for modular encoding: -1=default, 0-48=RCT
+  /** Reversible color transform for modular encoding: -1=default, 0-41=RCT
    * index, e.g. index 0 = none, index 6 = YCoCg.
    * If this option is set to a non-default value, the RCT will be globally
    * applied to the whole frame.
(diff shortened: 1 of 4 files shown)


veluca93 (commented):
Can we use named constants, maybe in the API headers?

jonsneyers:
> Can we use named constants, maybe in the API headers?

We don't really seem to do that in the API for the ranges of options. Not saying it's a bad idea, but then maybe we should do it for all options (or at least the ones with a more interesting range than [-1,1]), also e.g. for effort etc.

To be honest this particular option is probably not useful at all, since manually setting a specific global RCT is very likely to produce worse results than doing the default thing (which is trying different RCTs per group). The same is true for the option to manually set a specific global predictor. Perhaps we should deprecate / get rid of such not-very-useful options now we still can, to reduce clutter in the API?

veluca93:
> > Can we use named constants, maybe in the API headers?
> 
> We don't really seem to do that in the API for the ranges of options. Not saying it's a bad idea, but then maybe we should do it for all options (or at least the ones with a more interesting range than [-1,1]), also e.g. for effort etc.
> 
> To be honest this particular option is probably not useful at all, since manually setting a specific global RCT is very likely to produce worse results than doing the default thing (which is trying different RCTs per group). The same is true for the option to manually set a specific global predictor. Perhaps we should deprecate / get rid of such not-very-useful options now we still can, to reduce clutter in the API?

Sounds good to me.

veluca93 (approved):

PR #1430, opened 2022-05-19
Update AUTHORS check

mo271:
This makes sure that not only the last commit in a chain of commits is checked for authors.

At the moment this does not optimize by reducing the number of authors by removing duplicates before running the check; it basically runs the old check for each commit since the common ancestor commit.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1413,11 +1413,14 @@ cmd_bump_version() {
 # Check that the AUTHORS file contains the email of the committer.
 cmd_authors() {
   merge_request_commits
-  # TODO(deymo): Handle multiple commits and check that they are all the same
-  # author.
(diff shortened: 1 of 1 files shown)


mo271:
I intentionally added an fake author in order to test this. 
Once the authors test fails, I will squash it with a real author only...

mo271:
> I intentionally added an fake author in order to test this. Once the authors test fails, I will squash it with a real author only...

Seems to work: https://github.com/libjxl/libjxl/runs/6506304282?check_suite_focus=true

```
Checking name 'Defininitely Not A.N. Author' with email 'fake@email.com' ...
User Defininitely Not A.N. Author <fake@email.com> not found, please add yourself to the AUTHORS file
Individual authors are sorted alphabetically.
+ retcode=1
```

PR #1431, opened 2022-05-19
Debug build for Windows

sboukortt:


jonsneyers:
I guess even with `CI:full` this new build doesn't get triggered, since it's part of something that only gets done on pushes to main. Shall we just try merging this?
(alternatively we could build artifacts also for PRs, or at least for CI:full ones?)

jonsneyers (approved):

sboukortt:
> I guess even with `CI:full` this new build doesn't get triggered, since it's part of something that only gets done on pushes to main. Shall we just try merging this? (alternatively we could build artifacts also for PRs, or at least for CI:full ones?)

We managed to work around this by pushing to the main branch of a fork and creating a release: https://github.com/sboukortt/libjxl/releases/tag/windebug

But the debugging symbols are for Visual Studio, and I also forgot to check whether they were embedded in the executables or written as separate `.pdb` files that should be included in the archive as well. (I still don’t know.)

mo271:
Can this be merged, @sboukortt?

sboukortt:
I don’t remember whether we actually wanted to. I think this might have been meant as a one-off to get binaries with such debug info from the CI. @veluca93, do you remember?

PR #1435, opened 2022-05-23
Add command line flag to log adaptive quantization search states.

szabadka:


tools/benchmark/benchmark_codec_jxl.cc
```diff
@@ -85,7 +86,10 @@ Status AddCommandLineOptionsJxlCodec(BenchmarkArgs* args) {
   args->AddOverride(&jxlargs->patches, "patches",
                     "Enable(1)/disable(0) patch dictionary.");
 
-  args->AddString(
+  args->AddFlag(&jxlargs->log_search_state, "log_search_state",
+                "Print out debug info for tortoise mode AQ loop.", false);
+
+ args->AddString(
```
sboukortt, line 92: It seems that the indentation got accidentally changed here.

szabadka: Fixed.

PR #1436, opened 2022-05-23
Using jxl::extras::Encoder in djxl_ng

mo271:
This replaces the direct call to jxl::extras::EncodeImagePNM.

### DIFF
--- a/tools/djxl_ng_main.cc
+++ b/tools/djxl_ng_main.cc
@@ -22,6 +22,7 @@
 #include "jxl/thread_parallel_runner_cxx.h"
 #include "jxl/types.h"
 #include "lib/extras/dec/decode.h"
+#include "lib/extras/enc/encode.h"
 #include "lib/extras/enc/pnm.h"
 #include "lib/extras/packed_image.h"
 #include "lib/jxl/base/printf_macros.h"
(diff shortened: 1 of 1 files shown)


tools/djxl_ng_main.cc
```diff
@@ -425,52 +425,44 @@ int main(int argc, char** argv) {
...
-        JXL_WARNING("For color images, the filename should end with .ppm.\n");
-      }
-      if (num_channels == 3 && ppf.info.num_color_channels == 1) {
-        JXL_WARNING(
-            "For grayscale images, the filename should end with .pgm.\n");
-      }
-      if (ppf.info.bits_per_sample > 16) {
-        JXL_WARNING("PPM only supports up to 16 bits per sample");
-      }
+    if (!jxl::extras::SelectFormat(encoder->AcceptedFormats(), ppf.info,
```
sboukortt, line 439: What is `format` used for outside of `DecompressJxlToPackedPixelFile`? Do we need this?

tools/djxl_ng_main.cc
```diff
@@ -425,52 +425,44 @@ int main(int argc, char** argv) {
...
+    }
+    jxl::extras::EncodedImage encoded_image;
+    JXL_RETURN_IF_ERROR(encoder->Encode(ppf, &encoded_image));
+    if (ppf.frames.size() != encoded_image.bitstreams.size()) {
+      fprintf(stderr,
+              "didn't receive expected number of frames: %" PRIuS
+              " (expected) versus %" PRIuS "(received)\n",
+              ppf.frames.size(), encoded_image.bitstreams.size());
+      return EXIT_FAILURE;
     }
```
sboukortt, line 452: I am not completely sure it’s for this function to check that `Encode` returned the correct number of bitstreams (which could be 1 if the format supports animations). I think I would favour trusting `Encode` to return an appropriate number of bitstreams or an error if one occurs, and replacing the instances of `ppf.frames.size()` below with `encoded_image.bitstreams.size()`. What do you think?

tools/djxl_ng_main.cc
```diff
@@ -425,52 +425,44 @@ int main(int argc, char** argv) {
...
-      if (ppf.info.bits_per_sample > 16) {
-        JXL_WARNING("PPM only supports up to 16 bits per sample");
-      }
+    if (!jxl::extras::SelectFormat(encoder->AcceptedFormats(), ppf.info,
+                                   &format)) {
+      fprintf(stderr, "SelectFormat failed\n");
+      return EXIT_FAILURE;
+    }
+    jxl::extras::EncodedImage encoded_image;
+    JXL_RETURN_IF_ERROR(encoder->Encode(ppf, &encoded_image));
```
sboukortt, line 445: If this fails, `JXL_RETURN_IF_ERROR` will return `false`, which will mean a return code equal to `EXIT_SUCCESS`.

tools/djxl_ng_main.cc
mo271: We don't need it; removed.

tools/djxl_ng_main.cc
mo271: changed the check

sboukortt (approved):

PR #1438, opened 2022-05-23
cmake: Do not add static suffix with static libraries in mingw

Biswa96:
This fixes static linking using pkgconfig files in mingw.
Otherwise, linker can not find static library showing this

ld.exe: cannot find -ljxl:No such file or directory
ld.exe: cannot find -ljxl_threads: No such file or directory

Signed-off-by: Biswapriyo Nath <nathbappai@gmail.com>


sboukortt (approved):

sboukortt:
Thanks!

PR #1444, opened 2022-05-25
pkgconfig: link -lstdc++ to Libs.private

shinchiro:
When static linking libjxl with other program like ffmpeg will throw error:
```
../lib/libjxl.a(decode.cc.obj):decode.cc:(.text+0x376): undefined reference to `operator new(unsigned int)'
../lib/libjxl.a(decode.cc.obj):decode.cc:(.text+0x421): undefined reference to `operator delete(void*)'
../lib/libjxl.a(decode.cc.obj):decode.cc:(.text+0x483): undefined reference to `std::__throw_length_error(char const*)'
../lib/libjxl.a(decode.cc.obj):decode.cc:(.text+0x4a1): undefined reference to `__gxx_personality_sj0'
../lib/libjxl.a(decode.cc.obj):decode.cc:(.text+0x5df): undefined reference to `operator delete(void*)'
```
Adding `-lstdc++` to Libs.private fixed it.

### DIFF
--- a/lib/jxl/libjxl.pc.in
+++ b/lib/jxl/libjxl.pc.in
@@ -8,5 +8,5 @@ Description: Loads and saves JPEG XL files
 Version: @JPEGXL_LIBRARY_VERSION@
 Requires.private: @JPEGXL_LIBRARY_REQUIRES@
 Libs: -L${libdir} -ljxl
-Libs.private: -lm
+Libs.private: -lm -lstdc++
 Cflags: -I${includedir}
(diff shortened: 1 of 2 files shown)


Biswa96:
I think c or c++ libraries should not be in pkgconfig file. For example, here libstdc++ is valid for gcc based toolchain. But for clang it should be libc++. Hence, while static linking users have to add those standard libraries explicitly.

If standard c/c++ libraries were required in pkgconfig file then every projects would have those.

PR #1452, opened 2022-05-27
Update HWY

eustas:


ci.sh
```diff
@@ -46,7 +46,7 @@ SANITIZER="none"
 if [[ "${BUILD_TARGET%%-*}" == "x86_64" ||
     "${BUILD_TARGET%%-*}" == "i686" ]]; then
   # Default to building all targets, even if compiler baseline is SSE4
-  HWY_BASELINE_TARGETS=${HWY_BASELINE_TARGETS:-HWY_SCALAR}
+  HWY_BASELINE_TARGETS=${HWY_BASELINE_TARGETS:-HWY_EMU128}
```
sboukortt, line 49: I am not sure I understand the implications of this line change? It now seems to contradict the comment above unless I am missing something.

eustas: Taken from original @jan-wassenberg PR.
Jan, is `HWY_EMU128` our new "baseline"? In my understanding, this is scalar + 4x unroll (for float32)...

jan-wassenberg: That's right. Both do not use intrinsics, but EMU128 has the advantage that it supports all ops and does not require special-case code like in convolve-inl.h. Thus it is meant to replace/supersede scalar.

The comment is trying to explain the following: say the baseline copts allow SSE4. Then S-SSE3 and scalar/emu128 will not be included because they are worse than SSE4. By overriding the baseline to the minimum (which is now changing to emu128), we prevent skipping any targets. Does that make sense?

sboukortt (approved):

Issue #1454, opened 2022-05-27
`ci.sh lint` is broken


eustas:
Run: `ci.sh lint`
Result:
```
fatal: no upstream configured for branch 'exta'
Warning, not tracking any branch, using the last commit in HEAD.
```
However, it worked before.

PR #1455, opened 2022-05-27
fix linting

mo271:
This was caused by the (fairly recent) change https://github.com/llvm/llvm-project/issues/53220


mo271:
closes #1454

Issue #1079, opened 2022-01-08
libjxl should not consume input beyond the end of a JPEG XL file


thebombzen:
libjxl's current API reports how much was consumed and how much was not consumed by libjxl, but it will consume past the end of a codestream and simply do nothing with this information. For raw JPEG XL Codestream input into libjxl's decoder, it should refuse to consume input after the end of the codestream. For libjxl's container input, it should refuse to consume input after the last box, and if a `jxlc` or `jxlp` box appears with size `0`, it should refuse to consume input after the codestream ends.

jonsneyers:
I think this one might be important for ffmpeg, so perhaps we should prioritize it a bit.

PR #1458, opened 2022-05-30
Stop consuming input in the decoder after codestream is finished.

szabadka:
This fixes two cases in Issue #1079, the raw codestream case and the
unbounded jxlc/jxlp box at the end of the input case.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -593,6 +593,9 @@ struct JxlDecoderStruct {
   // all bytes at once, to save memory. Find alternative to std::vector doubling
   // strategy to prevent some memory usage.
   std::vector<uint8_t> codestream_copy;
+  // Number of bytes at the end of codestream_copy that were not yet consumed
+  // by calling AdvanceInput().
+  size_t codestream_unconsumed;
   // Position in the actual codestream, which codestream_copy.begin() points to.
   // Non-zero once earlier parts of the codestream vector have been erased.
   // TODO(lode): use this variable to allow pruning codestream_copy
@@ -749,6 +752,7 @@ void JxlDecoderRewindDecodingState(JxlDecoder* dec) {
   dec->frame_header.reset(new jxl::FrameHeader(&dec->metadata));
 
   dec->codestream_copy.clear();
+  dec->codestream_unconsumed = 0;
 
   dec->frame_stage = FrameStage::kHeader;
   dec->frame_start = 0;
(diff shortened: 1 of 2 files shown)


lib/jxl/decode.cc
```diff
@@ -2041,9 +2045,10 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
       if (have_copy) {
         // TODO(lode): prune the codestream_copy vector if the codestream
         // decoder no longer needs data from previous frames.
-        dec->codestream_copy.insert(dec->codestream_copy.end(), dec->next_in,
+        dec->codestream_copy.insert(dec->codestream_copy.end(),
+                                    dec->next_in + dec->codestream_unconsumed,
                                     dec->next_in + avail_codestream);
-        dec->AdvanceInput(avail_codestream);
+        dec->codestream_unconsumed = avail_codestream;
```
eustas, line 2051: Do we assume `dec->codestream_unconsumed == 0`; if yes, lets add `JXL_CHECK` for that.

lib/jxl/decode_test.cc
```diff
@@ -1685,6 +1685,41 @@ TEST(DecodeTest, PixelTestOpaqueSrgbLossyNoise) {
...
+  size_t xsize = 123, ysize = 77;
+  size_t num_pixels = xsize * ysize;
+  std::vector<uint8_t> pixels = jxl::test::GetSomeTestImage(xsize, ysize, 3, 0);
+  jxl::CompressParams cparams;
+  CodeStreamBoxFormat box_formats[3] = {
+      kCSBF_None,
+      kCSBF_Single_Zero_Terminated,
+      kCSBF_Multi_Zero_Terminated,
+  };
+  for (int i = 0; i < 3; ++i) {
```
eustas, line 1698: Perhaps for-in would be more readable here.

eustas (approved):

lib/jxl/decode.cc
szabadka: dec->codestream_unconsumed is not necessarily zero here, only if the last call to ProcessInput returned with JXL_DEC_NEED_MORE_INPUT, otherwise we keep copying from where we left off.

lib/jxl/decode_test.cc
szabadka: Done.

lib/jxl/decode.cc
eustas: But `AdvanceInput` was cumulative... and rewriting is not...

eustas (approved):

lib/jxl/decode.cc
szabadka: It is still cumulative, because if we do not consume the input via AdvanceInput, the next time the same input will still be there, possibly with some more bytes at the end (this is required by the API), and then avail_codestream will increase with that same number of bytes. But if we return NEED_MORE_INPUT, then we advance and codestream_unconsumed will be zero.

eustas: Got it. Thanks for the clarification.

PR #1460, opened 2022-05-30
trying to fix failing cross compile

mo271:


mo271:
#1459 has a problem running the cross-compiling tests

mo271:
> #1459 has a problem running the cross-compiling tests

now it seems to work...

PR #1463, opened 2022-05-31
Fix crash in decoder due to calling Flush() before dc was ready.

szabadka:
This was found by running the test case in Issue #919


veluca93:
That seems to be going in the wrong direction, we probably should instead create the RP or something like that.

szabadka:
> That seems to be going in the wrong direction, we probably should instead create the RP or something like that.

Could be, I was looking at the public API function, JxlDecoderFlushImage() that also refuses to flush without dc.

veluca93:
> > That seems to be going in the wrong direction, we probably should instead create the RP or something like that.
> 
> Could be, I was looking at the public API function, JxlDecoderFlushImage() that also refuses to flush without dc.

We probably should fix that too then (or decide we don't want to support it)

szabadka:
I created a post1.0 issue #1465 to implement the Flush support before dc is ready, but before that is done, we should not be crashing in the decoder.

lib/jxl/dec_frame.cc
```diff
@@ -984,20 +984,15 @@ Status FrameDecoder::FinalizeFrame() {
...
+    // can not call Flush() yet.
+    return JXL_FAILURE("FinalizeFrame called before DC was fully decoded");
   }
   if (!HasEverything() && !allow_partial_frames_) {
     return JXL_FAILURE(
         "FinalizeFrame called before the frame was fully decoded");
   }
 
-  if (!finalized_dc_) {
-    JXL_ASSERT(allow_partial_frames_);
```
eustas, line 997: Perhaps we should move this assert remaining `if (!finalized_dc_)`.

eustas (approved):

PR #1467, opened 2022-06-01
remove unreachable line

mo271:


veluca93:
This will cause extra warnings on compilers that don't recognize [[no return]], of which IIRC we had at least one of :)

mo271:
When compiling chrome with default settings, I got:

```
../../third_party/libjxl/src/lib/jxl/render_pipeline/stage_epf.cc:510:10: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
```
When you say "we had at least one of", do you mean in our pipeline? Is that compiler still used?
Let me run the full pipeline...

veluca93:
> When compiling chrome with default settings, I got:
> 
> ```
> ../../third_party/libjxl/src/lib/jxl/render_pipeline/stage_epf.cc:510:10: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
> ```
> 
> When you say "we had at least one of", do you mean in our pipeline? Is that compiler still used? Let me run the full pipeline...

I don't remember where we got those...

If we have both compilers that complain with the line, and compilers that don't, then we have a bit of a problem :) let's hope not...

mo271:
full pipeline seems to pass

veluca93 (approved):

PR #1471, opened 2022-06-02
prevent infinite loop in box decoding

mo271:
After the commit https://github.com/libjxl/libjxl/commit/5a024e5bf5b9b93da698c7f6cb06abb874b73c4c, the jxl decoder in chrome goes into an infinite loop when trying to decoder a raw jxl image. 
This happens in `HandleBoxes` and more precisely this happens when `dec->avail_in == 0` and` dec->box_stage` switches back and forth between `BoxStage::kHeader` and `BoxStage::kCodestream`.

The proposed changed stops the infinite loop and makes the chrome jxl decoder work again. 
(I haven't run any tests yet..)

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -2092,7 +2092,8 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
           dec->AdvanceInput(avail_codestream);
         }
 
-        if (dec->file_pos == dec->box_contents_end) {
+        if (dec->file_pos == dec->box_contents_end &&
+            !dec->box_contents_unbounded) {
           dec->box_stage = BoxStage::kHeader;
           continue;
         }
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -21,6 +21,7 @@
 #include "lib/extras/enc/jpg.h"
 #include "lib/jxl/base/byte_order.h"
 #include "lib/jxl/base/file_io.h"
+#include "lib/jxl/base/padded_bytes.h"
 #include "lib/jxl/base/span.h"
 #include "lib/jxl/base/status.h"
 #include "lib/jxl/common.h"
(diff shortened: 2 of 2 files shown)


veluca93 (dismissed):

lib/jxl/decode.cc
```diff
@@ -2094,7 +2094,7 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
 
         if (dec->file_pos == dec->box_contents_end) {
           dec->box_stage = BoxStage::kHeader;
-          continue;
+          return status;
```
szabadka, line 2097: Do you need this return status; line here? It seems that just deleting the continue; is enough, because we do have return status; later at the end of this block.

mo271: That is even better, because less confusing! Thanks

mo271:
This now works in chrome, perhaps we could a test similar to  
`ExtraBytesAfterCompressedStream` to check if everything works as intended with `file_pos == 0`, and doesn't go in an infinite loop in that case.

lib/jxl/decode.cc
```diff
@@ -2092,7 +2092,7 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
           dec->AdvanceInput(avail_codestream);
         }
 
-        if (dec->file_pos == dec->box_contents_end) {
+        if (dec->file_pos == dec->box_contents_end && dec->box_size != 0) {
```
szabadka, line 2095: To be consistent with the rest of the file, please use !dec->box_contents_unbounded instead of dec->box_size != 0

szabadka:
> This now works in chrome, perhaps we could a test similar to `ExtraBytesAfterCompressedStream` to check if everything works as intended with `file_pos == 0`, and doesn't go in an infinite loop in that case.

Yes, it would be helpful to add a test case to this pull request that would trigger the infinite loop without this fix.

lib/jxl/decode.cc
mo271: Done

mo271:
> > This now works in chrome, perhaps we could a test similar to `ExtraBytesAfterCompressedStream` to check if everything works as intended with `file_pos == 0`, and doesn't go in an infinite loop in that case.
> 
> Yes, it would be helpful to add a test case to this pull request that would trigger the infinite loop without this fix.

Added a test which closely mimics the order things are done in the chrome decoder. Without the fix, this test triggers the infinite loop.

lib/jxl/decode_test.cc
```diff
@@ -1710,6 +1711,42 @@ TEST(DecodeTest, PixelTestOpaqueSrgbLossyNoise) {
...
+    CodeStreamBoxFormat box_format = (CodeStreamBoxFormat)i;
+    printf("Testing empty input with box format %d\n", (int)box_format);
+    jxl::PaddedBytes compressed = jxl::CreateTestJXLCodestream(
+        jxl::Span<const uint8_t>(pixels.data(), pixels.size()), xsize, ysize, 3,
+        cparams, box_format, JXL_ORIENT_IDENTITY, /*add_preview=*/false,
+        /*add_intrinsic_size=*/false);
+    const int events =
+        JXL_DEC_BASIC_INFO | JXL_DEC_FULL_IMAGE | JXL_DEC_COLOR_ENCODING;
+    EXPECT_EQ(JXL_DEC_SUCCESS,
+              JXL_DEC_SUCCESS != JxlDecoderSubscribeEvents(dec, events));
```
szabadka, line 1731: Why the != here?

lib/jxl/decode_test.cc
```diff
@@ -1710,6 +1711,42 @@ TEST(DecodeTest, PixelTestOpaqueSrgbLossyNoise) {
...
+    EXPECT_EQ(JXL_DEC_SUCCESS,
+              JxlDecoderImageOutBufferSize(dec, &format, &buffer_size));
+    JxlBasicInfo info;
+    EXPECT_EQ(JXL_DEC_SUCCESS, JxlDecoderGetBasicInfo(dec, &info));
+    const size_t remaining = JxlDecoderReleaseInput(dec);
+    EXPECT_LE(remaining, compressed.size());
+    EXPECT_EQ(JXL_DEC_NEED_MORE_INPUT, JxlDecoderProcessInput(dec));
+    EXPECT_EQ(JXL_DEC_SUCCESS,
+              JxlDecoderSetInput(dec, compressed.data(), compressed.size()));
+  }
```
szabadka, line 1747: dec should be released here

lib/jxl/decode_test.cc
```diff
@@ -1710,6 +1711,42 @@ TEST(DecodeTest, PixelTestOpaqueSrgbLossyNoise) {
...
+    size_t buffer_size;
+    EXPECT_EQ(JXL_DEC_SUCCESS,
+              JxlDecoderImageOutBufferSize(dec, &format, &buffer_size));
+    JxlBasicInfo info;
+    EXPECT_EQ(JXL_DEC_SUCCESS, JxlDecoderGetBasicInfo(dec, &info));
+    const size_t remaining = JxlDecoderReleaseInput(dec);
+    EXPECT_LE(remaining, compressed.size());
+    EXPECT_EQ(JXL_DEC_NEED_MORE_INPUT, JxlDecoderProcessInput(dec));
+    EXPECT_EQ(JXL_DEC_SUCCESS,
+              JxlDecoderSetInput(dec, compressed.data(), compressed.size()));
```
szabadka, line 1746: This is technically a violation of the API contract, since the buffer should start with whatever was released by JxlDecoderReleaseInput(). But anyway, since there is nothing tested after, maybe this can be removed.

szabadka (changes requested):

lib/jxl/decode_test.cc
mo271: oops, this was just leftover somehow. Fixed.

lib/jxl/decode_test.cc
mo271: removed the `JxlDecoderSetInput`, since the infinity loop is triggered earlier anyhow.

lib/jxl/decode_test.cc
mo271: removed, see comment above

lib/jxl/decode_test.cc
szabadka: JxlDecoderDestory should be still called here

mo271: done

szabadka (approved):

PR #1472, opened 2022-06-02
Fix progressive decoding for streaming input.

szabadka:
Define a progression detail level for each new frame right after
we decode the frame header and save the state of what progression
levels were already signaled in the internal JxlDecoderStruct.

Added byte-by-byte progression to the ProgressiveEventTest.


mo271 (approved):
Nice work!

PR #1473, opened 2022-06-03
Implement kLastPasses progressive detail in the decoder.

szabadka:
Added test with 4 intermediate passes, 2 for each downsampling
target. For this I had to fix a bug in progressive_splitter that
filled in the last_pass array of the frame header incorrectly.


mo271 (approved):
looks perfect!
... and nice fix for progressive_split

PR #1474, opened 2022-06-03
slightly larger float error in a test

mo271:
This is currently causing errors when Cross-compiling i686-linux-gnu.


mo271:
Cross-compiling i686-linux-gnu still failing, but due to other tests, which I propose to handle in a different pr.

szabadka (approved):

PR #1482, opened 2022-06-08
Deprecate the --allow_progressive_steps flag from djxl.

szabadka:
This change has the same effect as if setting --allow_progressive_steps
to the same value as --allow_partial_files.

If the --allow_partial_files is set to true, then we will always
decode and render as much detail as possible.


mo271 (approved):
LGTM

PR #1485, opened 2022-06-08
Implement CMake exports

amyspark:
This PR implements installable Config Modules for libjxl, based on the namespace defined by Sony Interactive Entertainment in WebKit [1] . For reference, my version of the Find Module [2] adds lookup for jxl_threads; these need to be patched here just as was done in #1414.

I'm open to any suggestions, especially as regards definition of `JPEGXL_LIBRARIES` and `JPEGXL_INCLUDE_DIRS`.

[1]: https://github.com/WebKit/WebKit/blob/main/Source/cmake/FindJPEGXL.cmake

[2]: https://invent.kde.org/graphics/krita/-/blob/master/cmake/modules/FindJPEGXL.cmake


amyspark:
I've been reviewing this and, due to the way libjxl builds third_party components, it is not possible to use CMake exports with it. See this bug: https://gitlab.kitware.com/cmake/cmake/-/issues/15415

The main blocker is that libjxl, for libraries that are built from source, directly embeds the project instead of going through ExternalProject_Add + the relevant Find Module that would import the library (e.g. [OpenColorIO](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/1d26d58c31cedf65c91954f319f5e2fc8ea429b5/share/cmake/modules/Findpystring.cmake)). This will cause affected dependencies to not be exported, and at present CMake is unable to pierce past it.

PR #1487, opened 2022-06-09
correct extended box size handling

### LINKED ISSUE #1478: libjxl cannot read extended ISOBMFF box sizes
**Describe the bug**
A file that has an extended box size fails to be read by either libjxl or djxl.

**To Reproduce**
An extended box size is when an ISOBMFF box has `size == 1`. When this happens, the size field is immediately followed by an 8-byte extended-size field, which contains a 64-bit true size. These are permitted even if the extended size is less than 4G. 

Take any JXL file with an extended box size and attempt to decode it with djxl or libjxl. It will not decode properly.

**Expected behavior**
The file should decode properly.

**Other information**
I believe the fault here is in `decode.cc` line 1768. https://github.com/libjxl/libjxl/blob/ab775e2298b1ef25229e19ad3f4dba4b06000197/lib/jxl/decode.cc#L1768
The extended box size field is read after the box tag field, but it should be *immediately* after the box size field, and before the box tag field. Contrast the existing behavior with the FFmpeg `mov.c` decoder: https://github.com/ffmpeg/FFmpeg/blob/35ec5c819b0f472536eb48e8f3871eb5ca222da6/libavformat/mov.c#L7813

mo271:
closes #1478

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -1783,14 +1783,15 @@ static JxlDecoderStatus ParseBoxHeader(const uint8_t* in, size_t size,
   size_t box_start = pos;
   // Box size, including this header itself.
   *box_size = LoadBE32(in + pos);
-  memcpy(type, in + pos + 4, 4);
-  pos += 8;
+  pos += 4;
   if (*box_size == 1) {
     *header_size = 16;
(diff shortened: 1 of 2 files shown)


mo271:
Not sure if we should add a test for this. It seems in principle possible to have a test even for a files with boxes with size less than $2^{32}$ bytes.

PR #1490, opened 2022-06-09
Add tests for rendering partially decoded streams with API.

szabadka:
The tests only document the current state of API decoding, and
show that rendering on partial input produces new details when
we have a whole group available, but does not produce new details
if we only have part of a group available.
This is the case for both lossy and lossless modes.


mo271 (approved):
LGTM!

PR #1493, opened 2022-06-10
Allow libjxl to tone map HDR images when decoding them

sboukortt:


szabadka:
Shouldn't you also change djxl.cc? At least to remove the --preserve_saturation flag, since it will not be supported by API.

sboukortt:
> Shouldn't you also change djxl.cc? At least to remove the --preserve_saturation flag, since it will not be supported by API.

As discussed offline, it is probably fine to keep this for now.

lib/extras/tone_mapping.cc
```diff
@@ -34,42 +34,13 @@ Status ToneMapFrame(const std::pair<float, float> display_nits,
...
-        MulAdd(t_b_3 + MulAdd(Set(df, -2), t_b_2, t_b), Set(df, 1) - ks,
-               MulAdd(Set(df, -2), t_b_3, Set(df, 3) * t_b_2) * max_lum));
-  };
-
-  const V inv_max_display_nits = Set(df, 1 / display_nits.second);
+  Rec2408ToneMapper<decltype(df), V> tone_mapper(
+      df,
+      {ib->metadata()->tone_mapping.min_nits,
+       ib->metadata()->IntensityTarget()},
+      display_nits, /*red_Y=*/0.2627f, /*green_Y=*/0.6780f, /*blue_Y=*/0.0593f);
```
veluca93, line 41: Can we get a comment on where the constants come from? :)

lib/jxl/dec_tone_mapping-inl.h
```diff
@@ -0,0 +1,203 @@
...
+#include <hwy/highway.h>
+
+#include "lib/jxl/transfer_functions-inl.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace jxl {
+namespace HWY_NAMESPACE {
+namespace {
+
+template <typename D, typename V>
```
veluca93, line 22: I don't think we need to pass both D and V, we can get V as hwy::Vec<D>, no?

lib/jxl/dec_tone_mapping-inl.h
```diff
@@ -0,0 +1,203 @@
...
+class Rec2408ToneMapper {
+ public:
+  explicit Rec2408ToneMapper(D df, std::pair<float, float> source_range,
```
veluca93, line 25: Is there a reason to pass a df instead of constructing one where needed?

veluca93: (one reason could be, if this was not a constructor but a free function, to get deduction for the template parameter)

veluca93, line 25: Any specific reason for the `explicit`? (also below)

lib/jxl/dec_tone_mapping-inl.h
```diff
@@ -0,0 +1,203 @@
...
+                      :  // inverse gamma for the inverse OOTF
+                      (1 / 1.2f) *
+                              std::pow(1.111f,
+                                       -std::log2(target_luminance / 1000.f)) -
+                          1),
+        red_Y_(red_Y),
+        green_Y_(green_Y),
+        blue_Y_(blue_Y) {}
+
+  template <typename D, typename V>
```
veluca93, line 133: Also here we probably don't need the V.

lib/jxl/dec_tone_mapping-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  const float exponent_;
+  const bool apply_ootf_ = exponent_ < -0.01f || 0.01f < exponent_;
+  const float red_Y_;
+  const float green_Y_;
+  const float blue_Y_;
+};
+
+template <typename D, typename V>
+void GamutMap(D df, V* red, V* green, V* blue, float red_Y, float green_Y,
+              float blue_Y, float preserve_saturation = 0.1f) {
```
veluca93, line 158: supernit: I'd think a "preserve_saturation" parameter to be a bool

veluca93 (changes requested):

lib/jxl/dec_tone_mapping-inl.h
sboukortt: Good point, updated.

lib/jxl/dec_tone_mapping-inl.h
sboukortt: Changed the other way around as discussed.

lib/jxl/dec_tone_mapping-inl.h
sboukortt: I prefer to always have it rather than have to think about it and accidentally miss it when it becomes useful (e.g. when giving certain parameters default values).

(Also, for what it’s worth, it also disables implicit construction from initializer lists.)

lib/jxl/dec_tone_mapping-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  explicit HlgOOTF(float source_luminance, float target_luminance, float red_Y,
+                   float green_Y, float blue_Y)
+      : exponent_(
+            std::pow(1.111f, std::log2(target_luminance / source_luminance)) -
+            1),
+        red_Y_(red_Y),
+        green_Y_(green_Y),
+        blue_Y_(blue_Y) {}
+
+  // If one end of the conversion is scene light
```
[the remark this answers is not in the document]
sboukortt: Done, thanks.

PR #1495, opened 2022-06-13
Implement --pixels_to_jpeg mode in djxl_ng.

szabadka:
This change also wires up the --jpeg_quality and --use_sjpeg flags
and fixes --num_reps > 1 mode.

Added an end-to-end test for djxl_ng.

### DIFF
--- a/tools/roundtrip_test.sh
+++ b/tools/roundtrip_test.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# End-to-end roundtrip tests for cjxl and djxl tools.
+
+MYDIR=$(dirname $(realpath "$0"))
+JPEGXL_TEST_DATA_PATH="${MYDIR}/../testdata"
+
+set -eux
+
+# Temporary files cleanup hooks.
+CLEANUP_FILES=()
+cleanup() {
+  if [[ ${#CLEANUP_FILES[@]} -ne 0 ]]; then
+    rm -rf "${CLEANUP_FILES[@]}"
+  fi
+}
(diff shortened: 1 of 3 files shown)


tools/roundtrip_test.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Conformance test tooling test. This is not the JPEG XL conformance test
+# runner. This test that the tooling to generate the conformance test and the
+# conformance test runner work together.
```
mo271, line 9: Let's adapt the doc string from tooling test.

mo271 (commented):
Great to have roundtrip testing. I was planning on doing this with python script instead of a shell script, in order to be able to easily compare pixels of images, but using ssimulacra for this purpose is also nice.

tools/roundtrip_test.sh
szabadka: Done.

mo271 (approved):

tools/roundtrip_test.sh
```diff
@@ -0,0 +1,89 @@
...
+trap 'retcode=$?; { set +x; } 2>/dev/null; cleanup' INT TERM EXIT
+
+roundtrip_test() {
+  local infn="${JPEGXL_TEST_DATA_PATH}/$1"
+  local encargs="$2"
+  local maxdist="$3"
+  
+  local encoder="${build_dir}/tools/cjxl"
+  local decoder="${build_dir}/tools/djxl_ng"
+  local comparator="${build_dir}/tools/ssimulacra_main"
```
mo271, line 30: In some build, like making the deb packages, `/tools/ssimulacra_main` is not build, see https://github.com/libjxl/libjxl/runs/6902801608?check_suite_focus=true

szabadka:
#1545 fixes this. For some reason I thought the jpeg reconstruction box always comes before basic info and in the test here it was the case, but apparently not always.

PR #1500, opened 2022-06-15
relax assert after new highway release

mo271:
@jan-wassenberg: would this be a good new constant? The first new failure after a highway release on cross-compile i686 in an emu128 test was occuring at a value for which `std::abs(approx[0] - approx[i])` was 2.3...E-7, so I approximately doubled that to have some wiggle room for future updates.


lib/jxl/enc_xyb.cc
```diff
@@ -86,13 +86,12 @@ void TestCubeRoot() {
 
     // All lanes are same
     for (size_t i = 1; i < Lanes(d); ++i) {
-      JXL_ASSERT(std::abs(approx[0] - approx[i]) <= 1.2E-7f);
+      JXL_ASSERT(std::abs(approx[0] - approx[i]) <= 5E-7f);
```
veluca93, line 89: Maybe I'm missing something here, but shouldn't this always be 0? the input to CubeRootAndAdd is the same for all lanes, and I don't see any lane crossing operation in CubeRootAndAdd -- it seems rather fishy to me if this fails (possibly indicative of a bug in hwy?)

For that matter, even the version with 1.2E-7f as an error bound seems suspicious.

veluca93:
as a more coding style level comment, we probably should move the whole function to something like fast_math-inl.h and test it in a _test.cc file with gtest assertions, rather than recycling JXL_ASSERTs for the purpose.

lib/jxl/enc_xyb.cc
mo271: I can confirm that even before the tests would not have passed with `0` instead of `1.2E-7f`.

mo271:
> as a more coding style level comment, we probably should move the whole function to something like fast_math-inl.h and test it in a _test.cc file with gtest assertions, rather than recycling JXL_ASSERTs for the purpose.

Sounds good to me. Should the fast_math-inl.h then only contain CubeRootAndAdd or more functions?

lib/jxl/enc_xyb.cc
veluca93: That seems very surprising to me. Do you know on which targets it happens?

@jan-wassenberg do you have an idea why this might be happening?

veluca93:
> > as a more coding style level comment, we probably should move the whole function to something like fast_math-inl.h and test it in a _test.cc file with gtest assertions, rather than recycling JXL_ASSERTs for the purpose.
> 
> Sounds good to me. Should the fast_math-inl.h then only contain CubeRootAndAdd or more functions?

fast_math-inl.h already exists I believe (perhaps it is named in a slightly different way)

lib/jxl/fast_math_test.cc
```diff
@@ -107,6 +107,26 @@ HWY_NOINLINE void TestFastErf() {
...
+    const float expected = cbrtf(x);
+    HWY_ALIGN float approx[MaxLanes(d)];
+    Store(CubeRootAndAdd(Set(d, x), Zero(d)), d, approx);
+
+    // All lanes are same
+    for (size_t i = 1; i < Lanes(d); ++i) {
+      EXPECT_LE(std::abs(approx[0] - approx[i]), 5E-7f);
+    }
+
+    const float err = std::abs(approx[0] - expected);
```
veluca93, line 124: Can we instead EXPECT_LE err, instead of computing max_err? This would tell us where things fail to satisfy the bound if we also << x.

lib/jxl/fast_math_test.cc
```diff
@@ -107,6 +107,26 @@ HWY_NOINLINE void TestFastErf() {
...
+  float max_err = 0.0f;
+  for (uint64_t x5 = 0; x5 < 2000000; x5++) {
+    const float x = x5 * 1E-5f;
+    const float expected = cbrtf(x);
+    HWY_ALIGN float approx[MaxLanes(d)];
+    Store(CubeRootAndAdd(Set(d, x), Zero(d)), d, approx);
+
+    // All lanes are same
+    for (size_t i = 1; i < Lanes(d); ++i) {
+      EXPECT_LE(std::abs(approx[0] - approx[i]), 5E-7f);
```
veluca93, line 121: https://google.github.io/googletest/reference/assertions.html#EXPECT_NEAR should give better error messages.

Can we see what failures we get with 0 here?

mo271: Let's take a look...

lib/jxl/fast_math_test.cc
veluca93: yep, absolutely

lib/jxl/fast_math_test.cc
mo271: So, the result are errors like this:

```
The difference between approx[0] and approx[i] is 2.384185791015625e-07, which exceeds 0, where
approx[0] evaluates to 2.635331392288208,
approx[i] evaluates to 2.6353311538696289, and
0 evaluates to 0.
```
I will set the tolerance to 5E-7f again, for greener tests.

veluca93 (approved):

jan-wassenberg:
I agree that magnitude of epsilon is fine, and can be expected to happen from replacing SCALAR with EMU128.

mo271:
> Just to clarify, EMU128 then does *not* guarantee that when running non-lane-crossing operations with inputs being vectors with identical lanes then the result will also have identical lanes? If so, where is this behaviour documented? It seems at best very surprising.


@jan-wassenberg can probably answer those questions, any thoughts here, especially in context of https://github.com/google/highway/issues/765?

PR #1512, opened 2022-06-17
detect CI:full differently

mo271:


mo271:
This works now, i.e. the tests are actually run when adding the CI:full label. However, since currently Build/Test / Windows MSYS2 / i686 (pull_request) if failing I cannot submit the pull request, because that is a required tests. I should be able to submit the pull request after removing the CI:full label and triggering the workflows again.

veluca93 (approved):

PR #1513, opened 2022-06-17
Increase stack size limit of the tests.

szabadka:
This seems to fix the asan test failure on ubuntu.


mo271 (approved):

mo271:
It will be impossible to commit with when only running the CI:full commit queue, since after the recent change, https://github.com/libjxl/libjxl/pull/1512, when the CI:full actually runs the tests
Build/Test / Windows MSYS2 / i686 fails because of emu128 stuff. So in order to get this submitted, I suggested first to wait for the CI:full to be completed (and check that this fixes the asan issue) and then remove the CI:full label, retrigger a run of the CI and get it submitted.

mo271 (approved):

PR #1501, opened 2022-06-15
Improved input handling in the decoder API implementation.

szabadka:
For JXL_DEC_NEED_MORE_INPUT events, we always consume everything,
except if we are in a signature, box header or the first 4 bytes of
brob/ftyp/jxlp boxes.

For JXL_DEC_BASIC info, we consume input until the end of ImageMetadata,
together with extensions, and we no longer return JXL_DEC_EXTENSIONS.

For JXL_DEC_COLOR_ENCODING, we consume input until end of all headers.

For JXL_DEC_PREVIEW_IMAGE, we consume input until end of preview frame.

For JXL_DEC_FRAME, we consume input until end of frame header.

Still missing is the input handling for the JXL_DEC_FRAME_PROGRESSION
events.

PR #1514, opened 2022-06-17
Improved input handling in the decoder API (part 2).

szabadka:
This is a continuation of PR #1501.

For JXL_DEC_FRAME_PROGRESSION events, we stop at the end of the
section that was required to produce the progression event.

If the caller of the API does not subscribe to the full image output,
the decoder now stops and returns JXL_DEC_SUCCESS at the last event
that was produced from all of the subscribed ones.

Added tests for input handling for one-shot and streaming input.

The decoder now clears the codestream_copy when it advances the
codestream, so we now only use it as a temporary buffer instead of
the copy of the whole codestream.


mo271 (approved):
LGTM

PR #1526, opened 2022-06-22
Fix cjxl_ng for inputs with an ICC profile

sboukortt:
`JxlEncoderSetICCProfile` was accidentally called twice.


mo271 (approved):
nice catch!

PR #1529, opened 2022-06-23
djxl_ng: check extension

mo271:


tools/djxl_ng_main.cc
```diff
@@ -460,6 +460,10 @@ int main(int argc, char** argv) {
   if (decode_to_pixels) {
     std::unique_ptr<jxl::extras::Encoder> encoder =
         jxl::extras::Encoder::FromExtension(extension);
+  if (encoder == nullptr) {
```
szabadka, line 463: indentation is off here

PR #1533, opened 2022-06-27
Use djxl_ng in the conformance tests.

szabadka:
This removes the code duplication of the decoding loop between
djxl_conformance.cc and djxl_ng.cc and adds more test coverage
to the djxl_ng tool.

The numpy array and json metadata generating code was moved from
djx_conformance.cc to a new kind of encoder in lib/extras/enc/npy.cc

This change also removed the internal jxl library from the djxl_ng tool,
we only link the shared jxl_dec and jxl_threads libraries now.


mo271:
cool, this also closes https://github.com/libjxl/libjxl/pull/1225, I guess...

szabadka:
> cool, this also closes #1225, I guess...

No, that seems to be a NumPy decoder, while here I moved the NumPy encoder from djxl_conformance

mo271 (approved):

PR #1536, opened 2022-06-27
Add informative flags to djxl_ng and make output file optional.

szabadka:
This change implements the --quiet and --print_read_bytes flag from
djxl and in case of missing output filename it calls the decoder
loop but discards the results.

### DIFF
--- a/tools/djxl_ng_main.cc
+++ b/tools/djxl_ng_main.cc
@@ -14,7 +14,6 @@
 #include <string>
 #include <vector>
 
-#include "gflags/gflags.h"
 #include "jxl/codestream_header.h"
 #include "jxl/decode.h"
 #include "jxl/decode_cxx.h"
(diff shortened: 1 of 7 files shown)


tools/djxl_ng_main.cc
```diff
@@ -621,50 +635,73 @@ bool DecompressJxlToPackedPixelFile(
...
-                           std::to_string((version / 1000) % 1000) + "." +
-                           std::to_string(version % 1000));
-  // TODO(firsching): rethink --help handling
+  std::string version = jpegxl::tools::CodecConfigString(JxlDecoderVersion());
+  gflags::SetVersionString(version);
+  gflags::SetUsageMessage(
+      "JPEG XL decoder\n\n"
+      "Usage: djxl_ng INPUT [OUTPUT] [FLAGS]");
   gflags::ParseCommandLineNonHelpFlags(&argc, &argv, /*remove_flags=*/true);
+  // Disable full help listing, only show flags from this file.
```
mo271, line 656: Potential problems with that approach:

 - the `-help` flag and the `-version` flag are not shown, when running `-help`.
 - personally, I like it when both `--help` and `-h` work for help, perhaps this can be done somehow.
 - (Other than `-h`, the three abbreviations we are dropping are `-s`, `-q` and `-j` and `-V`, which is fine, I guess?!)

szabadka: To get the exact same command-line interface as with djxl, we would need to keep using our own flag parser, since gflags does not have aliases for flags:

From the gflags github page:
"Note that flags do not have single-letter synonyms, like they do in the getopt library, nor do we allow "combining" flags behind a single dash, as in ls -la."

I would be happy to replace the gflags dependency with CommandLineParser.

szabadka: Now I use the same command line parser as djxl, we have -h, -s, -q, -j and -V

mo271 (approved):

PR #1538, opened 2022-06-27
encode: use -1 to let encoder decide level automatically

thebombzen:
Change the codestream level mechanics in the Codestream API to
allow the encoder to automatically choose level 10 codestreams
based on the information in the basic info. This should be ABI
compatible with the previous behavior, as explicitly setting the
level to 10 should still work, and letting it default to 5 will
work in every case it did before.

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -952,8 +952,8 @@ JXL_EXPORT JxlEncoderStatus
 JxlEncoderStoreJPEGMetadata(JxlEncoder* enc, JXL_BOOL store_jpeg_metadata);
 
 /** Sets the feature level of the JPEG XL codestream. Valid values are 5 and
- * 10. Keeping the default value of 5 is recommended for compatibility with all
- * decoders.
+ * 10, or -1 (to choose automatically). Using the minimum required level, or
+ * level 5 in most cases, is recommended for compatibility with all decoders.
  *
  * Level 5: for end-user image delivery, this level is the most widely
  * supported level by image decoders and the recommended level to use unless a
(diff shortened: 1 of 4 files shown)


lib/include/jxl/encode.h
```diff
@@ -969,16 +969,17 @@ JxlEncoderStoreJPEGMetadata(JxlEncoder* enc, JXL_BOOL store_jpeg_metadata);
...
- * explicitly set to 10, the encoder will not automatically enable it. If
- * incompatible parameters such as too high image resolution for the current
- * level are set, the encoder will return an error. For internal coding tools,
- * the encoder will only use those compatible with the level setting.
+ * The default value is -1. This means the encoder will automatically choose
+ * between level 5 and level 10 based on what information is inside the @ref
+ * JxlBasicInfo structure. Do note that some level 10 features, particularly
+ * those used by animated JPEG XL codestreams, might require level 10, even
+ * though the @ref JxlBasicInfo only suggests level 5. In this case, the level
+ * must be explicitly set to 10.
```
jonsneyers, line 977:
```suggestion
 * must be explicitly set to 10, otherwise the encoder will return an error.
 * The encoder will restrict internal encoding choices to those compatible with
 * the level setting.
```

thebombzen:
I force-pushed a change in the encode-test, since it was failing. The reason the test was failing is it expected a `1 << 30` x 64 image to return JXL_ENC_ERROR on `JxlSetBasicInfo` but this should not fail anymore, so I changed it to EXPECT_EQ(SUCCESS).

sboukortt:
> I force-pushed a change in the encode-test, since it was failing. The reason the test was failing is it expected a `1 << 30` x 64 image to return JXL_ENC_ERROR on `JxlSetBasicInfo` but this should not fail anymore, so I changed it to EXPECT_EQ(SUCCESS).

Given the possible intent of the test, it might make sense to keep expecting a failure but set the codestream to 5 explicitly first, or something along those lines. What do you think?

thebombzen:
> Given the possible intent of the test, it might make sense to keep expecting a failure but set the codestream to 5 explicitly first, or something along those lines. What do you think?

Done.

sboukortt (approved):

PR #1543, opened 2022-06-28
Fix build when asciidoc is wrapped using bash

SuperSandro2000:
Basically #1541 but with signed CLA

Description copied over from the other PR:

In https://github.com/libjxl/libjxl/pull/805 the build was fixed for the case of asciidoc being wrapped in a shell script with /bin/sh shebang. Some distros use bash instead.
cc @deymo


SuperSandro2000:
Added myself to the AUTHORS file so that the github action hopefully becomes green.

mo271 (approved):

PR #1548, opened 2022-06-29
Remove unused fields from DecompressParams.

szabadka:
This is in preparation to implementing DecodeFile using API.


szabadka:
Abandoning this in favor of #1558.

PR #1552, opened 2022-06-29
Fix unaligned load/store (when frame_origin is not 0)

eustas:


lib/jxl/render_pipeline/stage_xyb.cc
```diff
@@ -29,6 +29,7 @@ class XYBStage : public RenderPipelineStage {
     PROFILER_ZONE("UndoXYB");
 
     const HWY_FULL(float) d;
+    // TODO(eustas): check if it should be 'xsize + xextra'
```
veluca93, line 32: We can assert xextra to be 0 here.

eustas: Cool, will add it soon.

lib/jxl/render_pipeline/stage_xyb.cc
```diff
@@ -39,18 +40,20 @@ class XYBStage : public RenderPipelineStage {
     msan::UnpoisonMemory(row0 + xsize, sizeof(float) * (xsize_v - xsize));
     msan::UnpoisonMemory(row1 + xsize, sizeof(float) * (xsize_v - xsize));
     msan::UnpoisonMemory(row2 + xsize, sizeof(float) * (xsize_v - xsize));
+    // TODO(eustas): investigate if it is allowed to start 'earlier' for
```
veluca93, line 43: You can use up to kPipelinePaddingX (I don't remember the name) bytes before and after the data

eustas:
In regular use (without frame origin) addresses are aligned -> no performance penalty.
Could improve performance with frame origin somewhere in the future.

eustas:
@veluca93 PTAL. Let's postpone proper alignment until it is necessary. In the most common use-case address is aligned -> no penalty.

szabadka (approved):

PR #1553, opened 2022-06-29
Replace ICC profiles with an enum representation when appropriate

sboukortt:
Legacy `cjxl` already does this. This brings this behavior to the API and `cjxl_ng` as well.


sboukortt:
> RoundtripTest.TestICCProfile seems to be a new failure

Indeed, thanks. It should be fixed now.

szabadka (approved):

szabadka:
Coverage test was fixed, so if you rebase it should be possible to submit.

szabadka (approved):

PR #1558, opened 2022-07-01
Use djxl's decoding loop in benchmark and tests.

szabadka:
This caught a bug in the rendering pipeline where pixel callbacks were not
reset between frames and the patch-frame of the second displayed frame was
still writing to the pixel callback set on the first frame.

Added preview frame handling to the ppf->codec io conversion function.

The reference one-shot decoding loop is in lib/extras/dec/jxl.cc now and it is
used in both djxl and the benchmark, lib/jxl/dec_file.cc was removed and
DecodeFile was added to test_utils.h and uses the same decoding function as
djxl.

A custom API-free DecodeFile had to be added back to render pipeline test, since
we don't want to expose the internal slow rendering pipeline through the API.


mo271 (approved):

mo271:
this might interfere with https://github.com/libjxl/libjxl/pull/1559 (or even solve the .deb problem...)

szabadka:
> this might interfere with #1559 (or even solve the .deb problem...)

Now decoder is linked statically together with extras, so the .deb problem could be solved for now. Of course it would be nice if the whole extras would depend only through public library on libjxl/

PR #1562, opened 2022-07-01
Disable failing hwy tests on Windows build.

szabadka:


mo271:
we could also add the "DISABLED_" prefix to the test (and a TODO, of course)

mo271:
> we could also add the "DISABLED_" prefix to the test (and a TODO, of course)

ahh no, that would not work, this is not a normal test...

mo271 (approved):

PR #1563, opened 2022-07-01
Add clang32/64 MSYS environments

eustas:
Update HWY to ToT.

Also re-enable AVX targets and CLink tests

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -231,16 +231,21 @@ jobs:
         ./ci.sh gbench --benchmark_min_time=0
 
   windows_msys:
-    name: Windows MSYS2 / ${{ matrix.arch }}
+    name: Windows MSYS2 / ${{ matrix.msystem }}
     runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
         include:
-          - arch: x86_64
-            msystem: mingw64
-          - arch: i686
-            msystem: mingw32
+         - msystem: mingw64
+         - msystem: clang64
+         - msystem: mingw32
+           # TODO(eustas): investigate HWY Mul failures
+           disable_tests: HwyMulTestGroup/HwyMulTest\.TestAllMulHigh/Emu128|HwyMulTestGroup/HwyMulTest\.TestAllMulFixedPoint15/Emu128
+         - msystem: clang32
+           # TODO(eustas): investigate HWY Sort and JXL ANS failures
+           disable_tests: SortTestGroup/SortTest\.TestAllSort/.*|ANSTest\.RandomUnbalancedStreamRoundtrip3|ANSTest\.RandomUnbalancedStreamRoundtripBig
+
     defaults:
       run:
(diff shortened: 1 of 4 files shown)


kmilos:
Please try the CLANG64 environment/toolchain instead of just switching compiler in the MINGW64 environment:

https://www.msys2.org/docs/environments/

https://github.com/msys2/setup-msys2/blob/main/examples/cmake.yml 

Also, instead of installing the entire "toolchain" metapackage w/ pacboy, to reduce CI time one can do only the "cc" metapackage (gives you the default compiler for the given msys environment) and other basic requirements.

kmilos:
Same difference, you could be trying out CLANG32 then in addition to or vs MINGW32.

.github/workflows/build_test.yml
```diff
@@ -259,24 +268,24 @@ jobs:
...
-            mingw-w64-${{ matrix.arch }}-gtest
-            mingw-w64-${{ matrix.arch }}-libavif
-            mingw-w64-${{ matrix.arch }}-libjpeg-turbo
-            mingw-w64-${{ matrix.arch }}-libpng
-            mingw-w64-${{ matrix.arch }}-libwebp
-            mingw-w64-${{ matrix.arch }}-ninja
+          pacboy: >-
+            brotli:p
+            cmake:p
+            ${{ matrix.compiler }}:p
```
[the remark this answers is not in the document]
eustas: Cool! Thanks.

.github/workflows/build_test.yml
```diff
@@ -259,24 +264,23 @@ jobs:
...
+            cmake:p
+            gflags:p
+            giflib:p
+            gtest:p
+            libavif:p
+            libjpeg-turbo:p
+            libpng:p
+            libwebp:p
+            ninja:p
+            toolchain:p
```
kmilos, line 279: Nice, a much more intuitive action now 👍

Just out of curiousity, why was the entire `toolchain:p` needed? Usually just `cc:p` and (for C++ code) `gcc-libs:p` suffice, and might give some time savings by not installing all the toolchain packages... (Note that there is a `gcc-libs` metapackage available in clang environments for this reason, which will pull in libc++ and libunwind for you; there might missing corner cases though like openmp and libssp that are not part of that metapackage, do that's why I ask...)

eustas: `clang64` environment and with `cc` was missing libs. Will try with `gcc-libs` (well, package name is counterintuitive)

kmilos: Indeed, hence the lengthy explanation 😉 It was one way to extend the legacy mingw environments...

eustas: What's suspicious is that it tries to link 32-bit libs...

kmilos: The "32" is just the Win32 API, not the architecture...

Try adding `crt:p` to the list as well?

mo271 (dismissed):

PR #1566, opened 2022-07-02
Simplify CMake

kleisauke:
See the individual commits for more details.


kleisauke:
This will probably fix a few bugs as well. For example, the [`release:clang-5`](https://github.com/libjxl/libjxl/runs/7162119239) failed with:
```
CMake Error at CMakeLists.txt:162 (message):
  Minimum Clang version required is Clang 6, please update.
```

Which would imply that the previous `${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS *` checks were always returning `false`.

szabadka (approved):
Thanks for the cleanup!

PR #1573, opened 2022-07-04
document rewind

### LINKED ISSUE #632: Using JxlDecoderSetPreferredColorProfile with JxlDecoderRewind
Hello,

I use `JxlDecoderSetPreferredColorProfile` in my programs.

I observed that after `JxlDecoderRewind`, the preferred color profile is forgotten, and the image data decoded later corresponds to the image's default color profile (like the JxlDecoderSetPreferredColorProfile call was never used).

I have to subscribe for the `JXL_DEC_COLOR_ENCODING` event again, because that's the only place where JxlDecoderSetPreferredColorProfile could be set again now.

Is it desired behavior?
If it is desired, please add info to documentation.
Or could be the JxlDecoderRewind call be made to remember the color profile preference?

mo271:
Fixes #632

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -366,6 +366,15 @@ typedef enum {
  * events that were already handled before, such as @ref JXL_DEC_BASIC_INFO
  * and @ref JXL_DEC_COLOR_ENCODING, since they will provide the same information
  * as before.
+ * The difference to @ref JxlDecoderReset is that some state is kept, namely
+ * settings set by a call to
+ *  - @ref JxlDecoderSetCoalescing,
+ *  - @ref JxlDecoderSetDesiredIntensityTarget,
+ *  - @ref JxlDecoderSetDecompressBoxes,
+ *  - @ref JxlDecoderSetKeepOrientation,
+ *  - @ref JxlDecoderSetRenderSpotcolors,
+ *  - @ref JxlDecoderSkipFrames, and
+ *  - @ref JxlDecoderSubscribeEvents.
  *
  * @param dec decoder object
  */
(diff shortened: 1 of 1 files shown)


lib/include/jxl/decode.h
```diff
@@ -366,6 +366,15 @@ typedef enum {
...
  * and @ref JXL_DEC_COLOR_ENCODING, since they will provide the same information
  * as before.
+ * The difference to @ref JxlDecoderReset is that some state is kept, namely
+ * settings set by a call to
+ *  - @ref JxlDecoderSetCoalescing,
+ *  - @ref JxlDecoderSetDesiredIntensityTarget,
+ *  - @ref JxlDecoderSetDecompressBoxes,
+ *  - @ref JxlDecoderSetKeepOrientation,
+ *  - @ref JxlDecoderSetRenderSpotcolors,
+ *  - @ref JxlDecoderSkipFrames, and
```
szabadka, line 376: dec->skip_frames counter is reset when we rewind.

lib/include/jxl/decode.h
```diff
@@ -366,6 +366,15 @@ typedef enum {
...
+ *  - @ref JxlDecoderSubscribeEvents.
```
szabadka, line 377: JxlDecoderSetParallelRunner should be added to the list.

szabadka (changes requested):

lib/include/jxl/decode.h
mo271: done

lib/include/jxl/decode.h
mo271: done

szabadka (approved):

PR #1574, opened 2022-07-05
Update butteraugli_main.cc

aous72:
This is a small bug fix for butteraugli_main.cc.
1. The original code did not increment the arguments counter when intensity_target is supplied.
2. The short help menu did not print a usage message to let the end user know that there is a "--pnorm <pth norm>" option.  If you do not want this to be a user option, please ignore this change in usage message.


mo271:
Thanks for your contribution. In addition to signing the CLA, it would also be great if you could add your name to the [AUTHORS](https://github.com/libjxl/libjxl/blob/main/AUTHORS) file.

mo271:
Great! The list of authors needs to be in alphabetical order, but all other tests seem to pass...

mo271 (approved):

PR #1577, opened 2022-07-05
Allow conversion between RGB and Grayscale in djxl.

szabadka:
Removed the limitation from JxlDecoderSetPreferredColorProfile
that the "grayscaleness" of the original and requested colorspaces
must match. Instead we check that the requested output format has
enough channels for color output.

DecodeImageJXL now sets the ppf.info.num_color_channels to match
the requested output color space (which is the original color space
by default), and clears ppf.info.alpha_bits if the accepted formats
can not handle an alpha channel.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -429,6 +429,9 @@ struct JxlDecoderStruct {
   std::vector<ExtraChannelOutput> extra_channel_output;
 
   jxl::CodecMetadata metadata;
+  // Same as metadata.m, except for the color_encoding, which is set to the
+  // output encoding.
+  jxl::ImageMetadata image_metadata;
   std::unique_ptr<jxl::ImageBundle> ib;
 
   std::unique_ptr<jxl::PassesDecoderState> passes_state;
@@ -725,6 +728,7 @@ void JxlDecoderRewindDecodingState(JxlDecoder* dec) {
 
   dec->ib.reset();
   dec->metadata = jxl::CodecMetadata();
+  dec->image_metadata = dec->metadata.m;
   dec->frame_header.reset(new jxl::FrameHeader(&dec->metadata));
 
   dec->codestream_copy.clear();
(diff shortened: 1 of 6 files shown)


lib/jxl/decode.cc
```diff
@@ -2303,13 +2309,12 @@ JxlDecoderStatus JxlDecoderGetColorAsEncodedProfile(
   return JXL_DEC_SUCCESS;
 }
 
-JxlDecoderStatus JxlDecoderGetICCProfileSize(const JxlDecoder* dec,
-                                             const JxlPixelFormat* format,
-                                             JxlColorProfileTarget target,
-                                             size_t* size) {
+JxlDecoderStatus JxlDecoderGetICCProfileSize(
+    const JxlDecoder* dec, const JxlPixelFormat* unused_format,
```
eustas, line 2313: Should reflect in `decode.h` - clang-tidy does not like when params have different names in declaration and definiton.

lib/jxl/jxl_test.cc
```diff
@@ -405,6 +405,43 @@ TEST(JxlTest, RoundtripMultiGroup) {
...
+  CodecInOut io2;
+  EXPECT_FALSE(io.Main().IsGray());
+  EXPECT_LE(Roundtrip(&io, cparams, dparams, &pool, &io2), 55000u);
+  EXPECT_TRUE(io2.Main().IsGray());
+
+  // Do a very crude conversion to grayscale and compare with butteraugli.
+  Image3F* color = io.Main().color();
+  for (size_t y = 0; y < color->ysize(); ++y) {
+    float* row_r = color->PlaneRow(0, y);
+    float* row_g = color->PlaneRow(0, y);
```
eustas, line 431: use planes 0, 1, 2?

szabadka: Thanks for catching this! The test passed because both the test and the implementation just dropped green and blue ... Now it works and butteraugli distance also improved.

lib/jxl/decode.cc
szabadka: Done.

eustas (dismissed):

eustas (approved):

PR #1580, opened 2022-07-06
Fix render pipeline for frames completely outside full image.

szabadka:
This fixes an uninitialized memeory access discovered by fuzzers.


eustas:
Isn't it rollback of https://github.com/libjxl/libjxl/pull/1357?

eustas (approved):

PR #1581, opened 2022-07-07
Fixes for fuzzer bugs.

szabadka:
* Initialize full image area with RenderPadding if frame is outside image
* Use unaligned store/load in all stages
* Only run per-group render pipeline if modular could decode something

### DIFF
--- a/lib/jxl/render_pipeline/low_memory_render_pipeline.cc
+++ b/lib/jxl/render_pipeline/low_memory_render_pipeline.cc
@@ -787,38 +787,54 @@ void LowMemoryRenderPipeline::ProcessBuffers(size_t group_id,
     RectT<ssize_t> full_image_rect(0, 0, full_image_xsize_, full_image_ysize_);
     group_rect = group_rect.Translate(frame_origin_.x0, frame_origin_.y0);
     image_rect = image_rect.Translate(frame_origin_.x0, frame_origin_.y0);
-    group_rect =
-        group_rect.Intersection(image_rect).Intersection(full_image_rect);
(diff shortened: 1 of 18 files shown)


eustas (changes requested):
Wait, it will cause full-image-padding for each group?

szabadka:
> Wait, it will cause full-image-padding for each group?

Good point, I changed it to do the full padding only for group 0. PTAL

lib/jxl/render_pipeline/low_memory_render_pipeline.cc
```diff
@@ -787,38 +787,54 @@ void LowMemoryRenderPipeline::ProcessBuffers(size_t group_id,
...
-        gx + 1 == frame_dimensions_.xsize_groups) {
-      RenderPadding(thread_id, Rect(x1, y1, full_image_xsize_ - x1,
-                                    full_image_ysize_ - y1));
+    JXL_DEBUG_V(6,
+                "Rendering padding for full image rect %s "
+                "outside group rect %s",
+                Description(full_image_rect).c_str(),
+                Description(group_rect).c_str());
+
+    if (group_id == 0 && (image_rect.xsize() == 0 || image_rect.ysize() == 0)) {
```
eustas, line 802: It should be per-thread, I believe.

szabadka: Whichever thread gets group-id 0, will do the initialization. This is analogous to how RenderPadding works in the other cases below. How else should I do it?

PR #1584, opened 2022-07-07
Fix return value of JxlDecoderSetPreferredColorProfile.

### LINKED ISSUE #1578: JxlDecoderSetPreferredColorProfile regression when uses_original_profile=true
**Describe the bug**
I have a JPEG->JXL reconstructed file. `uses_original_profile = true` because it's a JPEG reconstruction, and it's got a simple sRGB enum set with no attached ICC profile. The following results in a JXL_ERROR being returned:
```
JxlColorEncoding jxl_color;
JxlDecoderGetColorAsEncodedProfile(decoder, NULL, JXL_COLOR_PROFILE_TARGET_ORIGINAL, &jxl_color);
ret = JxlDecoderSetPreferredColorProfile(decoder, &jxl_color);
```
After the following `ret == JXL_DEC_ERROR`, which is incorrect, since the code requested the exact original profile returned. If any other profile was requested, this should return false, but this specific one being requested without modification should return `JXL_DEC_SUCCESS`.

I believe this was introduced in [acc405da4471afd7a66c0d8fcd545dd4bfadee7c](https://github.com/libjxl/libjxl/commit/acc405da4471afd7a66c0d8fcd545dd4bfadee7c).

Sample of the type of file to trigger the bug is attached as a `.jxl.jpg` but it's actually a JXL file.

szabadka:
Return JXL_DEC_SUCCESS if requested color space is the same as original,
even if the image is not XYB encoded.

Fixes #1578

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -2795,10 +2795,11 @@ JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
   jxl::ColorEncoding c_out;
   JXL_API_RETURN_IF_ERROR(
       ConvertExternalToInternalColorEncoding(*color_encoding, &c_out));
-  JXL_API_RETURN_IF_ERROR(
-      dec->passes_state->output_encoding_info.MaybeSetColorEncoding(c_out));
-  dec->image_metadata.color_encoding =
-      dec->passes_state->output_encoding_info.color_encoding;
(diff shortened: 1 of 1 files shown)


mo271 (changes requested):
accidentally added the test.jpg image?

mo271 (approved):

PR #1608, opened 2022-07-13
Reduce the number of jobs running for PR/push

eustas:
Run all jobs on the schedule instead.


mo271 (dismissed):

mo271:
Is this still up-to-date? Should we rebase and get it merged, @eustas?

eustas:
It seems to be not implementable in a clear way. There is solution, but I think it is not worth it...

PR #1610, opened 2022-07-14
[no ci] Remove fuzzer_test from lib.gni

eustas:
Fuzzers require special linking, so do not belong to common sources list.


eustas:
Oh, perhaps we should rename it (likely it reacts on _test.cc suffix).

eustas:
PTAL

szabadka (dismissed):

PR #1617, opened 2022-07-15
JxlEncoderFrameSettingsSetOption to take int64_t

mo271:
This change widens the type of the `value` argument
of `JxlEncoderFrameSettingsSetOption`, from
`int32_t` to `long long`. This allows setting the
center_x and center_y arguments.


veluca93 (approved):
FWIW, I suspect I was wrong about saying that using fixed-width integers is bad for ABI compatibility :)

mo271:
> FWIW, I suspect I was wrong about saying that using fixed-width integers is bad for ABI compatibility :)

So do you suggest to go with `int64_t` instead of `long long` after all?

veluca93:
Yeah that seems like a reasonable solution...

mo271:
> Yeah that seems like a reasonable solution...


done

veluca93 (approved):

PR #1624, opened 2022-07-18
Handle unpremultiply the same way in image out buffer and callback.

szabadka:
For the image out callback case we did not unpremultiply alpha, and
there was a test to check that, but for the image buffer case we
always unpremultiplied alpha.

I added an API function similar to keep/undo orientation where the
user can specify the desired behavior.


mo271 (commented):
Perhaps mention `JxlDecoderSetUnpremultiplyAlpha` in the CHANGELOG.md?!

szabadka:
Added changelog entry

PR #1628, opened 2022-07-19
Do not unpremultiply-alpha by default

eustas:
This is generally bad idea to have alpha premultiplied (in storage format).
Unpremultiplying can cause very low precision results, and the result for
pixels with alpha == 0 is a separate discussion topic.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -521,8 +521,7 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL skip_reorientation);
  * The colors will be unpremultiplied based on the alpha channel. This function
  * has no effect if the image does not have an associated alpha channel.
  *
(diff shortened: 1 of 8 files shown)


lib/include/jxl/decode.h
```diff
@@ -521,8 +521,7 @@ JxlDecoderSetKeepOrientation(JxlDecoder* dec, JXL_BOOL skip_reorientation);
  * The colors will be unpremultiplied based on the alpha channel. This function
  * has no effect if the image does not have an associated alpha channel.
  *
- * By default, this option is enabled, and the returned pixel data is
- * unpremultiplied.
+ * By default, this option is disables, and the returned pixel data "as is".
```
szabadka, line 524: disables->disabled

szabadka (approved):
The conformance test case for premultiplied alpha will need to be updated before (or soon after) merging.

szabadka (approved):

PR #1635, opened 2022-07-20
Fix float32 small/large cube implicit palette value scaling.

szabadka:
This is analogous to how implicit delta palette value scaling is
defined for float32 palette.

This fixes an integer overflow bug found by oss-fuzz that was
introduced by #1341.


mo271 (approved):

eustas (approved):

jyrkialakuijala:
Thank you!

PR #1637, opened 2022-07-20
tools/cjxl_main: correctly report modular mode for lossless

### LINKED ISSUE #1627: cjxl reports VarDCT for lossless (but correctly uses modular)
**Describe the bug**
Running `cjxl -d 0 input.png output.jxl` causes cjxl to report something like: `Encoding [VarDCT, lossless, effort: 7],` which is not correct. It does actually encode it losslessly, which forces modular mode and the original profile, as it should, so this is only an error in the human-readable report.

**To Reproduce**
Run `cjxl -d 0 input.png output.jxl` on any PNG.

**Environment**
 - OS: Arch Linux x86_64
 - Compiler version: gcc 12.1.0
 - cjxl/djxl version string: libjxl/libjxl@66afb51b

**Additional Information**
As far as I'm aware, this bug is here. It doesn't take distance or quality into account: https://github.com/libjxl/libjxl/blob/66afb51bb4f132f205ba16d236993ebf8fdba89f/tools/cjxl_main.cc#L488-L492
Thanks to The_Decryptor#7970 for finding this.

On lines https://github.com/libjxl/libjxl/blob/66afb51bb4f132f205ba16d236993ebf8fdba89f/tools/cjxl_main.cc#L1096-L1101 you can see that it is properly setting Lossless mode though, and jxlinfo reports modular mode for the resulting file.

thebombzen:
cjxl should correctly report that it is using modular mode when a
user specifies distance=0 or quality=100. It already uses modular,
but until this commit it reported VarDCT incorrectly.

Fixes #1627.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -487,7 +487,9 @@ struct CompressArgs {
 
 const char* ModeFromArgs(const CompressArgs& args) {
   if (args.lossless_jpeg) return "JPEG";
-  if (args.modular == jxl::Override::kOn) return "Modular";
+  if (args.modular == jxl::Override::kOn || args.distance == 0 ||
+      args.quality == 100)
+    return "Modular";
   return "VarDCT";
 }
 


mo271 (approved):
LGTM, thanks @thebombzen for the contribution!

PR #1638, opened 2022-07-21
Add tighter memory usage limits on patch dictionary.

szabadka:
The vector of patch positions is broken up into three
vectors: reference positions, patch positions (with indexes into
the reference positions vector) and blending infos. Each vector's
size is limited in the decoder, giving a combined upper bound on
the memory usage (currently 1.5GB).

I will fix the memory usage of the structures used for looking up
patches that insersect a certain row in a separate PR.

### DIFF
--- a/lib/jxl/dec_patch_dictionary.cc
+++ b/lib/jxl/dec_patch_dictionary.cc
@@ -35,8 +35,6 @@
 
 namespace jxl {
 
-constexpr size_t kMaxPatches = 1 << 24;
-
 Status PatchDictionary::Decode(BitReader* br, size_t xsize, size_t ysize,
                                bool* uses_extra_channels) {
   positions_.clear();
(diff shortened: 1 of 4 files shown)


veluca93:
@jonsneyers what does the spec say about this? (and what _should_ it say?)

szabadka:
> Does it make sense to make the limits proportional to the image area?

We could make the limit num_pixels/4 for reference positions, num_pixels for positions and num_pixels * 4 for blending infos, which would require at most 66 bytes per pixel, does that sound reasonable?

veluca93:
> > Does it make sense to make the limits proportional to the image area?
> 
> We could make the limit num_pixels/4 for reference positions, num_pixels for positions and num_pixels * 4 for blending infos, which would require at most 66 bytes per pixel, does that sound reasonable?

SGTM except I'd do an affine limit, like 1024 + num_pixels/4 or something.

veluca93 (approved):
LGTM but I'll leave the final word to @jonsneyers wrt spec

PR #1642, opened 2022-07-22
Fix typo in build file; should fix sve/rvv build with gcc

eustas:


eustas:
Thanks to @jan-wassenberg for noting the problem!

mo271 (approved):

PR #1646, opened 2022-07-25
SSIMULACRA 2

jonsneyers:
Major new version of the SSIMULACRA metric.
Changes compared to the original version:

- works in XYB color space
- uses 1-norm and 4-norm (instead of 1-norm and max-norm-after-downscaling)
- penalizes both smoothing and ringing artifacts (instead of only penalizing ringing but not smoothing)
- removed specific blockiness detection

The score weights were obtained by running Nelder-Mead simplex search, optimizing to minimize MSE and maximize Kendall and Pearson correlation for training data consisting of 17611 subjective quality scores, validated on separate validation data consisting of 4292 scores.

Results for the full dataset (almost 22k subjective scores):

Metric | Kendall correlation | Spearman | Pearson |
-- | -- | -- |-- 
PSNR | 0.34677 | 0.49926 | 0.48178
SSIM | 0.41612 | 0.58939 | 0.52543
SSIMULACRA | -0.5214 | -0.7124 | -0.6891
VMAF | 0.61720 | 0.81549 | 0.77781
Butteraugli 2-norm | -0.6572 | -0.8461 | -0.8103
SSIMULACRA 2 | 0.74185 | 0.91695 | 0.90244

Results for just the validation set (4292 subjective scores):

Metric | Kendall correlation | Spearman | Pearson |
-- | -- | -- |-- 
SNR | 0.35089 | 0.50092 | 0.50157
SSIM | 0.45002 | 0.62777 | 0.55710
SSIMULACRA | -0.5881 | -0.7845 | -0.7808
VMAF | 0.58799 | 0.78798 | 0.75031
Butteraugli 2-norm | -0.6226 | -0.8127 | -0.7834
SSIMULACRA 2 | 0.72163 | 0.90354 | 0.89504

Computing the mean absolute error between opinion scores (on a scale of 0 to 100) and SSIMULACRA 2 results, on the full set the MAE is 4.274 and on the validation set the MAE is 4.470.

### DIFF
--- a/tools/ssimulacra2.cc
+++ b/tools/ssimulacra2.cc
@@ -1,0 +1,424 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+SSIMULACRA 2 - Structural SIMilarity Unveiling Local And Compression Related
+Artifacts
+
+Perceptual metric developed by Jon Sneyers (Cloudinary) in July 2022.
+Design:
+- XYB color space (X+0.5, Y, Y-B+1.0)
+- SSIM map
+- 'blockiness/ringing' map (distorted has edges where original is smooth)
+- 'smoothing' map (distorted is smooth where original has edges)
+- error maps are computed at 6 scales (1:1 to 1:32) for each component (X,Y,B)
+- for all 6*3*3=54 maps, two norms are computed: 1-norm (mean) and 4-norm
+- a weighted sum of these 54*2=108 norms leads to the final score
+- weights were tuned based on a large set of subjective scores for images
+  compressed with JPEG, JPEG 2000, JPEG XL, WebP, AVIF, and HEIC.
+*/
+
+#include "tools/ssimulacra2.h"
+
+#include <stdio.h>
+
+#include <cmath>
+
+#include "lib/jxl/enc_color_management.h"
+#include "lib/jxl/enc_xyb.h"
+#include "lib/jxl/gauss_blur.h"
+#include "lib/jxl/image_ops.h"
+
+namespace {
+
+using jxl::Image3F;
+using jxl::ImageF;
+
+static const float kC1 = 0.0001f;
+static const float kC2 = 0.0003f;
+static const int kNumScales = 6;
+
+Image3F Downsample(const Image3F& in, size_t fx, size_t fy) {
+  const size_t out_xsize = (in.xsize() + fx - 1) / fx;
+  const size_t out_ysize = (in.ysize() + fy - 1) / fy;
+  Image3F out(out_xsize, out_ysize);
+  const float normalize = 1.0f / (fx * fy);
+  for (size_t c = 0; c < 3; ++c) {
+    for (size_t oy = 0; oy < out_ysize; ++oy) {
+      float* JXL_RESTRICT row_out = out.PlaneRow(c, oy);
+      for (size_t ox = 0; ox < out_xsize; ++ox) {
+        float sum = 0.0f;
+        for (size_t iy = 0; iy < fy; ++iy) {
+          for (size_t ix = 0; ix < fx; ++ix) {
+            const size_t x = std::min(ox * fx + ix, in.xsize() - 1);
+            const size_t y = std::min(oy * fy + iy, in.ysize() - 1);
+            sum += in.PlaneRow(c, y)[x];
+          }
+        }
+        row_out[ox] = sum * normalize;
+      }
+    }
+  }
+  return out;
+}
+
+void Multiply(const Image3F& a, const Image3F& b, Image3F* mul) {
+  for (size_t c = 0; c < 3; ++c) {
+    for (size_t y = 0; y < a.ysize(); ++y) {
+      const float* JXL_RESTRICT in1 = a.PlaneRow(c, y);
+      const float* JXL_RESTRICT in2 = b.PlaneRow(c, y);
+      float* JXL_RESTRICT out = mul->PlaneRow(c, y);
+      for (size_t x = 0; x < a.xsize(); ++x) {
+        out[x] = in1[x] * in2[x];
+      }
+    }
+  }
+}
+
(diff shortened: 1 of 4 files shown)


tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+  }
+}
+
+// Temporary storage for Gaussian blur, reused for multiple images.
+class Blur {
+ public:
+  Blur(const size_t xsize, const size_t ysize)
+      : rg_(jxl::CreateRecursiveGaussian(1.5)), temp_(xsize, ysize) {}
+
+  void operator()(const ImageF& in, ImageF* JXL_RESTRICT out) {
```
veluca93, line 92: Let's please not do overloaded `operator()` :) (arguably also not operator() at all, but that might be more OK)

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+      const float* JXL_RESTRICT row_s11 = s11.PlaneRow(c, y);
+      const float* JXL_RESTRICT row_s22 = s22.PlaneRow(c, y);
+      const float* JXL_RESTRICT row_s12 = s12.PlaneRow(c, y);
+      for (size_t x = 0; x < m1.xsize(); ++x) {
+        float mu1 = row_m1[x];
+        float mu2 = row_m2[x];
+        float mu11 = mu1 * mu1;
+        float mu22 = mu2 * mu2;
+        float mu12 = mu1 * mu2;
+        float nom_m = 2 * mu12 + kC1;
```
veluca93, line 132: nit: s/nom/num/

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+      const float* JXL_RESTRICT rowm2 = mu2.PlaneRow(c, y);
+      for (size_t x = 0; x < img1.xsize(); ++x) {
+        double d1 = (1.0 + std::abs(row2[x] - rowm2[x])) /
+                        (1.0 + std::abs(row1[x] - rowm1[x])) -
+                    1.0;
+        // d1 > 0: distorted has an edge where original is smooth
+        //         (indicating ringing, color banding, blockiness, etc)
+        // d1 < 0: original has an edge where distorted is smooth
+        //         (indicating smoothing, blurring, smearing, etc)
+        double d2 = std::max(d1, 0.0);
```
veluca93, line 167: call this smoothness, and line 172 blockiness?

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+  if (ssim > 0) {
+    ssim = 100.0 - 10.0 * pow(ssim, 0.6940375837127916);
+
+  } else {
+    ssim = 100.0;
+  }
+  return ssim;
+}
+
+void BminusY(jxl::Image3F& img) {
```
veluca93, line 345: Misleading name :)

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+
+The weights were obtained by running Nelder-Mead simplex search,
+optimizing to minimize MSE and maximize Kendall and Pearson correlation
+for training data consisting of 17611 subjective quality scores,
+validated on separate validation data consisting of 4292 scores.
+*/
+double Msssim::Score() const {
+  double ssim = 0.0;
+
+  const double weight[108] = {-0.4887721343447775,
```
veluca93, line 206: static constexpr

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,435 @@
...
+
+  Msssim msssim = ComputeSSIMULACRA2(orig, dist);
+  printf("%.8f\n", msssim.Score());
+  return 0;
+}
+
+}  // namespace
+}  // namespace ssimulacra2
+
+int main(int argc, char** argv) { return ssimulacra2::Run(argc, argv); }
```
veluca93, line 435: main and run really should be in a different file if the intention is to allow using this as a library.

If not, then the .h file is not particularly useful :)

jonsneyers: Right. I guess having it as a library could be useful (say to be used in benchmark_xl), so I'll split it up.

tools/ssimulacra2.cc
jonsneyers: Something descriptive could in any case be useful, yes.

What about 'detail_lost' and 'artifacts' or something like that?

veluca93: Sure

tools/ssimulacra2.cc
jonsneyers: I don't disagree, but this part of the code was just copied from ssimulacra.cc, so it probably should be changed in both places (or be put in some shared auxiliary .h or something).

jonsneyers:
Added a separate main file, added some more info to the help screen, and made it work for images with alpha too.

tools/ssimulacra2.cc
```diff
@@ -0,0 +1,429 @@
...
+   compressed with JPEG, JPEG 2000, JPEG XL, WebP, AVIF, and HEIC.
+*/
+
+#include "tools/ssimulacra2.h"
+
+#include <stdio.h>
+
+#include <cmath>
+
+//#include "lib/extras/codec.h"
```
veluca93, line 32: remove?

veluca93 (approved):

veluca93 (approved):

jonsneyers:
> @jonsneyers I see that you are using coefficients correlations to evaluate metrics. You might be interested to learn that there is a new coefficient correlation in town, with very interesting properties and therefore maybe that this new meta-metric could help you compare and improve compressions metrics? https://news.ycombinator.com/item?id=29687613

Thanks for the pointer! Sounds interesting indeed.

Note however that most metrics require a nonlinear mapping to be used as MOS predictors, so in general, rank correlation (like Kendall's tau) is more useful than Pearson correlation to compare the performance of various metrics. But I have to read that paper...

LifeIsStrange:
I'm not an expert but I found this visualization quite impressive https://twitter.com/adad8m/status/1474754752193830912?s=21

PR #1649, opened 2022-07-25
Split convolve.cc

eustas:
In some builds it takes enormous amount of time to compile.
Let it compile at least in parallel.


eustas:
convolve.cc was compiling 1800s. Now the biggest item is convolve_separable7 (960s) and compilation time reduced to 18min. I believe we could improve further, after understanding why separable7 is different.

szabadka (approved):

PR #1651, opened 2022-07-26
Use an interval tree for looking up patches that intersect a row.

szabadka:
This fixes a potential OOM attack by reducing the memory usage of
the patch dictionary from ```O(N*ysize)``` to ```O(N)``` where ```N``` is the number
of patches. Since ```N``` is restricted to be at most ```O(xsize*ysize)``` by
the decoder, the worst-case memory usage decreases from ```O(xsize*ysize^2)```
to ```O(xsize*ysize)```.

### DIFF
--- a/lib/jxl/dec_patch_dictionary.cc
+++ b/lib/jxl/dec_patch_dictionary.cc
@@ -165,7 +165,7 @@ Status PatchDictionary::Decode(BitReader* br, size_t xsize, size_t ysize,
     return JXL_FAILURE("ANS checksum failure.");
   }
 
-  ComputePatchCache();
+  ComputePatchTree();
   return true;
 }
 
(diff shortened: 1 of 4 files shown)


lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,138 @@ int PatchDictionary::GetReferences() const {
...
+
+void PatchDictionary::ComputePatchTree() {
+  patch_tree_.clear();
+  num_patches_.clear();
+  sorted_patches_y0_.clear();
+  sorted_patches_y1_.clear();
+  if (positions_.empty()) {
+    return;
   }
-  // The relative order of patches that affect the same pixels is preserved.
```
veluca93, line 192: Is this still true?

lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,138 @@ int PatchDictionary::GetReferences() const {
...
+    }
+    ++next;
+  }
+}
+
+std::vector<size_t> PatchDictionary::GetPatchesForRow(size_t y) const {
+  std::vector<size_t> result;
+  if (y < num_patches_.size() && num_patches_[y] > 0) {
+    result.resize(num_patches_[y]);
+    size_t* next_result = &result[0];
```
veluca93, line 281: Why not reserve + push_back?

lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,138 @@ int PatchDictionary::GetReferences() const {
...
+  root.start = 0;
+  root.num = intervals.size();
+  patch_tree_.push_back(root);
+  size_t next = 0;
+  while (next < patch_tree_.size()) {
+    auto& node = patch_tree_[next];
+    size_t start = node.start;
+    size_t end = node.start + node.num;
+    // Choose the y_center for this node to be the median of interval starts.
+    sort_by_y0(start, end);
```
veluca93, line 231: If we only need the median here, it might make sense to use nth_element instead.

lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,138 @@ int PatchDictionary::GetReferences() const {
...
-  if (positions_.empty()) return;
-  std::vector<std::pair<size_t, size_t>> sorted_patches_y;
-  for (size_t i = 0; i < positions_.size(); i++) {
-    const PatchPosition& pos = positions_[i];
-    const PatchReferencePosition& ref_pos = ref_positions_[pos.ref_pos_idx];
-    for (size_t y = pos.y; y < pos.y + ref_pos.ysize; y++) {
-      sorted_patches_y.emplace_back(y, i);
-    }
+namespace {
+struct PatchInterval {
```
veluca93, line 181: Does it make sense to extract the interval tree in a class?

lib/jxl/dec_patch_dictionary.cc
szabadka: Not sure, probably not, should we add a sort at the end of GetPatchesForRow then?

lib/jxl/dec_patch_dictionary.cc
szabadka: Maybe if we use it for something else too.

lib/jxl/dec_patch_dictionary.cc
szabadka: We need to iterate a bit further away the median in case there are more intervals starting at y_center.

lib/jxl/dec_patch_dictionary.cc
szabadka: Done.

lib/jxl/dec_patch_dictionary.cc
veluca93: Probably yes, as patch application is not necessarily commutative :)

szabadka: Done.

lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,141 @@ int PatchDictionary::GetReferences() const {
...
+    }
+    sort_by_y1(start, right_start);
+    size_t left_end = right_start;
+    while (left_end > start && intervals[left_end - 1].y1 > node.y_center) {
+      --left_end;
+    }
+    // Fill in sorted_patches_y0_ and sorted_patches_y1_ for the current node.
+    node.num = right_start - left_end;
+    node.start = sorted_patches_y0_.size();
+    for (int i = (int)right_start - 1; i >= (int)left_end; --i) {
```
veluca93, line 250: right_start and `left_end` are `size_t`s, so I'd either use `int64_t`, make them `int`s too if we don't believe more than an int will be useful, or do the loop with `size_t` somehow making sure it works out fine :)

lib/jxl/dec_patch_dictionary.h
```diff
@@ -125,20 +127,23 @@ class PatchDictionary {
...
-  // The relative order of patches that affect the same pixels is the same -
-  // important when applying patches is noncommutative.
-
-  // Compute patches_by_y_ after updating positions_.
-  void ComputePatchCache();
+  // Interval tree on the y coordinates of the patches.
+  struct PatchTreeNode {
+    int left_child;
+    int right_child;
+    size_t y_center;
```
veluca93, line 134: What's the rationale with y_center being size_t, but left_child and right_child not?

lib/jxl/dec_patch_dictionary.cc
szabadka: Changed to int64_t

lib/jxl/dec_patch_dictionary.h
szabadka: I use -1 for the case that there is no left/right child

veluca93: Then it should be int64_t/ssize_t, no? (or the size_t should be uint32_t / unsigned int)

szabadka: Ok, I made these ssize_t and also tree_idx in the lookup and the int64_ts in the reverse-loop for consistency.

lib/jxl/dec_patch_dictionary.cc
```diff
@@ -177,44 +177,141 @@ int PatchDictionary::GetReferences() const {
...
+    }
+    sort_by_y1(start, right_start);
+    size_t left_end = right_start;
+    while (left_end > start && intervals[left_end - 1].y1 > node.y_center) {
+      --left_end;
+    }
+    // Fill in sorted_patches_y0_ and sorted_patches_y1_ for the current node.
+    node.num = right_start - left_end;
+    node.start = sorted_patches_y0_.size();
+    for (ssize_t i = (ssize_t)right_start - 1; i >= (ssize_t)left_end; --i) {
```
veluca93, line 250: last nit: static_cast<ssize_t>()

szabadka: Done.

veluca93 (approved):

veluca93 (approved):

PR #1655, opened 2022-07-27
Always add cICP chucks to djxl generated pngs if possible.

szabadka:


lib/extras/enc/apng.cc
```diff
@@ -235,36 +269,8 @@ Status EncodePackedPixelFileToAPNG(const PackedPixelFile& ppf, ThreadPool* pool,
...
-        cicp_data[3] = 1;
-        cicp_chunk.data = cicp_data;
-        cicp_chunk.size = sizeof(cicp_data);
-        cicp_chunk.location = PNG_HAVE_PLTE;
-        memcpy(cicp_chunk.name, "cICP", 5);
-        png_set_keep_unknown_chunks(
-            png_ptr, 3, reinterpret_cast<const png_byte*>("cICP"), 1);
-        png_set_unknown_chunks(png_ptr, info_ptr, &cicp_chunk, 1);
-      } else {
+      MaybeAddCICP(ppf.color_encoding, png_ptr, info_ptr);
```
eustas, line 272: result ignored?

szabadka: Changed it to a void function. Writing a PNG should not fail if we cannot set cICP.

lib/extras/enc/apng.cc
```diff
@@ -115,6 +115,42 @@ class BlobsWriterPNG {
...
+    cicp_data[0] = static_cast<png_byte>(c_enc.primaries);
+  } else {
+    return;
+  }
+  if (c_enc.transfer_function == JXL_TRANSFER_FUNCTION_UNKNOWN ||
+      c_enc.transfer_function == JXL_TRANSFER_FUNCTION_GAMMA) {
+    return;
+  }
+  cicp_data[1] = static_cast<png_byte>(c_enc.transfer_function);
+  cicp_data[3] = 1;
```
kmilos, line 144: How about explicitly setting cicp_data[2] = 0 (identity matrix for RGB) for clarity and consistency?

szabadka: Done.

eustas (approved):

kmilos:
Did anyone manage to actually read this back? @veluca93 @sboukortt 

[I'm getting zero chunks from `png_get_unknown_chunks()`...](https://github.com/darktable-org/darktable/pull/12237#issuecomment-1245957706)

veluca93:
> 

The chunk is there according to `pngcheck`. How do you call set_keep_unknown_chunks? I remember it took me some convincing to get it working in Chrome: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/png/png_image_reader.cc;l=102?q=png_set_keep_unknown_chunks&sq=&ss=chromium

kmilos:
> The chunk is there according to `pngcheck`.

Yes, I verified it using `exiftool` as well (and used the tool to remove the ICC profile).

> How do you call set_keep_unknown_chunks?

Thanks for the pointer - it seems indeed critical to call it **_before_** `info_ptr` is created!

PR #1663, opened 2022-07-30
tsc_timer.h: sync with nanobenchmark.cc from Highway

kleisauke:
This file was originally originated from nanobenchmark.cc on the
Highway project. Re-syncing this file ensures that the improvements
from commit google/highway@a884393 and google/highway@12ca8ff are
incorporated.

---

Note that I did not sync commit google/highway@1ac50ad, as
`JPEGXL_ENABLE_PROFILER` is not enabled by default. I think(?) we
can safely assume that `RDTSCP` on x86 is available, when users
explicitly enable that option.

/cc @jan-wassenberg

### DIFF
--- a/lib/profiler/tsc_timer.h
+++ b/lib/profiler/tsc_timer.h
@@ -101,6 +101,9 @@ static HWY_INLINE HWY_MAYBE_UNUSED Ticks TicksBefore() {
   Ticks t;
 #if HWY_ARCH_PPC && defined(__GLIBC__)
   asm volatile("mfspr %0, %1" : "=r"(t) : "i"(268));
+#elif HWY_ARCH_ARM_A64 && !HWY_COMPILER_MSVC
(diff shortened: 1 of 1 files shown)


mo271 (dismissed):
LGTM, what do you think, @jan-wassenberg?

kleisauke:
Updated to also include commit https://github.com/google/highway/commit/12ca8ff15390ccc4ef0a257ec095bb50e8b288c0.

kleisauke:
CI failures looks unrelated to this PR.

> Updated to also include commit [google/highway@12ca8ff](https://github.com/google/highway/commit/12ca8ff15390ccc4ef0a257ec095bb50e8b288c0).

IIUC, this is only necessary when building with `-DJPEGXL_ENABLE_SIZELESS_VECTORS=true` (disabled by default), I can omit that change if needed.

kleisauke:
Closing in favor of PR #2511.

PR #1672, opened 2022-08-03
cjxl: take exif orientation into account

mo271:
This makes sure that the exif orientation is taken into account except
for the jpeg reconstruction case (which is planned next).


lib/jxl/enc_file.cc
```diff
@@ -139,6 +136,7 @@ Status WriteHeaders(CodecMetadata* metadata, BitWriter* writer,
   return true;
 }
 
+// TODO(firsching): remove this function and all tests that use it.
```
szabadka, line 139: "all tests that use it" is basically all roundtrip tests, I don't think we want to remove those

The plan should be to re-implement this using API functions and move it to test_utils.h, same as with DecodeFile

mo271:
Closed in favour of https://github.com/libjxl/libjxl/pull/1677

PR #1673, opened 2022-08-03
explain tags for branch-off points

mo271:


eustas:
Let's do that differently. "describe" should say what we are looking at. Starting from 0.8 we will add a "descriptive" tag that refers to the next version we are planning to release next: `0.8-snapshot`. Whenever we decide that next version is not 0.8, but 1.0, we will add extra tag `1.0-snapshot`...

eustas (approved):

PR #1677, opened 2022-08-03
get orientation from exif

mo271:


veluca93 (approved):

lib/jxl/encode.cc
```diff
@@ -1425,12 +1426,15 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
     }
   }
 
-  if (!frame_settings->enc->basic_info_set) {
+  if (!frame_settings->enc->basic_info_set || !io.blobs.exif.empty()) {
```
szabadka, line 1429: I don't think we should overwrite basic info just because we have exif. In that case we should expect the orientation to be already set correctly.

mo271: only overwriting orientation now

lib/jxl/encode.cc
```diff
@@ -1437,6 +1438,11 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
                            "Error setting basic info");
     }
   }
+  if (!io.blobs.exif.empty()) {
+    JxlOrientation orientation;
+    jxl::InterpretExif(io.blobs.exif, &orientation);
```
szabadka, line 1443: This fails the tests, you should pass &frame_settings->enc->metadata.m.orientation directly

mo271: I had tried that before, but `frame_settings->enc->metadata.m.orientation` is uint32_t and InterpretExif expects JxlOrientation...
Initializing should fix the test, hopefully.

lib/jxl/encode.cc
```diff
@@ -1439,7 +1439,7 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
     }
   }
   if (!io.blobs.exif.empty()) {
-    JxlOrientation orientation;
+    JxlOrientation orientation = JXL_ORIENT_IDENTITY;
```
szabadka, line 1442: I don't know how relevant it is, but now what could happen is that basic info was set outside the API with some non-identity orientation, and the jpeg contains an exif, but without an orientation tag. In this case we would overwrite the user provided orientation with identity. So to be sure we could initialize this with frame_settings->enc->metadata.m.orientation.

Also I just noticed that there is already an if(!io.blobs.exif.empty()) a few lines below, maybe these should be merged.

mo271: done

szabadka (approved):

PR #1695, opened 2022-08-11
Use low memory pipeline for swap-endian float callback output.

szabadka:


lib/jxl/render_pipeline/stage_write.cc
```diff
@@ -337,7 +339,15 @@ class WriteToPixelCallbackStage : public RenderPipelineStage {
         // TODO(szabadka) SIMDify (possibly in a separate pipeline stage).
         UnpremultiplyAlpha(temp, ix);
       }
-      pixel_callback_.run(run_opaque_, thread_id, xpos + x0, ypos, ix, temp);
+      uint8_t* temp_u8 = reinterpret_cast<uint8_t*>(temp);
```
veluca93, line 342: Nitpick time: this is not guaranteed to not be UB, we should use `char` instead.

That said, I'd consider adding a bswap function in base/bits or the like that calls __builtin_bswap32 etc if present, and otherwise does this.

veluca93 (dismissed):

lib/jxl/render_pipeline/stage_write.cc
szabadka: I added a BSwapFloat function to byte_order.h and use that here.

veluca93 (approved):

PR #1700, opened 2022-08-12
Undo orientation in WriteToPixelCallbackStage.

szabadka:


lib/jxl/dec_frame.h
```diff
@@ -223,13 +223,14 @@ class FrameDecoder {
   // orientation. Performing this operation is not yet supported, so this
```
veluca93, line 223: out of date :)

lib/jxl/render_pipeline/stage_write.cc
```diff
@@ -334,17 +339,36 @@ class WriteToPixelCallbackStage : public RenderPipelineStage {
...
+        for (size_t i = 0; i < num; i += num_channels_) {
+          for (size_t c = 0; c < num_channels_; ++c) {
+            std::swap(temp[i + c], temp[last - i + c]);
+          }
+        }
+        xstart = width_ - xstart - xlen;
+      }
+      if (transpose_) {
+        for (size_t i = 0, j = 0; i < xlen; ++i, j += num_channels_) {
+          pixel_callback_.run(run_opaque_, thread_id, ypos, xstart + i, 1,
```
veluca93, line 366: While this works, it is *rather slow*. Maybe we should (eventually) think of faster ways to do it.

lib/jxl/dec_frame.h
szabadka: updated

lib/jxl/render_pipeline/stage_write.cc
szabadka: There are some tricks to do NxN transpositions with SIMD, so maybe we could buffer some lines here, and output N pixels at each call. (But transposition is an inherently slow operation and the implementation we have currently does not seem to be much better.)

veluca93: Buffering lines wouldn't necessarily work unless we keep large buffers, it is probably better to buffer i.e. 8x8 chunks and guarantee in the RP that dimensions satisfy this-and-that constraint... Anyway, we don't need to do this now, just adding a TODO is fine.

szabadka: added the TODO

veluca93 (approved):

veluca93 (approved):

eustas:
The following tests started failing:
```
The following tests FAILED:
	1869 - DecodeTest/DecodeTestParam.PixelTest/280x12RGBAtoRGBAf32BECallbackEarlyBufferO2  # GetParam() = 280x12RGBAtoRGBAf32BECallbackEarlyBufferO2 (Failed)
	2037 - DecodeTest/DecodeTestParam.PixelTest/280x12RGBAtoRGBAf32BECallbackEarlyBufferO3  # GetParam() = 280x12RGBAtoRGBAf32BECallbackEarlyBufferO3 (Failed)
	2709 - DecodeTest/DecodeTestParam.PixelTest/280x12RGBAtoRGBAf32BECallbackEarlyBufferO7  # GetParam() = 280x12RGBAtoRGBAf32BECallbackEarlyBufferO7 (Failed)
	2877 - DecodeTest/DecodeTestParam.PixelTest/280x12RGBAtoRGBAf32BECallbackEarlyBufferO8  # GetParam() = 280x12RGBAtoRGBAf32BECallbackEarlyBufferO8 (Failed)
```

PR #1711, opened 2022-08-22
tests: Make sure libjxl0.7 can still be found

malaterre:
In some cases user may set CMAKE_SKIP_RPATH:BOOL=ON in which case we
must indicate explicitly where to find libjxl dynamic library.


malaterre:
For instance in Debian:

* https://buildd.debian.org/status/fetch.php?pkg=jpeg-xl&arch=amd64&ver=0.7.0%7Egit20220805.980c90f-1&stamp=1660127268&raw=0

eustas (approved):

eustas (approved):

eustas:
Oops! This should have been landed to `main` first...

PR #1724, opened 2022-08-30
Use low-memory render pipeline for all callback data types.

szabadka:


lib/jxl/render_pipeline/stage_write.cc
```diff
@@ -392,10 +325,15 @@ class WriteToPixelCallbackStage : public RenderPipelineStage {
     run_opaque_ =
         pixel_callback_.Init(num_threads, /*num_pixels=*/kMaxPixelsPerCall);
     JXL_RETURN_IF_ERROR(run_opaque_ != nullptr);
-    temp_.resize(num_threads);
-    for (CacheAlignedUniquePtr& temp : temp_) {
+    tempf_.resize(num_threads);
```
veluca93, line 328: IIRC it should be fine if you only create the array for the float case and use it for uint16_t too.

You could also create it for whichever case is actually going to trigger...

veluca93 (approved):

lib/jxl/render_pipeline/stage_write.cc
szabadka: For unsigned and float16 I currently use both, but for floats there is no need, so I fixed that.

veluca93 (approved):

PR #1725, opened 2022-09-01
Add codespace that use ubuntu and emsdk container.

dlemstra:
This PR adds two GitHub Codespaces. The default (in the `.decontainer`) is an ubuntu container. And the other one uses an emsdk container to make it possible to test emsdk changes. People with access to GitHub Codespaces could use this to work on this library and quickly test a change in the browser without requiring to set up a local development environment. The `onCreateCommand` can be part of the created container by configuring a pre-build codespace. And then when the codespace starts the commands are already executed. For the emsdk codespace I was unable to use `./ci.sh debug` because this resulted in a SIGKILL so I used release for now. The `onCreateCommand` part is optional but I thought it would be nice that the binaries are already pre-build in the codespace. This can then be used by someone to check a possible bug with the pre-build tools.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,26 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:ubuntu-20.04
+
+RUN apt-get update && \
+    apt install -y \
+        ccache \
+        clang-7 \
+        cmake \
+        doxygen \
+        libbenchmark-dev \
+        libbenchmark-tools \
+        libbrotli-dev \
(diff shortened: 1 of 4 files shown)


mo271 (approved):

mo271:
LGTM, what do you think, @eustas?

eustas:
Hi, Dirk. Thanks for the PR.

Currently codespaces are not enabled on this repo / org. Our team will discuss soon that topic.
On the parallel track - I'm preparing docker images for the very same purpose. Likely they will be published in github-packages =)

PR #1727, opened 2022-09-02
Implement decoding to a scaled XYB space.

szabadka:
Add the XYB profile created by sboukortt@ and veluca93@ that describes
a transformed version of the XYB space that is inside the unit cube for
sRGB colors.

Implement decoding into this modified XYB space by the API and thus
enable saving JXL files to XYB jpegs or PNGs.


veluca93:
As mentioned online, we probably should stop libjpeg from YCbCr-ing the XYB data :)

szabadka:
> As mentioned online, we probably should stop libjpeg from YCbCr-ing the XYB data :)

Done.

veluca93 (approved):

PR #1733, opened 2022-09-08
Unify the pixel callback and image buffer write stages.

szabadka:
This enables low memory code-path for all image buffer cases with the exception of extra channels.

### DIFF
--- a/lib/jxl/dec_frame.h
+++ b/lib/jxl/dec_frame.h
@@ -172,73 +172,60 @@ class FrameDecoder {
                                          : std::numeric_limits<size_t>::max());
   }
 
-  void MaybeSetUnpremultiplyAlpha(bool unpremul_alpha) {
+  // Sets the pixel callback or image buffer where the pixels will be decoded.
+  // This is not supported for all images. If it succeeds, HasRGBBuffer() will
+  // return true.
+  // If it does not succeed, the image is decoded to the ImageBundle passed to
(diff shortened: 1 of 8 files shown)


lib/jxl/dec_frame.h
```diff
@@ -172,73 +172,48 @@ class FrameDecoder {
...
-        decoded_->metadata()->Find(jxl::ExtraChannel::kAlpha);
-    if (alpha && alpha->alpha_associated && unpremul_alpha) {
-      dec_state_->unpremul_alpha = true;
-    }
-  }
-
-  // Sets the buffer to which uint8 sRGB pixels will be decoded. This is not
-  // supported for all images. If it succeeds, HasRGBBuffer() will return true.
+  // Sets the pixel callback or image buffer where the pixels will be decoded.
+  // This is not supported for all images. If it succeeds, HasRGBBuffer() will
```
veluca93, line 176: I believe this comment is stale :)

veluca93 (dismissed):

veluca93 (dismissed):

veluca93 (approved):

lib/jxl/dec_frame.h
szabadka: I wanted to update it only after extra channels are also handled and then I can remove HasRGBBuffer() and also JxlDecoderOutputImage()

PR #1740, opened 2022-09-13
Handle extra channels in WriteToOutputStage.

szabadka:


lib/jxl/dec_frame.h
```diff
@@ -173,28 +173,21 @@ class FrameDecoder {
...
   void SetImageOutput(const PixelCallback& pixel_callback, void* image_buffer,
-                      size_t xsize, size_t ysize, JxlPixelFormat format,
-                      bool unpremul_alpha, bool undo_orientation) const {
-    dec_state_->pixel_callback = pixel_callback;
-    dec_state_->image_buffer = image_buffer;
+                      size_t image_buffer_size, size_t xsize, size_t ysize,
+                      JxlPixelFormat format, bool unpremul_alpha,
+                      bool undo_orientation) const {
     dec_state_->width = xsize;
     dec_state_->height = ysize;
```
veluca93, line 185: maybe silly question, but how do these relate to image metadata / frame header sizes?

veluca93 (approved):

lib/jxl/dec_frame.h
szabadka: They are image (or preview) dimensions, unless coalescing is disabled, in which case they are frame dimensions. The sizes are calculated by GetCurrentDimensions in decode.cc

veluca93 (approved):

PR #1717, opened 2022-08-25
tests: Make sure libjxl0.7 can still be found

eustas:
In some cases user may set CMAKE_SKIP_RPATH:BOOL=ON in which case we
must indicate explicitly where to find libjxl dynamic library.

Reverse cherry-pick from v0.7.x (PR #1711)

PR #1749, opened 2022-09-15
tests: Make sure libjxl0.7 can still be found

mo271:
In some cases user may set CMAKE_SKIP_RPATH:BOOL=ON in which case we must indicate explicitly where to find libjxl dynamic library.

Reverse cherry-pick from v0.7.x (PR #1711)


mo271:
https://github.com/libjxl/libjxl/pull/1717 was stuck because of the authors check and perhaps because it was cherry-picking too much.

eustas (approved):

PR #1753, opened 2022-09-16
ICC: generate cicp tag when appropriate

sboukortt:


lib/jxl/color_management.cc
```diff
@@ -339,6 +339,44 @@ Status CreateICCChadTag(float chad[9], PaddedBytes* JXL_RESTRICT tags) {
   return true;
 }
 
+void MaybeCreateICCCICPTag(const ColorEncoding& c,
```
kmilos, line 342: Just an idea: perhaps extract common code w/ https://github.com/libjxl/libjxl/blob/d7e18a998e8c28051fe6c95b4d10cae86186cc55/lib/extras/enc/apng.cc#L147 into a general method, so there is no potential for divergence?

sboukortt: We considered it, but `enc/apng` is in `extras` whereas this is in the core library, so we thought that it might be less trouble to duplicate this small amount of code than to figure out how to cleanly share it.

kmilos: Ok, thanks

Issue #1686, opened 2022-08-08
brotli linking issues on macOS


adamjstewart:
**Describe the bug**

When building from source, I encounter linker issues on macOS.
```console
$ otool -L lib/libjxl.dylib 
lib/libjxl.dylib:
	/Users/ajstewart/spack/opt/spack/darwin-monterey-m1/apple-clang-13.1.6/libjxl-main-swqqeydkhwe6vivbmo6aigloq352xq2v/lib/libjxl.0.7.dylib (compatibility version 0.7.0, current version 0.7.0)
	libbrotlidec.1.dylib (compatibility version 1.0.0, current version 1.0.7)
	libbrotlienc.1.dylib (compatibility version 1.0.0, current version 1.0.7)
	libbrotlicommon.1.dylib (compatibility version 1.0.0, current version 1.0.7)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```
As you can see, the linker is unable to find `libbrotli*`.

**To Reproduce**

On macOS:
```console
$ git clone https://github.com/libjxl/libjxl.git
$ cd libjxl
$ git submodule update --init --recursive
$ mkdir build
$ cd build
$ cmake .. -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON
$ make
$ make install
```

**Expected behavior**

I would expect libjxl.dylib to know where to find its dependencies from RPATH.

**Environment**

 - OS: macOS 12.5
 - Compiler version: Apple Clang 13.1.6
 - CPU type: arm64
 - cjxl/djxl version string: ?

**Additional context**

Full build logs:

* [build log](https://github.com/libjxl/libjxl/files/9285579/spack-build-02-build-out.txt)
* [build env](https://github.com/libjxl/libjxl/files/9285581/spack-build-env.txt)

Trying to build libjxl as a GDAL driver. Added a package for libjxl to Spack in https://github.com/spack/spack/pull/31983, let me know if anyone would like to help maintain this build recipe.

P.S. Is it possible to build libjxl with external packages instead of git submodules?

adamjstewart:
Yep, that seems to solve my problem. I'll let you decide whether to close this issue or not, but the git submodule build is definitely still broken and the only way I've found to build libjxl on macOS is with an external brotli installation.

eustas:
[...]

PR #1754, opened 2022-09-16
Set install_name properly when building brotli on macOS

tmkk:
This will fix issue #1686.


mo271:
Thanks for the patch, @tmkk! Could you sing the CLA (instruction in the message above from the google-cla bot ...)?

mo271:
Also you would need to add yourself to the [AUTHORS](https://github.com/libjxl/libjxl/blob/main/AUTHORS) file

mo271:
I rebased instead of merged to fix the authors check...

mo271 (approved):

PR #1760, opened 2022-09-19
cjxl: fix check for resampling flag

### LINKED ISSUE #1755: cjxl not accepting --resampling=2
cjxl doesn't accept --resampling=2, it complains it is an invalid resampling value. Other resampling values do work.
```
d:\>cjxl lena_std.png lena.jxl --resampling=2
JPEG XL encoder v0.8.0 7c0ce23 [SSE4,SSSE3,Unknown]
Invalid flag value for --resampling: Valid values are {-1, 1, 2, 4, 8}.

d:\>cjxl lena_std.png lena.jxl --resampling=4
JPEG XL encoder v0.8.0 7c0ce23 [SSE4,SSSE3,Unknown]
Read 512x512 image, 473243 bytes, 139.0 MP/s
Encoding [VarDCT, d1.000, effort: 7],
Compressed to 5684 bytes (0.173 bpp).
512 x 512, 14.85 MP/s [14.85, 14.85], 1 reps, 4 threads.
```
cjxl should allow using --resampling=2 according to documentation.

It looks like there is a mistake in cjxl_main.cc
https://github.com/libjxl/libjxl/blob/7c0ce231165db47a6a065e01afd7d5083e8605a6/tools/cjxl_main.cc#L834
The valuie 2 is missing in the check.

It looks like the same problem may affect ec_resampling as well as there is also value 2 missing in the check.

mo271:
Fixes #1755 both for --resampling and --ec_resampling.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -831,14 +831,14 @@ int main(int argc, char** argv) {
       process_flag("resampling", args.resampling,
                    JXL_ENC_FRAME_SETTING_RESAMPLING,
                    [](int64_t x) -> std::string {
-                     return (x == -1 || x == 1 || x == 4 || x == 8)
+                     return (x == -1 || x == 1 || x == 2 || x == 4 || x == 8)
                                 ? ""
                                 : "Valid values are {-1, 1, 2, 4, 8}.\n";
                    });
       process_flag("ec_resampling", args.ec_resampling,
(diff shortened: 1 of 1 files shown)


ziemek99:
Do you think it should be merged into v0.7.x branch?

mo271:
Sure, there is no risk that this introduces new bugs...

PR #1762, opened 2022-09-20
remove discontinuity in DC approach at d2

jonsneyers:
Currently we use a different approach for DC at d >= 2 than at d < 2. Most metrics don't notice this, but according to ssimulacra2, this causes a discontinuity at d2; effectively the DC precision in slow gradients drops a bit at d2-epsilon since the effective deadzone doubles when extra_levels goes from 1 to 0 and we switch from 'near-lossless' DC encoding with half-sized quantization buckets to losslessly encoding DC coefficients with full-sized quantization buckets.

This pull request proposes to remove the discontinuity by using the d >= 2 approach also at d < 2, and adjusting the dc quantization factors accordingly to stay close to the original behavior.

Butteraugli 3-norm slightly doesn't like this (small regression in bpp*pnorm of ~0.5%) though max-norm does like it. I have no idea to what extent Butteraugli is reliable for subtle DC adjustments like this though.

Manual inspection of some images where it could make a difference (stp.png and stp2.png) revealed no visible change at 2x zoom in the d1.9 and d2.0 images before and after.

According to ssimulacra2, this is a ~5% improvement at d1.5-2. Red line is before, dark red line is after:
Screen Shot 2022-09-20 at 10 09 06 AM

It's somewhat hard to tell if this is really an improvement or a regression, unless someone else can see any difference upon manual inspection. If we have to go by the metrics, it all depends on which metric to believe. In general I'd argue though that it's better to avoid discontinuities in encoder behavior.

Generally speaking I think we're probably somewhat overly generous anyway in our bit allocation to DC, which is probably OK — better to be safe than sorry. But this does make it hard to assess the impact of subtle changes to DC.

### DIFF
--- a/lib/jxl/decode_test.cc
+++ b/lib/jxl/decode_test.cc
@@ -4798,7 +4798,7 @@ TEST_P(DecodeProgressiveTest, ProgressiveEventTest) {
         // Verify that the returned pass image is actually not the
         // same as the next pass image, by checking that it has a bit
         // worse butteraugli score.
-        EXPECT_LT(distances[next_p] * 1.2f, distances[p]);
+        EXPECT_LT(distances[next_p] * 1.1f, distances[p]);
         p = next_p;
       }
     }
--- a/lib/jxl/enc_adaptive_quantization.cc
+++ b/lib/jxl/enc_adaptive_quantization.cc
@@ -733,8 +733,8 @@ ImageF TileDistMap(const ImageF& distmap, int tile_size, int margin,
   return tile_distmap;
 }
 
-constexpr float kDcQuantPow = 0.57f;
-static const float kDcQuant = 1.12f;
+constexpr float kDcQuantPow = 0.66f;
+static const float kDcQuant = 1.0f;
 static const float kAcQuant = 0.8294f;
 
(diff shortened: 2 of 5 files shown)


jyrkialakuijala:
In the past I have done all adjustments to DC allocation with manual verification -- metrics are not reliable here.

Particularly one needs to be concerned about red-green striping 

There was no intention of the heuristics being discontinuous, but the 2.9 distance was somewhat carefully chosen back in the day

could you make images available for viewing?

jonsneyers:
Example images for viewing:
Before:
https://jon-cld.s3.amazonaws.com/test_images/003/jxl-e6-q80.png
After:
https://jon-cld.s3.amazonaws.com/test_images/003/jxl-nldc-e6-q80.png
Original:
https://jon-cld.s3.amazonaws.com/test_images/reference/003.png

You can replace 003 by 001-049 for other images, e6 by other effort settings (e3-e8), q80 by other quality settings.

jyrkialakuijala (dismissed):

jyrkialakuijala (approved):

jyrkialakuijala (dismissed):

jyrkialakuijala (approved):

PR #1771, opened 2022-09-21
build pixbufloader-jxl as module

novomesk:
These changes are important for MSYS2 (Windows).

gdk-pixbuf plugins are built as modules in other projects too. For example:
https://github.com/AOMediaCodec/libavif/blob/2e8fe112b97ea9b7a0f00e7ae67ba53128f5ee99/contrib/gdk-pixbuf/CMakeLists.txt#L11


mo271 (approved):
thanks!

PR #1775, opened 2022-09-22
Add a ppf based roundtrip test framework.

szabadka:
Start replacing CodecInOut based roundtrip tests in jxl_test.cc


mo271 (approved):
Great change, nice tests!

PR #1798, opened 2022-09-29
Add a jpeg decoding function that is based on libjxl internals.

szabadka:
This is now equivalent to the lossless-jpeg roundtrip version, but is faster because we do not create the actual jxl bitstream.

### DIFF
--- a/lib/extras/decode_jpeg.cc
+++ b/lib/extras/decode_jpeg.cc
@@ -1,0 +1,396 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "lib/extras/decode_jpeg.h"
+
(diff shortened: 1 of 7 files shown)


jonsneyers:
Nice!

Should we add something like this to the libjxl decoder so you can take a jpeg bitstream as input and decode it as if it's a recompressed jxl file? Doing that in a streaming/progressive way will be tricky and quite a bit of work, but if it works already in the simple way (full input bitstream available), that's already quite useful for many use cases: basically it can mean that applications can drop a dependency on libjpeg-turbo or whatever they're currently using to decode jpeg, and use libjxl for both jpeg and jxl decoding...

szabadka:
Yes, that would be the long-term plan. I did not want to add it to djxl yet, as it is still a work in progress and there are no tests yet, I am only testing it through the benchmark with a few input files.

mo271 (approved):
Looks great!

PR #1803, opened 2022-09-30
Also include version.h in the encode header file.

dlemstra:
In the PR #834 the version header file was added to `decode.h` because `decode.cc` uses these defines. Can this also be added to the `encode.h` header file since `encode.cc` also uses these defines?


mo271 (approved):
LGTM, thanks!

Issue #1763, opened 2022-09-20
Guidance for lossy 16bit compression


tbonfort:
I'm implementing support for lossy 16bit RGB compression with JXL and am encountering an issue. However I can reproduce my issue with cjxl.

Test image (16bit png, so doesn't render in gihub):
![extract](https://user-images.githubusercontent.com/358469/191227466-2cde4a16-9945-4464-b5c0-c2a622b0f1ec.png)

Test setup: `cjxl extract.png extract.jxl` , i.e. expecting compression with distance=1, visually lossless

Expected result: visually lossless jxl
Observed result: 1.5kb jxl image, severlely lossy and barely recognisable

Attached screenshots reconverted to 8bits for visualization:
Original:
![Screenshot from 2022-09-20 11-57-12](https://user-images.githubusercontent.com/358469/191228382-625919f7-b0e1-4fc4-ab94-d05139ce4dcf.png)
JXL:
![Screenshot from 2022-09-20 11-57-17](https://user-images.githubusercontent.com/358469/191228499-b5ce2bdc-6a81-46a7-b18f-c46e7d5f3e1c.png)

Is this configuration expected to be supported? If so how?

jyrkialakuijala:
Lots of guesswork in my response, so feel free to ignore. Consider it as an early uneducated response:

Usually 8 bit to 16 bit conversion is done by multiplying the 8 bit values by 257 -- not just writing zeros into the higher bits.

This is bit of a stretch but would it work if you add --intensity_target=15893440 to cjxl (perhaps there is a max that limits this)

(15893440 would be from 80 for srgb intensity, then multiplied with 256**2.2 more for the information being in the lower 8 bits only)

jonsneyers:
Since the png looks black, I assume you only use the low part of the range — e.g. the actual data is 10-bit, and you encode white as 1023 instead of 65535 when representing it in 16 bits. This is not how you're supposed to represent 10-bit data in PNG, so both cjxl and any png viewer will get it wrong.
Two options: 1) use 10-bit ppm as input instead, 2) make a 16-bit png with padding in the low bits instead of padding in the high bits.

Basically the poor visual result is caused by making libjxl think that the image looks exactly like how your browser will render that png: extremely dark, e.g. if it's 10-bit it's as if all colors are between 8-bit rgb #000000 and #030303. This is obviously not how you intend to view the image.

tbonfort:
Thanks Jon for the insights.
[...]

PR #1812, opened 2022-10-05
Add API functions for setting input/output bit depth.

szabadka:
This addresses part of the issue in #1763.

For the encoder the two choices are using the bit depth of the pixel format (a.k.a. repeating higher bits in lower bits), or to use the bit deth from basic info (a.k.a. zero-padding higher bits). For the decoder there is an additional option to use a custom bit depth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+ - encoder API: new function `JxlEncoderSetFrameBitDepth` to set the bit depth
+   of the input buffer.
+ - decoder API: new function `JxlDecoderSetImageBitDepth` to set the bit depth
+   of the output buffer.
+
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+ - encoder API: new function `JxlEncoderSetFrameBitDepth` to set the bit depth
+   of the input buffer.
+ - decoder API: new function `JxlDecoderSetImageBitDeth` to set the bit depth
```
mo271, line 13:
```suggestion
 - decoder API: new function `JxlDecoderSetImageBitDepth` to set the bit depth
```

mo271 (dismissed):

CHANGELOG.md
szabadka: Thanks for catching this! Fixed.

mo271 (approved):

PR #1814, opened 2022-10-05
Add support for progressive jpegs in libjpeg benchmark.

szabadka:


lib/extras/enc/jpg.cc
```diff
@@ -253,28 +257,26 @@ class JPEGEncoder : public Encoder {
...
+        chroma_subsampling = it.second;
+      } else if (it.first == "jpeg_encoder") {
+        if (it.second == "libjpeg") {
+          jpeg_encoder = JpegEncoder::kLibJpeg;
+        } else if (it.second == "sjpeg") {
+          jpeg_encoder = JpegEncoder::kSJpeg;
+        } else {
+          return JXL_FAILURE("unknown jpeg encoder \"%s\"", it.second.c_str());
+        }
+      } else if (it.first == "progressive") {
```
jyrkialakuijala, line 278: we might want to try with several scan scripts -- Moritz knows which ones are popular

jyrkialakuijala: I like something like this:
dc-first for all channels, 
then 5 (or 9) next coefficients possibly Y channel only, but without last bit
all ac, but without last bit
all ac, last bit only

jyrkialakuijala (approved):

lib/extras/enc/jpg.cc
szabadka: Yes, I can add those next. This change was to get a general idea of how much slower progressive decoding is in libjpeg, and for this default scan-script progressive decoding is 2.7x slower.

PR #1815, opened 2022-10-05
jxlinfo: avoiding div by zero for blending

mo271:


mo271:
For blending frames, this triggers undefined behavior, because `info.animation.tps_numerator` is zero. We avoid that by not computing it for non-animation frames

jonsneyers (approved):

PR #1818, opened 2022-10-06
cjxl: avoid div-by-zero

mo271:


mo271:
Running cjxl with lossless transcode under msan gives 
`runtime error: division by zero`. This fixes it.

szabadka (approved):

PR #1830, opened 2022-10-11
Remove dependence on libjxl's render pipeline for jpeg decoding.

szabadka:
We decode the image one MCU-row at a time (8 or 16 rows) instead of by 256x256 tiles.

Benchmark before (single-threaded):
```
Encoding                 kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
-----------------------------------------------------------------------------------------------------------------------
jpeg:q95:djxl8             13270  7112433    4.2876819  55.107  93.795   1.63272071   0.44628983  1.913548813406      0
jpeg:q95:yuv420:djxl8      13270  5230427    3.1531274  83.334 128.316   6.06236982   0.78961182  2.489746632976      0
```

Bnenchmark after:
```
Encoding                 kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
-----------------------------------------------------------------------------------------------------------------------
jpeg:q95:djxl8             13270  7112433    4.2876819  54.919 104.071   1.63272071   0.44628983  1.913548813406      0
jpeg:q95:yuv420:djxl8      13270  5230427    3.1531274  82.888 142.408   6.06236982   0.78961033  2.489741937504      0
```

Benchmark for libjpeg-turbo:
```
Encoding           kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
-----------------------------------------------------------------------------------------------------------------
jpeg:q95             13270  7112433    4.2876819  53.544 139.849   1.71719611   0.48347738  2.072997220807      0
jpeg:q95:yuv420      13270  5230427    3.1531274  81.365 209.858   6.00716972   0.80121163  2.526322308500      0
```


lib/extras/dec_group_jpeg.cc
```diff
@@ -89,47 +100,137 @@ void DequantBlock(const int16_t* JXL_RESTRICT qblock, size_t c,
...
-      }
+void DecodeJpegBlock(const int16_t* qblock, size_t c,
+                     const float* JXL_RESTRICT dequant_matrices,
+                     float* JXL_RESTRICT scratch_space,
+                     float* JXL_RESTRICT output, size_t output_stride) {
+  HWY_ALIGN float* const block = scratch_space + kDCTBlockSize;
+  DequantBlock(qblock, c, dequant_matrices, kDefaultQuantBias, scratch_space);
+  // JPEG XL transposes the DCT, JPEG doesn't.
+  Transpose<8, 8>::Run(DCTFrom(scratch_space, 8), DCTTo(block, 8));
+  TransformToPixels(AcStrategy::DCT, block, output, output_stride,
```
veluca93, line 111: We could also allocate the scratch space on the stack here since IIRC we don't need more than 64 floats or so.

veluca93 (approved):

lib/extras/dec_group_jpeg.cc
szabadka: Here block and scratch_space are both pre-allocated temp space so we would need 128 floats, but that could still be ok. I will try this out in a different PR.

PR #1837, opened 2022-10-17
Compute optimal dequant biases assuming Laplacian AC distribution.

szabadka:
Based on article by J. R. Price and M. Rabbani from 2000.

Benchmark before:
```
Encoding                 kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
-----------------------------------------------------------------------------------------------------------------------
jpeg:q95:djxl8             13270  7112433    4.2876819  54.804 104.038   1.63272071   0.44628983  1.913548813406      0
jpeg:q95:yuv420:djxl8      13270  5230427    3.1531274  81.308 140.788   6.06236982   0.78961033  2.489741937504      0
jpeg:libjxl:djxl8          13270  3197015    1.9272988  19.081  84.911   2.19547820   0.70263982  1.354196861688      0
```

Benchmark after:
```
Encoding                 kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
-----------------------------------------------------------------------------------------------------------------------
jpeg:q95:djxl8             13270  7112433    4.2876819  54.394 101.044   1.41126788   0.42871893  1.838210386950      0
jpeg:q95:yuv420:djxl8      13270  5230427    3.1531274  81.098 140.151   6.03857374   0.77685715  2.449529533303      0
jpeg:libjxl:djxl8          13270  3197015    1.9272988  19.056  83.773   2.23333597   0.70076511  1.350583733701      0
```

### DIFF
--- a/lib/extras/dec_group_jpeg.cc
+++ b/lib/extras/dec_group_jpeg.cc
@@ -28,14 +28,10 @@ namespace jxl {
 namespace HWY_NAMESPACE {
 
 // These templates are not found via ADL.
-using hwy::HWY_NAMESPACE::And;
-using hwy::HWY_NAMESPACE::AndNot;
-using hwy::HWY_NAMESPACE::ApproximateReciprocal;
+using hwy::HWY_NAMESPACE::Abs;
 using hwy::HWY_NAMESPACE::Clamp;
 using hwy::HWY_NAMESPACE::Gt;
-using hwy::HWY_NAMESPACE::IfThenElse;
 using hwy::HWY_NAMESPACE::IfThenElseZero;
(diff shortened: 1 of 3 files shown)


jonsneyers:
Can/should this also be applied to jxl dequant?

szabadka:
> Can/should this also be applied to jxl dequant?

My experience was that this quantization bias has much less impact on jxl. I do not fully understand why this is, perhaps the XYB colorspace changes the AC distribution significantly (XYB jpegs also do not benefit much from this), or because there are more post-processing steps.

PR #1842, opened 2022-10-21
djxl: fix decoding to custom bit depth via --bits_per_sample

### LINKED ISSUE #1840: cjxl --bits_per_sample 16 not working
**Describe the bug**

```
> djxl test.jxl --bits_per_sample 16 test3.png
JPEG XL decoder v0.8.0 60a019fb [AVX2,SSE4,SSSE3,Unknown]
Read 259 compressed bytes.
JxlDecoderSetImageOutBitDepth failed
```

**To Reproduce**
Encode any image into any kind of jxl, and decode into png with --bits_per_sample 16

**Expected behavior**
It should produce a 16 bit png

**Environment**
 - OS: Ubuntu 22.04
 - Compiler version: clang 14
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL decoder v0.8.0 60a019fb [AVX2,SSE4,SSSE3,Unknown]

thebombzen:
Using --bits_per_sample should not fail if the requested bits per sample disagrees with the tagged bits per sample, provided that the output format actually supports the depth that is requested. This commit fixes this discrepancy so 8-bit-tagged JXL files can still be decoded as 16-bit PNGs, and similar.

Fixes #1840.

### DIFF
--- a/lib/extras/dec/jxl.cc
+++ b/lib/extras/dec/jxl.cc
@@ -288,6 +288,30 @@ bool DecodeImageJXL(const uint8_t* bytes, size_t bytes_size,
         fprintf(stderr, "SelectFormat failed\n");
         return false;
       }
+      // Update selected format to forced bit depth.
+      // if we don't set the bit depth here, the JxlPixelFormat
+      // will match the tagged format, regardless of what bit
+      // depth the user requests, which will fail verification
+      switch (dparams.output_bitdepth.bits_per_sample) {
+        case 0:
+          // output_bitdepth == 0 means default
+          break;
+        case 8:
+          format.data_type = JXL_TYPE_UINT8;
+          break;
+        case 16:
+          format.data_type = JXL_TYPE_UINT16;
+          break;
+        case 32:
+          format.data_type = JXL_TYPE_FLOAT;
+          break;
+        default:
+          fprintf(
+              stderr,
+              "Invalid Bit Depth Requested: %d. Only 8, 16, and 32 supported\n",
+              dparams.output_bitdepth.bits_per_sample);
+          return false;
+      }
       bool have_alpha = (format.num_channels == 2 || format.num_channels == 4);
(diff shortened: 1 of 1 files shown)


lib/extras/dec/jxl.cc
```diff
@@ -288,6 +288,30 @@ bool DecodeImageJXL(const uint8_t* bytes, size_t bytes_size,
...
+      // depth the user requests, which will fail verification
+      switch (dparams.output_bitdepth.bits_per_sample) {
+        case 0:
+          // output_bitdepth == 0 means default
+          break;
+        case 8:
+          format.data_type = JXL_TYPE_UINT8;
+          break;
+        case 16:
+          format.data_type = JXL_TYPE_UINT16;
```
kmilos, line 303: Isn't float16 a possibility as well?

thebombzen: The user doesn't have a way to request float16 on the command line, they only get a bits per sample value. None of the output formats support float16 that don't also support int16 so this is safer.

kmilos: > The user doesn't have a way to request float16 on the command line

Ah, therein lies the bug perhaps.



> None of the output formats support float16 that don't also support int16 so this is safer.

I beg to differ: [uint16 will clip, so it is less "safe"](https://github.com/libjxl/libjxl/blob/99b0721251b51028927c56a81657cf76de7ea320/lib/include/jxl/codestream_header.h#L167-L185); float16 is very popular/default format for EXRs that are used a lot in ACES workflows (movies, CGI, VFX...).

thebombzen: The bug is that the option is not respected. Ideally the user should have an option to request the pixel format, but at the very least the existing option of bits-per-sample should actually work.

thebombzen:
Looks like it was fixed by #1850.

PR #1848, opened 2022-10-26
Ssimulacra2: linear downscaling and fixed SSIM

jonsneyers:
This makes two changes to the SSIMULACRA 2 algorithm:

- Downscaling (for the 1:2, 1:4 etc scales) is done in linear RGB. This is perceptually more correct, and allows the metric to notice any artifacts caused by codecs typically working in a gamma-compressed domain, causing e.g. the DC values to not actually correspond exactly to the correct color for a 1:8 image.
- The SSIM formula was fixed. There was a conceptual error in the original SSIM formula: it is effectively applying gamma correction twice: once by working in a gamma-compressed domain (with e.g. the sRGB transfer curve in typical SSIM implementations, or the XYB transfer curve in case of SSIMULACRA 2), and then again by having a division by mean luma which effectively penalizes errors in the darks more than in the brights, which does not make sense if gamma compression was already performed beforehand. This also makes the resulting error map make sense when applied to chroma channels (since it definitely does not make sense to e.g. penalize errors in the reds more than in the greens).

Weights have been re-tuned. On the validation set (not used in tuning), these are the correlations:

Metric | Kendall | Spearman | Pearson
-----|-----|------|-----
PSNR | 0.34911 | 0.49951 | 0.50134
SSIM | 0.45788 | 0.63796 | 0.56530
VMAF | 0.58789 | 0.78829 | 0.75012
DSSIM | -0.6806 | -0.8721 | -0.8219
Butteraugli max-norm | -0.5499 | -0.7408 | -0.6832
Butteraugli 3-norm | -0.6102 | -0.7938 | -0.7449
SSIMULACRA | -0.5939 | -0.7912 | -0.7862
SSIMULACRA 2 before | 0.71108 | 0.89225 | 0.88503
SSIMULACRA 2 after | 0.70330 | 0.88541 | 0.87448

The correlations are a bit lower than before, which I believe is caused by the linear downscaling making it harder for the tuning to fit the data in a non-generalizing, codec-specific (in particular jxl-specific) way. So while it does not look like an improvement based on just the correlation numbers, I do think it's an actual improvement in terms of robustness and actual perceptual relevance.

### DIFF
--- a/tools/ssimulacra2.cc
+++ b/tools/ssimulacra2.cc
@@ -4,16 +4,17 @@
 // license that can be found in the LICENSE file.
 
 /*
-SSIMULACRA 2 - Structural SIMilarity Unveiling Local And Compression Related
-Artifacts
+SSIMULACRA 2
+Structural SIMilarity Unveiling Local And Compression Related Artifacts
 
 Perceptual metric developed by Jon Sneyers (Cloudinary) in July 2022.
 Design:
 - XYB color space (X+0.5, Y, Y-B+1.0)
-- SSIM map
+- SSIM map (with correction: no double gamma correction)
 - 'blockiness/ringing' map (distorted has edges where original is smooth)
 - 'smoothing' map (distorted is smooth where original has edges)
 - error maps are computed at 6 scales (1:1 to 1:32) for each component (X,Y,B)
+- downscaling is done in linear RGB
 - for all 6*3*3=54 maps, two norms are computed: 1-norm (mean) and 4-norm
 - a weighted sum of these 54*2=108 norms leads to the final score
 - weights were tuned based on a large set of subjective scores for images
@@ -26,6 +27,7 @@ Design:
 
 #include <cmath>
 
+#include "lib/jxl/base/printf_macros.h"
 #include "lib/jxl/enc_color_management.h"
 #include "lib/jxl/enc_xyb.h"
 #include "lib/jxl/gauss_blur.h"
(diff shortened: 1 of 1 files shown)


mo271 (dismissed):

mo271 (approved):

jyrkialakuijala:
The summation of pixels can happen physically when you look far apart.

If you zoom out, usual systems would sum the pixels in their compressed space (I think).

There is some summation of pixels that happens after receptors. When compression happens in the receptors, these neural sums in the retina or visual cortex will relate better to the compressed sums.

I don't think it is necessarily better to compute these sums in the linear space.

jonsneyers:
> The summation of pixels can happen physically when you look far apart.
> 
> If you zoom out, usual systems would sum the pixels in their compressed space (I think).
> 
> There is some summation of pixels that happens after receptors. When compression happens in the receptors, these neural sums in the retina or visual cortex will relate better to the compressed sums.
> 
> I don't think it is necessarily better to compute these sums in the linear space.

You can look at the disagreements between this version (with linear downscaling, called modelD) and the previous version (with nonlinear downscaling, called modelB) here: https://jon-cld.s3.amazonaws.com/test/disagreement_SSIMULACRA_2_modelD_vs_SSIMULACRA_2_modelB.html


The biggest disagreements were related to distorted version of this image: https://jon-cld.s3.amazonaws.com/test/images/1028637.png

When I looked at these disagreements, I tended to agree with the modelD scores.

Also when testing on this example: https://twitter.com/jonsneyers/status/1580457266503114752?s=20
modelB says the white background one is better while modelD says the black background one is better.

So for now, I think doing it in linear space is the best choice.

I do agree with you though that modeling physical light versus post-receptor aggregation should probably be done differently, i.e. at some point viewing distance has to be taken into account in order to estimate which kind of modeling is most appropriate at which scale. As a first approximation, perhaps doing downscaling in linear space for the first 1-2 downscales and in compressed space for the next scales could give better results.

PR #1863, opened 2022-11-04
plugins/gimp: Support for decoding animations

ctrlcctrlv:
animated_ico_in_gimp

(edited by @mo271)

### DIFF
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -347,7 +347,7 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
         char *temp_frame_name = nullptr;
         bool must_free_frame_name = false;
         if (frame_name_len == 0) {
-          temp_frame_name = g_strdup_printf("Frame %lu", layer_idx);
+          temp_frame_name = g_strdup_printf("Frame %lu", layer_idx + 1);
           must_free_frame_name = true;
         } else {
           temp_frame_name = frame_name;
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -75,8 +75,9 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
     g_printerr(LOAD_PROC " Error: JxlDecoderSetParallelRunner failed\n");
     return false;
   }
-
-  if (JXL_DEC_SUCCESS != JxlDecoderSetCoalescing(dec.get(), JXL_FALSE)) {
+  // TODO: make this work with coalescing set to false, while handling frames
+  // with duration 0 and references to earlier frames correctly.
+  if (JXL_DEC_SUCCESS != JxlDecoderSetCoalescing(dec.get(), JXL_TRUE)) {
     g_printerr(LOAD_PROC " Error: JxlDecoderSetCoalescing failed\n");
     return false;
   }
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -59,10 +59,9 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
 
   auto dec = JxlDecoderMake(nullptr);
   if (JXL_DEC_SUCCESS !=
-      JxlDecoderSubscribeEvents(dec.get(), JXL_DEC_BASIC_INFO |
-                                               JXL_DEC_COLOR_ENCODING |
-                                               JXL_DEC_FULL_IMAGE |
-                                               JXL_DEC_FRAME)) {
+      JxlDecoderSubscribeEvents(dec.get(),
+                                JXL_DEC_BASIC_INFO | JXL_DEC_COLOR_ENCODING |
+                                    JXL_DEC_FULL_IMAGE | JXL_DEC_FRAME)) {
     g_printerr(LOAD_PROC " Error: JxlDecoderSubscribeEvents failed\n");
     return false;
   }
@@ -101,8 +100,7 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
       }
 
       JxlResizableParallelRunnerSetThreads(
-          runner.get(),
-          JxlResizableParallelRunnerSuggestThreads(xsize, ysize));
+          runner.get(), JxlResizableParallelRunnerSuggestThreads(xsize, ysize));
     } else if (status == JXL_DEC_COLOR_ENCODING) {
       // check for ICC profile
       size_t icc_size = 0;
@@ -298,11 +296,11 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
 
       // create new image
       if (is_linear) {
-        *image_id = gimp_image_new_with_precision(
-            xsize, ysize, image_type, GIMP_PRECISION_FLOAT_LINEAR);
+        *image_id = gimp_image_new_with_precision(xsize, ysize, image_type,
+                                                  GIMP_PRECISION_FLOAT_LINEAR);
       } else {
-        *image_id = gimp_image_new_with_precision(
-            xsize, ysize, image_type, GIMP_PRECISION_FLOAT_GAMMA);
+        *image_id = gimp_image_new_with_precision(xsize, ysize, image_type,
+                                                  GIMP_PRECISION_FLOAT_GAMMA);
       }
 
       if (profile_int) {
@@ -327,15 +325,15 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
       }
     } else if (status == JXL_DEC_FULL_IMAGE) {
       // create and insert layer
-      gchar* layer_name;
+      gchar *layer_name;
       if (layer_idx == 0 && !info.have_animation) {
         layer_name = g_strdup_printf("Background");
       } else {
         layer_name = g_strdup_printf("Frame %lu (%.1fms)", layer_idx,
                                      frame_duration * tick_duration / 1000.0);
       }
-      layer = gimp_layer_new(*image_id, layer_name, xsize, ysize,
-                             layer_type, /*opacity=*/100,
+      layer = gimp_layer_new(*image_id, layer_name, xsize, ysize, layer_type,
+                             /*opacity=*/100,
                              gimp_image_get_default_new_layer_mode(*image_id));
 
       gimp_image_insert_layer(*image_id, layer, /*parent_id=*/-1,
@@ -361,8 +359,8 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
       babl_process(babl_fish(source_format, destination_format),
                    pixels_buffer_1, pixels_buffer_2, xsize * ysize);
 
-      gegl_buffer_set(buffer, GEGL_RECTANGLE(0, 0, xsize, ysize), 0,
-                      nullptr, pixels_buffer_2, GEGL_AUTO_ROWSTRIDE);
+      gegl_buffer_set(buffer, GEGL_RECTANGLE(0, 0, xsize, ysize), 0, nullptr,
+                      pixels_buffer_2, GEGL_AUTO_ROWSTRIDE);
       gimp_item_transform_translate(layer, crop_x0, crop_y0);
 
       g_clear_object(&buffer);
@@ -370,9 +368,10 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
       layer_idx++;
     } else if (status == JXL_DEC_FRAME) {
       JxlFrameHeader frame_header;
-      if (JxlDecoderGetFrameHeader(dec.get(), &frame_header) != JXL_DEC_SUCCESS) {
-          g_printerr(LOAD_PROC " Error: JxlDecoderSetImageOutBuffer failed\n");
-          return false;
+      if (JxlDecoderGetFrameHeader(dec.get(), &frame_header) !=
+          JXL_DEC_SUCCESS) {
+        g_printerr(LOAD_PROC " Error: JxlDecoderSetImageOutBuffer failed\n");
+        return false;
       }
       xsize = frame_header.layer_info.xsize;
       ysize = frame_header.layer_info.ysize;
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -20,7 +20,7 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
   long crop_x0 = 0, crop_y0 = 0;
   size_t layer_idx = 0;
   uint32_t frame_duration = 0;
-  double tps_factor = 1.0;
+  double tps_denom = 1.f, tps_numer = 1.f;
 
   gint32 layer;
 
@@ -99,9 +99,8 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
       ysize = info.ysize;
       if (info.have_animation) {
         animation = info.animation;
-        tps_factor = animation.tps_numerator / animation.tps_denominator;
-        // GIMP expects animation durations to be expressed in ms.
-        tps_factor /= 1000.0;
+        tps_denom = animation.tps_denominator;
+        tps_numer = animation.tps_numerator;
       }
 
       JxlResizableParallelRunnerSetThreads(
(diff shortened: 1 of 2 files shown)


ctrlcctrlv:
Test file:
* Russian Propaganda.jxl
    ```bash
    zip -Z store Russian\ Propaganda.jxl.zip Russian\ Propaganda.jxl
    ```
    * [Russian Propaganda.jxl.zip](https://github.com/libjxl/libjxl/files/9936207/Russian.Propaganda.jxl.zip)

Follows up to work done by @xiota and others.

mo271:
Thanks a lot for the contribution, @ctrlcctrlv!

I tested plugin with an animation from our conformace repo [animation_icos4d/input.jxl](https://github.com/libjxl/conformance/blob/43d8135b50e53167ee6fee0b842b9eb15cfc4aa2/testcases/animation_icos4d/input.jxl)
Seems to work like a charm.
In order to comply with [github-acceptable-use-policies](https://docs.github.com/en/site-policy/acceptable-use-policies/github-acceptable-use-policies),
I’ll edit your comments to replace the screenshot and link by a screenshot of and link to the icosahedron image.

Looking forward to a pull request for encoding animation, like you propose in #1864, if you come around doing that.

mo271:
I fixed the failing authors check and linting https://github.com/libjxl/libjxl/actions/runs/3392132321/jobs/5643700386, taking the name from the commit.

mo271 (dismissed):

plugins/gimp/file-jxl-load.cc
```diff
@@ -333,12 +357,28 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
...
+          JXL_DEC_SUCCESS) {
+        g_printerr(LOAD_PROC " Error: JxlDecoderSetImageOutBuffer failed\n");
+        return false;
+      }
+      xsize = frame_header.layer_info.xsize;
+      ysize = frame_header.layer_info.ysize;
+      crop_x0 = frame_header.layer_info.crop_x0;
+      crop_y0 = frame_header.layer_info.crop_y0;
+      frame_duration = frame_header.duration;
+
```
sboukortt, line 381: A very superficial comment: perhaps we could remove this empty line at the end of the block.

sboukortt (approved):
Great, thanks!

plugins/gimp/file-jxl-load.cc
```diff
@@ -81,9 +92,15 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
         return false;
       }
 
+      xsize = info.xsize;
+      ysize = info.ysize;
+      if (info.have_animation) {
+        animation = info.animation;
+        tick_duration = animation.tps_numerator / animation.tps_denominator;
```
alistair7, line 99: Looks like this will need a cast to avoid integer division.

thebombzen: It would be denominator / numerator, and the units are in seconds. I'm not sure the units on `tick_duration` but double check to see if it's in seconds or milliseconds or something.

ctrlcctrlv: It's a bad variable name. I'm not using it as a duration.

ctrlcctrlv:
I apologize for the offensive image originally used. It was an art project, and I wasn't aware there was an existing animation I could use as a test.

ctrlcctrlv:
Besides handling review comments, 21adf35 handles blend modes equivalently to how it is done in the GIF loading plugin. Therefore, it is possible to export directly as GIF now.

input

Screenshot from 2022-11-05 02-01-58
Screenshot from 2022-11-05 02-01-50

plugins/gimp/file-jxl-load.cc
alistair7: In that case, when creating the layer name, I think you need to divide by `tps_factor` rather than multiplying.  See attached for example.  Each frame should be 2000ms.
[rgb.jxl.zip](https://github.com/libjxl/libjxl/files/9944131/rgb.jxl.zip)

ctrlcctrlv: Thanks, that's right.

ctrlcctrlv: Fixed by [f150585](https://github.com/libjxl/libjxl/pull/1863/commits/f150585111c933a8396cb53013c4e47e1375baad).

Screenshot from 2022-11-05 13-01-01

ctrlcctrlv:
Thanks for being so careful @alistair7. It was working for me with my test files because they all had the same denominator/numerator. My original commit (f150585) didn't quite fix it, it only fixed it for your files and not mine, so I based b81ff91 on `tools/jxlinfo.c`, as that program understands both.

Screenshot from 2022-11-05 13-24-17

That's `Wikipedia.jxl`:
[Wikipedia.jxl.7z.001.zip](https://github.com/libjxl/libjxl/files/9944398/Wikipedia.jxl.7z.001.zip)
[Wikipedia.jxl.7z.002.zip](https://github.com/libjxl/libjxl/files/9944396/Wikipedia.jxl.7z.002.zip)

alistair7:
No problem. Nice work!

ctrlcctrlv:
78d0a94 omits frames of zero duration, which _shouldn't_ cause issues, but did indeed cause flickering in GIMP's animation playback function.

There are no more blank frames in `Wikipedia.jxl`.

Screenshot from 2022-11-05 19-20-36

mo271 (approved):

sboukortt:
> [78d0a94](https://github.com/libjxl/libjxl/commit/78d0a94a3aa29059ca9b9a7ec898e7bb8aa2d8f2) omits frames of zero duration, which _shouldn't_ cause issues, but did indeed cause flickering in GIMP's animation playback function.

Should it really do that? Those frames could conceivably play a part in rendering due to blending.

ctrlcctrlv:
@sboukortt Yes it should. If the display time is 0, then that means no display. How could it not?

ctrlcctrlv:
I think it's a cjxl bug that those frames even get added, by the way. I can remove my workaround, but unless the spec itself clarifies this issue I think my reading is logical, and will result in a less buggy plugin for users as cjxl is the most common encoder.

mo271:
> @sboukortt Yes it should. If the display time is 0, then that means no display. How could it not?

What @sboukortt presumably means by "Those frames could conceivably play a part in rendering due to blending" is that you could have a frame of duration 0, that is a new background and then the next frame is some image that has alpha on top of that which comes with a non-zero duration. You never want to see a frame that is just the new background, but put something on it immediately. For this or in slightly more involved blending situations it is useful to allow frames of duration 0, no?

ctrlcctrlv:
@mo271 I know what they mean, but cjxl doesn't do this. It (seemingly randomly) adds transparent layers of duration 0. I think that this should be accepted, if it matters to someone they should investigate why cjxl does this and get it to stop, then the two commits [78d0a94](https://github.com/libjxl/libjxl/pull/1863/commits/78d0a94a3aa29059ca9b9a7ec898e7bb8aa2d8f2) and [2193f94](https://github.com/libjxl/libjxl/pull/1863/commits/2193f94325fc90b48c640786131df2a667da480f) can be reverted.

ctrlcctrlv:
> > [78d0a94](https://github.com/libjxl/libjxl/commit/78d0a94a3aa29059ca9b9a7ec898e7bb8aa2d8f2) omits frames of zero duration, which _shouldn't_ cause issues, but did indeed cause flickering in GIMP's animation playback function.
> 
> Should it really do that? Those frames could conceivably play a part in rendering due to blending.

Maybe c28e3588a95d8a4f1282e613d26dd8b7ebc70248 will make you happier. It's more hacky, and looks terrible, lol, but on the off chance someone creates an image without cjxl that requires this strange combining edge case, I am now doing the combine in GIMP via `gimp_image_merge_down(image,layer)` instead of just dismissing the layer.

mo271:
I noticed that there was `JxlDecoderSetCoalescing`, and I think when you set this to `JXL_TRUE` or not call it at all, (since then the default value is true), then the problem with the 0-duration frames will go away. 
In fact I checked just that in this branch:
https://github.com/libjxl/libjxl/compare/main...mo271:libjxl:coalescing_gimp
with the wikipedia image, and I get 248 frames just as expected. 
So I suggest to do it that way. What do you think?

mo271:
https://libjxl.readthedocs.io/en/latest/api_decoder.html#_CPPv423JxlDecoderSetCoalescingP10JxlDecoder8JXL_BOOL

ctrlcctrlv:
Sorry @mo271 but I feel like I need the coalescing. It makes the output much more useful to actually edit, especially for my original example, Russian Propaganda.jxl.

mo271:
> Sorry @mo271 but I feel like I need the coalescing. It makes the output much more useful to actually edit, especially for my original example, Russian Propaganda.jxl.

When you say "I need the coalescing" do you mean "I need set coalescing to be true" or "I need set coalescing to be false"?

I checked and it seems that the two frames in that image work exactly as intended with the coalescing set to be true, which is what is done in https://github.com/libjxl/libjxl/compare/main...mo271:libjxl:coalescing_gimp.

In what way does it make the output more useful to actually edit? (I'm not sure I'm understanding this correctly...)

thebombzen:
Sounds like they're saying that they want the separate layers to remain separate so they can be treated as layers in GIMP.

mo271:
> Sounds like they're saying that they want the separate layers to remain separate so they can be treated as layers in GIMP.

Even with coalescing set to true, each frame in the animation will be separate layers in GIMP. Not sure what is going on with the empty frames of duration zero when we turn off coalescing, perhaps GIMP can't handle properly frames of different dimensions.

Will open a bug for further understand of what is going on.

ctrlcctrlv:
>Even with coalescing set to true, each frame in the animation will be separate layers in GIMP.

Yes, but with the entire background. Not as easy to edit at all.

alistair7:
Another complication with having coalescing disabled is that a frame's `blend_source` doesn't necessarily refer to the previous frame. That's probably not a concept that Gimp's combine/replace tags can deal with, so the exported results won't look right in those cases.  I'm not sure whether cjxl will ever produce a file like that from a converted GIF/APNG, but they certainly exist.  Silly example attached: [face_skip_frame.jxl.zip](https://github.com/libjxl/libjxl/files/9974744/face_skip_frame.jxl.zip)

I don't know the best approach to dealing with this (if at all).  I guess at worst, you could consider throwing an error if we find a frame where `blend_info.source != previous_save_as_reference`, where `previous_save_as_reference` is stored from the previous frame's `layer_info.save_as_reference` ...in a similar way to how it currently rejects the more exotic blend modes.

Or would it make any sense for the plugin to *try* decoding without coalescing, but on finding any non-trivial blending, reset the decoder and try again with coalescing enabled?  Maybe that's overcomplicating it...

jonsneyers:
I think this points to a gap in our current API, since there are currently only two options:
- Coalescing (default): gives only the blended frames, so you lose the crop region and transparency (i.e. only returns full-canvas sized frames with nonzero duration)
- No coalescing: gives all frames as they are, without blending, including zero-duration frames (i.e. layers) that are kRegularFrames (the kReferenceOnly and kLfFrames ones are never returned because they are intended strictly to be used only as coding tools, not as semantically meaningful objects by themselves).

What would be useful here is an option to merge the zero-duration frames but not to blend the nonzero-duration frames, i.e. return only nonzero-duration frames but still have them as they are before blending and making them canvas-sized.

The question is if this can even be done in general, since jxl's frame blending model is not simply a linear stack where every next frame is "on top" of the one before, but frame blending can take other "source frames" than the usual one that is the current canvas state. This model is strictly richer than what can be expressed in GIF/APNG and in Gimp layers, so in the general case there is no way to map a jxl frame sequence to that.

For the specific case of transcoded GIF/APNG/AWebP though, there should be a way to map the jxl frame sequence to a suitable Gimp layer stack. I suspect the dummy empty frames cjxl generates are just a way to circumvent the lack of dispose modes in jxl, or something like that.

jonsneyers:
Not related to this, but another coalescing mode that would be good to have for animation display, is one where the returned frames are blended but only the crop box containing modified pixels gets returned (so it only returns kReplace frames like the default coalescing, but not necessarily canvas-sized ones). This has obvious advantages in terms of buffer sizes and would allow e.g. browsers to have a higher likelihood of being able to keep the entire animation in memory.

So I think we basically need one boolean setting to indicate whether zero-duration frames should be blended or not, and another boolean setting to indicate whether cropped frames should be returned or canvas-sized ones. We currently only have an option where it is (true,false)  (the default coalescing==true) and another where it is (false,true)  (this is what coalescing==false does).

And then we need to find a way for dealing with exotic blend source choices; probably this is something most applications will not want to deal with so it's probably best if by default anything that is not a simple "stack of layers" gets converted in something that is (we can always do the blending and turn anything into a kReplace frame that overwrites the whole canvas), and only when some new option is set, you can get the original thing which may be quite exotic.
Also it would be good to have a way to tell the decoder not to return 'exotic' blend modes (i.e. not kReplace and not kBlend), since in many cases (like Gimp) the application will already know about replace and blend but not about the other blend modes.

mo271:
So how to best proceed here? 
With the current API, I see three options

1. set coalescing to true 
2. set coalescing to false and not merge 0 duration frames
3. set coalescing to false and merge 0 duration frame 

The problem for 2. and 3. is that there will be animated jxl files that blend source will not be the previous frame, which gimp cannot seemt to handle, like @alistair7 pointed out. 

Might it be possible to make the user select an option during the opening of a file, if there are frames of duration 0?

I think right now the cleanest approach would be 1., because it seems to work in all cases, however I can see that it would be nicer if the structure of the layers would be preserved.

Really great that you implemented the decoding (and encoding!) of animated JXLs in the gimp plugin, @ctrlcctrlv, showing that the API with respect to coalescing can possibly be improved. The source of the problem comes from the fact that jxl is more with frames and layers than gimp...

Would the problem with solution 2. other than the problem with further-back-reference, be that flickering in GIMP's animation playback function? Perhaps one could get that resolved in GIMP somehow.

jonsneyers:
Zero-duration frames in Gimp will not only be an issue for its own playback, but it will be even worse if you then save it as GIF/APNG/WebP where typical players impose a minimum duration per frame because of legacy content that relies on that.

ctrlcctrlv:
I think wrong assumptions about GIMP's blending modes are being made. GIMP has many blending modes. I just didn't implement this because the plugin as it existed before my patches was already so incomplete I viewed my patchset(s) as incremental improvements, but it seems that I'm being held to a higher standard than the original patches. In any event, GIMP has plenty of support.

Screenshot_20221110_135443

But, the API issue _remains_, because GIMP is not the only editor out there. And, as @jonsneyers correctly notes, in other formats there's a minimum delay, so the Coalescing API needs fixed anyway.

mo271:
I agree; I think having the ability to decode at least some (not too exotic) animations is clearly an improvement over the current state of the gimp plugin. Do you prefer to get the version in https://github.com/libjxl/libjxl/commit/c28e3588a95d8a4f1282e613d26dd8b7ebc70248 (with the hack to remove 0 duration frames in gimp, while setting coalescing to false) or the versio in
https://github.com/mo271/libjxl/tree/coalescing_gimp (setting coalescing to true) merged?
I have a slight preference for the 2nd, since it will likely work for more files (having the drawback of not getting the layers correct for some of them).
@ctrlcctrlv

jonsneyers:
Coalescing frames has the disadvantage that going gif -> jxl -> gif will potentially produce a way larger gif, when the original had many frames that update only a small area. Then again I think Gimp does have an "optimize for gif" thing that can trim the frames again.

So if we want to get something that works in all cases first and make it better later, then coalescing=true is the way to go for now, I think. But we should expand the api to get a way to avoid coalescing while not returning exotic blend sources/modes — i.e. have a way to select which blend modes you want to keep in output and which ones you want libjxl to convert to kReplace, and also have a way to get only "blend source for current frame is current canvas state". This would be useful in browsers/viewers (to save memory and repaint by avoiding all frames to be full canvas sized), where you would want to get only kReplace, and also for Gimp, where you want to get both kReplace and kBlend.

For animations, merging zero-duration frames makes sense, but for layered images, keeping them as seperate layers makes more sense. So I think this is an orthogonal thing that should be configurable. For this plugin, it would make sense to merge zero-duration frames iff it is an animation, and to keep them as separate layers if it is a layered still image. The case of animations with layered frames is not something Gimp can currently handle and it is not something that will exist in the wild since there are no existing formats that can do that, or authoring tools that produce them, so merging the layers within each frame is OK in this case.

ctrlcctrlv:
I'm OK with coalescing=true for now. Users of the GIMP format already are well aware of its many shortcomings and know that they may have to use e.g. IM with `-layers Optimize` after a convert from an exotic format. GIMP's plugin for that may even provide such a function.

mo271 (approved):

sboukortt (approved):

mo271:
Many thanks again, @ctrlcctrlv!

ctrlcctrlv:
My pleasure :o)

PR #1867, opened 2022-11-04
tools/libjxl_test.c: avoid function declaration without prototype

thebombzen:
Avoid declaring int main() without a prototype as this is deprecated in all versions of C.


mo271 (approved):

mo271:
Thanks!

PR #1868, opened 2022-11-05
fast_lossless: rename FastLosslessEncode

### LINKED ISSUE #1855: experimental/fast_lossless doesn't anon-namespace its BitWriter
I can't combine `libjxl/experimental/fast_lossless` and the `github.com/veluca93/fpnge` encoder code in the same program (e.g. to compare the two encoders' speeds), because both fjxl and fpnge declare `struct BitWriter` in the top-level namespace, with the same name (and both have a `Write(uint32_t, uint64_t)` method) but different field layout (see links below). Linking both of them leads to a crash, presumably because the fjxl code mistakenly calls the fpnge BitWriter or vice versa.

- https://github.com/libjxl/libjxl/blob/506714edaa3dd670133b9fe094d5efb94f6d0f6e/experimental/fast_lossless/fast_lossless.cc#L46
- https://github.com/veluca93/fpnge/blob/48020025e197318ab3c695903a4577e3d9565ddb/fpnge.cc#L374

I believe one solution would be to wrap a C++ "anonymous namespace" around all of `fast_lossless.cc` code except for the final `FastLosslessEncode` function.

@veluca93

nigeltao:
Adding a "Jxl" prefix disambiguates it when combining it with e.g. a fast PNG encoder or fast WebP encoder. Some lines were subsequently re-formatted to fit within 80 columns.

Also wrap an "extern C" around that function.

Also wrap an anonymous namespace around the implementation, so that the fjxl BitWriter doesn't conflict with e.g. fpnge's BitWriter.

Also fix some -Wunknown-pragmas and -Wsign-compare g++ warnings.

Fixes #1855

### DIFF
--- a/experimental/fast_lossless/fast_lossless.cc
+++ b/experimental/fast_lossless/fast_lossless.cc
@@ -20,6 +20,8 @@
 #error "system not known to be little endian"
 #endif
 
+namespace {
+
 struct BitWriter {
   void Allocate(size_t maximum_bit_size) {
     assert(data == nullptr);
@@ -143,10 +145,10 @@ struct PrefixCode {
   static void ComputeCodeLengths(uint64_t* freqs, size_t n, size_t limit,
                                  uint8_t* nbits) {
     if (n <= 1) return;
-    assert(n <= (1 << limit));
+    assert(n <= (1u << limit));
     assert(n <= 32);
-    int parent[64] = {};
-    int height[64] = {};
+    unsigned int parent[64] = {};
+    unsigned int height[64] = {};
     using QElem = std::pair<uint64_t, size_t>;
     std::priority_queue<QElem, std::vector<QElem>, std::greater<QElem>> q;
     // Standard Huffman code construction. On failure (i.e. if going beyond the
@@ -1088,8 +1090,8 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
   encoder.code = &code;
   int16_t p[4][32 + 1024] = {};
   uint8_t prgba[4];
-  int i = 0;
-  int have_zero = 0;
+  size_t i = 0;
+  size_t have_zero = 0;
   if (palette[pcolors_real - 1] == 0) have_zero = 1;
   for (; i < pcolors; i++) {
     if (i < pcolors_real) {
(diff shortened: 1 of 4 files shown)


nigeltao:
@veluca93

experimental/fast_lossless/fast_lossless.cc
```diff
@@ -1297,7 +1299,9 @@ size_t LLEnc(const unsigned char* rgba, size_t width, size_t stride,
     PrepareDCGlobalPalette(onegroup, width, height, hcode, palette, pcolors,
                            &group_data[0][0]);
   }
+#ifdef __clang__
```
veluca93, line 1302: I don't think this is the right #define - iirc gcc -fopenmp is a thing too.

experimental/fast_lossless/fast_lossless.cc
```diff
@@ -1324,9 +1328,16 @@ size_t LLEnc(const unsigned char* rgba, size_t width, size_t stride,
   return writer.bytes_written;
 }
 
-size_t FastLosslessEncode(const unsigned char* rgba, size_t width,
-                          size_t stride, size_t height, size_t nb_chans,
-                          size_t bitdepth, int effort, unsigned char** output) {
+}  // namespace
+
+size_t JxlFastLosslessEncode(const unsigned char* rgba,
```
veluca93, line 1333: doesn't this need an extern "C" too then?

experimental/fast_lossless/fast_lossless.cc
nigeltao: Changed to `#ifdef _OPENMP`.

experimental/fast_lossless/fast_lossless.cc
nigeltao: Technically, I don't think it's necessary (https://stackoverflow.com/a/1380926). But I put an extern here too.

veluca93: uh, TIL. Still useful to have it for documentation and possible shenanigans.

veluca93 (approved):

veluca93 (approved):

Issue #1864, opened 2022-11-04
GIMP plugin cannot encode animations


ctrlcctrlv:
I solved the _decoding_ of animations in #1863. Probably I will do this too.

mo271:
Great!

PR #1874, opened 2022-11-07
plugins/gimp: Add support for encoding JPEG XL animations

ctrlcctrlv:
[plugins/gimp: Add support for encoding JPEG XL animations](https://github.com/libjxl/libjxl/commit/63cb827ee7e33843b87153dade25e711b8883e93)

Closes https://github.com/libjxl/libjxl/issues/1864.

Because I use the same semantics as the GIMP's existing GIF and WebP plugins, one can open a WebP and then immediately save a JPEG XL. E.g.:

Screenshot from 2022-11-07 00-37-29

**Note**: This assumes the acceptance of my other PR's for the GIMP plugin, #1863 and #1870. This should be merged only when those are merged. It became too difficult to try to maintain separate histories given how intertwined some of my (necessary) changes are.

### DIFF
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -348,16 +348,13 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
           last_frame_blank = true;
           continue;
         }
-        const GString *blend_null_flag = g_string_new("");
-        const GString *blend_replace_flag = g_string_new(" (replace)");
-        const GString *blend_combine_flag = g_string_new(" (combine)");
         GString *blend;
         if (blend_mode == JXL_BLEND_REPLACE || last_frame_blank) {
-          blend = (GString *)blend_replace_flag;
+          blend = (GString *)GIMP_BLEND_REPLACE_FLAG;
         } else if (blend_mode == JXL_BLEND_BLEND) {
-          blend = (GString *)blend_combine_flag;
+          blend = (GString *)GIMP_BLEND_COMBINE_FLAG;
         } else {
-          blend = (GString *)blend_null_flag;
+          blend = (GString *)GIMP_BLEND_NULL_FLAG;
         }
         char *temp_frame_name = nullptr;
         bool must_free_frame_name = false;
--- a/plugins/gimp/file-jxl-load.cc
+++ b/plugins/gimp/file-jxl-load.cc
@@ -36,6 +36,7 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
   GimpColorProfile *profile_icc = nullptr;
   GimpColorProfile *profile_int = nullptr;
   bool is_linear = false;
+  bool last_frame_blank = false;
   unsigned long xsize = 0, ysize = 0;
   long crop_x0 = 0, crop_y0 = 0;
   size_t layer_idx = 0;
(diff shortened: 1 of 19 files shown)


jonsneyers:
Could you rebase so it's easier to see what's new?

ctrlcctrlv:
https://github.com/libjxl/libjxl/pull/1874/commits/fb8948eda09b50beaaa039dfbb5ee342c14f3f04

mo271:
I took the liberty to make a squashed rebase, so that only the 4 files that have been changed show up in the review here...

ctrlcctrlv:
Thanks, wasn't sure how

mo271:
rebasing once more, because https://github.com/libjxl/libjxl/pull/1903 just landed and that modifies the same files (this one was trivial to rebase, though..)

plugins/gimp/file-jxl-save.cc
```diff
@@ -643,6 +675,52 @@ bool JpegXlSaveOpts::SetPrecision(int gimp_precision) {
...
+    }
+    gchar* found;
+    gsize len = g_ref_string_length(*remainder);
+    found = g_strstr_len(*remainder, len, GIMP_BLEND_REPLACE_FLAG->str);
+    if (found != nullptr) {
+      frame_header.layer_info.blend_info.blendmode = JXL_BLEND_REPLACE;
+    }
+    found = g_strstr_len(*remainder, len, GIMP_BLEND_COMBINE_FLAG->str);
+    if (found != nullptr) {
+      frame_header.layer_info.blend_info.blendmode = JXL_BLEND_ADD;
```
jonsneyers, line 702: I think this should be JXL_BLEND_BLEND. That's the normal alpha blending that should correspond to "combine".

JXL_BLEND_ADD numerically adds the RGBA sample values (including alpha values), which is not what you want to do here.

ctrlcctrlv: thank you, i'm sure you're right

plugins/gimp/file-jxl-load.cc
```diff
@@ -36,6 +36,7 @@ bool LoadJpegXlImage(const gchar *const filename, gint32 *const image_id) {
   GimpColorProfile *profile_icc = nullptr;
   GimpColorProfile *profile_int = nullptr;
   bool is_linear = false;
+  bool last_frame_blank = false;
```
jonsneyers, line 39: What is the purpose of this variable? It seems to be never-read.

plugins/gimp/file-jxl-save.cc
```diff
@@ -643,6 +675,52 @@ bool JpegXlSaveOpts::SetPrecision(int gimp_precision) {
...
+    JxlFrameHeader frame_header;
+    JxlEncoderInitFrameHeader(&frame_header);
+    gchar** remainder = &layer_name;
+    LayerNameToJxlLayerName(layer_name, remainder);
+    if (g_str_has_prefix(*remainder, " (")) {
+      gchar* remainder2;
+      gchar** endptr = &remainder2;
+      gint64 ms_len = g_ascii_strtoll(*remainder + 2, endptr, 10);
+      if (ms_len == 0) {
+        ms_len = 1;
```
jonsneyers, line 690: Why force frames to have a duration of at least 1 ms? JXL does support zero-duration frames and specifies that all consecutive zero-duration frames should be blended before showing a frame. While GIF/WebP do not support this (they have a minimum frame duration for historical reasons, which is longer than 1ms btw), JXL does and it seems useful to allow using this feature.

mo271:
@jonsneyers I think we should move forward on this and even accept a version that can later be improved. Better to have slighly buggy animation support than none...

mo271:
I plan to rebase and resovle the conflicts after https://github.com/libjxl/libjxl/pull/2190 is merged...

PR #1875, opened 2022-11-07
XYB-jpeg: jpeg-specific clustering and more progression.

szabadka:
Benchmark before:
```
Encoding            kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------------
jpeg:libjxl:d1.0      13270  3197015    1.9272988   5.469  29.993   2.19093180   0.70477101  1.358304300765      0
```

Benchmark after:
```
Encoding            kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm        pnorm       BPP*pnorm   Bugs
------------------------------------------------------------------------------------------------------------------
jpeg:libjxl:d1.0      13270  3163678    1.9072018   4.031  25.939   2.19093180   0.70477101  1.344140529099      0
```

### DIFF
--- a/lib/extras/encode_jpeg.cc
+++ b/lib/extras/encode_jpeg.cc
@@ -193,9 +193,34 @@ void AddJpegScanInfos(const std::vector<ProgressiveScan>& scans,
   }
 }
 
+float HistogramCost(const Histogram& histo) {
+  std::vector<uint32_t> counts(jpeg::kJpegHuffmanAlphabetSize + 1);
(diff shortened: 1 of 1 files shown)


jyrkialakuijala (approved):
slowdown is substantial here on both encoding and decoding, it is unclear if this is a good compromise -- let's keep this in mind as an option to reverse later if we need more speed

PR #1890, opened 2022-11-11
For PQ and HLG, make it possible to generate ICC profiles that perform tone mapping

sboukortt:
The tone mapping can be via 1D curves (smaller profile) or a 3D look-up table (more flexible).
    
For PQ, the 1D version will significantly desaturate highlights whereas the 3D version will preserve their color better.
    
For HLG, the 1D version assumes a system gamma of 1 (which, in a reference environment, would be a ~300-cd/m² monitor), which is likely to result in an image that looks quite dark on many SDR monitors or viewing environments. The 3D version is able to perform the rendering corresponding to a lower gamma, thereby lifting the shadows and mid-tones, without distorting the colors as altering the 1D curves would.

### DIFF
--- a/lib/jxl/color_management.cc
+++ b/lib/jxl/color_management.cc
@@ -585,13 +585,11 @@ Status MaybeCreateProfile(const ColorEncoding& c,
       switch (c.tf.GetTransferFunction()) {
         case TransferFunction::kHLG:
           CreateICCCurvCurvTag(
-              HWY_DYNAMIC_DISPATCH(CreateTableCurve)(4096, ExtraTF::kHLG),
-              &tags);
+              HWY_DYNAMIC_DISPATCH(CreateTableCurve)(64, ExtraTF::kHLG), &tags);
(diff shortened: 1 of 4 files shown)


lib/jxl/color_management.cc
```diff
@@ -25,21 +25,105 @@
...
+#endif
+
 HWY_BEFORE_NAMESPACE();
 namespace jxl {
 namespace HWY_NAMESPACE {
 
-// NOTE: this is only used to provide a reasonable ICC profile that other
-// software can read. Our own transforms use ExtraTF instead because that is
-// more precise and supports unbounded mode.
+#if JXL_ENABLE_3D_ICC_TONEMAPPING
```
veluca93, line 42: Why are these functions #ifdef-ed out?

sboukortt: Not the case anymore.

veluca93 (approved):

PR #1900, opened 2022-11-16
Update dec_xyb.cc to fix wasm decode issue

jimbo2150:
Changed OR to AND. In WASM, images that were not XYB encoded would fail to decode, as `CanOutputToColorEncoding(c_desired)` is never called.


jimbo2150:
Meat to fix https://github.com/libjxl/libjxl/issues/1902

mo271 (dismissed):

mo271 (approved):

PR #1904, opened 2022-11-18
Add demo site generator

eustas:


jonsneyers (approved):

tools/wasm_demo/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set_target_properties(jxl_decoder PROPERTIES LINK_FLAGS "\
   -s DISABLE_EXCEPTION_CATCHING=1 \
   -s MODULARIZE=1 \
   -s FILESYSTEM=0 \
+  -s NODERAWFS=1 \
```
kleisauke, line 29: Was this setting specified on purpose? AFAIK, this should not be necessary when building with `-sFILESYSTEM=0`.

When filesystem support is enabled, it also causes the binary to be usable only by Node.js, see:
https://github.com/emscripten-core/emscripten/blob/2e02d0eba92c9c7489bed38eff9bd1ee61e93229/src/library_noderawfs.js#L28

eustas: Was trying to fix test. Seems it did not help. Though it runs fine locally. Going to roll-back this and look for another possible solution.
Thanks for the heads-up.

PR #1913, opened 2022-11-22
Support 16 bit output buffer in libjpegli decoder.

szabadka:
This can be enabled by the application by setting

```
  cinfo.quantize_colors = FALSE;
  cinfo.desired_number_of_colors = 65536;
```

In this case a scanline buffer should be twice as long.

Other changes in the commit:

  * Remove code duplication between input reading API functions.

  * Use the existing cinfo fields for marker parsing and keeping internal state, remove redundant fields from ```jpeg_decomp_master```.

  * Add trivial implementation of ```jpeg_calc_output_dimensions``` and use ```cinfo->output_{width,height}```, 
     and ```cinfo->out_color_components``` where appropiate.

### DIFF
--- a/lib/jpegli/decode_api.cc
+++ b/lib/jpegli/decode_api.cc
@@ -20,7 +20,111 @@
 #include "lib/jpegli/source_manager.h"
 #include "lib/jxl/base/status.h"
 
+#define JPEGLI_STATE_READ_HEADER 1
+#define JPEGLI_STATE_START_DECOMPRESS 2
+#define JPEGLI_STATE_READ_SCANLINES 3
+
 typedef jpeg_decomp_master::State State;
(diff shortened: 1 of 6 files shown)


jyrkialakuijala (approved):
JPEGs overshoot, i.e., quantization makes some pixel values to be negative and some above the maximum. If we had 14 bit dynamics in 16 bit signed integers, then we would have some reserve in both ends -- and this could possibly slightly reduce generation loss in these conditions.

For now full 16 bit is ok, but let's keep an open mind if we want to switch to 14 bit dynamics [0..16383] in a 16 bit signed int [-32768..32767] -- I'm not sure what is the best way forward, just want to spread awareness of this idea

PR #1918, opened 2022-11-23
Add 16 bit jpeg output mode to benchmark_xl.

szabadka:


tools/cjxl_main.cc
```diff
@@ -584,7 +584,7 @@ jxl::Status GetPixeldata(const std::vector<uint8_t>& image_data,
 #endif
 #if JPEGXL_ENABLE_JPEG
     if (jxl::extras::DecodeImageJPG(encoded, color_hints, size_constraints,
-                                    &ppf)) {
+                                    /*output_bit_depth=*/8, &ppf)) {
```
jonsneyers, line 587: Why not 16 here?

szabadka: Right, it can be 16 now. I just forgot that I made it work with the old library in the meantime.

jonsneyers (approved):

jyrkialakuijala:
could you post the results of the benchmark in the reviewlog -- people can have interest on how fast it is

jyrkialakuijala (commented):
nice.

szabadka:
> could you post the results of the benchmark in the reviewlog -- people can have interest on how fast it is

I edited the PR description to include benchmark results. There is about a 4% speed difference between 8-bit and 16-bit output.

PR #1919, opened 2022-11-23
remove deprecated or dead code

jonsneyers:
Remove the deprecated DCOutBuffer stuff from the decode API — this was deprecated a long time ago and didn't work anymore anyway, so no need to keep it in.

Also remove some code that is no longer used. Move the speed tier name parsing to benchmark_xl (which is the only thing that still uses that). The function to convert speed settings to animal name strings was only used to produce pretty names for the tests in `speed_tier_test.cc`, so removing that too. Some overloads of `UnpremultiplyAlpha` and `ConvertToExternal` are not used by anything anymore, so removing them too.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -168,16 +168,6 @@ typedef enum {
    */
   JXL_DEC_NEED_PREVIEW_OUT_BUFFER = 3,
 
-  /** The decoder is able to decode a DC image and requests setting a DC output
-   * buffer using @ref JxlDecoderSetDCOutBuffer. This occurs if @ref
-   * JXL_DEC_DC_IMAGE is requested and it is possible to decode a DC image from
-   * the codestream and the DC out buffer was not yet set. This event re-occurs
-   * for new frames if there are multiple animation frames.
-   * @deprecated The DC feature in this form will be removed. For progressive
-   * rendering, @ref JxlDecoderFlushImage should be used.
-   */
-  JXL_DEC_NEED_DC_OUT_BUFFER = 4,
-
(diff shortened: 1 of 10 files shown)


lib/include/jxl/decode.h
```diff
@@ -263,15 +253,10 @@ typedef enum {
...
-   * indicates we're past this point in the codestream. This event occurs max
-   * once per frame and always later than @ref JXL_DEC_FRAME and other header
-   * events and earlier than full resolution pixel data.
-   *
-   * @deprecated The DC feature in this form will be removed. For progressive
-   * rendering, @ref JxlDecoderFlushImage should be used.
+   * it does it will do so before outputting the full frame. This return status
+   * only indicates we're past this point in the codestream. This event occurs
+   * max once per frame and always later than @ref JXL_DEC_FRAME and other
+   * header events and earlier than full resolution pixel data.
```
szabadka, line 259: We should say here that we never return this event any more.

jonsneyers: Maybe better to just remove this from the enum. If an application is still subscribing to this event, it's probably better to give them a compile error so they can update their code to do progressive rendering correctly, than it is to keep things 'working' but never actually produce the event...

jonsneyers:
Updated: removed JXL_DEC_DC_IMAGE completely (it already didn't do anything for a long time, but now applications that still subscribe to it will get a compile error).

lib/include/jxl/decode.h
szabadka: Yes, I agree, it is better to remove it, since it is marked as deprecated anyway.

szabadka (approved):
Thanks for the cleanup!

PR #1924, opened 2022-11-28
cjpeg_hdr: reuse EncodeJpeg

sboukortt:


lib/extras/encode_jpeg.cc
```diff
@@ -75,7 +76,9 @@ void ComputeDCTCoefficients(const Image3F& opsin, const ImageF& qf,
             block[ix * 8 + iy] = cc;
           }
         }
-        block[0] = std::round((2040 * dct[0] - 1024) * qmc[0]);
+        if (xyb) {
+          block[0] = std::round((2040 * dct[0] - 1024) * qmc[0]);
```
szabadka, line 80: please add a comment explaining the different behavior, something like "ToXYB does not create zero-centered sample values like RgbToYcbcr does, so we apply an offset to the DC values instead"

sboukortt: Thanks, done.

tools/cjpeg_hdr.cc
```diff
@@ -243,53 +27,11 @@ int HBDJPEGMain(int argc, const char* argv[]) {
...
-  HWY_DYNAMIC_DISPATCH(FillJPEGData)
-  (ycbcr, io.metadata.m.color_encoding.ICC(), qf, frame_dim,
-   out.Main().jpeg_data.get());
-  jxl::PaddedBytes output;
-  if (!jxl::jpeg::EncodeImageJPGCoefficients(&out, &output)) {
-    return 1;
-  }
+  jxl::extras::JpegSettings settings;
+  settings.xyb = false;
+  settings.distance = 0.0f;
```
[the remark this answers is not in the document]
sboukortt: Empirically, it seems fine with 0.

szabadka (dismissed):

veluca93 (approved):

PR #1927, opened 2022-11-29
premultiply alpha in BufferToImageBundle

### LINKED ISSUE #1869: Encoding with premultiplied alpha is not supported
**Describe the bug**

`Jxl::ImageBundle::SetAlpha` calls `Jxl::Abort` on the following line when encoding with premultiplied alpha.
https://github.com/libjxl/libjxl/blob/a0204edcaf9288b62a0fd1a708c97456ec717b02/lib/jxl/image_bundle.cc#L107

The issue appears to be caused by `BufferToImageBundle`  setting `alpha_is_premultiplied` to `false` instead of using the value from the alpha channel info.

 https://github.com/libjxl/libjxl/blob/a0204edcaf9288b62a0fd1a708c97456ec717b02/lib/jxl/enc_external_image.cc#L271

**To Reproduce**

Set the alpha channel to premultiplied using `JxlEncoderSetExtraChannelInfo`.
Call `JxlEncoderAddImageFrame` with the image data.

**Expected behavior**

The image encodes correctly.

**Screenshots**
N/A

**Environment**
 - OS: Windows
 - Compiler version: MSVC 19.33.31630
 - CPU type:  x86_64
 - cjxl/djxl version string: N/A, I am using libjxl version 0.7.0 directly.

**Additional context**

The stack trace that Visual Studio shows when hitting this issue:

```
jxl::Abort() Line 136	C++
jxl::ImageBundle::SetAlpha(jxl::Plane<float> && alpha, bool alpha_is_premultiplied) Line 107	C++
jxl::ConvertFromExternal(jxl::Span<unsigned char const> bytes, unsigned __int64 xsize, unsigned __int64 ysize, const jxl::ColorEncoding & c_current, unsigned __int64 channels, bool alpha_is_premultiplied, unsigned __int64 bits_per_sample, JxlEndianness endianness, jxl::ThreadPool * pool, jxl::ImageBundle * ib, bool float_in, unsigned __int64 align) Line 379	C++
jxl::BufferToImageBundle(const JxlPixelFormat & pixel_format, unsigned int xsize, unsigned int ysize, const void * buffer, unsigned __int64 size, jxl::ThreadPool * pool, const jxl::ColorEncoding & c_current, jxl::ImageBundle * ib) Line 418	C++
JxlEncoderAddImageFrame(const JxlEncoderFrameSettingsStruct * frame_settings, const JxlPixelFormat * pixel_format, const void * buffer, unsigned __int64 size) Line 1626	C++
```

mo271:
fixes #1869

### DIFF
--- a/lib/jxl/enc_external_image.cc
+++ b/lib/jxl/enc_external_image.cc
@@ -208,9 +208,8 @@ Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
 }
 Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
                            size_t ysize, const ColorEncoding& c_current,
-                           bool alpha_is_premultiplied, size_t bits_per_sample,
-                           JxlPixelFormat format, ThreadPool* pool,
-                           ImageBundle* ib) {
+                           size_t bits_per_sample, JxlPixelFormat format,
+                           ThreadPool* pool, ImageBundle* ib) {
   const size_t color_channels = c_current.Channels();
   bool has_alpha = format.num_channels == 2 || format.num_channels == 4;
   if (format.num_channels < color_channels) {
@@ -238,13 +237,13 @@ Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
     JXL_RETURN_IF_ERROR(
         ConvertFromExternal(bytes, xsize, ysize, bits_per_sample, format,
                             format.num_channels - 1, pool, &alpha));
-    ib->SetAlpha(std::move(alpha), alpha_is_premultiplied);
+    ib->SetAlpha(std::move(alpha), ib->AlphaIsPremultiplied());
   } else if (!has_alpha && ib->HasAlpha()) {
     // if alpha is not passed, but it is expected, then assume
     // it is all-opaque
     ImageF alpha(xsize, ysize);
     FillImage(1.0f, &alpha);
-    ib->SetAlpha(std::move(alpha), alpha_is_premultiplied);
+    ib->SetAlpha(std::move(alpha), ib->AlphaIsPremultiplied());
   }
 
   return true;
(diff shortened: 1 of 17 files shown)


jonsneyers (dismissed):

lib/jxl/enc_external_image.cc
```diff
@@ -268,7 +268,8 @@ Status BufferToImageBundle(const JxlPixelFormat& pixel_format, uint32_t xsize,
   JXL_RETURN_IF_ERROR(ConvertFromExternal(
       jxl::Span<const uint8_t>(static_cast<const uint8_t*>(buffer), size),
       xsize, ysize, c_current,
-      /*alpha_is_premultiplied=*/false, bitdepth, pixel_format, pool, ib));
+      /*alpha_is_premultiplied=*/ib->AlphaIsPremultiplied(), bitdepth,
```
szabadka, line 271: we could read out ib->AlphaIsPremultiplied inside the function, since we are passing ib as parameter anyway

mo271: I tried to remove the argument `alpha_is_premultiplied` from the function `ConvertFromExternal`, but here it seems that it is needed:

https://github.com/libjxl/libjxl/blob/8632a76da0884ca685b71cdb76756bdef4417f60/lib/extras/packed_image_convert.cc#L59-L62

What do you think?

szabadka: I don't think that it is really needed there either, since bundle->AlphaIsPremultiplied() should return the same as info.alpha_premultiplied.

mo271: Please check that this makes sense...

lib/jxl/enc_frame.cc
```diff
@@ -895,6 +895,8 @@ class LossyFrameEncoder {
     for (PassesEncoderState::PassData& pass : enc_state_->passes) {
       pass.ac_tokens.resize(shared.frame_dim.num_groups);
     }
+    fprintf(stderr, "enc_stat_->passes.size(): %zu\n",
```
szabadka, line 898: should be removed

lib/jxl/enc_external_image.cc
```diff
@@ -238,13 +237,13 @@ Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
...
                             format.num_channels - 1, pool, &alpha));
-    ib->SetAlpha(std::move(alpha), alpha_is_premultiplied);
+    ib->SetAlpha(std::move(alpha), ib->AlphaIsPremultiplied());
   } else if (!has_alpha && ib->HasAlpha()) {
     // if alpha is not passed, but it is expected, then assume
     // it is all-opaque
     ImageF alpha(xsize, ysize);
     FillImage(1.0f, &alpha);
-    ib->SetAlpha(std::move(alpha), alpha_is_premultiplied);
+    ib->SetAlpha(std::move(alpha), ib->AlphaIsPremultiplied());
```
szabadka, line 246: now ImageBundle::SetAlpha could be simplified to not take the is_premultiplied argument

lib/jxl/enc_frame.cc
mo271: Done

lib/jxl/enc_external_image.cc
mo271: Makes sense, done.

szabadka (approved):

PR #1938, opened 2022-12-01
Refactor fast_lossless to support 16-bit data.

veluca93:
>8 bit paths are not SIMDfied yet.


veluca93:
> Can't spot a problem by eyeballing the code. Would be good to add some test scripts or code to see if it works on all bit depths.

I tested this manually on noise at a bunch of bit depths (4 6 8 10 11 12 13 14 16) and a 8 and 16 bit nonphotographic image. I'll leave it to @sboukortt to do more in depth tests :)

PR #1946, opened 2022-12-05
Add an --update_on_failure flag to the conformance test runner.

szabadka:
Also update the conformance test set commit to make the tests pass.


mo271 (approved):
very convenient!

PR #1957, opened 2022-12-09
Do not calculate Adler32 (->faster)

eustas:
Try to employ WASM compiled code caching; it works sometimes.

### DIFF
--- a/tools/wasm_demo/no_png.cc
+++ b/tools/wasm_demo/no_png.cc
@@ -63,6 +63,22 @@ void AdlerCopy(const uint8_t* src, uint8_t* dst, size_t length, uint32_t* s1,
 constexpr size_t kMaxDeflateBlock = 65535;
 constexpr uint32_t kIhdrSize = 13;
 
+void WriteU8(uint8_t*& dst, uint8_t value) { *(dst++) = value; }
+
+void WriteU16(uint8_t*& dst, uint16_t value) {
+  memcpy(dst, &value, 2);
+  dst += 2;
+}
+
+void WriteU32(uint8_t*& dst, uint32_t value) {
+  memcpy(dst, &value, 4);
+  dst += 4;
+}
+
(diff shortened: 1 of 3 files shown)


tools/wasm_demo/no_png.cc
```diff
@@ -148,7 +148,8 @@ uint8_t* WrapPixelsToPng(size_t width, size_t height, size_t bit_depth,
       bytes_to_next_row -= bytes_to_copy;
     }
   }
-  uint32_t adler = (adler_s2 << 8) | adler_s1;
+  // Fake Adler works well in Chrome; so let's not waste CPU cycles.
+  uint32_t adler = 0; // (adler_s2 << 8) | adler_s1;
   *reinterpret_cast<uint32_t*>(dst) = __builtin_bswap32(adler);
```
veluca93, line 153: this is undefined behaviour, we should probably memcpy instead.

tools/wasm_demo/service_worker.js
```diff
@@ -182,6 +184,8 @@
   };
 
   const serviceWorkerMain = () => {
+    leak = WebAssembly.compileStreaming(fetch("jxl_decoder.wasm"));
```
jimbo2150, line 187: Why not build with streaming compile through emscripten rather than trying to add it as a side-load? I suspect it only works some of the time since it's likely racing emscripten's own wasm loader.
https://emsettings.surma.technology/#MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION

eustas: As you can see, variable is named `leak` - we stream-compile decoder in the context of service-worker and never use; that is done only to enable compiled WASM code caching.

tools/wasm_demo/no_png.cc
eustas: Fixed. PTAL

veluca93 (approved):

PR #1958, opened 2022-12-11
Fix support for reading/writing from/to stdin/stdout in cjxl

dalbani:
There has been work in the past to support reading/writing from/to stdin/stdout for the various programs in libjxl.  
In particular:
- https://github.com/libjxl/libjxl/pull/554
- https://github.com/libjxl/libjxl/pull/594

It's not clear to me what happened, but if I try with libjxl v0.7.0:
```
$ cat file.jpg | cjxl - - > out.jxl               
JPEG XL encoder v0.7.0 0.7.0 [AVX2,SSE4,SSSE3]
Reading image data failed.
```

Whereas it now behaves as such with this PR:
```
$ cat file.jpg | ../build/tools/cjxl - - > out.jxl
JPEG XL encoder v0.8.0 995d56f5 [AVX2,SSE4,SSSE3]
Note: Implicit-default for JPEG is lossless-transcoding. To silence this message, set --lossless_jpeg=(1|0).
Read JPEG image with 65275 bytes.
Compressed to 41397 bytes including container
```

I tried applying the same fix for `djxl`  as well, but I didn't know how to solve the issue related to the choice of encoder using the extension:
```
$ cat file.jxl | ../build/tools/djxl - - > file.jpg                       
JPEG XL decoder v0.8.0 995d56f5 [AVX2,SSE4,SSSE3]
Read 41397 compressed bytes.
can't decode to the file extension ''
```

I suppose other programs shipped with libjxl would need the same treatment, as in to rely on `ReadFile` and `WriteFile` from `lib/jxl/base/file_io.h`, instead of the functions in `tools/file_io.{c,h}` (why, by the way??). Or, worse, direct use of `fopen()` in `jxlinfo` for example.

But for now, given that I'm not familiar with the code base, here's my first proposal to at least fix `cjxl` 😄

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -37,6 +37,7 @@
 #include "lib/extras/dec/pnm.h"
 #include "lib/extras/enc/jxl.h"
 #include "lib/extras/time.h"
+#include "lib/jxl/base/file_io.h"
 #include "lib/jxl/base/override.h"
 #include "lib/jxl/base/printf_macros.h"
 #include "lib/jxl/base/status.h"
@@ -45,7 +46,6 @@
 #include "tools/args.h"
 #include "tools/cmdline.h"
 #include "tools/codec_config.h"
-#include "tools/file_io.h"
 #include "tools/speed_stats.h"
 
 namespace jpegxl {
@@ -942,7 +942,7 @@ int main(int argc, char** argv) {
   jxl::extras::Codec codec = jxl::extras::Codec::kUnknown;
   double decode_mps = 0;
   size_t pixels = 0;
-  if (!jpegxl::tools::ReadFile(args.file_in, &image_data)) {
+  if (!jxl::ReadFile(args.file_in, &image_data)) {
     std::cerr << "Reading image data failed." << std::endl;
     exit(EXIT_FAILURE);
   }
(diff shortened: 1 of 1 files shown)


dalbani:
Can I do something to get this issue move forward?
Is there a consensus on what should be the fix?

jonsneyers:
So the suggested alternative fix is to move the code that deals with stdin/stdout from `lib/jxl/base/file_io.*` to `tools/file_io.*`. I think the idea is that ideally cjxl/djxl shouldn't be using any code from `lib/jxl/` (I know we're not exactly there yet, but let's not move in the wrong direction).

For djxl let's find a solution in a separate pull request, but I would suggest to add a syntax in ImageMagick style to specify the format explicitly, e.g. `png:filename` would produce a file called `filename` in the png format, without requiring the matching filename extension. Then for decoding to stdout, just `-` could default to something (say png) but you could also do `jpg:-` or `ppm:-`.
Another option is to add flags for every output format; that's how djpeg and dwebp do it:  `--pnm`, `--bmp`, etc. I don't particularly like that approach because it means adding one flag per format, so we'd currently need --ppm, --pfm, --pgm, --pam, --png, --jpeg, --exr and --numpy, and maybe in the future more output formats could be added. I prefer a generic approach where `[fmt:]filename.ext` sets the format to `fmt` and if it's not given sets it to `ext`.

jonsneyers:
@dalbani are you still around to update this PR?

jonsneyers:
Closing, see #2560

PR #1975, opened 2022-12-14
Use fast-lossless for effort 1.

veluca93:
Before:
```
Compressed to 2148068 bytes (2.731 bpp).
3072 x 2048, geomean: 23.99 MP/s [21.01, 24.47], 100 reps, 0 threads.
```

After:
```
Compressed to 2166489 bytes (2.755 bpp).
3072 x 2048, geomean: 375.76 MP/s [175.02, 377.97], 100 reps, 0 threads.
```

### DIFF
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -88,8 +88,7 @@ struct JxlFastLosslessFrameState {
   uint64_t bit_buffer = 0;
 };
 
-size_t JxlFastLosslessMaxRequiredOutput(
-    const JxlFastLosslessFrameState* frame) {
+size_t JxlFastLosslessOutputSize(const JxlFastLosslessFrameState* frame) {
   size_t total_size_groups = 0;
   for (size_t i = 0; i < frame->group_data.size(); i++) {
     size_t sz = 0;
@@ -100,7 +99,12 @@ size_t JxlFastLosslessMaxRequiredOutput(
     sz = (sz + 7) / 8;
     total_size_groups += sz;
   }
-  return frame->header.bytes_written + total_size_groups + 32;
+  return frame->header.bytes_written + total_size_groups;
+}
+
+size_t JxlFastLosslessMaxRequiredOutput(
+    const JxlFastLosslessFrameState* frame) {
+  return JxlFastLosslessOutputSize(frame) + 32;
 }
 
 void JxlFastLosslessPrepareHeader(JxlFastLosslessFrameState* frame,
@@ -741,11 +745,11 @@ void EncodeChunkAVX2(const uint16_t* residuals, const PrefixCode& code,
     for (size_t i = 0; i < 4; i++) {
       output.buffer |= bits_simd[i] << output.bits_in_buffer;
       memcpy(output.data.get() + output.bytes_written, &output.buffer, 8);
-      // If >> 64, next_buffer is unused.
-      uint64_t next_buffer = bits_simd[i] >> (64 - output.bits_in_buffer);
+      uint64_t shift = 64 - output.bits_in_buffer;
       output.bits_in_buffer += nbits_simd[i];
       // This `if` seems to be faster than using ternaries.
       if (output.bits_in_buffer >= 64) {
+        uint64_t next_buffer = bits_simd[i] >> shift;
         output.buffer = next_buffer;
         output.bits_in_buffer -= 64;
         output.bytes_written += 8;
@@ -948,7 +952,7 @@ FJXL_INLINE void StoreNeonAbove14(const uint32_t* nbits_tok,
 
 void EncodeHybridUint000(uint32_t value, uint32_t* token, uint32_t* nbits,
                          uint32_t* bits) {
-  uint32_t n = 31 - __builtin_clz(value);
+  uint32_t n = value ? 31 - __builtin_clz(value) : 32;
   *token = value ? n + 1 : 0;
   *nbits = value ? n : 0;
   *bits = value ? value - (1 << n) : 0;
@@ -1275,7 +1279,7 @@ void PrepareDCGlobal(bool is_single_group, size_t width, size_t height,
 void EncodeHybridUint404_Mul16(uint32_t value, uint32_t* token_div16,
                                uint32_t* nbits, uint32_t* bits) {
   // NOTE: token in libjxl is actually << 4.
-  uint32_t n = 31 - __builtin_clz(value);
+  uint32_t n = value ? 31 - __builtin_clz(value) : 32;
   *token_div16 = value < 16 ? 0 : n - 3;
   *nbits = value < 16 ? 0 : n - 4;
   *bits = value < 16 ? 0 : (value >> 4) - (1 << *nbits);
@@ -1398,8 +1402,8 @@ void ProcessImageArea(const unsigned char* rgba, size_t x0, size_t y0,
   using pixel_t = typename BitDepth::pixel_t;
   using upixel_t = typename BitDepth::upixel_t;
 
-  // Could use nb_chans, but clang-tidy complains otherwise.
-  pixel_t group_data[4][2][256 + kPadding * 2] = {};
+  std::vector<std::array<std::array<pixel_t, 256 + kPadding * 2>, 2>>
+      group_data(nb_chans);
   upixel_t allzero[4] = {};
   upixel_t allone[4];
   auto get_pixel = [&](size_t x, size_t y, size_t channel) {
@@ -1561,7 +1565,7 @@ void ProcessImageAreaPalette(const unsigned char* rgba, size_t x0, size_t y0,
                              Processor* processors) {
   constexpr size_t kPadding = 16;
 
-  int16_t group_data[2][256 + kPadding * 2] = {};
+  std::vector<std::array<int16_t, 256 + kPadding * 2>> group_data(2);
   Processor& row_encoder = processors[0];
 
   for (size_t y = 0; y < ys; y++) {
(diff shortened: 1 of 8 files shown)


jonsneyers:
Probably the largest speedup I ever saw in a single libjxl pull request :)

lib/jxl/enc_fast_lossless.h
```diff
@@ -50,6 +50,11 @@ void JxlFastLosslessPrepareHeader(JxlFastLosslessFrameState* frame,
 // JxlFastLosslessPrepareHeader.
 size_t JxlFastLosslessMaxRequiredOutput(const JxlFastLosslessFrameState* frame);
 
+// Actual size of the frame once it is encoded. This is not identical to
+// JxlFastLosslessOutputSize because JxlFastLosslessWriteOutput may require
```
szabadka, line 54: I think you meant "not identical to JxlFastLosslessMaxRequiredOutput"

lib/jxl/encode.cc
```diff
@@ -9,8 +9,12 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <cstdint>
 #include <cstring>
 
+#include "base/data_parallel.h"
+#include "enc_fast_lossless.h"
```
szabadka, line 16: this should be lib/jxl/enc_fast_lossless.h and lib/jxl/enc_params.h

lib/jxl/encode.cc
```diff
@@ -438,122 +452,151 @@ JxlEncoderStatus JxlEncoderStruct::RefillOutputByteQueue() {
 
   // Choose frame or box processing: exactly one of the two unique pointers (box
   // or frame) in the input queue item is non-null.
-  if (input.frame) {
+  if (input.frame || input.fast_lossless_frame) {
     jxl::MemoryManagerUniquePtr<jxl::JxlEncoderQueuedFrame> input_frame =
         std::move(input.frame);
+    if (input.fast_lossless_frame) {
+      output_fast_frame_queue.push_back(std::move(input.fast_lossless_frame));
+    } else {
```
szabadka, line 460: remove?

lib/jxl/encode.cc
```diff
@@ -1617,6 +1710,52 @@ JxlEncoderStatus JxlEncoderAddImageFrame(
...
+      return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_API_USAGE,
+                           "provided image buffer too small");
+    }
+    const bool big_endian =
+        pixel_format->endianness == JXL_BIG_ENDIAN ||
+        (pixel_format->endianness == JXL_NATIVE_ENDIAN && !IsLittleEndian());
+
+    auto runner = +[](void* void_pool, void* opaque, void fun(void*, size_t),
+                      size_t count) {
+      auto* pool = (jxl::ThreadPool*)void_pool;
```
szabadka, line 1743: this should be reinterpret_cast

lib/jxl/enc_fast_lossless.cc
```diff
@@ -948,7 +952,7 @@ FJXL_INLINE void StoreNeonAbove14(const uint32_t* nbits_tok,
 
 void EncodeHybridUint000(uint32_t value, uint32_t* token, uint32_t* nbits,
                          uint32_t* bits) {
-  uint32_t n = 31 - __builtin_clz(value);
+  uint32_t n = value ? 31 - __builtin_clz(value) : 32;
```
jonsneyers, line 955: I think there's a function that does exactly this in base/bits.h, might be nicer to use that one?

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
 }
 
+static bool CanDoFastLossless(const JxlEncoderFrameSettings* frame_settings,
+                              const JxlPixelFormat* pixel_format,
+                              bool has_alpha) {
+  if (!frame_settings->values.lossless) {
+    return false;
+  }
+  if (frame_settings->values.frame_index_box) {
+    return false;
```
jonsneyers, line 1644: Maybe put a TODO here because I don't think there's an inherent reason why this combination cannot work.

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  }
+  if (frame_settings->values.cparams.speed_tier != jxl::SpeedTier::kLightning) {
+    return false;
```
jonsneyers, line 1647: Not for this PR, but I would eventually use fast lossless at both e1 and e2 (with different fast effort settings, e.g. no palette check and cheaper histogram sampling at e1), and then make e3 do something that also tries the fast path since it can very well give better results for non-photo with big run lengths.

I'm not sure if we still need the current e2 except as a fallback option for cases not covered by the fast path.

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  }
+  if (frame_settings->values.image_bit_depth.type ==
+          JxlBitDepthType::JXL_BIT_DEPTH_CUSTOM &&
+      frame_settings->values.image_bit_depth.bits_per_sample !=
+          frame_settings->enc->metadata.m.bit_depth.bits_per_sample) {
+    return false;
```
jonsneyers, line 1653: Maybe this could also get a TODO — as long as it's <= 16-bit, we could still implement this in principle.

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  }
+  if (!frame_settings->values.frame_name.empty()) {
+    return false;
```
jonsneyers, line 1656: I don't understand why the frame name needs to be empty. At least add a TODO here, because I think the main use case for fast lossless encoding is in authoring workflows (gimp, krita, photoshop etc) where you may very well have named layers. Maybe just make it work already :)

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  }
+  if (!(has_alpha && frame_settings->enc->metadata.m.num_extra_channels == 1) &&
+      frame_settings->enc->metadata.m.num_extra_channels != 0) {
+    return false;
```
jonsneyers, line 1660: This condition is hard to understand, a comment would be welcome. I suppose you want the input buffer to have alpha iff the frame should be encoded with alpha, or something like that? It's not very clear to me what fjxl precondition this condition addresses and how it does that.

lib/jxl/encode.cc
```diff
@@ -1588,6 +1634,53 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
...
+  }
+  if (frame_settings->enc->metadata.m.bit_depth.floating_point_sample ||
+      frame_settings->enc->metadata.m.bit_depth.bits_per_sample > 16) {
+    return false;
```
jonsneyers, line 1664: Perhaps the fast path can also be used for the lossless float case (for storing half-floats). It's not like the current e1 will be doing anything better...

lib/jxl/encode.cc
```diff
@@ -1617,6 +1710,52 @@ JxlEncoderStatus JxlEncoderAddImageFrame(
...
+          pool, 0, count, jxl::ThreadPool::NoInit,
+          [&](size_t i, size_t) { fun(opaque, i); }, "Encode fast lossless"));
+    };
+    QueueFastLosslessFrame(
+        frame_settings,
+        JxlFastLosslessPrepareFrame(
+            (const unsigned char*)buffer, xsize, row_size, ysize,
+            pixel_format->num_channels,
+            frame_settings->enc->metadata.m.bit_depth.bits_per_sample,
+            big_endian, /*effort=*/2, frame_settings->enc->thread_pool.get(),
```
jonsneyers, line 1754: OK for now. Eventually I'd go for something like effort=1 at libjxl e1 and effort=8 or so at libjxl e2.

lib/jxl/enc_fast_lossless.h
veluca93: haha, yep!

lib/jxl/enc_fast_lossless.cc
veluca93: There is such a function, but I'd like enc_fast_lossless to be usable in a drop-in way, so I'd prefer to avoid using it.

lib/jxl/encode.cc
veluca93: I put a general TODO in the function; it should be relatively easy to add support for a few of those, follow-up PRs are welcome ;)

lib/jxl/enc_fast_lossless.cc
jonsneyers: Fair enough.

szabadka (approved):

alantudyk:
@veluca93,

I see linear scalability of decoding (per core, not per thread, 8C/16T):
```
14557 x 8418, geomean:  33.99 MP/s [ 33.54,  34.49], 100 reps,  1 threads.
14557 x 8418, geomean: 234.08 MP/s [221.79, 273.27], 100 reps, 16 threads.
```
But encoding is limited to 4x:
```
14557 x 8418, geomean:  312.91 MP/s [237.69,  319.53], 100 reps,  1 threads.
14557 x 8418, geomean: 1197.58 MP/s [512.70, 1222.38], 100 reps, 16 threads.
```
Is RAM a bottleneck? (2x2R DDR4-3200)

veluca93:
That's certainly possible - consider that 1200MP/s is already ~3.6GB/s of memory transfer just for reading the input, and while internal processing likely uses mostly the cache, it still does about 4x that amount of memory transfers.

Also, some parts of encoding are not parallelized (yet?) - I never spent significant effort optimizing fjxl for multiple threads, I figured it is already fast enough :) unless you have a use case in mind for which it would be beneficial to further increase multithreaded speed, I don't think I'll try to improve this in the near future :)

veluca93:
That's certainly possible - consider that 1200MP/s is already ~3.6GB/s of memory transfer just for reading the input, and while internal processing likely uses mostly the cache, it still does about 4x that amount of memory transfers.

Also, some parts of encoding are not parallelized (yet?) - I never spent significant effort optimizing fjxl for multiple threads, I figured it is already fast enough :) unless you have a use case in mind for which it would be beneficial to further increase multithreaded speed, I don't think I'll try to improve this in the near future :)

PR #1950, opened 2022-12-06
Build release package for Ubuntu 22.04

dalbani:
This PR adds Ubuntu 22.04 to the existing list of Ubuntu 18.04 and 20.04 for which a package is generated for releases.

PR #1985, opened 2022-12-17
Add Ubuntu 22.04 based tasks to Build/Test workflow

dalbani:
I'm not absolutely sure I understand the CI setup correctly, but I suppose the Build/Test workflow should now include tasks for `gcc-10` and `clang-14`, given that packages for Ubuntu 22.04 are to be included in releases as per https://github.com/libjxl/libjxl/pull/1950.


mo271 (approved):

eustas (approved):

dalbani:
I've just rebased the code on `main`.

eustas (approved):

eustas:
To land this we need to:
 - fix move-in-return-statement
 - remove "gcc8" required workflow and add "gcc-8" after PR is landed

PR #1992, opened 2022-12-21
relax rendering intent check

### LINKED ISSUE #1973: JxlEncoderAddJPEGFrame() failed
**Describe the bug**
I'm transcoding a large (~1.3M) image dataset of JPEGs to JXL using the lossless transcoding. For 864 images I get the following error:

```
JPEG XL encoder v0.8.0 091b60e5 [AVX3,AVX2,SSE4,SSSE3,Unknown]                                                                                                                                                                   
Read JPEG image with 43279 bytes.
JxlEncoderAddJPEGFrame() failed.
EncodeImageJXL() failed.
```

running in verbose mode does not give any extra information. 

I've attached an example inside a zip file (in case github tries to mess with the image)

[n03476991_52119.zip](https://github.com/libjxl/libjxl/files/10230108/n03476991_52119.zip)

**To Reproduce**

```
cjxl --lossless_jpeg=1 n03476991_52119.JPEG ./test.jxl 
```
should reproduce the failure

**Expected behavior**
A JXL trancoding is produced

**Environment**
 - OS: Linux
 - Compiler version: n/a
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.8.0 091b60e5 [AVX3,AVX2,SSE4,SSSE3,Unknown]

mo271:
Closes #1973.
Because there appear to be images with faulty rendering intent in their ICC profiles in the wild, we relax the check for valid rendering intent to a warning if we are able to deduce a plausible value for the rendering for the last byte of the rendering intent of the ICC profiles, i.e. the byte at offset 67, even if the other bytes (at offsets 64, 65 and 66) are non-zero. In that case we print a warning.

### DIFF
--- a/lib/jxl/cms/jxl_cms.cc
+++ b/lib/jxl/cms/jxl_cms.cc
@@ -971,14 +971,31 @@ JXL_BOOL JxlCmsSetFieldsFromICC(void* user_data, const uint8_t* icc_data,
   JXL_RETURN_IF_ERROR(skcms_Parse(icc_data, icc_size, &profile));
 
   // skcms does not return the rendering intent, so get it from the file. It
-  // is encoded as big-endian 32-bit integer in bytes 60..63.
-  uint32_t rendering_intent32 = icc_data[67];
-  if (rendering_intent32 > 3 || icc_data[64] != 0 || icc_data[65] != 0 ||
-      icc_data[66] != 0) {
-    return JXL_FAILURE("Invalid rendering intent %u\n", rendering_intent32);
+  // should be encoded as big-endian 32-bit integer in bytes 60..63.
+  uint32_t big_endian_rendering_intent = icc_data[67] + (icc_data[66] << 8) +
+                                         (icc_data[65] << 16) +
+                                         (icc_data[64] << 24);
+  // Some files encode rendering intent as little endian, which is not spec
+  // compliant. However we accept those with a warning.
+  uint32_t litte_endian_rendering_intent = (icc_data[67] << 24) +
+                                           (icc_data[66] << 16) +
+                                           (icc_data[65] << 8) + icc_data[64];
+  uint32_t candidate_rendering_intent =
+      std::min(big_endian_rendering_intent, litte_endian_rendering_intent);
+  if (candidate_rendering_intent == litte_endian_rendering_intent) {
+    JXL_WARNING(
+        "Invalid rendering intent bytes: [0x%02X 0x%02X 0x%02X 0x%02X], "
+        "assuming %u was meant",
+        icc_data[64], icc_data[65], icc_data[66], icc_data[67],
+        candidate_rendering_intent);
+  }
+  if (candidate_rendering_intent > 3) {
+    return JXL_FAILURE("Invalid rendering intent %u\n",
+                       candidate_rendering_intent);
   }
   // ICC and RenderingIntent have the same values (0..3).
-  c_enc.rendering_intent = static_cast<RenderingIntent>(rendering_intent32);
+  c_enc.rendering_intent =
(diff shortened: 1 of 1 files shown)


sboukortt:
The fact that the profile that prompted this happens to contain `rendering intent = [0x01 0x00 0x00 0x00]` is a bit suspicious and gives the impression that the mistake was that the integer was encoded as little-endian instead of big-endian, suggesting that we should guess 1, not 0. How about trying both endiannesses instead?

jonsneyers:
TIFF/Exif allow both endiannesses. I don't know the ICC specification, I guess it doesn't allow that? But I agree that this could be an endianness mistake.

mo271:
Reading the ICC spec, it becomes clear that it doesn't allow little-endianness. 
So how to proceed here in case the byte at offset 64 is non-zero? I see two options when actually assuming that there was an endianness mistake during encoding,
1. ... and correct it and print a warning (this would mean rendering intent = 1 in the case of bug #1973)
2. ... print no warning
2. ... fail

I think 1. is the best option and will prepare a fix in that direction.

jonsneyers:
Agreed that 1) is the best option.

lib/jxl/enc_color_management.cc
```diff
@@ -887,11 +887,31 @@ Status ColorEncoding::SetFieldsFromICC() {
...
+    }
+    rendering_intent32 = icc_[64];
+    JXL_WARNING(
+        "Invalid rendering intent bytes: [0x%02X 0x%02X 0x%02X 0x%02X], "
+        "assuming %u was meant",
+        icc_[64], icc_[65], icc_[66], icc_[67], rendering_intent32);
+  }
+  uint32_t combined_rending_intent = rendering_intent32 + (icc_[66] << 8) +
+                                     (icc_[65] << 16) + (icc_[64] << 24);
+  if (combined_rending_intent > 3) {
```
sboukortt, line 914: This is empty. Perhaps the similar check above could be moved here? Also, in the big endian case, `combined_rendering_intent` ends up with `icc_[64] + icc_[64] << 24` as far as I can tell.

Perhaps clearer would be to remove all this and just:

1. compute `big_endian_rendering_intent` from all four bytes (no “just take the last byte and check that the other bytes are zero” trick);
2. compute `little_endian_rendering_intent` from all four bytes in the other direction;
3. take the min and warn if that min is the little-endian version;
4. return an error if it’s more than 3.

What do you think?

sboukortt (changes requested):

lib/jxl/enc_color_management.cc
mo271: I like that approach, will do it like this!

sboukortt (dismissed):

lib/jxl/enc_color_management.cc
```diff
@@ -887,14 +887,26 @@ Status ColorEncoding::SetFieldsFromICC() {
...
-      icc_[66] != 0) {
-    return JXL_FAILURE("Invalid rendering intent %u\n", rendering_intent32);
+  // should be encoded as big-endian 32-bit integer in bytes 60..63.
+  uint32_t big_endian_rendering_intent =
+      icc_[67] + (icc_[66] << 8) + (icc_[65] << 16) + (icc_[64] << 24);
+  // Some files encode rendering intent as little endian, which is not spec
+  // compliant. However we accept those with a warning.
+  uint32_t litte_endian_rendering_intent =
+      (icc_[67] << 24) + (icc_[66] << 16) + (icc_[65] << 8) + icc_[64];
+  uint32_t cand_rendering_intent =
```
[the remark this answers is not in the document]
mo271: done

sboukortt (approved):

sboukortt (approved):

PR #1995, opened 2022-12-22
Update software_support.md

mo271:


jonsneyers:
Maybe also add the Serif Affinity software suite and Adobe Camera Raw?

mo271:
> Maybe also add the Serif Affinity software suite and Adobe Camera Raw?

Done

jonsneyers (approved):

PR #1997, opened 2022-12-27
Make extras depend on extras_codec instead of duplicating source files

sboukortt:
This ensures that everything that uses them has them built with the same flags.


veluca93 (dismissed):

mo271:
I think this has been superseded  by other changes to the build system

PR #2002, opened 2022-12-28
WIP: cms stage

mo271:
Work towards fixing #1880.

This is currently only a rough draft.

Co-authored-by: Zoltan Szabadka <szabadka@google.com>

PR #2012, opened 2022-12-31
cjxl: add --use_original_profile option

Traneptora:
Enable --use_original_profile on the command-line tool cjxl, which makes libjxl avoid the XYB transform and encode the image in the same color space as the original image. This is potentially useful if the original ICC profile needs to be preserved.

### DIFF
--- a/lib/extras/enc/jxl.cc
+++ b/lib/extras/enc/jxl.cc
@@ -91,7 +91,7 @@ bool EncodeImageJXL(const JXLCompressParams& params, const PackedPixelFile& ppf,
         std::max<uint32_t>(num_alpha_channels, ppf.info.num_extra_channels);
     basic_info.num_color_channels = ppf.info.num_color_channels;
     const bool lossless = params.distance == 0;
-    basic_info.uses_original_profile = lossless;
+    basic_info.uses_original_profile = lossless || params.uses_original_profile;
     if (params.override_bitdepth != 0) {
       basic_info.bits_per_sample = params.override_bitdepth;
       basic_info.exponent_bits_per_sample =
--- a/lib/extras/enc/jxl.h
+++ b/lib/extras/enc/jxl.h
@@ -47,6 +47,8 @@ struct JXLCompressParams {
   // Upper bound on the intensity level present in the image in nits (zero means
   // that the library chooses a default).
   float intensity_target = 0;
+  // Override uses_original_profile
+  bool uses_original_profile = false;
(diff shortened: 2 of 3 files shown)


jyrkialakuijala:
I am not sure of the implications -- will this actually store the colors in original or still in xyb and only recover the original color profile.

possibly the flag name could make explicit which one it is

if it stores colors in the original profile, it likely breaks all psychovisual optimizations (such as adaptive quantization, smoothing and ac-strategy) since they all silently assume xyb

in our internal xyb coding we have the luma component in index 1, whereas most codings would have it likely at 0 -- and some color codings don't have luma at all

thebombzen:
> I am not sure of the implications -- will this actually store the colors in original or still in xyb and only recover the original color profile.

As far as I'm aware, `uses_original_profile` in the basic info struct of the libjxl API disables XYB-encoding. So the option would basically disable XYB and pass through the original profile image.

The reason you might want to do this is that as far as I'm aware, if the original ICC is something weird (e.g. the green bench), then libjxl doesn't let you recover the original profile and also request pixels in the space described by that profile. 

And yes this generally would not be desired unless you specifically needed to recover pixels in the space described by the original profile, which is why I hid it behind --verbose. That being said, I could add better documentation. I could also rename the option --disable_xyb or something of that form, if that's a bit more clear.

jonsneyers:
I think that what you want here will be possible in a different and better way when https://github.com/libjxl/libjxl/pull/2002 is done.

In my opinion, not doing lossy compression in XYB just to circumvent the issue that the current decoder cannot handle arbitrary ICC profiles (while the current encoder can) seems like a bad idea. We should just make it so that the decoder can give you the pixels in the original colorspace, regardless of whether that's an enum space or a space that requires a full CMS. In the lightweight decoder-only dependency-free `libjxl_dec` this will not work (unless you bring your own cms and plug it in via the interface we provide), but in the full `libjxl` we depend on a cms anyway (for the encoder) so there is no reason why we can't just make it work on the decode side too.

That's exactly what https://github.com/libjxl/libjxl/pull/2002 is all about and it's one of the most important final pieces of functionality/polishing still missing for version 1.0, imo.

This said, there might be other reasons to not use XYB, e.g. if the image data is not visual at all or optimizing perceptually is not desired. This could be an interesting option to add for that reason. I'd indeed put the option in a somewhat hidden spot (maybe even behind `--allow_expert_options`), since this is not something you'd generally want to do when doing lossy compression on images aimed at human consumption.

PR #2015, opened 2023-01-02
[fast-lossless] Speedup and enable (if supported) AVX512 by default 

veluca93:
With the RLE improvements and the changes here, AVX512 is about 50% faster than AVX2 on a i7-11850H.

```
Before, AVX2:
build/fast_lossless /tmp/noise-4.ppm
   245.864 MP/s
    14.668 bits/pixel
build/fast_lossless /tmp/noise-6.ppm
   236.304 MP/s
    20.813 bits/pixel
build/fast_lossless /tmp/noise-8.ppm
   241.526 MP/s
    26.878 bits/pixel
build/fast_lossless /tmp/noise-10.ppm
   164.410 MP/s
    32.878 bits/pixel
build/fast_lossless /tmp/noise-11.ppm
   161.683 MP/s
    35.897 bits/pixel
build/fast_lossless /tmp/noise-12.ppm
   159.196 MP/s
    38.912 bits/pixel
build/fast_lossless /tmp/noise-13.ppm
   159.249 MP/s
    41.925 bits/pixel
build/fast_lossless /tmp/noise-14.ppm
   141.446 MP/s
    48.705 bits/pixel
build/fast_lossless /tmp/noise-15.ppm
   117.555 MP/s
    62.039 bits/pixel
build/fast_lossless /tmp/noise-16.ppm
   119.843 MP/s
    66.092 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb8/artificial.ppm
   401.069 MP/s
     2.677 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb16/artificial.ppm
   185.927 MP/s
    10.235 bits/pixel



Before, AVX512:
build/fast_lossless /tmp/noise-4.ppm
   288.877 MP/s
    14.668 bits/pixel
build/fast_lossless /tmp/noise-6.ppm
   274.973 MP/s
    20.813 bits/pixel
build/fast_lossless /tmp/noise-8.ppm
   268.379 MP/s
    26.878 bits/pixel
build/fast_lossless /tmp/noise-10.ppm
   176.508 MP/s
    32.878 bits/pixel
build/fast_lossless /tmp/noise-11.ppm
   173.567 MP/s
    35.897 bits/pixel
build/fast_lossless /tmp/noise-12.ppm
   167.652 MP/s
    38.912 bits/pixel
build/fast_lossless /tmp/noise-13.ppm
   164.720 MP/s
    41.925 bits/pixel
build/fast_lossless /tmp/noise-14.ppm
   147.509 MP/s
    48.705 bits/pixel
build/fast_lossless /tmp/noise-15.ppm
   131.819 MP/s
    62.039 bits/pixel
build/fast_lossless /tmp/noise-16.ppm
   133.364 MP/s
    66.092 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb8/artificial.ppm
   492.252 MP/s
     2.828 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb16/artificial.ppm
   215.241 MP/s
    10.298 bits/pixel




After, AVX512:
build/fast_lossless /tmp/noise-4.ppm
   401.756 MP/s
    14.668 bits/pixel
build/fast_lossless /tmp/noise-6.ppm
   380.501 MP/s
    20.813 bits/pixel
build/fast_lossless /tmp/noise-8.ppm
   359.672 MP/s
    26.878 bits/pixel
build/fast_lossless /tmp/noise-10.ppm
   232.566 MP/s
    32.878 bits/pixel
build/fast_lossless /tmp/noise-11.ppm
   228.881 MP/s
    35.897 bits/pixel
build/fast_lossless /tmp/noise-12.ppm
   221.053 MP/s
    38.912 bits/pixel
build/fast_lossless /tmp/noise-13.ppm
   224.517 MP/s
    41.925 bits/pixel
build/fast_lossless /tmp/noise-14.ppm
   216.443 MP/s
    48.705 bits/pixel
build/fast_lossless /tmp/noise-15.ppm
   196.745 MP/s
    62.039 bits/pixel
build/fast_lossless /tmp/noise-16.ppm
   198.877 MP/s
    66.092 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb8/artificial.ppm
   598.934 MP/s
     2.828 bits/pixel
build/fast_lossless /data/jpeg_xl_data/imagecompression.info.rgb16/artificial.ppm
   305.824 MP/s
    10.298 bits/pixel
```

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -46,7 +46,7 @@ jobs:
           # Build scalar-only hwy instructions.
           - name: scalar
             mode: release
-            cxxflags: -DHWY_COMPILE_ONLY_SCALAR
+            cxxflags: -DHWY_COMPILE_ONLY_SCALAR -DFJXL_ENABLE_AVX2=0 -DFJXL_ENABLE_AVX512=0
           # Disabling optional features to speed up msan build a little bit.
           - name: msan
             skip_install: true
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -18,10 +18,47 @@
 #include <memory>
 #include <vector>
 
+// Enable NEON and AVX2/AVX512 if not asked to do otherwise and the compilers
+// support it.
 #if defined(__aarch64__) || defined(_M_ARM64)
 #include <arm_neon.h>
+
+#ifndef FJXL_ENABLE_NEON
+#define FJXL_ENABLE_NEON 1
+#endif
+
 #elif defined(__x86_64__) || defined(_M_X64)
 #include <immintrin.h>
+
+// TODO(veluca): MSVC support for dynamic dispatch.
+#if defined(__clang__) || defined(__GNUC__)
+
+#ifndef FJXL_ENABLE_AVX2
+#define FJXL_ENABLE_AVX2 1
+#endif
+
+#ifndef FJXL_ENABLE_AVX512
+// On clang-7 or earlier, and gcc-8 or earlier, AVX512 seems broken.
+#if (defined(__clang__) && __clang_major__ > 7) || \
+    (defined(__GNUC__) && __GNUC__ > 8)
+#define FJXL_ENABLE_AVX512 1
+#endif
+#endif
+
+#endif
+
+#endif
+
+#ifndef FJXL_ENABLE_NEON
+#define FJXL_ENABLE_NEON 0
+#endif
+
+#ifndef FJXL_ENABLE_AVX2
+#define FJXL_ENABLE_AVX2 0
+#endif
+
+#ifndef FJXL_ENABLE_AVX512
+#define FJXL_ENABLE_AVX512 0
 #endif
 
 namespace {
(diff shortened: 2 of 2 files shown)


jonsneyers:
While avx512 now makes a lot more sense to use by default, I am not entirely comfortable with the idea of getting slightly worse compression just because you have a cpu that can do wider SIMD.
In particular, I can imagine that you'd still want to have the slightly better compression and use avx2 only, which is still much faster than e2.

Is there a way to make avx512 and avx2 produce the same output? If not, then maybe we should add an e0 that does the simd-width-dependent thing, and keep e1 like it is now, independent of simd width (even if that means not using avx512, or at least not where it makes a difference in compression).

veluca93:
> While avx512 now makes a lot more sense to use by default, I am not entirely comfortable with the idea of getting slightly worse compression just because you have a cpu that can do wider SIMD. In particular, I can imagine that you'd still want to have the slightly better compression and use avx2 only, which is still much faster than e2.
> 
> Is there a way to make avx512 and avx2 produce the same output? If not, then maybe we should add an e0 that does the simd-width-dependent thing, and keep e1 like it is now, independent of simd width (even if that means not using avx512, or at least not where it makes a difference in compression).

You already get this behaviour anyway with avx2 Vs no-avx (which get somewhat better compression still). Given that the difference is minimal (a couple of % in rle-heavy areas), I wouldn't be too worried.

jyrkialakuijala:
I'm with Luca -- I strongly believe that especially with fast lossless and other fastest encoding options we need to match the coding decisions to the simd width, but we may want to do that even in slower modes, depending on the density/speed impact.

In the fast lossless the weight on density is pretty minimal and we just need a fast coding for the use case of loading an album of images, applying a color correction/processing and saving them back. Think 500 studio quality photographs processed this way. This kind of speed improvements are the thing that put JPEG XL on the map for raw-like storage for photo editing tools. This is not a fully theoretical use case, but one where JPEG XL already is the best performer -- so making it stronger will further increase the appeal and utility.

jyrkialakuijala (approved):
Ideally there would some more guiding comments on the purpose of the functions.

jonsneyers:
For the artificial.ppm image, the compression difference is a bit over 5%. This is an image that benefits from rle, but maybe for some screenshot images the difference is even larger.

Of course 600 Mp/s instead of 400 MP/s is very nice, but for most 'normal' use cases, both are just 'more than fast enough' / 'image saves instantly'. In that sense I think 5% compression gain can be preferable over 50% speed improvement.

It's mostly just confusing ux, imo, that doing the same thing from the user pov (saving an image at e1) can produce a noticeably larger file after getting a better computer, while the time to encode is probably not noticeably different (going from 10ms to 7ms is not something humans can generally appreciate).

I am not opposed to doing that anyway by default, but there should at least be an option (preferably a runtime option) to get the 'consistent result' behavior as opposed to the 'best trade-off' behavior.

veluca93:
> Ideally there would some more guiding comments on the purpose of the functions.

git and GitHub made an absolute mess at rendering the diff - in practice there are only three new functions (`AppendBytesWithBitOffset`, `StoreToWriter` and `StoreToWriterAVX512`), whose name and type signature I think make the purpose sufficiently clear.

veluca93:
> For the artificial.ppm image, the compression difference is a bit over 5%. This is an image that benefits from rle, but maybe for some screenshot images the difference is even larger.
> 
> Of course 600 Mp/s instead of 400 MP/s is very nice, but for most 'normal' use cases, both are just 'more than fast enough' / 'image saves instantly'. In that sense I think 5% compression gain can be preferable over 50% speed improvement.
> 
> It's mostly just confusing ux, imo, that doing the same thing from the user pov (saving an image at e1) can produce a noticeably larger file after getting a better computer, while the time to encode is probably not noticeably different (going from 10ms to 7ms is not something humans can generally appreciate).
> 
> I am not opposed to doing that anyway by default, but there should at least be an option (preferably a runtime option) to get the 'consistent result' behavior as opposed to the 'best trade-off' behavior.

The way I see it, -e 1 = absolute fastest you can get with decent compression, and -e 2 is where you start making more meaningful tradeoffs between speed and compression. There is I think some space for a fancier encoding method at effort 2, for example by doing very simple generic lz77 and having a simple decision tree.

jyrkialakuijala:
Jon, you can relatively easily increase consistency by not using the avx/avx2/avx512 code paths, but I'd consider it a rare need. I believe in some compilations we generate the code for several simd paths, so it would be technically possible to dynamically choose weaker codes when needed.

Consider discussing this with Jan on highway -- possibly it should be solved and supported at highway level rather than just hacked into libjxl.

jyrkialakuijala (approved):

jyrkialakuijala (approved):

PR #2018, opened 2023-01-03
Fix big-endian test for float16

malaterre:
The following commit allow the
`DecodeTest/DecodeTestParam.PixelTest/301x33*f16*` test family to pass on big-endian architectures.


mo271:
fyi: The failure in Cross-compiling was there before and is in the process of being fixed.

szabadka (approved):

PR #2025, opened 2023-01-04
remove TODO about SCALAR

mo271:
Should be done after #2023.


szabadka (commented):
I just realized that the reason I could not tighten the expectation is that the baseline is wrong. Could you also update the expectation to 40018 and lower the bound to 10?

mo271:
Done

szabadka (approved):

PR #2028, opened 2023-01-04
Add an -e10 option for best lossless compression

veluca93:


lib/jxl/enc_modular.cc
```diff
@@ -1417,6 +1417,7 @@ Status ModularFrameEncoder::PrepareStreamParams(const Rect& rect,
       case SpeedTier::kKitten:
         nb_rcts_to_try = 9;
         break;
+      case SpeedTier::kGlacier:
```
jonsneyers, line 1420: Could consider to try even more RCTs. IIRC, the 19 are the unique ones up to permutation, but with the extra properties, permutations could make a difference. Probably not worth it but it might be interesting to check how much of a difference it can make...

jonsneyers (approved):

PR #2029, opened 2023-01-05
Improve parallelization of e10.

veluca93:


mo271 (approved):

lib/jxl/enc_frame.cc
```diff
@@ -1152,8 +1140,34 @@ Status EncodeFrame(const CompressParams& cparams_orig,
...
+        },
+        "Compress kGlacier"));
+    JXL_RETURN_IF_ERROR(num_errors.load(std::memory_order_relaxed) == 0);
+
+    size_t best_idx = 0;
+    for (size_t i = 1; i < all_params.size(); i++) {
+      if (size[best_idx] > size[i]) {
+        best_idx = i;
+      }
+    }
```
sboukortt, line 1167: How about:

```suggestion
    size_t best_idx = std::distance(size.begin(), std::max_element(size.begin(), size.end()));
```

PR #2044, opened 2023-01-10
Avoid EnumValid as much as possible

mo271:
We change the order of evaluations in some if statements in order to avoid calling EnumValid when it is in fact not valid, e.g. when color_primaries==12.

### DIFF
--- a/lib/jxl/enc_color_management.cc
+++ b/lib/jxl/enc_color_management.cc
@@ -849,9 +849,9 @@ bool ApplyCICP(const uint8_t color_primaries,
 
   const auto primaries = static_cast<Primaries>(color_primaries);
   const auto tf = static_cast<TransferFunction>(transfer_characteristics);
-  if (!EnumValid(tf) || tf == TransferFunction::kUnknown) return false;
-  if (!(EnumValid(primaries) || color_primaries == 12) ||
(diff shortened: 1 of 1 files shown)


lib/jxl/enc_color_management.cc
```diff
@@ -849,9 +849,9 @@ bool ApplyCICP(const uint8_t color_primaries,
 
   const auto primaries = static_cast<Primaries>(color_primaries);
   const auto tf = static_cast<TransferFunction>(transfer_characteristics);
-  if (!EnumValid(tf) || tf == TransferFunction::kUnknown) return false;
-  if (!(EnumValid(primaries) || color_primaries == 12) ||
-      primaries == Primaries::kCustom) {
+  if (tf == TransferFunction::kUnknown || !EnumValid(tf)) return false;
```
sboukortt, line 852: Do we need the swapping for `kUnknown` and `kCustom`? Are they not considered valid by `EnumValid`?

sboukortt (approved):

lib/jxl/enc_color_management.cc
mo271: No, they are considered valid, I just thought it would be nice to switch those statements around as well, in order to make the coder easier to read.

PR #2093, opened 2023-01-23
Allow fast-lossless for 16-bit float input

veluca93:


jonsneyers:
Maybe a test to check that it roundtrips would be good. It would be interesting to see if it roundtrips also on out-of-range values.

veluca93:
> Maybe a test to check that it roundtrips would be good. It would be interesting to see if it roundtrips also on out-of-range values.

It does - we support u16, which is more values than f16.

As for a test, I don't see a way to generate f16 values...

mo271 (approved):

kmilos:
> I don't see a way to generate f16 values...

Casting to/from u16? Just need to maybe take extra care to exclude the special values (+/-Inf, qNaN, sNaN) from the tests if they're not supported in the f32 case either?

veluca93:
> > I don't see a way to generate f16 values...
> 
> [Casting to/from u16?](https://github.com/AcademySoftwareFoundation/Imath/blob/fcb8a32bb23a5467a4b8e4c51890b85041a4b1c0/src/Imath/half.h#L263-L269) Just need to maybe take extra care to exclude the special values (+/-Inf, qNaN, sNaN) from the tests if they're not supported in the f32 case either?

I meant in our testing infra ;)

kmilos:
> I meant in our testing infra ;)

Yep, sorry, I assumed random numbers are needed only, in which case random u16 could do... Otherwise, [convert from f32](https://gist.github.com/rygorous/2156668)...

PR #2114, opened 2023-01-27
Compare pixels to libjpeg output in decode_api_test.cc

szabadka:
This way we can have tighter bounds that are not so much dependent on the quality of the input image.


mo271:
should the brotli changes be here? Perhaps rebase?

mo271 (approved):

Issue #2111, opened 2023-01-26
Inefficient alpha compression in lossy JXL


jendalinda:
When encoding lossy JXL, distance/quality of alpha channel cannot be adjusted separately and depends on the overall distance/quality setting. This may result in inefficient compression of alpha channel and larger file size.

In the example bellow, there are three JXL files. All of them were cretaed using the recent version of cjxl.
1. The lone alpha channel encoded as grayscale image, arguments: "-d 0 -e 9". The file size is **1091 bytes**.
2. The RGB image, arguments: "-e 9". The file size is **14861 bytes**.
3. The RGB+Alpha image combinng the previous images,  arguments: "-e 9". The file size is **19594 bytes**.

Apparetly the image 3 is notieably larger than images 1 and 2 added together. The reason is the lossless encoding, in this case, compresses the alpha channel much more efficiently than lossy encoding.

**My suggested solution is to allow specifying independent distance/quality for alpha channel.** Perhaps the default could be set to lossless alpha. Lossless compression is very efficient for alpha consisting mostly of fully transparent and fully opaque areas.

![1-only_alpha.jxl](https://user-images.githubusercontent.com/32044039/214905884-ab1cf1c1-bcb1-47e5-a4a4-8957e3325dcf.jpg)
![2-only_rgb.jxl](https://user-images.githubusercontent.com/32044039/214905890-c38f8ebb-5f20-46be-920e-6bed9b611803.jpg)
![3-rgb+alpha.jxl](https://user-images.githubusercontent.com/32044039/214905891-3ed29b75-c48c-43e6-a1cf-13a268a993ca.jpg)

jendalinda:
Adding the "-a" option for setting alpha distance is a big improvement. However, RGB+Alpha image is still bigger than RGB and Alpha components encoded as separate images.
Tried "cjxl -d 1 -a 0 -e 9", the result is  17513 bytes.
Is it some overhead or is the compression of extra channels less efficient than plain grayscale image?

jonsneyers:
Can you share the input image so I can try to understand what's happening here?

jendalinda:
Sorry I made a mistake, I used RGB pixels that were already lossily compressed. The difference is that fresh pixels produce even bigger file.
These are the same PNG filess that were used to produce the original set of JXL files. RGB pixels, alpha as grayscale and RGB+Alpha combined respectively:

[...]

PR #2117, opened 2023-01-27
Alpha distance

jonsneyers:
Adds an option to the API to set the distance for extra channels separately, addressing #2111 .

This e.g. allows doing lossless alpha while the color is lossy, or in general to set the alpha quality independently of the color quality. We might want to experiment with this and set more sensible defaults. (Note: we can't really rely on the metrics though, since they don't really look at alpha, they just use it to blend the image to a solid background.)

Adds a parameter `D` to benchmark_xl and `--alpha_distance` to cjxl.

Flyby fix: correct the modular quantization adjustment to compensate for the actual bitdepth of extra channels (instead of assuming that all channels have the same bitdepth).

```
Encoding       kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
--------------------------------------------------------------------------------------------------------------------------
jxl:d0.5          4312  1326476    2.4608889   1.247   9.371   1.46619773  93.29027781   0.32387099  0.797010517600      0
jxl:d0.5:D0       4312  1535745    2.8491264   1.296   9.113   1.52147913  93.17293088   0.29785035  0.848613285350      0
jxl:d1            4312   923797    1.7138356   1.526   9.811   2.18971467  88.25578150   0.54898038  0.940862091175      0
jxl:d1:D0         4312  1262339    2.3419014   1.246   8.648   2.03960395  88.48580332   0.50947993  1.193151755443      0
jxl:d1:D2         4312   824846    1.5302609   1.487  10.698   2.32580400  88.42132275   0.64015614  0.979605903443      0
jxl:d2            4312   630823    1.1703079   1.507   9.813   3.26052332  82.49513662   0.88900277  1.040406932615      0
jxl:d2:D4         4312   558300    1.0357626   1.391   9.768   3.80601859  82.32916682   1.01417959  1.050449298471      0
jxl:d3            4312   491437    0.9117178   1.425   7.479   4.17905045  78.08833075   1.16395135  1.061195208251      0
jxl:d3:D6         4312   434096    0.8053384   1.468   8.684   4.96663904  77.92568643   1.30559641  1.051446871010      0
Aggregate:        4312   808263    1.4994954   1.395   9.222   2.62748584  85.65005370   0.65969720  0.989212882981      0
```

For some images, doing alpha losslessly results in smaller files than doing it lossy, e.g.:

```
Encoding       kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
--------------------------------------------------------------------------------------------------------------------------
jxl:d0.5           409    47889    0.9353320   0.999  10.693   0.76884204  97.97876012   0.29019442  0.271428139716      0
jxl:d0.5:D0        409    46700    0.9121094   1.014  12.442   0.76884204  98.26795927   0.29005663  0.264563367873      0
jxl:d1             409    31685    0.6188477   1.017  10.603   1.19141257  91.69366061   0.47456421  0.293682948235      0
jxl:d1:D0          409    31095    0.6073242   1.045  11.505   1.19293487  91.82356930   0.47412265  0.287946168915      0
jxl:d2             409    20969    0.4095508   1.021  10.966   2.02522826  86.19662382   0.75935998  0.310996471061      0
jxl:d2:D0          409    20971    0.4095898   1.039  11.696   2.02522278  86.03497192   0.75843025  0.310645329176      0
jxl:d3             409    16549    0.3232227   1.004   9.977   2.59280419  82.55876856   0.97508168  0.315168489608      0
jxl:d3:D0          409    16891    0.3299023   0.985   8.871   2.64150572  82.59064968   0.97371519  0.321230921912      0
Aggregate:         409    26859    0.5245902   1.015  10.794   1.48461947  89.45138936   0.56480591  0.296291627848      0
```

So maybe we should reconsider the default behavior, and do alpha losslessly when doing high quality lossy, and only make alpha lossy when the quality is low enough. To be investigated later...

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -306,7 +306,7 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
     : frame_dim_(frame_header.ToFrameDimensions()), cparams_(cparams_orig) {
   size_t num_streams =
       ModularStreamId::Num(frame_dim_, frame_header.passes.num_passes);
-  if (cparams_.IsLossless()) {
+  if (cparams_.ModularPartIsLossless()) {
     switch (cparams_.decoding_speed_tier) {
       case 0:
         break;
@@ -331,7 +331,7 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
     }
   }
   if (cparams_.decoding_speed_tier >= 1 && cparams_.responsive &&
-      cparams_.IsLossless()) {
+      cparams_.ModularPartIsLossless()) {
     cparams_.options.tree_kind =
         ModularOptions::TreeKind::kTrivialTreeNoPredictor;
     cparams_.options.nb_repeats = 0;
@@ -341,7 +341,7 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
   // use a sensible default if nothing explicit is specified:
   // Squeeze for lossy, no squeeze for lossless
   if (cparams_.responsive < 0) {
-    if (cparams_.IsLossless()) {
+    if (cparams_.ModularPartIsLossless()) {
       cparams_.responsive = 0;
     } else {
       cparams_.responsive = 1;
@@ -428,7 +428,7 @@ ModularFrameEncoder::ModularFrameEncoder(const FrameHeader& frame_header,
     delta_pred_ = cparams_.options.predictor;
     if (cparams_.lossy_palette) cparams_.options.predictor = Predictor::Zero;
   }
-  if (!cparams_.IsLossless()) {
+  if (!cparams_.ModularPartIsLossless()) {
     if (cparams_.options.predictor == Predictor::Weighted ||
         cparams_.options.predictor == Predictor::Variable ||
         cparams_.options.predictor == Predictor::Best)
@@ -637,8 +637,7 @@ Status ModularFrameEncoder::ComputeEncodingData(
         cparams_.level, max_bitdepth, level_max_bitdepth);
 
   // Set options and apply transformations
-
-  if (cparams_.butteraugli_distance > 0) {
+  if (!cparams_.ModularPartIsLossless()) {
     if (cparams_.palette_colors != 0) {
       JXL_DEBUG_V(3, "Lossy encode, not doing palette transforms");
     }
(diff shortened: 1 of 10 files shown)


veluca93:
Can you split the all_color_aware commit in another PR?

jonsneyers:
> Can you split the all_color_aware commit in another PR?

Yes, that one is already merged (https://github.com/libjxl/libjxl/pull/2116).

Rebased and fixed the merge conflict.

lib/jxl/enc_modular.cc
```diff
@@ -785,20 +784,30 @@ Status ModularFrameEncoder::ComputeEncodingData(
...
+      quantizer *= 0.1f;
     }
+    float bitdepth_correction = 1.f;
     if (cparams_.color_transform != ColorTransform::kXYB) {
-      quality *= maxval / 255.f;
+      bitdepth_correction = maxval / 255.f;
+    }
+    std::vector<float> quantizers;
+    float dist = cparams_.butteraugli_distance;
+    for (size_t i = 0; i < 3; i++)
```
veluca93, line 802: {} please

lib/jxl/encode.cc
```diff
@@ -1031,6 +1031,27 @@ JxlEncoderStatus JxlEncoderSetFrameDistance(
...
+  }
+  if (distance < 0.f || distance > 25.f) {
+    return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_API_USAGE,
+                         "Distance has to be in [0.0..25.0] (corresponding to "
+                         "quality in [0.0..100.0])");
+  }
+  if (distance > 0.f && distance < 0.01f) {
+    distance = 0.01f;
+  }
+  frame_settings->values.cparams.ec_distance.resize(
```
veluca93, line 1048: I don't like the resize happening here - I think we should maintain the invariant that cparams.ec_distanze.size() == number of extra channels.

lib/jxl/enc_modular.cc
jonsneyers: sigh, when will clang-format automatically do that when it adds a newline to a simple loop...

lib/jxl/encode.cc
jonsneyers: That would also make sense. The advantage of doing it this way is that setting color distance first and then setting alpha distance does the same thing as setting alpha distance first and then setting color distance, while not setting alpha distance always makes it use the color distance. If we want to maintain that invariant, then the ec_distance needs to get values before color distance is set, so then it would mean changing the color distance would need to overwrite the ec_distances (so not setting alpha distance makes it use the right default) so you would need to first set color distance and then alpha distance.

Then again this approach also isn't invariant to the order of setting things in the case that there are multiple extra channels and some of them get an explicit distance while others use the default. Maybe let's do something where distance -1 means "default", then we can have the invariant and also avoid requiring a specific order of setting things.

mo271:
Could you add an entry to the changelog?

lib/jxl/encode.cc
jonsneyers: Fully maintaining the invariant is annoying, because cparams doesn't know how many extra channels there are. I'll make it so that in our use of cparams in encode.cc, we do maintain the invariant that ec_distance.size() == nb_extra_chans, but in enc_modular.cc, if ec_distance.size() < nb_extra_chans, it just assumes the missing values are default (-1, same as color distance).

jonsneyers:
> Could you add an entry to the changelog?

Done, thanks for the reminder!

lib/jxl/encode.cc
jonsneyers: PTAL if you like it better this way.

veluca93 (approved):

PR #2120, opened 2023-01-30
fix palette bug

### LINKED ISSUE #2118: Fast lossless can produce invalid/incorrect JXLs
cjxl in fast lossless mode produces invalid JXLs from the attached PNGs.

**To Reproduce**
Encode the attached PNGs with `cjxl -e 1 -d 0` and attempt to decode.

[failure.png](https://user-images.githubusercontent.com/16582685/215329162-c94173e3-8f5e-4050-a95e-726a16ba997f.png)
```
$ cjxl failure.png failure.jxl -d 0 -e 1 && djxl failure.jxl --disable_output
JPEG XL encoder v0.9.0 20ca234f [AVX2,SSE4,SSSE3,Unknown]
./lib/extras/dec/color_hints.cc:54: No color_space/icc_pathname given, assuming sRGB
Read 2x2 image, 84 bytes, 0.3 MP/s
Encoding [Modular, lossless, effort: 1], 
Compressed to 183 bytes (366.000 bpp).
2 x 2, 0.00 MP/s [0.00, 0.00], 1 reps, 8 threads.
JPEG XL decoder v0.9.0 20ca234f [AVX2,SSE4,SSSE3,Unknown]
Read 183 compressed bytes.
./lib/jxl/modular/encoding/encoding.cc:566: JXL_FAILURE: Truncated input
./lib/jxl/modular/encoding/encoding.cc:597: JXL_RETURN_IF_ERROR code=1: dec_status
./lib/jxl/dec_modular.cc:257: JXL_RETURN_IF_ERROR code=1: dec_status
./lib/jxl/decode.cc:1139: frame out of bounds
Failed to decode image
```

Decoder backtrace to the first failure:

```
#0  jxl::ModularDecode (br=0x5555555f4760, image=..., header=..., group_id=0, options=0x7fffffffd1e0, global_tree=0x5555555fb748, global_code=0x5555555fb760, global_ctx_map=0x5555555fb7f0, allow_truncated_group=false)
    at ./lib/jxl/modular/encoding/encoding.cc:566
#1  0x00007ffff7b5b9bd in jxl::ModularGenericDecompress (br=0x5555555f4760, image=..., header=0x5555555fb808, group_id=0, options=0x7fffffffd1e0, undo_transforms=false, tree=0x5555555fb748, code=0x5555555fb760, ctx_map=0x5555555fb7f0, allow_truncated_group=false)
    at ./lib/jxl/modular/encoding/encoding.cc:595
#2  0x00007ffff7adba31 in jxl::ModularFrameDecoder::DecodeGlobalInfo (this=0x5555555fb640, reader=0x5555555f4760, frame_header=..., allow_truncated_group=false) at ./lib/jxl/dec_modular.cc:252
#3  0x00007ffff7a4c3d0 in jxl::FrameDecoder::ProcessDCGlobal (this=0x5555555fb330, br=0x5555555f4760) at ./lib/jxl/dec_frame.cc:292
#4  0x00007ffff7a4f14e in jxl::FrameDecoder::ProcessSections (this=0x5555555fb330, sections=0x5555555fbf70, num=1, section_status=0x5555555fbf90) at ./lib/jxl/dec_frame.cc:619
#5  0x00007ffff7afa4e1 in jxl::(anonymous namespace)::JxlDecoderProcessSections (dec=0x5555555f4bd0) at ./lib/jxl/decode.cc:1123
#6  0x00007ffff7afbc1d in jxl::(anonymous namespace)::JxlDecoderProcessCodestream (dec=0x5555555f4bd0) at ./lib/jxl/decode.cc:1438
#7  0x00007ffff7afd682 in HandleBoxes (dec=0x5555555f4bd0) at ./lib/jxl/decode.cc:1898
#8  0x00007ffff7afdd42 in JxlDecoderProcessInput (dec=0x5555555f4bd0) at ./lib/jxl/decode.cc:2049
#9  0x0000555555570546 in jxl::extras::DecodeImageJXL (bytes=0x5555555f4af0 "\377\nA\300J\b\b\020", bytes_size=201, dparams=..., decoded_bytes=0x7fffffffdd40, ppf=0x7fffffffe230, jpeg_bytes=0x0) at ./lib/extras/dec/jxl.cc:205
#10 0x000055555555e7f3 in (anonymous namespace)::DecompressJxlToPackedPixelFile (args=..., compressed=std::vector of length 201, capacity 201 = {...}, accepted_formats=std::vector of length 0, capacity 0, runner=0x5555555f3ae0, ppf=0x7fffffffe230, decoded_bytes=0x7fffffffdd40, 
    stats=0x7fffffffde30) at ./tools/djxl_main.cc:283
#11 0x000055555555f378 in main (argc=3, argv=0x7fffffffe568) at ./tools/djxl_main.cc:437
```

The second PNG attached here is a crop of the first one - it appears to work correctly, but decodes to the wrong pixels.

[wrongpixels.png](https://user-images.githubusercontent.com/16582685/215329196-362ccb34-8553-45ac-92f5-4683eb492a31.png)

```
$ cjxl wrongpixels.png wrongpixels.jxl -d 0 -e 1
JPEG XL encoder v0.9.0 20ca234f [AVX2,SSE4,SSSE3,Unknown]
./lib/extras/dec/color_hints.cc:54: No color_space/icc_pathname given, assuming sRGB
Read 1x2 image, 75 bytes, 0.1 MP/s
Encoding [Modular, lossless, effort: 1], 
Compressed to 182 bytes (728.000 bpp).
1 x 2, 0.00 MP/s [0.00, 0.00], 1 reps, 8 threads.

$ magick identify -format '%#\n' wrongpixels.png wrongpixels.jxl
3674e32a54f125bc23b069ab31af8043f8f17a02ae27f6f22ff19ffdb0961318
c31c8fa31580112a8f7f8aaca994ec1ef63511a32600dca28e2ec02aebfe6a36

$ magick identify -verbose wrongpixels.png wrongpixels.jxl  | grep Colors:
  Colors: 2
  Colors: 1
```

In this case the result is fully transparent - a slightly wider crop around the same pixels resulted in some strange artifacts, as if a section of the image had been rotated 90°.

I haven't been able to reproduce it with inputs larger than a few pixels.

**Expected behavior**
Successful encoding and decoding.

**Environment**
 - OS: Linux
 - Compiler version: clang 14.0.0
 - CPU type: x86_64
 - cjxl/djxl version string: v0.9.0 20ca234f [AVX2,SSE4,SSSE3,Unknown]

**Additional context**
The symptoms are similar to #2039 but the example PNG from that bug doesn't cause a problem now.

mo271:
fixes #2118
Co-authored-by: Luca Versari <veluca@google.com>

### DIFF
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -3425,7 +3425,7 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
   row_encoder.t = &encoder;
   encoder.output = output;
   encoder.code = &code[0];
-  int16_t p[4][32 + 2048] = {};
+  int16_t p[4][32 + 1024] = {};
   uint8_t prgba[4];
   size_t i = 0;
   size_t have_zero = 0;
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -3355,7 +3355,7 @@ void WriteACSectionPalette(const unsigned char* rgba, size_t x0, size_t y0,
 
   row_encoder.t = &encoder;
   encoder.output = &output;
-  encoder.code = &code[0];
+  encoder.code = &code[1];
   ProcessImageAreaPalette<
       ChannelRowProcessor<ChunkEncoder<UpTo8Bits>, UpTo8Bits>>(
       rgba, x0, y0, xs, 0, ys, row_stride, lookup, nb_chans, &row_encoder);
@@ -3374,8 +3374,8 @@ void CollectSamples(const unsigned char* rgba, size_t x0, size_t y0, size_t xs,
         row_sample_collectors[4];
     for (size_t c = 0; c < nb_chans; c++) {
       row_sample_collectors[c].t = &sample_collectors[c];
-      sample_collectors[c].raw_counts = raw_counts[c];
-      sample_collectors[c].lz77_counts = lz77_counts[c];
+      sample_collectors[c].raw_counts = raw_counts[1];
+      sample_collectors[c].lz77_counts = lz77_counts[1];
     }
     ProcessImageAreaPalette<
         ChannelRowProcessor<ChunkSampleCollector<UpTo8Bits>, UpTo8Bits>>(
@@ -3425,7 +3425,7 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
   row_encoder.t = &encoder;
   encoder.output = output;
   encoder.code = &code[0];
-  int16_t p[4][32 + 1024] = {};
+  int16_t p[4][32 + 2048] = {};
   uint8_t prgba[4];
   size_t i = 0;
   size_t have_zero = 0;


lib/jxl/enc_fast_lossless.cc
```diff
@@ -3425,7 +3425,7 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
   row_encoder.t = &encoder;
   encoder.output = output;
   encoder.code = &code[0];
-  int16_t p[4][32 + 1024] = {};
+  int16_t p[4][32 + 2048] = {};
```
jonsneyers, line 3428: Is this needed? I don't remember exactly how this code works but I assume the size of this array should be something like `kMaxColors + kChunkSize` (and possibly the `+ kChunkSize` was only needed there because of potential padding, which is now no longer needed).

veluca93: Ah, you are right, 1024 would have been enough - I think we should instead static_assert that kMaxColors < 1024 then.

veluca93 (approved):

lib/jxl/enc_fast_lossless.cc
jonsneyers: Any reason why we can't just change it to this?
```suggestion
  int16_t p[4][kMaxColors] = {};
```

That way we don't need any static_asserts, and we don't allocate an array that is bigger than needed (`kMaxColors` is 512...

veluca93: This, certainly not. kMaxColors + 32 should work fine though :)

jonsneyers: Where does the 32 come from?

veluca93: kChunkSize * 2, but this might actually not be 32... Then again we also have a +16 below, which should be +kChunkSize instead.

mo271: ok, I'm changing back to 1024 as suggested below. The change of the size of `p` and the `+16` below is somewhat unrelated to the current fix, so perhaps lets do this in another pr?!

veluca93: Sounds good!

mo271: Can you fix this in another pr, @jonsneyers?

jonsneyers: sure

jonsneyers (approved):

PR #2134, opened 2023-02-02
Allow smooth transition to namespace hwy::

malaterre:


eustas (approved):

kleisauke:
It looks like the `TARGET_EXISTS` generator expression was introduced in [CMake 3.12](https://cmake.org/cmake/help/latest/release/3.12.html#generator-expressions), so this currently fails on the `.deb packages / ubuntu:18.04` CI job (since [Ubuntu 18.04 ships with CMake 3.10.2](https://packages.ubuntu.com/bionic/cmake)).
```
CMake Error at lib/jxl_extras.cmake:144 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_EXISTS:hwy::hwy>

  Expression did not evaluate to a known generator expression
```
https://github.com/libjxl/libjxl/actions/runs/4084058635/jobs/7040311736#step:15:1161

PR #2165, opened 2023-02-08
Iso noise

### LINKED ISSUE #2149: 'photon_noise' cannot be set to 'ISO0' despite it being the default.
**Describe the bug**
According to the built in help in cjxl (`-v -h`), the `--photon_noise` parameter works as follows:
> --photon_noise=ISO3200
    Adds noise to the image emulating photographic film noise. The higher the given number, the grainier the image will be. As an example, a value of 100 gives low noise whereas a value of 3200 gives a lot of noise. The default value is 0.

However, if this is set to the 'default' of `0`, the following error appears:
> Error parsing flag --photon_noise=ISO0

**To Reproduce**
Run `cjxl` like so: `cjxl foo.png foo.jxl --photon_noise=ISO0`

**Expected behavior**
Ideally, the image would still be encoded, just with no photon noise at all, even though 'ISO0' doesn't make a whole lot of sense naming-wise.

**Environment**
 - OS: Linux
 - Compiler version: `clang version 14.0.6`
 - CPU type: x86_64
 - cjxl/djxl version string: `cjxl v0.8.1 [AVX2,SSE4,SSSE3,Unknown]`

mo271:
fixes #2149 

Also moving the "ISO" from the argument to the name of the flag.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -54,7 +54,8 @@ namespace tools {
 
 namespace {
 inline bool ParsePhotonNoiseParameter(const char* arg, float* out) {
-  return strncmp(arg, "ISO", 3) == 0 && ParseFloat(arg + 3, out) && *out >= 0;;
+  return ParseFloat(arg, out) && *out >= 0;
+  ;
 }
 inline bool ParseIntensityTarget(const char* arg, float* out) {
   return ParseFloat(arg, out) && *out > 0;
@@ -249,7 +250,7 @@ struct CompressArgs {
                            &disable_output, &SetBooleanTrue, 1);
 
     cmdline->AddOptionValue(
-        '\0', "photon_noise", "ISO3200",
+        '\0', "photon_noise_iso", "3200",
         "Adds noise to the image emulating photographic film noise. "
         "The higher the given number, the grainier the image will be. "
         "As an example, a value of 100 gives low noise whereas a value "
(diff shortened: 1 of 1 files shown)


tools/cjxl_main.cc
```diff
@@ -54,7 +54,8 @@ namespace tools {
 
 namespace {
 inline bool ParsePhotonNoiseParameter(const char* arg, float* out) {
-  return strncmp(arg, "ISO", 3) == 0 && ParseFloat(arg + 3, out) && *out > 0;
+  return ParseFloat(arg, out) && *out >= 0;
+  ;
```
szabadka, line 58: remove?

szabadka (approved):

tools/cjxl_main.cc
mo271, line 58:
```suggestion
  
```

tools/cjxl_main.cc
```diff
@@ -54,7 +54,8 @@ namespace tools {
 
 namespace {
 inline bool ParsePhotonNoiseParameter(const char* arg, float* out) {
-  return strncmp(arg, "ISO", 3) == 0 && ParseFloat(arg + 3, out) && *out > 0;
+  return ParseFloat(arg, out) && *out >= 0;
+
```
mo271, line 58:
```suggestion
  return ParseFloat(arg, out) && *out >= 0;
```

tools/cjxl_main.cc
mo271: done

szabadka (approved):

PR #2169, opened 2023-02-09
don't run some workflows on markdown-only changes

mo271:


mo271:
Seems like this will need extra work to play well with the required checks:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

mo271:
In the cases there are only markdown changes, the workflow foo_md will run.
In the cases there are only non-markdown changes the workflow foo will run.

In the case there are both markdown and non-markdown changes, both foo and foo_md will be run, I suppose

eustas:
Later we would be able to generate those "extra" workflows for markdown files...

I wonder, if we should move some of the MD files to built-in wiki?

eustas (approved):

PR #2172, opened 2023-02-09
Remove unused reference

eustas:


eustas:
The comment mentioned that frame can not reference CodecInOut images. What was the plan to use it?
(in my other unpublished PR I change CodecInOut::frames to store unique_ptr's...)

veluca93 (dismissed):
This only made sense when we were decoding to a vector of frames, it is not very useful now :)

veluca93 (approved):

PR #2174, opened 2023-02-09
Update CHANGELOG.md

mo271:
leftover from #2165


mo271:
Trying out the simplified workflows for markdown-only changes

mo271:
simplified workflows for markdown-only change seems to work now. (Still needs to confirm with a changes that touches both markdown and non-markdown files....

szabadka (approved):

PR #2178, opened 2023-02-09
djxl: fix segmentation fault when JPEG is disabled

jolivain:
When libjxl is compiled without JPEG support, by configuring for example with:

    cmake -DCMAKE_DISABLE_FIND_PACKAGE_JPEG=TRUE ...

djxl crashes with a segmentation fault at:
https://github.com/libjxl/libjxl/blob/v0.8.1/tools/djxl_main.cc#L367

The crash can be reproduced with the sequence:

    gm convert IMAGE:LOGO ref.ppm
    cjxl ref.ppm enc.jxl
    djxl enc.jxl dec.ppm

The crash happen because opt_jpeg_quality_id does not get initialized at:
https://github.com/libjxl/libjxl/blob/v0.8.1/tools/djxl_main.cc#L107

This commit fixes the crash by testing opt_jpeg_quality_id only if libjxl is compiled with JPEG.

Signed-off-by: Julien Olivain <ju.o@free.fr>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,6 +37,7 @@ gi-man
 Heiko Becker <heirecka@exherbo.org>
 Jim Robinson <jimbo2150@gmail.com>
 Jon Sneyers <jon@cloudinary.com>
+Julien Olivain <ju.o@free.fr>
 Kai Hollberg <Schweinepriester@users.noreply.github.com>
 Kleis Auke Wolthuizen <github@kleisauke.nl>
 L. E. Segovia
--- a/tools/cmdline.h
+++ b/tools/cmdline.h
@@ -19,7 +19,7 @@ namespace tools {
 
 class CommandLineParser {
  public:
-  typedef size_t OptionId;
+  typedef ssize_t OptionId;
 
   // An abstract class for defining command line options.
   class CmdOptionInterface {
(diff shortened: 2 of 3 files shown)


eustas:
Hello, Julien.

Those "optional" args are intentionally accessed via `opt_xxx_id`, and initial value for it is `-1`. So, cleaner solution here would be not "if-def", but extra check that `opt_jpeg_quality_id != -1`.

jolivain:
Hi Eugene,

Thanks for the review.  I'll send an updated patch including your proposal.

jyrkialakuijala (dismissed):

mo271:
Thanks for the contribution, @jolivain! Please add yourself to the AUTHORS.md file

mo271:
> Thanks for the contribution, @jolivain! Please add yourself to the AUTHORS.md file

fixed it...

mo271:
rebased

eustas (approved):

ffontaine:
It seems that this issue has been fixed by https://github.com/libjxl/libjxl/commit/c3a4f9ca89ae59c6265a2f1bf2a6d2a87a71fc16

mo271:
> It seems that this issue has been fixed by [c3a4f9c](https://github.com/libjxl/libjxl/commit/c3a4f9ca89ae59c6265a2f1bf2a6d2a87a71fc16)

PR #2184, opened 2023-02-12
Decoder API: remember subscribed events after JxlDecoderRewind

alistair7:
According to decode.h, `JxlDecoderRewind` should keep the previously subscribed decoder events, but it actually forgets them.  I think this fixes it.


mo271 (approved):

jonsneyers:
I didn't really understand why we need `orig_events_wanted` and `events_wanted` in the first place. Looks like `events_wanted` should really be named `events_still_left` or something, and then the `orig_` could be dropped. The current variable names confused me :)

mo271:
thanks for the fix, @alistair7

PR #2190, opened 2023-02-16
plugins/gimp/file-jxl-save: dont submit sRGB-tagged linear samples

thebombzen:
After the last patch, the JPEG XL exporter no longer crashes upon lossless exports but it converts sRGB images to linear before feeding them to libjxl, but it tells libjxl that they are sRGB, resulting in incorrect colors. This patch fixes that issue.

### DIFF
--- a/plugins/gimp/file-jxl-save.cc
+++ b/plugins/gimp/file-jxl-save.cc
@@ -831,11 +831,7 @@ bool SaveJpegXlImage(const gint32 image_id, const gint32 drawable_id,
     g_clear_object(&buffer);
 
     // use babl to fix gamma mismatch issues
-    if (jxl_save_opts.icc_attached) {
-      jxl_save_opts.SetModel(jxl_save_opts.is_linear);
-    } else {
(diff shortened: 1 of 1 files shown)


jonsneyers:
I think for lossless you generally want to save the image in a gamma-compressed space, not in linear space. At least when using uint8...

jonsneyers:
The Gimp plugin code is a bit tricky for me to understand. Is the way you did the fix causing images to be saved in linear space and tagged as linear space, or saved in sRGB space and tagged as sRGB?

thebombzen:
It causes it to feed sRGB data to the encoder and tell the encoder that the data is sRGB. Without the patch it linearizes and then tells libjxl that it's sRGB anyway, which is incorrect. This prevents linearization first.

jyrkialakuijala (approved):

jyrkialakuijala:
I'm approving this without fully understanding consequences and hoping that thebombzen is correct (as usual).

PR #2194, opened 2023-02-18
make sure jpeglib.h is found when building jpegli/libjpeg_wrapper.cc

### LINKED ISSUE #2167: jpegli doesn't compile on OpenBSD
jpegli doesn't compile on OpenBSD because it includes `<jpeglib.h>` which is not in the default include path. In OpenBSD, the default include path is _/usr/include_, but package headers are installed to _/usr/local/include_.

The other BSDs have the same detail, NetBSD in particular uses _/usr/pkg/include_, so it's very likely it doesn't work on the other BSDs too, but I haven't tried it out.

The solution would be to use pkg-config to get the CFLAGS but not the LDFLAGS needed to use libjpeg. This might have to be propagated to libjxl's own pkg-config files.

landryb:
allows to build 0.8.1 on OpenBSD and probably NetBSD/FreeBSD - fixes #2167

note that i haven't checked yet consumers (my interest is for gdal on OpenBSD) so i dunno yet if this has to be "propagated to libjxl's own pkg-config files" as suggested by @guijan in #2167

### DIFF
--- a/lib/jpegli.cmake
+++ b/lib/jpegli.cmake
@@ -85,6 +85,7 @@ set(JPEGLI_LIBJPEG_OBJ_COMPILE_DEFINITIONS
 )
 
 add_library(jpegli-libjpeg-obj OBJECT "${JPEGXL_INTERNAL_JPEGLI_WRAPPER_SOURCES}")
+target_include_directories(jpegli-libjpeg-obj PUBLIC "${JPEG_INCLUDE_DIRS}")
 target_compile_options(jpegli-libjpeg-obj PRIVATE ${JPEGXL_INTERNAL_FLAGS})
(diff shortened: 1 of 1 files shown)


landryb:
fwiw i'm not that much interested in CLA nightmares, so my contribution is in the public domain. Feel free to reassign ownership for my patch.

jyrkialakuijala:
could you file an issue instead? taking a code change without the cla is very difficult for us

landryb:
> could you file an issue instead? taking a code change without the cla is very difficult for us

there's already #2167 which will be fixed by the above commit. Anybody having already signed google's CLA is free to take ownership of the oneliner, its in the public domain. CLAs are made by lawyers for lawyers, and i'm not interested in signing anything with google...

PR #2203, opened 2023-02-20
Cut benchmark->test dependency

eustas:


lib/extras/tone_mapping_gbench.cc
```diff
@@ -7,38 +7,33 @@
...
 
 static void BM_ToneMapping(benchmark::State& state) {
-  CodecInOut image;
-  const PaddedBytes image_bytes = ReadTestData("jxl/flower/flower.png");
-  JXL_CHECK(SetFromBytes(Span<const uint8_t>(image_bytes), &image));
+  Image3F color(2268, 1512);
+  FillImage(0.5f, &color);
 
-  // Convert to linear Rec. 2020 so that `ToneMapTo` doesn't have to and we
+  // Use linear Rec. 2020 so that `ToneMapTo` doesn't have to and we
```
sboukortt, line 17: “Doesn’t have to” now refers to “use” instead of “convert”:
```suggestion
  // Use linear Rec. 2020 so that `ToneMapTo` doesn't have to convert to it and we
```

sboukortt (approved):

PR #2238, opened 2023-02-28
CMake: Export jxl and jxl_threads targets.

FtZPetruska:
## Description

This enables downstream projects to consume libjxl through CMake's `find_package` without relying on pkg-config or Find modules:

```cmake
find_package(Jxl)
target_link_libraries(main PRIVATE Jxl::jxl Jxl::jxl_threads)
```

The only targets exported in this PR are jxl(-static) and jxl_threads(-static) in order to match what is provided with pc files. The `jxl_includes` target is also exported as `Jxl::headers` to handle the include directories for the other targets.

When doing a static-only build (`BUILD_SHARED_LIBS=OFF`), the `Jxl::jxl` and `Jxl::jxl_threads` targets are aliases of their respective static target. This makes it so all four targets are always available.

When Highway, Brotli, LCMS2, or SKCMS are vendored (but not bundled), they are also installed under the `JxlDeps::` namespace. This is required by CMake, as all dependencies of an exported library which are not imported must be exported as well.

When using system dependencies, the associated Find modules are also installed since they are required for static linking.

## Testing

The tests conducted include:
- Building with and without the shared library (ensures all targets are always always and working)
- Using system dependencies + bundled SKCMS (ensures no extraneous targets are installed when not needed)
- Using vendored dependencies with LCMS2 (ensures the `JxlDeps::` targets work properly)
- Using vendored dependencies and non-bundled SKCMS (same as above)

For every configuration, the package was searched by setting either `CMAKE_PREFIX_PATH` to the installed prefix or `Jxl_DIR` to the directory containing `JxlConfig.cmake` (by default `<prefix>/lib/cmake/Jxl`). Then an example executable was built and linked against the exported targets.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,6 +51,7 @@ Misaki Kasumi <misakikasumi@outlook.com>
 Nicholas Hayes <0xC0000054@users.noreply.github.com>
 Nigel Tao <nigeltao@golang.org>
 Petr Diblík
+Pierre Wendling <pierre.wendling.4@gmail.com>
 Pieter Wuille
 roland-rollo
 Samuel Leong <wvvwvvvvwvvw@gmail.com>
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -154,6 +154,52 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/jxl
 install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/jxl
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 
+# Destination of CMake files
+if(NOT DEFINED JPEGXL_INSTALL_CMAKEDIR)
+  set(JPEGXL_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/JPEGXL")
+endif()
+
+# Find modules also need to be installed for non-vendored libraries
+if(JPEGXL_FORCE_SYSTEM_BROTLI)
+  list(APPEND JPEGXL_FIND_MODULES "${PROJECT_SOURCE_DIR}/cmake/FindBrotli.cmake")
+endif()
+if(JPEGXL_FORCE_SYSTEM_HWY)
+  list(APPEND JPEGXL_FIND_MODULES "${PROJECT_SOURCE_DIR}/cmake/FindHWY.cmake")
+endif()
+if(JPEGXL_FORCE_SYSTEM_LCMS2 AND NOT JPEGXL_ENABLE_SKCMS)
+  list(APPEND JPEGXL_FIND_MODULES "${PROJECT_SOURCE_DIR}/cmake/FindLCMS2.cmake")
+endif()
+
+# Prepare and install the CMake config files
+include(CMakePackageConfigHelpers)
+configure_package_config_file("${CMAKE_CURRENT_SOURCE_DIR}/JPEGXLConfig.cmake.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/JPEGXLConfig.cmake"
+  INSTALL_DESTINATION "${JPEGXL_INSTALL_CMAKEDIR}")
+
+write_basic_package_version_file("${CMAKE_CURRENT_BINARY_DIR}/JPEGXLConfigVersion.cmake"
+  VERSION "${JPEGXL_LIBRARY_VERSION}"
+  COMPATIBILITY SameMinorVersion)
+
+install(FILES
+  "${CMAKE_CURRENT_BINARY_DIR}/JPEGXLConfig.cmake"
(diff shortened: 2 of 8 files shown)


kmilos:
See also https://github.com/libjxl/libjxl/pull/1485 @amyspark

amyspark:
Looks good at first glance to me, **except** the namespace. SCEI already designated it as `JPEGXL`. See https://github.com/WebKit/WebKit/blob/main/Source/cmake/FindJPEGXL.cmake and https://github.com/kde/krita/blob/master/cmake/modules/FindJPEGXL.cmake.

I'd need to do deeper testing to ensure the new targets are also correctly exported, in particular `JPEGXL_ENABLE_JPEGLI_LIBJPEG` was giving us enough trouble that we had to disable it.

FtZPetruska:
Thank you for the feedback!

> Looks good at first glance to me, **except** the namespace. SCEI already designated it as `JPEGXL`.

Got it, I will move the targets into `JPEGXL::` and the vendored dependencies into `JPEGXLDeps::` for consistency.

> I'd need to do deeper testing to ensure the new targets are also correctly exported, in particular `JPEGXL_ENABLE_JPEGLI_LIBJPEG` was giving us enough trouble that we had to disable it.

As it currently stands, only the libraries which were already installed are exported. I tried adding the following statement in jpegli.cmake: (`JPEGLI_TARGETS` contains `jpegli-static` and, if it exists, `jpeg`)
```cmake
# file: lib/jpegli.cmake
install(TARGETS ${JPEGLI_TARGETS} EXPORT JxlTargets
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
This worked out of the box, where linking `JPEGXL::jpeg` was a drop-in replacement for libjpeg, and `JPEGXL::jpegli` links properly to a simple executable using `jpegli::FormatString`. There is still some work to be done, most notably installing jpegli's headers, and probably generate a pkg-config file, which are out of scope for this PR, but this seems promising.

I have also noticed that when `JPEGXL_ENABLE_PROFILER` is enabled, installing the `jxl_profiler` target is necessary. This is not surprising since it becomes a dependency of `jxl`, and is would be needed at link time with the static jxl library. Appending `install(TARGETS jxl_profiler EXPORT JxlTargets)` in `jxl_profiler.cmake` solves this issue and installs the library along the others. The caveat is that no header is installed for this target and is just there as a usage dependency of jxl-static. Do you think it is preferable to also install the jxl_profiler's headers when the library is install? Assuming the headers should be installed in `<prefix>/include/jxl`, this would consist in the following statements:
```cmake
# file: lib/jxl_profiler.cmake
target_include_directories(jxl_profiler PUBLIC "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
install(FILES 
  "${CMAKE_CURRENT_LIST_DIR}/profiler/profiler.h"
  "${CMAKE_CURRENT_LIST_DIR}/profiler/tsc_timer.h"
  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/jxl")
```

amyspark:
> The caveat is that no header is installed for this target and is just there as a usage dependency of jxl-static.

Unsure if this target is meant to be consumed externally (as opposed to merely linked), we'd need a maintainer input here?

FtZPetruska:
> > The caveat is that no header is installed for this target and is just there as a usage dependency of jxl-static.
> 
> Unsure if this target is meant to be consumed externally (as opposed to merely linked), we'd need a maintainer input here?

Got it, I appreciate the feedback 👍 

For now, I rebased on main, applied the namespace changes you suggested, fixed an issue where TCmalloc was not searched for in the installed config. And for the time being, only install the profiler library without its headers, so at the very least the proposed change is functional.

FtZPetruska:
CI failure:

```
[0/1] Install the project...
-- Install configuration: "Release"
-- Installing: /usr/local/lib/libhwy.a
CMake Error at third_party/cmake_install.cmake:50 (file):
  file INSTALL cannot find
  "/home/runner/work/libjxl/libjxl/build/third_party/liblcms2.a": No such
  file or directory.
```

This was because LCMS2 may be enabled even when libjxl does not use it. So we just guard the installation to only happen when it is explicitly needed:

```
if(NOT JPEGXL_ENABLE_SKCMS)
  list(APPEND VENDORED_DEPENDENCIES lcms2)
endif()
```

FtZPetruska:
Rebased on main.

Also since https://github.com/libjxl/libjxl/commit/6c76806eb812299e33cf98b0e2f4dba83c7deb64, Highway exports its targets even when vendored, so it no longer needs to be exported manually in `JPEGXLDeps::`.

FtZPetruska:
I am going to close this PR for now. This branch is getting too out of sync with `main` and I no longer have the time to work on it.

Moreover, there would need to be other changes to the current install logic (handling sanitizers, etc) for this PR to work better.

PR #2248, opened 2023-03-01
MIME plugins not necessary anymore

mo271:


plugins/mime/README.md
```diff
@@ -1,3 +1,20 @@
+## :warning: Not needed anymore
+
+As `image/jxl` is now supported by [shared-mine-info 2.2](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/releases/2.2), it should not be necessary anymore to install this plugin.
+
+You can test if your system correctly understand the MIME type of JPEG XL image by obtaining an JPEG XL image, e.g. with
+```bash
+wget https://github.com/libjxl/conformance/blob/master/testcases/bicycles/input.jxl
```
szabadka, line 7: I tried it and this URL gets you a HTML page, not a jxl file.

To make it work, there needs to be "?raw=true" at the end of the url and then the flag -O input.jxl needs to be added, so the final command-line is:wget https://github.com/libjxl/conformance/blob/master/testcases/bicycles/input.jxl?raw=true -O input.jxl

mo271: good point
changed it to 
https://raw.githubusercontent.com/libjxl/conformance/master/testcases/bicycles/input.jxl
which also works...

szabadka (approved):

plugins/mime/README.md
```diff
@@ -1,3 +1,20 @@
+## :warning: Not needed anymore
+
+As `image/jxl` is now supported by [shared-mine-info 2.2](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/releases/2.2), it should not be necessary anymore to install this plugin.
```
ziemek99, line 3: You probably meant `shared-mime-info 2.2`...

veluca93: That'd make for an interesting package name :)

kmilos:
Why not default it OFF as well?

PR #2250, opened 2023-03-01
Fix coverage build

eustas:


lib/CMakeLists.txt
```diff
@@ -124,11 +124,16 @@ else()  # WIN32
 endif()  # WIN32
 
 # Internal flags for coverage builds:
+set(JPEGXL_COVERAGE_FLAGS)
```
mo271, line 127: `JPEGXL_COVERAGE_FLAGS`

is this flag used anywhere?

eustas: Not really. This seems to be a bug in clang - sometimes it picks wrong compiler-rt, when not asked explicitly.

mo271 (approved):

PR #2259, opened 2023-03-03
disable installation of MIME file

### LINKED ISSUE #2183: Installation of plugin mime and thumbnailer files doesn't respect DATAROOTDIR
If the build is done on a Linux distribution where the DATADIR is in a location other than `${CMAKE_INSTALL_PREFIX}/share`, then the data files for plugins - the mime `packages/image-jxl.xml` file, and thumbnailers `jxl.thumbnailer` file in particular - will be installed in the wrong location.

An example is in [Exherbo](https://www.exherbo.org/), where our standard cmake path setup uses `-DCMAKE_INSTALL_PREFIX:PATH=/usr/x86_64-pc-linux-gnu -DCMAKE_INSTALL_DATAROOTDIR:PATH=/usr/share/`. The data files get installed to `/usr/x86_64-pc-linux-gnu/share/` instead of the desired `/usr/share/`.

The fix is that these files should be installed using the [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs) `${CMAKE_INSTALL_DATADIR}` variable instead of a hardcoded `share` path.

mo271:
fixes #2183

### DIFF
--- a/debian/libjxl-gdk-pixbuf.install
+++ b/debian/libjxl-gdk-pixbuf.install
@@ -1,3 +1,2 @@
 usr/lib/*/gdk-pixbuf-*/*/loaders/*
-usr/share/mime/packages/image-jxl.xml
 usr/share/thumbnailers/jxl.thumbnailer
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -15,7 +15,4 @@ if(JPEGXL_ENABLE_PLUGIN_GIMP210)
   add_subdirectory(gimp)
 endif()
 
-option(JPEGXL_ENABLE_PLUGIN_MIME "Enable image/jxl declaration for shared-mime-info" ON)
-if(JPEGXL_ENABLE_PLUGIN_MIME)
-  add_subdirectory(mime)
-endif()
+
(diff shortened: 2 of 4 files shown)


andrius4669:
depends on distro, eg https://packages.debian.org/bullseye/shared-mime-info and https://packages.ubuntu.com/jammy/shared-mime-info

kmilos:
> depends on distro, eg https://packages.debian.org/bullseye/shared-mime-info and https://packages.ubuntu.com/jammy/shared-mime-info

libjxl is not available (and is unlikely to be) on bullseye nor jammy, so no problem?

andrius4669:
you can still build it on these systems

kmilos:
The readme has instructions how to add it manually.

kmilos:
I guess the alternative would be to default the option to OFF and have it enabled explicitly in deb recipes for these older systems...

mo271:
In light of the issue mentioned (#2183), I think it would be best to merge this as is.

veluca93 (approved):

eustas (approved):

PR #2263, opened 2023-03-03
Fix Is444, Is420, Is440, Is422.

veluca93:
Previous implementation was wrong in many ways (i.e. taking 420 frames to be 422).

This version is more obviously correct.


lib/jxl/frame_header.h
```diff
@@ -115,37 +115,32 @@ struct YCbCrChromaSubsampling : public Fields {
...
 
   bool Is444() const {
-    for (size_t c : {0, 2}) {
-      if (channel_mode_[c] != channel_mode_[1]) {
-        return false;
-      }
-    }
-    return true;
+    return HShift(0) == 0 && VShift(0) == 0 && HShift(2) == 0 &&
+           VShift(2) == 0 && HShift(1) == 0 && VShift(1) == 0;
```
jonsneyers, line 119:
```suggestion
//     Cb                Y                 Cr
return HShift(0) == 0 && HShift(1) == 0 && HShift(2) == 0 &&
       VShift(0) == 0 && VShift(1) == 0 && VShift(2) == 0;
```

lib/jxl/frame_header.h
```diff
@@ -115,37 +115,36 @@ struct YCbCrChromaSubsampling : public Fields {
...
+  }
+
+  std::string DebugString() const {
+    if (Is444()) return "444";
+    if (Is420()) return "420";
+    if (Is422()) return "422";
+    if (Is440()) return "440";
+    return "cs" + std::to_string(channel_mode_[0]) +
+           std::to_string(channel_mode_[1]) + std::to_string(channel_mode_[2]);
   }
```
sboukortt, line 148: Are we going to use this / are we somehow already?

sboukortt (approved):

lib/jxl/frame_header.h
veluca93: mostly it's just useful for debugging :)

PR #2280, opened 2023-03-08
GitHub Action to lint Python code

cclauss:


mo271:
Thanks for the contribution! We only have very little Python code, so I don't know if it it worth running the linter in the CI. Perhaps you could remove the addition of the workflow file, while keeping the two lint edits or at least change the workflow so that it only runs when changes to `.py` files are made.
Also, please edit the AUTHORS file to include your name

PR #2300, opened 2023-03-15
Support progress monitor callback in the jpegli encoder.

szabadka:


lib/jpegli/bitstream.cc
```diff
@@ -740,6 +740,16 @@ static JXL_INLINE void EmitMarker(JpegBitWriter* bw, int marker) {
   bw->data[bw->pos++] = marker;
 }
 
+void ProgressMonitorEncodePass(j_compress_ptr cinfo, size_t scan_index,
+                               size_t mcu_y) {
+  if (cinfo->progress != nullptr) {
```
eustas, line 745: nit: would be more readable with "early return", i.e. if nullptr, then return, and no else branch (here and 3 more places)

eustas (approved):

lib/jpegli/bitstream.cc
szabadka: Done.

eustas (approved):

PR #2302, opened 2023-03-16
Break up DecodeAPITest into smaller tests, refactor test_util

szabadka:


lib/jpegli/decode_api_test.cc
```diff
@@ -109,434 +64,229 @@ class ChunkedSourceManager : public SourceManager {
...
+    if (dparams.out_color_space == JCS_UNKNOWN) {
+      cinfo->jpeg_color_space = JCS_UNKNOWN;
+    }
   }
+}
 
-  bool LoadNextChunk() {
-    if (pos_ >= len_) {
-      return false;
+void ReadOutputImage(const DecompressParams& dparams, j_decompress_ptr cinfo,
```
eustas, line 86: Perhaps `ReadCroppedOutputImage`?
There is another `ReadOutputImage` that looks much shorter...

eustas (dismissed):

lib/jpegli/decode_api_test.cc
szabadka: This is sort-of intentional, all of the decoding tests have a ReadOutputImage that uses the decompress parameters relevant for that specific test. It is there to deduplicate this code for the different test cases within the test.cc file.

mo271:
should this be tried with CI:full?

PR #2303, opened 2023-03-16
Fix Clang32 build

eustas:


mo271 (approved):

eustas:
Seems to work now. PTAL

PR #2310, opened 2023-03-17
Run robot overnight

eustas:


.github/workflows/test_new_highway.yml
```diff
@@ -7,7 +7,9 @@
 
 name: Update branch with updated submodule on libjxl
 
-on: repository_dispatch
+on:
+  schedule:
+    - cron: '37 2 * * *' # Daily on 02:37
```
mo271, line 12:
```suggestion
    - cron: '37 2 * * *' # Daily on 02:37 UTC
```
That is the time zone, right?

.github/workflows/test_new_highway.yml
```diff
@@ -7,7 +7,9 @@
 
 name: Update branch with updated submodule on libjxl
 
-on: repository_dispatch
+on:
```
mo271, line 10:
```suggestion
on:
  workflow_dispatch:
```
Let's also add an easy way to trigger it manually..

mo271:
fyi: I made the corresponding change to highway to remove the workflow triggering libjxl there.

.github/workflows/test_new_highway.yml
eustas: Correct.

.github/workflows/test_new_highway.yml
eustas: 👍

.github/workflows/test_new_highway.yml
veluca93: Isn't it better to do something like 3:30AM? it's more comfortably away from the summer time change :P

mo271 (approved):

PR #2319, opened 2023-03-20
Use the same {width,height}_in_block definition as in libjpeg.

szabadka:
drive-by: fix msan failure.


mo271 (approved):
Great drive-by fix!

PR #2338, opened 2023-03-28
Better default output color space if the image has an ICC profile

sboukortt:


lib/extras/dec/jxl.cc
```diff
@@ -372,6 +372,30 @@ bool DecodeImageJXL(const uint8_t* bytes, size_t bytes_size,
...
+          const bool float_output = format.data_type == JXL_TYPE_FLOAT ||
+                                    format.data_type == JXL_TYPE_FLOAT16;
+          color_encoding.transfer_function = float_output
+                                                 ? JXL_TRANSFER_FUNCTION_LINEAR
+                                                 : JXL_TRANSFER_FUNCTION_SRGB;
+          color_encoding.rendering_intent = JXL_RENDERING_INTENT_RELATIVE;
+          color_encoding.primaries = float_output ? JXL_PRIMARIES_SRGB
+                                     : format.data_type == JXL_TYPE_UINT8
+                                         ? JXL_PRIMARIES_P3
+                                         : JXL_PRIMARIES_2100;
```
kmilos, line 391: Though valid under CICP/H.273, the Rec2020/2100 primaries + sRGB transfer function is a bit of strange combination, and I doubt many 3rd party SW would understand/implement it (as e.g. built-in ICC profiles to map to). Rec2020 goes w/ Rec709 curve usually (if not Rec2100 w/ HDR curves), [but only for encoding, not for decoding.](https://en.wikipedia.org/wiki/Rec._2020#Transfer_characteristics)

sboukortt: This code would mainly be used by `djxl` (it is not part of libjxl itself) which, when outputting to PNG, generates an appropriate ICC profile to go with it. (And a cICP chunk.)

The Rec. 709 TRC would be more likely to create inconsistencies – Chrome just replaces it with the sRGB one: https://source.chromium.org/chromium/chromium/src/+/main:ui/gfx/color_space.cc;l=911-925;drc=c69b50ac19eba2d9f9e3ef405ee8121c0b23c4c1

kmilos: > This code would mainly be used by `djxl` (it is not part of libjxl itself)

Ah, makes sense then, thanks.

PR #2340, opened 2023-03-29
include cmath

mo271:


mo271:
attempting to fix https://github.com/libjxl/libjxl/actions/runs/4549757262/jobs/8022138395?pr=2269

eustas:
Yeah, that happens after partial includes sanation. Still thinking of making a great raid to make all includes right and straight...

Issue #2260, opened 2023-03-03
Compilation errror enc_fast_lossless.cc


kpeo:
**Describe the bug**
Compilation error at enc_fast_lossless.cc, preventing installation with brew on MacOS 10.13.6.

[ 55%] Building CXX object lib/CMakeFiles/jxl_enc-obj.dir/jxl/enc_noise.cc.o
[...]

PR #2383, opened 2023-04-15
Check version of Apple clang correctly

jmroot:
The `__clang_major__` macro provides a marketing version which is not required to be consistent between vendors[[1]]. And indeed, Apple uses a versioning scheme for its Clang releases that is unrelated to that of the llvm.org releases on which they are based. The correspondences between the versions can however be determined[[2]].

So, checking whether `__apple_build_version__` is defined is necessary to determine whether an Apple version of Clang is being used, so that the correct version check can be done. Checking the value of that macro is simpler for the purposes of this code than checking (a different value of) `__clang_major__`, because distinguishing between the relevant versions of Apple Clang would also require checking `__clang_minor__` and `__clang_patchlevel__`.

Closes: #2260

[1]: https://clang.llvm.org/docs/LanguageExtensions.html#builtin-macros
[2]: https://en.wikipedia.org/wiki/Xcode#Toolchain_versions

### DIFF
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -52,8 +52,9 @@ _mm512_cvtsi512_si32(__m512i __A) {
 
 #ifndef FJXL_ENABLE_AVX512
 // On clang-7 or earlier, and gcc-10 or earlier, AVX512 seems broken.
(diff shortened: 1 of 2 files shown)


mo271:
Thanks! I fix the format and rebased so that the AUTHORS check also works automatically...

mo271 (approved):

PR #2401, opened 2023-04-19
remove constructor for HuffmanTableEntry

mo271:
This fixes -Werror=class-memaccess an error introduced in https://github.com/libjxl/libjxl/commit/1112ef0e263600f9bc7f0cbc735fcd0d004af33d
assuming the constructor is not used anymore

### DIFF
--- a/lib/jpegli/huffman.h
+++ b/lib/jpegli/huffman.h
@@ -21,10 +21,6 @@ constexpr int kJpegHuffmanRootTableBits = 8;
 constexpr int kJpegHuffmanLutSize = 758;
 
 struct HuffmanTableEntry {
-  // Initialize the value to an invalid symbol so that we can recognize it
-  // when reading the bit stream using a Huffman code with space > 0.
-  HuffmanTableEntry() : bits(0), value(0xffff) {}
-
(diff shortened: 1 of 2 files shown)


mo271:
Was failing here:
https://github.com/libjxl/libjxl/actions/runs/4738991172/jobs/8413410139?pr=2269

lib/jpegli/huffman.h
```diff
@@ -21,10 +21,6 @@ constexpr int kJpegHuffmanRootTableBits = 8;
 constexpr int kJpegHuffmanLutSize = 758;
 
 struct HuffmanTableEntry {
-  // Initialize the value to an invalid symbol so that we can recognize it
-  // when reading the bit stream using a Huffman code with space > 0.
-  HuffmanTableEntry() : bits(0), value(0xffff) {}
```
szabadka, line 26: Please set this value after the zero-initialization in decode.cc (e.g.  with simple for loop over all kAllHufLutSize entries)

mo271: done (if I understood correctly..)
I guess the `bits` don't have to be set to zero again

szabadka (approved):

PR #2402, opened 2023-04-19
fix apt issues

mo271:


mo271:
seems like the servers are reachable again without any fix...

PR #2427, opened 2023-04-26
Add jpegli decoder fuzzer and corpus generator.

szabadka:


szabadka:
CIFuzz will be fixed with a PR to google/oss-fuzz after this is submitted.

mo271 (approved):

PR #2428, opened 2023-04-26
splines: trigger limits earlier

mo271:
This makes sure that we first hit the `total_estimated_area_reached`, before calculating intermediate points and `points_to_draw`.

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -607,8 +607,7 @@ Status Splines::InitializeDrawCache(const size_t image_xsize,
     }
     splines.push_back(spline);
   }
-  for (Spline& spline: splines) {
-    Spline spline = splines[i];
+  for (Spline& spline : splines) {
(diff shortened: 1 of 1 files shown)


lib/jxl/splines.cc
```diff
@@ -589,6 +590,10 @@ Status Splines::InitializeDrawCache(const size_t image_xsize,
       return JXL_FAILURE(
           "identical successive control points in spline %" PRIuS, i);
     }
+    splines.push_back(spline);
+  }
+  for (size_t i = 0; i < splines_.size(); ++i) {
```
jonsneyers, line 595: Could also write it like this:

```suggestion
  for (Spline& spline: splines) {
```

mo271: sure!

jonsneyers (approved):

jonsneyers (approved):

PR #2443, opened 2023-04-28
add devtools to the artifacts

jonsneyers:
This should add more tools (such as butteraugli_main and ssimulacra2) to the Windows and Linux binaries in the CI artifacts and future releases.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -73,12 +73,13 @@ jobs:
           -DJPEGXL_ENABLE_VIEWERS=OFF \
           -DJPEGXL_ENABLE_PLUGINS=OFF \
           -DJPEGXL_ENABLE_OPENEXR=OFF \
+          -DJPEGXL_ENABLE_DEVTOOLS=ON \
 
     - name: Package release tarball
       run: |
         cd build
         tar -zcvf ${{ runner.workspace }}/release_file.tar.gz \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -73,12 +73,13 @@ jobs:
...
           -DJPEGXL_ENABLE_PLUGINS=OFF \
           -DJPEGXL_ENABLE_OPENEXR=OFF \
+          -DJPEGXL_ENABLE_DEVTOOLS=ON \
 
     - name: Package release tarball
       run: |
         cd build
         tar -zcvf ${{ runner.workspace }}/release_file.tar.gz \
-          LICENSE* tools/{cjxl,djxl,benchmark_xl}
+          LICENSE* tools/{cjxl,djxl,benchmark_xl,cjpegli,djpegli,jxlinfo,butteraugli_main,ssimulacra2}
```
sboukortt, line 82: Might some of the HDR tools (`pq_to_hlg`, `render_hlg`, `tone_map`, perhaps even `generate_lut_template` and `texture_to_cube`) make sense to add as well while we’re at it?

sboukortt (approved):

.github/workflows/release.yaml
jonsneyers: Sure, feel free to add those. Going to merge this one like it is just to see if conformance is OK now.

Issue #2391, opened 2023-04-17
pfm is not supported with cjxl: Only JXL_BIT_DEPTH_FROM_PIXEL_FORMAT is implemented for float types.


zhou13:
**Describe the bug**

I am trying to reproduce #2363 with cjxl, in which I need to convert a float32 buffer to jxl. Since OpenEXR is not correct supported anymore #1804, I tried pfm.

**To Reproduce**

I am using the following code to generate pfm files:
```
import cv2
import numpy as np

buffer = np.load("1.npy")
cv2.imwrite("1.pfm", buffer)
```

Running cjxl gives
```
$ cjxl 1.pfm 1.jxl
JPEG XL encoder v0.8.1 0.8.1 [AVX2,SSE4,SSSE3]
Read 1536x1536 image, 9437200 bytes, 568.5 MP/s
Encoding [VarDCT, d1.000, effort: 7], 
./lib/jxl/encode.cc:265: Only JXL_BIT_DEPTH_FROM_PIXEL_FORMAT is implemented for float types.
./lib/jxl/encode.cc:1848: Invalid input bit depth
JxlEncoderAddImageFrame() failed.
EncodeImageJXL() failed.
```

Example files are attached here:
[Archive.zip](https://github.com/libjxl/libjxl/files/11255630/Archive.zip)

**Expected behavior**
No Error.

**Environment**
 - OS: MacOS / Homebrew
 - CPU type: x86_64
 - cjxl/djxl version string: `cjxl v0.8.1 0.8.1 [AVX2,SSE4,SSSE3]`

**Additional context**

jonsneyers:
Looks like pfm indeed stopped working as an input format for cjxl.

Issue #430, opened 2021-08-11
Accept JPEG XL files as input to cjxl


Fraetor:
**Is your feature request related to a problem? Please describe.**
As JPEG XL supports bunch of stuff that makes it a great intermediate format, it would make sense if it was supported by cjxl for recompression. This would be especially useful for testing different compression levels, and particularly for testing generation loss.

**Describe the solution you'd like**
cjxl able to read in .jxl files. I think it would have to be read in as pixels for the lossy compression, but I may well be incorrect there, and something fancy (as is done for JPEG) could be done to make recompression quicker.

**Describe alternatives you've considered**
They can currently be converted to png by djxl then used in cjxl. While this works, it doesn't support images above the capabilities of PNG (multi layer / animated), and adds an extra step.

**Additional context**
It seems a bit silly that the encoder's own format isn't supported, especially if you are trying to encourage other programs to include JPEG XL support.

Fixed by #2444

Joachim-Otahal:
I second that request! After running cjxl on large collections (-d 0 -e 8) it sometimes depends whether -P 0 and / or -g 3 was included. If I could run it on already encoded .jxl files to check whether it could improve with those options it would save me getting the old original .gif etc back from the snapshot to rerun for optimization tests.
IMO somewhat related to https://github.com/libjxl/libjxl/issues/727 and https://github.com/libjxl/libjxl/issues/426

mo271:
also see https://github.com/libjxl/libjxl/issues/871

Issue #2193, opened 2023-02-17
cjxl can't read JXL it has just generated


PontiacCZ:
**Describe the bug**
I have converted a PNG into a lossless JXL. When I try to convert this JXL into a lossy one I get an error:
`Getting pixel data failed.`

**To Reproduce**
1. save the attached image and name it "logs.png"
2. run `cjxl logs.png logs_0.0.jxl -d 0.0` - you get lossless JXL (it is viewable via XnView on my system)
3. run `cjxl logs_0.0.jxl logs_1.0.jxl -d 1.0` - you get error: **Getting pixel data failed.**

**Expected behavior**
I expected no read error. :-)

**Screenshots**
If applicable, add screenshots or example input/output images to help explain your problem.

**Environment**
 - OS: Windows 10 Pro 22H2 x64
 - CPU type: AMD Ryzen 5 PRO 3500U
 - cjxl/djxl version string: JPEG XL encoder v0.8.1 c27d499 [AVX2,SSE4,SSSE3,Unknown]

![logs](https://user-images.githubusercontent.com/9153595/219710017-3e164cb8-cb3c-415a-9bfb-0ed46c1395a4.png)

mo271:
Thanks for the report. 
Regarding the issue with `cjxl`; currently reading of jxl files is not supported, running
`cjxl logs.png logs_1.0.jxl -d 1.0` should give the expected results.

see also https://github.com/libjxl/libjxl/issues/430

Conaclos:
> Regarding the issue with cjxl; currently reading of jxl files is not supported

In this case, how can we losslessly recreate a `jpeg` from a `jxl`?

TPS:
> In this case, how can we losslessly recreate a `jpeg` from a `jxl`?

That's a `djxl` option.

Traneptora:
Is this still an issue?

jonnyawsom3:
Works now

MasterInQuestion:
Try Pipe input from FFmpeg alike?
    [ https://github.com/libjxl/libjxl/issues/625#issuecomment-2460663870](https://github.com/libjxl/libjxl/issues/625#issuecomment-2460663870)

jonnyawsom3:
As Traneptora said, please open a separate issue for the failing PNGs, ideally with a sample we can test with.

MasterInQuestion:
For alike highly similar issues:
    Creating much duplicate tends to only bring bad segmentation.

[...]

PR #2444, opened 2023-04-28
cjxl: fix pfm input and allow jxl input

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/2391 and https://github.com/libjxl/libjxl/issues/2193

Allowing jxl input for cjxl can be useful if the input is lossless or higher quality than what is desired for the output.

This change also makes it possible to run butteraugli_main or ssimulacra2 directly on jxl files, which is both convenient and slightly more accurate than running it on jxl files decoded to png.


lib/extras/dec/decode.h
```diff
@@ -32,7 +32,8 @@ enum class Codec : uint32_t {
   kPGX,
   kJPG,
   kGIF,
-  kEXR
+  kEXR,
+  kJXL
```
Traneptora, line 36: Doesn't C++ support final trailing commas? I know C does.

jonsneyers: Yes it does, not sure what clang-format thinks about it though

kmilos: No problem AFAIK

Fraetor:
This also fixes #430

mo271 (approved):

PR #2445, opened 2023-04-29
Port the GUI tools to Qt6

sboukortt:


lib/jxl/frame_header.cc
veluca93, line 1: What do you think of splitting these changes in a separate commit/pr?

tools/icc_detect/icc_detect_x11.cc
```diff
@@ -3,16 +3,21 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// clang-format off
```
veluca93, line 6: I am not sure I want to know why this is needed...

veluca93 (approved):

lib/jxl/frame_header.cc
sboukortt: Done, thanks.

tools/icc_detect/icc_detect_x11.cc
sboukortt: Since the base name of the `.cc` is not quite the same as that of the `.h` (because of the `_x11`, `_win32` suffixes), clang-format thinks this is a “normal” header and wants to move it down.

veluca93 (approved):

tools/icc_detect/icc_detect_x11.cc
sboukortt: At least that is the explanation for the first one. The second one is because of that macro business in `X11/Xlib.h`…

PR #2453, opened 2023-05-02
Tool for local tone mapping

sboukortt:


tools/hdr/local_tone_map.cc
```diff
@@ -0,0 +1,378 @@
...
+  for (size_t y = 0; y < image.ysize(); ++y) {
+    const float* const JXL_RESTRICT row = laplacian.ConstRow(y);
+    for (size_t x = 0; x < image.xsize(); ++x) {
+      consumer(x, y, std::abs(row[x]));
+    }
+  }
+}
+
+template <typename Consumer>
+void Saturation(const jxl::Image3F& image, const Consumer& consumer) {
```
veluca93, line 59: stddev(r/g/b - average-rgb)?

This seems like a pretty primitive way to compute saturation - can it be improved?

tools/hdr/local_tone_map.cc
```diff
@@ -0,0 +1,378 @@
...
+                      const float midtoneness_sigma, ThreadPool* const pool) {
+  ImageF weights(image.xsize(), image.ysize());
+  FillImage(1.f, &weights);
+
+  if (contrast_weight > 0) {
+    Contrast(
+        image,
+        [&weights, contrast_weight](const size_t x, const size_t y,
+                                    const float weight) {
+          weights.Row(y)[x] = std::pow(weight, contrast_weight);
```
veluca93, line 105: A small optimization (also slightly numerically more sound) would be to fill with 0, add log(weight) * contrast_weight, and then exp in the end.

tools/hdr/local_tone_map.cc
```diff
@@ -0,0 +1,378 @@
...
+    if (image.xsize() != images.front().xsize() ||
+        image.ysize() != images.front().ysize()) {
+      return {};
+    }
+    weights.push_back(ComputeWeights(image, contrast_weight, saturation_weight,
+                                     midtoneness_weight, midtoneness_sigma,
+                                     pool));
+  }
+
+  std::vector<float*> rows(images.size());
```
veluca93, line 146: I'm not quite sure how this normalization is happening...

tools/hdr/local_tone_map.cc
jonsneyers: I would suggest something like the distance from origin in the Lab (a,b) plane or the XYB (scaledX,B-Y) plane, if that's not too expensive.

tools/hdr/local_tone_map.cc
```diff
@@ -0,0 +1,378 @@
...
+Image3F ExposureFusion(std::vector<Image3F> images, int num_levels,
+                       const float contrast_weight,
+                       const float saturation_weight,
+                       const float midtoneness_weight,
+                       const float midtoneness_sigma, ThreadPool* const pool) {
+  std::vector<ImageF> weights =
+      ComputeWeights(images, contrast_weight, saturation_weight,
+                     midtoneness_weight, midtoneness_sigma, pool);
+
+  std::vector<Image3F> pyramid(num_levels);
```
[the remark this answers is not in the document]
sboukortt: Is a link to the paper enough or should we rephrase what it does?

tools/hdr/local_tone_map.cc
```diff
@@ -0,0 +1,378 @@
...
+namespace {
+
+constexpr WeightsSeparable5 kPyramidFilter = {
+    {HWY_REP4(.375f), HWY_REP4(.25f), HWY_REP4(.0625f)},
+    {HWY_REP4(.375f), HWY_REP4(.25f), HWY_REP4(.0625f)}};
+
+// Expects sRGB input.
+// Will call consumer(x, y, contrast) for each pixel.
+template <typename Consumer>
+void Contrast(const jxl::Image3F& image, const Consumer& consumer,
```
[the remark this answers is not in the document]
sboukortt: Ah, and the callback was so that it wouldn’t have to construct an intermediary `ImageF` – but since it’s always used to add (a weighted log) to an `ImageF`, I suppose it could take a pointer to it and do the calculation directly?

tools/hdr/local_tone_map.cc
veluca93: sounds good as long as there is a todo :)

tools/hdr/local_tone_map.cc
veluca93: Can we write this in a comment? :)

tools/hdr/local_tone_map.cc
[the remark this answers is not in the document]
veluca93: if you can describe it in a paragraph it would be best

tools/hdr/local_tone_map.cc
sboukortt: Oops, I resolved and did the change locally but did not push. One moment…

tools/hdr/local_tone_map.cc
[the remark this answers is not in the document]
veluca93: that might be a good idea, but I wouldn't worry too much about it :)

tools/hdr/local_tone_map.cc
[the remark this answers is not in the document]
sboukortt: I gave it a try.

veluca93 (approved):

PR #2454, opened 2023-05-03
Add python script to easily apply simplex-search results.

szabadka:


tools/optimizer/apply_simplex.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+"""apply_simplex.py: Updates constants based on results of simplex search.
```
jyrkialakuijala, line 7: write a 25-line "users' manual" here how to use it and why

jyrkialakuijala (approved):

tools/optimizer/apply_simplex.py
szabadka: Done.

mo271 (approved):

PR #2455, opened 2023-05-04
give scipy.optimize a try

mo271:


jyrkialakuijala:
it would be even better if it was in a separate script, but this is also ok

jyrkialakuijala (approved):

PR #2466, opened 2023-05-10
fix frame_settingsTest

mo271:


mo271:
failing here for example: https://github.com/libjxl/libjxl/actions/runs/4927899421/jobs/8807677383

mo271:
I having some flexibility in the test was needed after https://github.com/libjxl/libjxl/pull/1618, and apparently the amount of tolerance we have right now leads to flaky test behaviour, hence the update of the tolerance.

veluca93 (approved):
I'm a bit scared of the implications of this being necessary but sure...

mo271:
> I'm a bit scared of the implications of this being necessary but sure...

Yeah, it is a bit peculiar, I don't have a better explanation...

PR #2491, opened 2023-05-22
PNG encoding: write sRGB, cHRM, gAMA chunks when appropriate

sboukortt:


lib/extras/enc/apng.cc
```diff
@@ -195,6 +196,86 @@ void MaybeAddCICP(JxlColorEncoding c_enc, png_structp png_ptr,
   png_set_unknown_chunks(png_ptr, info_ptr, &cicp_chunk, 1);
 }
 
+bool MaybeAddSRGB(JxlColorEncoding c_enc, png_structp png_ptr,
```
Traneptora, line 199: JxlColorEncoding is a struct, would we prefer to pass a pointer to it?

lib/extras/enc/apng.cc
```diff
@@ -195,6 +196,86 @@ void MaybeAddCICP(JxlColorEncoding c_enc, png_structp png_ptr,
...
+      wy = 32900;
+      break;
+    case JXL_WHITE_POINT_E:
+      wx = wy = 33333;
+      break;
+    case JXL_WHITE_POINT_DCI:
+      wx = 31400;
+      wy = 35100;
+      break;
+    case JXL_WHITE_POINT_CUSTOM:
```
Traneptora, line 230: I believe libjxl populates white_point_xy even if an enum value is used, although I'm not entirely sure. Double check

lib/extras/enc/apng.cc
```diff
@@ -195,6 +196,86 @@ void MaybeAddCICP(JxlColorEncoding c_enc, png_structp png_ptr,
...
+      break;
+    case JXL_PRIMARIES_2100:
+      png_set_cHRM_fixed(png_ptr, info_ptr, wx, wy, 70800, 29200, 17000, 79700,
+                         13100, 4600);
+      break;
+    case JXL_PRIMARIES_P3:
+      png_set_cHRM_fixed(png_ptr, info_ptr, wx, wy, 68000, 32000, 26500, 69000,
+                         15000, 6000);
+      break;
+    case JXL_PRIMARIES_CUSTOM:
```
Traneptora, line 249: Same comment as with white_point

Traneptora:
This causes the jpegli tools test to fail, I'm guessing it's comparing output, which is now different. Would we have to update the test as well?

lib/extras/enc/apng.cc
sboukortt: That’s a good point, thank you.

lib/extras/enc/apng.cc
sboukortt: Thanks, it does seem to be the case. The resulting cHRM chunks seem to have very slightly less precision, which bugs me a little, but likely not even close to an extent at which it matters.

sboukortt:
> This causes the jpegli tools test to fail, I'm guessing it's comparing output, which is now different. Would we have to update the test as well?

That sounds plausible. I’ll look into it.

lib/extras/enc/apng.cc
```diff
@@ -195,6 +195,49 @@ void MaybeAddCICP(JxlColorEncoding c_enc, png_structp png_ptr,
...
+bool MaybeAddSRGB(const JxlColorEncoding& c_enc, png_structp png_ptr,
+                  png_infop info_ptr) {
+  if (c_enc.color_space == JXL_COLOR_SPACE_RGB &&
+      c_enc.primaries == JXL_PRIMARIES_SRGB &&
+      c_enc.white_point == JXL_WHITE_POINT_D65 &&
+      c_enc.transfer_function == JXL_TRANSFER_FUNCTION_SRGB) {
+    png_set_sRGB(png_ptr, info_ptr, c_enc.rendering_intent);
+    png_set_cHRM_fixed(png_ptr, info_ptr, 31270, 32900, 64000, 33000, 30000,
+                       60000, 15000, 6000);
+    png_set_gAMA_fixed(png_ptr, info_ptr, 45455);
```
kmilos, line 207: Please don't store 2.2 gamma as sRGB transfer curve replacement as it leads to ambiguity. See e.g. recent discussion here: https://github.com/AOMediaCodec/libavif/pull/1389

As sRGB chunk is indicated, [both cHRM and gAMA are unnecessary](https://www.w3.org/TR/png/#11addnlcolinfo).

sboukortt: Sorry, I am not sure I see where the discussion references ambiguity as to this situation? The PNG specification that you link to is clear: if present, `sRGB` takes precedence over `gAMA` and `cHRM`, but it still recommends including them for compatibility with decoders that only recognise the latter. That discussion likewise starts with: “If the PNG file contains a gAMA chunk and a cHRM chunk, and does not contain an iCCP chunk or an sRGB chunk”, but these files would.

> It is recommended that a PNG encoder that writes the sRGB chunk also write a [gAMA](https://www.w3.org/TR/png/#11gAMA) chunk (and optionally a [cHRM](https://www.w3.org/TR/png/#11cHRM) chunk) for compatibility with decoders that do not use the sRGB chunk. Only the following values shall be used.

Traneptora: gAMA is a fallback to viewers which don't support sRGB. It's mostly there for legacy reasons as there's *very* few decoders that respect gAMA but don't respect sRGB, but it's also not harmful to have because it's explicitly ignored if sRGB, iCCP, or cICP are present.

kmilos: > it's also not harmful to have because it's explicitly ignored if sRGB, iCCP, or cICP are present

This, of course, assuming all the decoders are fully compliant.

jonsneyers: I suppose technically a compliant png decoder can ignore any chunk that has a name starting with a lowercase letter. In practice, I suppose the following types of png applications exist and are common enough to worry about them:

- ignores all colorspace info and treats everything as sRGB
- ignores all colorspace info and just sends pixel values to the screen (whatever the screen display space is)
- ignores anything except gAMA and cHRM
- ignores anything except iCCP
- understands everything except cICP but silently assumes gAMA+cHRM set to approximate sRGB (but without sRGB chunk and without iCCP) to be identical to sRGB
- understands everything except cICP (and applies precedence rules correctly, and correctly doesn't treat cHRM+gAMA-only as identical to sRGB)
- understands everything (and applies precedence correctly)

Sending all colorspace chunks all of the time seems to be the best approach to get things as right as possible as often as possible...

sboukortt:
> This causes the jpegli tools test to fail, I'm guessing it's comparing output, which is now different. Would we have to update the test as well?

It should be fixed now. Apparently, calling `png_set_cHRM` with invalid (all-zero) chromaticities makes libpng silently drop the previously-set iCCP chunk as well…

lib/extras/enc/apng.cc
sboukortt: It seems that libpng won’t write both `sRGB` and `iCCP`:

> ```
> libpng warning: profile matches sRGB but writing iCCP instead
> ```

The [PNG specification](https://w3c.github.io/PNG-spec/#11iCCP) also argues against it:

> Unless a [cICP](https://w3c.github.io/PNG-spec/#cICP-chunk) chunk exists, a PNG datastream should contain at most one embedded profile, whether specified explicitly with an iCCP or implicitly with an [sRGB](https://w3c.github.io/PNG-spec/#srgb-standard-colour-space) chunk.

So it seems that the behavior implemented by the current state of this PR matches the recommendation:

- sRGB: write `sRGB`, `cHRM`, `gAMA` (we _could_ add `cICP` but is there much point?);
- everything else: write `iCCP`, maybe `cICP`, maybe `cHRM`, maybe `gAMA`.

Traneptora: Specification allows `cICP` and `sRGB` and it says `cICP` explicitly overrides `sRGB`, but I don't believe there's any software that supports `cICP` that doesn't support `sRGB`. So if the color space is srgb I imagine that we'd prefer to just not write the `cICP` chunk and only write the `sRGB` chunk and fallback chunks.

PR #2500, opened 2023-05-25
remove docker

mo271:
We remove the docker build for now, since we don't use it any more in our CI.
At some point we might consider bringing it back


ziemek99:
It'd be nice to investigate if this change won't disrupt updating JPEG XL in [Squoosh](https://github.com/GoogleChromeLabs/squoosh).

mo271:
> It'd be nice to investigate if this change won't disrupt updating JPEG XL in [Squoosh](https://github.com/GoogleChromeLabs/squoosh).

I don't think they are using our docker

eustas (approved):

PR #2503, opened 2023-05-25
0.1 % quality improvement

jyrkialakuijala:
reduce quantization when large blocks are used with lots of quantization error, but not large coefficient values and the masking is low

```
Before:

Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm   PSNR   QABPP  SmallB  DCT4x8     AFV  DCT8x8    8x16    8x32      16   16x32      32   32x64      64       BPP*pnorm   Bugs
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
jxl:d0.1        19988 16609463    6.6477416   1.360   9.336   0.35353178  95.48321280   0.11547700  53.20   6.648 42.2069 18.2557  0.4579 27.5521  4.8560  0.0000  2.3809  0.8991  3.9345  0.1025  0.0820  0.767661287801      0
jxl:d0.5        19988  6418235    2.5688228   1.849  15.256   0.78360793  91.47222570   0.34784750  44.76   2.595 14.3282 25.2637  0.5715 12.2537 20.7266  0.0000 20.8162  1.2295  5.1077  0.1434  0.2869  0.893558611311      0
jxl:d0.8        19988  4749001    1.9007316   1.908  15.667   1.09299541  88.76910888   0.48042532  41.66   2.162 10.8798 21.9981  0.7060  9.0476 19.1608  0.0000 29.4831  1.5625  7.1928  0.2049  0.4918  0.913159593145      0
jxl:d1          19988  4097554    1.6399976   1.918  16.132   1.28714387  87.04977497   0.56373878  40.48   2.183  9.6156 20.1582  0.7457  7.8969 17.8032  0.0000 30.8395  2.5000 10.5535  0.1844  0.4303  0.924530273169      0
jxl:d1.1        19988  3858566    1.5443455   2.078  17.331   1.40323481  86.18913181   0.60289883  40.01   2.237  9.1302 19.3792  0.7675  7.4957 17.2929  0.0000 30.5423  3.2762 12.2082  0.2049  0.4303  0.931084108279      0
jxl:d1.2        19988  3643813    1.4583932   1.970  17.008   1.50159330  85.37331677   0.64052088  39.57   2.258  8.6874 18.7363  0.8069  7.1069 16.6313  0.0000 30.1850  4.0703 13.8476  0.1844  0.4713  0.934131264689      0
[...]

PR #2516, opened 2023-06-01
fix PixelTestWithICCProfileLossy

mo271:
caused by #2503, this fixes a problem when cross-compiling


szabadka (commented):
Should PixelTestOpaqueSrgbLossy also be updated?

mo271:
yes it should, let me push a fix for that

PR #2518, opened 2023-06-01
make e1 lossless also use palette for RGB images

jonsneyers:
Lossless e1 only detected palette on RGBA input, for no good reason.
This makes it detect palette also on RGB input.

For some random png8 image from discord:

Before:
```
JPEG XL encoder v0.9.0 c81a5a46 [SSE4,SSSE3,Unknown]
Read 3526x358 image, 627101 bytes, 277.8 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 2800470 bytes including container (17.748 bpp).
3526 x 358, geomean: 206.82 MP/s [72.86, 233.42], 100 reps, 4 threads.
```

After:
```
JPEG XL encoder v0.9.0 c0cfd188 [SSE4,SSSE3,SSE2]
Read 3526x358 image, 627101 bytes, 313.6 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 1194923 bytes including container (7.573 bpp).
3526 x 358, geomean: 65.45 MP/s [44.57, 69.96], 100 reps, 4 threads.
```

### DIFF
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -3414,14 +3414,23 @@ void CollectSamples(const unsigned char* rgba, size_t x0, size_t y0, size_t xs,
 }
 
 void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
-                            const PrefixCode code[4],
+                            size_t nb_chans, const PrefixCode code[4],
                             const std::vector<uint32_t>& palette,
                             size_t pcolors, BitWriter* output) {
   PrepareDCGlobalCommon(is_single_group, width, height, code, output);
   output->Write(2, 0b01);     // 1 transform
   output->Write(2, 0b01);     // Palette
   output->Write(5, 0b00000);  // Starting from ch 0
-  output->Write(2, 0b10);     // 4-channel palette (RGBA)
+  if (nb_chans == 1) {
+    output->Write(2, 0b00);  // 1-channel palette (Gray)
+  } else if (nb_chans == 3) {
+    output->Write(2, 0b01);  // 3-channel palette (RGB)
+  } else if (nb_chans == 4) {
+    output->Write(2, 0b10);  // 4-channel palette (RGBA)
+  } else {
+    output->Write(2, 0b11);
+    output->Write(13, nb_chans - 1);
+  }
   // pcolors <= kMaxColors + kChunkSize - 1
   static_assert(kMaxColors + kChunkSize < 1281,
                 "add code to signal larger palette sizes");
@@ -3471,6 +3480,32 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
   }
 }
 
+template <size_t nb_chans>
+bool detect_palette(const unsigned char* r, size_t width,
+                    std::vector<uint32_t>& palette) {
+  size_t x = 0;
+  bool collided = false;
+  // this is just an unrolling of the next loop
+  for (; x + 7 < width; x += 8) {
+    uint32_t p[8] = {}, index[8];
+    for (int i = 0; i < 8; i++) memcpy(&p[i], r + (x + i) * nb_chans, 4);
+    for (int i = 0; i < 8; i++) p[i] &= ((1llu << (8 * nb_chans)) - 1);
+    for (int i = 0; i < 8; i++) index[i] = pixel_hash(p[i]);
+    for (int i = 0; i < 8; i++) {
+      collided |= (palette[index[i]] != 0 && p[i] != palette[index[i]]);
+    }
+    for (int i = 0; i < 8; i++) palette[index[i]] = p[i];
+  }
+  for (; x < width; x++) {
+    uint32_t p = 0;
+    memcpy(&p, r + x * nb_chans, nb_chans);
+    uint32_t index = pixel_hash(p);
+    collided |= (palette[index] != 0 && p != palette[index]);
+    palette[index] = p;
+  }
+  return collided;
+}
+
 template <typename BitDepth>
 JxlFastLosslessFrameState* LLEnc(const unsigned char* rgba, size_t width,
                                  size_t stride, size_t height,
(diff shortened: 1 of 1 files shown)


veluca93:
This is almost 3x slower, what's the reason for that?

jonsneyers:
> This is almost 3x slower, what's the reason for that?

Probably there's room for speeding it up, but the palette detection is not particularly fast, especially when the pixels are not nicely uint32 like in 8-bit RGBA but annoyingly unaligned like 8-bit RGB. The detection aborts at the first collision though, so for non-palette images the slowdown should be small. And for palette images I would say it is worth it. But if you have ideas on how to speed the palette detection in this case, feel free to try to speed this up :)

veluca93:
> > This is almost 3x slower, what's the reason for that?
> 
> Probably there's room for speeding it up, but the palette detection is not particularly fast, especially when the pixels are not nicely uint32 like in 8-bit RGBA but annoyingly unaligned like 8-bit RGB. The detection aborts at the first collision though, so for non-palette images the slowdown should be small. And for palette images I would say it is worth it. But if you have ideas on how to speed the palette detection in this case, feel free to try to speed this up :)

Can you try loading a uint32 anyway (except in the last few pixels, I guess) and then masking out the invalid bits? :)

jonsneyers:
Tried making autovec doing a better job also in the case of nb_chans < 4.
Also fixed something I overlooked: the palette sorting was using alpha*luma which is not a good idea if alpha is not present and always equal to zero. Fixing that changes the ordering from basically random to something better, improving density.

New results:

Before (same as before of course):
```
JPEG XL encoder v0.9.0 c81a5a46 [SSE4,SSSE3,Unknown]
Read 3526x358 image, 627101 bytes, 376.1 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 2800470 bytes including container (17.748 bpp).
3526 x 358, geomean: 201.23 MP/s [93.69, 269.04], 100 reps, 4 threads.
```

After:
```
JPEG XL encoder v0.9.0 4e31aaa1 [SSE4,SSSE3,SSE2]
Read 3526x358 image, 627101 bytes, 278.3 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 948035 bytes including container (6.008 bpp).
3526 x 358, geomean: 216.46 MP/s [59.34, 265.84], 100 reps, 4 threads.
```

It's still worse than the nb_chans==4 case (which performs the same as before, even though the code is now slightly different), which looks like this:

```
JPEG XL encoder v0.9.0 4e31aaa1 [SSE4,SSSE3,SSE2]
Read 3526x358 image, 2078240 bytes, 86.6 MP/s
Encoding [Modular, lossless, effort: 1], 
Compressed to 947959 bytes (6.008 bpp).
3526 x 358, geomean: 365.15 MP/s [81.69, 498.52], 100 reps, 4 threads.
```

I'm assuming this is because of the benefit of being able to just read 32 bytes at a time as-is, as opposed to reading 24 bytes in a funky way to get it into a uint32_t p[8]. But at least for this image, the speed with palette is now similar to the speed without (maybe even slightly faster, but it's a bit hard to measure because even with 100 repetitions there is quite some fluctuation in the measured speeds), so at least the palette detection cost is compensated by the gains later on. Considering the density improvement from 17.7 bpp to 6 bpp, I would say it's a pretty clear win, at least in this case :)

veluca93:
If you want more consistent timings, you should disable multithreading :)

I am ok with the new timings regardless.

lib/jxl/enc_fast_lossless.cc
```diff
@@ -3471,10 +3480,38 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
...
+    palette[index] = p;
+  }
+  return collided;
+}
+
 template <typename BitDepth>
 JxlFastLosslessFrameState* LLEnc(const unsigned char* rgba, size_t width,
                                  size_t stride, size_t height,
                                  BitDepth bitdepth, size_t nb_chans,
+
```
veluca93, line 3514: Remove?

lib/jxl/enc_fast_lossless.cc
```diff
@@ -3471,10 +3480,38 @@ void PrepareDCGlobalPalette(bool is_single_group, size_t width, size_t height,
...
+template <size_t nb_chans>
+bool detect_palette(const unsigned char* r, size_t width,
+                    std::vector<uint32_t>& palette) {
+  size_t x = 0;
+  bool collided = false;
+  // this is just an unrolling of the next loop
+  for (; x + 7 < width; x += 8) {
+    uint32_t p[8] = {}, index[8];
+    for (int i = 0; i < 8; i++) memcpy(&p[i], r + (x + i) * nb_chans, 4);
+    if (nb_chans < 4)
```
veluca93, line 3492: This is probably unnecessary (the compiler will likely skip a & 0xFFFFFFFF on a 32-bit integer), but if you leave it, please add braces :)

veluca93 (approved):

lib/jxl/enc_fast_lossless.cc
jonsneyers: Right, it's unnecessary.

veluca93 (approved):

jonsneyers:
Timings for single-threaded encoding:

Before:
```
JPEG XL encoder v0.9.0 c81a5a46 [SSE4,SSSE3,Unknown]
Read 3526x358 image, 627101 bytes, 291.7 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 2800470 bytes including container (17.748 bpp).
3526 x 358, geomean: 106.90 MP/s [58.66, 113.01], 100 reps, 0 threads.
```

After:
```
JPEG XL encoder v0.9.0 5604406a [SSE4,SSSE3,SSE2]
Read 3526x358 image, 627101 bytes, 320.1 MP/s
Encoding [Container | Modular, lossless, effort: 1 | 14-byte Exif], 
Compressed to 948035 bytes including container (6.008 bpp).
3526 x 358, geomean: 162.45 MP/s [89.42, 173.53], 100 reps, 0 threads.
```

So here there's a clear speedup, as opposed to in the case with 4 threads, which is probably because the palette detection itself is not multithreaded. (it could probably be made multithreaded but that gets a bit tricky, probably you need to use nb_threads independent hash tables and then merge them and check for collisions that weren't detected yet)

PR #2523, opened 2023-06-02
improve palette ordering and encode speed

jonsneyers:
Use the luma * alpha palette sorting from e1 also at higher effort settings, as opposed to the lexicographic sorting that was there (which kind of makes sense for YCoCg pixels, but not really when there's alpha or when the data is still RGB).
Also use `unordered_set` to keep track of which colors have been seen (doesn't have to be ordered anymore since we're not using that order anymore anyway), and use an `unordered_map` to converts pixels to palette indices (instead of doing a naive linear search).

Net result is an encoder speedup and generally some density improvement, though that's not guaranteed on a per-image basis (palette ordering is weird).

On the jyrki31 corpus, converted to png8 with default pngquant:

Before:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d0:4        13270  7294204    4.3972613   0.678   5.189         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:6        13270  7105322    4.2833951   0.576   5.299         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:7        13270  6944452    4.1864157   0.399   4.992         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:8        13270  6856609    4.1334601   0.213   4.503         -nan 100.00000000   0.00000000  0.000000000000      0
Aggregate:      13270  7048184    4.2489501   0.427   4.986   0.00000000 100.00000000   0.00000000  0.000000000000      0
```

After:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d0:4        13270  7256193    4.3743466   1.116   6.540         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:6        13270  7052290    4.2514251   0.826   6.594         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:7        13270  6880832    4.1480628   0.585   6.223         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:8        13270  6791408    4.0941542   0.287   5.710         -nan 100.00000000   0.00000000  0.000000000000      0
Aggregate:      13270  6992942    4.2156477   0.627   6.257   0.00000000 100.00000000   0.00000000  0.000000000000      0
```

On some manga with alpha (also converted to png8 with pngquant):

Before:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d0:4         4945   527465    0.8532975   1.372   9.456         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:6         4945   496081    0.8025266   0.923   8.585         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:7         4945   450107    0.7281529   0.900   6.199         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:8         4945   422527    0.6835358   0.501   5.941         -nan 100.00000000   0.00000000  0.000000000000      0
Aggregate:       4945   472312    0.7640749   0.869   7.394   0.00000000 100.00000000   0.00000000  0.000000000000      0
```

After:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d0:4         4945   509826    0.8247623   1.813   8.655         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:6         4945   481004    0.7781360   1.107   8.565         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:7         4945   436016    0.7053574   1.035   8.038         -nan 100.00000000   0.00000000  0.000000000000      0
jxl:d0:8         4945   412052    0.6665901   0.545   7.538         -nan 100.00000000   0.00000000  0.000000000000      0
Aggregate:       4945   458149    0.7411619   1.031   8.186   0.00000000 100.00000000   0.00000000  0.000000000000      0
```

So on these images that's something like a 0.5% to 3% density improvement and something like a 10% (or more at the lower effort settings) encode speedup.

### DIFF
--- a/lib/jxl/modular/transform/enc_palette.cc
+++ b/lib/jxl/modular/transform/enc_palette.cc
@@ -268,12 +268,12 @@ Status FwdPaletteIteration(Image &input, uint32_t begin_c, uint32_t end_c,
       begin_c, end_c, nb_colors);
   nb_deltas = 0;
   bool delta_used = false;
-  std::set<std::vector<pixel_type>>
-      candidate_palette;  // ordered lexicographically
+  std::set<std::vector<pixel_type>> candidate_palette;
   std::vector<std::vector<pixel_type>> candidate_palette_imageorder;
   std::vector<pixel_type> color(nb);
   std::vector<float> color_with_error(nb);
   std::vector<const pixel_type *> p_in(nb);
+  std::map<std::vector<pixel_type>, size_t> inv_palette;
 
   if (lossy) {
     palette_iteration_data.FindFrequentColorDeltas(w * h, input.bitdepth);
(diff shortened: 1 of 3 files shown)


lib/jxl/modular/transform/enc_palette.cc
```diff
@@ -158,6 +160,17 @@ struct PaletteIterationData {
   }
 };
 
+struct pixel_hash {
```
veluca93, line 163: I don't like this hasher very much :) We should probably use a proper hasher.

veluca93 (changes requested):

lib/jxl/modular/transform/enc_palette.cc
jonsneyers: Agreed, it's probably a quite poor hash function. Best to keep it as simple as possible though — collisions are not as problematic here as in the e1 palette detection...

jonsneyers: Switched back to the (ordered) `set` and `map`, because it looks like the speedup wasn't coming from switching to unordered anyway (just using a map instead of linear search also does the job), and for some reason `unordered_map` was randomly causing `bad_alloc` in the CI and I can't be bothered to investigate what's going on there.

Speed and density results are similar to what I reported before, it's maybe even slightly faster this way.

veluca93 (dismissed):

jonsneyers:
Further simplified the code a bit to try global RGBA/RGB palette only once, before trying channel palette, as opposed to the previous strategy of first trying it with few max colors and then again after channel palette with more max colors.
This ensures that the palette ordering always makes sense (at least for global palettes). Doing channel palette first and _then_ RGB(A) palette can save some bytes in weird cases (like 16-bit images with ~1000 unique colors so palette can be done but the cost of signaling the palette is not so small), but usually it will not help much or even be counterproductive, and in any case it will make decoding slower (need to undo multiple palettes instead of just one) and makes the palette ordering weird since the channel ranges will be different (and the sorting kind of assumes they're the same).

jonsneyers:
Made some changes, PTAL @veluca93

PR #2534, opened 2023-06-07
faster jpeg reconstruction

jonsneyers:
Reconstruction of losslessly recompressed JPEG files was rather slow. A big reason for this was that a small initial output buffer size was used, leading to JXL_DEC_JPEG_NEED_MORE_OUTPUT, which with the current (non-streaming) implementation in `dec_jpeg_data_writer.cc` causes it to basically redo all of the work several times.
This gets particularly bad if the jpeg file is quite large — the initial buffer size was 16 kb and basically it would reconstruct the first 16 kb, then run out of output buffer, restart and reconstruct the first 32 kb, then 64 kb, and so on until the buffer size is large enough to store the entire jpeg file.

Besides the small change to set the initial output buffer size large enough to likely store the entire result on the first try, I also 'ported' some tricks from libjpeg-turbo to make the huffman coding itself a bit more efficient.

Before:
```
JPEG XL decoder v0.9.0 e11e728b [AVX3,AVX2,SSE4,SSSE3,SSE2]
Read 3487622 compressed bytes.
Reconstructed to JPEG.
2400 x 2400, geomean: 5.38 MP/s [5.26, 5.38], geomean: 4.13 MB/s [4.04, 4.13], 30 reps, 1 threads.
```

After:
```
JPEG XL decoder v0.9.0 e11e728b [AVX3,AVX2,SSE4,SSSE3,SSE2]
Read 3487622 compressed bytes.
Reconstructed to JPEG.
2400 x 2400, geomean: 22.39 MP/s [21.96, 22.49], geomean: 17.19 MB/s [16.85, 17.26], 30 reps, 1 threads.
```

### DIFF
--- a/lib/jxl/jpeg/dec_jpeg_data_writer.cc
+++ b/lib/jxl/jpeg/dec_jpeg_data_writer.cc
@@ -13,6 +13,7 @@
 #include <vector>
 
 #include "lib/jxl/base/bits.h"
+#include "lib/jxl/base/byte_order.h"
 #include "lib/jxl/common.h"
 #include "lib/jxl/image_bundle.h"
 #include "lib/jxl/jpeg/dec_jpeg_serialization_state.h"
@@ -76,18 +77,19 @@ static JXL_INLINE void Reserve(JpegBitWriter* bw, size_t n_bytes) {
  * space in the output buffer. Emits up to 2 bytes to buffer.
  */
 static JXL_INLINE void EmitByte(JpegBitWriter* bw, int byte) {
-  bw->data[bw->pos++] = byte;
-  if (byte == 0xFF) bw->data[bw->pos++] = 0;
+  bw->data[bw->pos] = byte;
+  bw->data[bw->pos + 1] = 0;
+  bw->pos += (byte != 0xFF ? 1 : 2);
 }
 
-static JXL_INLINE void DischargeBitBuffer(JpegBitWriter* bw) {
-  // At this point we are ready to emit the most significant 6 bytes of
-  // put_buffer_ to the output.
+static JXL_INLINE void DischargeBitBuffer(JpegBitWriter* bw, int nbits,
+                                          uint64_t bits) {
+  // At this point we are ready to emit the put_buffer to the output.
   // The JPEG format requires that after every 0xff byte in the entropy
   // coded section, there is a zero byte, therefore we first check if any of
-  // the 6 most significant bytes of put_buffer_ is 0xFF.
-  Reserve(bw, 12);
-  if (HasZeroByte(~bw->put_buffer | 0xFFFF)) {
+  // the 8 bytes of put_buffer is 0xFF.
+  bw->put_buffer |= (bits >> -bw->put_bits);
+  if (JXL_UNLIKELY(HasZeroByte(~bw->put_buffer))) {
     // We have a 0xFF byte somewhere, examine each byte and append a zero
     // byte if necessary.
     EmitByte(bw, (bw->put_buffer >> 56) & 0xFF);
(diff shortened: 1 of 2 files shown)


veluca93 (dismissed):
I can't comment on the porting-from-jpegli part, but the rest LGTM.

Also, maybe we should eventually figure out how to make WriteOutput not be all-or-nothing...

jonsneyers:
> I can't comment on the porting-from-jpegli part, but the rest LGTM.
> 
> Also, maybe we should eventually figure out how to make WriteOutput not be all-or-nothing...

I mean, I was initially assuming that it was the huffman writing itself that was causing the reconstruction to be so slow (since profiling showed a lot of time was spent in that), so I tried speeding that up by looking at the libjpeg-turbo code and used some of the ideas there plus some other stuff to avoid branching.
But it was then still a lot slower than libjpeg-turbo encoding (which doesn't need to entropy decode but it does need to do DCT), and eventually I figured out what was going on with the output buffer being too small, causing all the work to be repeated.

Making WriteOutput resumable would be nice but I don't think it's urgent — the size of a reconstructed jpeg is rather predictable (it will be hard to find cases where it isn't between 1.1x and 1.5x the size of the jxl) so it can be one-shotted quite well.

lib/jxl/jpeg/dec_jpeg_data_writer.cc
```diff
@@ -196,10 +191,31 @@ static JXL_INLINE void WriteSymbol(int symbol, HuffmanCodeTable* table,
   if (kOutputMode == OutputModes::kModeHistogram) {
     ++table->depth[symbol];
   } else {
+    // This is an optimization; if everything goes well,
+    // then |nbits| is positive; if non-existing Huffman symbol is going to be
+    // encoded, its length should be zero; later the encoder could check the
+    // "health" of JpegBitWriter.
+    // Not checking this can cause bad output without error on corrupt input,
+    // but that's OK. The check causes a noticeable slowdown.
+#if false
```
szabadka, line 200: shouldn't it at least be enabled in debug builds?

lib/jxl/jpeg/dec_jpeg_data_writer.cc
```diff
@@ -490,42 +506,115 @@ bool EncodeDCTBlockSequential(const coeff_t* coeffs, HuffmanCodeTable* dc_huff,
...
-    int symbol = (r << 4u) + ac_nbits;
-    WriteSymbol<kOutputMode>(symbol, ac_huff, bw);
-    WriteBits(bw, ac_nbits, temp2 & ((1 << ac_nbits) - 1));
-    r = 0;
+#endif
+  WriteBits(bw, dc_nbits, temp & ((1u << dc_nbits) - 1));
+  int16_t r = 0;
+
+  // Manually unfolded loop over kJPEGNaturalOrder[k] for k=1..63
+#define kloop(jpeg_natural_order_of_k)                            \
```
szabadka, line 526: how much speedup is the loop unrolling in itself?

jonsneyers: Without that unrolling (keeping everything else the same) it's about 3% slower, e.g. `21.74 MP/s [21.29, 21.80]` without unrolling vs `22.39 MP/s [21.96, 22.49]` with unrolling.

lib/jxl/jpeg/dec_jpeg_data_writer.cc
```diff
@@ -783,6 +872,9 @@ SerializationStatus JXL_NOINLINE DoEncodeScan(const JPEGData& jpg,
         ss.restarts_to_go = restart_interval;
         memset(ss.last_dc_coeff, 0, sizeof(ss.last_dc_coeff));
       }
+      // compressed size per block cannot be more than 256 bytes per component
+      Reserve(bw, 256 * scan_info.num_components);
```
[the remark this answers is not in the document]
jonsneyers: Right, will change it to blocks-per-MCU.

The 256 bytes per block bound I took from libjpeg-turbo. The worst-case scenario you describe would happen if the huffman codes and extra bits would both be all-1s, right? So I guess you're right, in theory it could be 512 bytes.

I don't see a convenient already-available variable for blocks-per-MCU so I'll just do one `Reserve` per component instead, it will not make much of a difference. It's just good to get the `Reserve` out of the inner loop (it used to be reserving only 12 bytes at a time).

szabadka (approved):

PR #2551, opened 2023-06-09
Fix an integer underflow bug in patch decoding.

veluca93:
When reading patches, the code doesn't properly check that all patches are in bounds.

In particular, if the start of a patch becomes negative (due to the delta coding of patch positions), but not negative *enough* to cause `start + size` to still be negative, the decoder happily accepted the patch. Such a patch would then lead the interval tree building code to run into an infinite loop.


eustas (approved):

jonsneyers (approved):

jonsneyers:
Can the infinite loop behavior be triggered in 0.8.1?

veluca93:
> Can the infinite loop behavior be triggered in 0.8.1?

Yep, we should release a patch release and/or backport this soon.

rfrohl:
FTR: this issue got CVE-2023-35790 assigned

PR #2552, opened 2023-06-11
Corrected resampling description and clarified size trade-offs

jonnyawsom3:
My first pull and first commit on Github, so thought I'd keep it simple to start.


mo271:
Please add yourself to the AUTHORS file and follow the instructions above to sign the cla

jonnyawsom3:
Hopefully that works, CLA was already signed too shortly after opening the pull

tools/cjxl_main.cc
```diff
@@ -274,14 +274,14 @@ struct CompressArgs {
 
     cmdline->AddOptionValue(
         '\0', "resampling", "-1|1|2|4|8",
-        "Resampling for extra channels. Default of -1 applies resampling only "
+        "Resampling for colour channels. Default of -1 applies resampling only "
```
mo271, line 277:
```suggestion
        "Resampling for color channels. Default of -1 applies resampling only "
```

tools/cjxl_main.cc
```diff
@@ -274,14 +274,14 @@ struct CompressArgs {
...
         "for low quality. Value 1 does no downsampling (1x1), 2 does 2x2 "
         "downsampling, 4 is for 4x4 downsampling, and 8 for 8x8 downsampling.",
         &resampling, &ParseInt64, 0);
 
     cmdline->AddOptionValue(
         '\0', "ec_resampling", "-1|1|2|4|8",
-        "Resampling for extra channels. Default of -1 applies resampling only "
+        "Resampling for extra channels (Such as Alpha). Default of -1 applies resampling only "
```
mo271, line 284:
```suggestion
        "Resampling for extra channels (such as alpha). Default of -1 applies resampling only "
```

tools/cjxl_main.cc
```diff
@@ -147,7 +147,7 @@ struct CompressArgs {
     cmdline->AddOptionValue(
         'e', "effort", "EFFORT",
         "Encoder effort setting. Range: 1 .. 9.\n"
-        "     Default: 7. Higher number is more effort (slower).",
+        "     Default: 7. Higher number is more effort (slower) for smaller size.",
```
monad0, line 150: In lossless, this is broadly the goal, but there is no assurance that higher effort produces smaller file sizes. In lossy, the semantics are different: higher effort aims to more accurately and consistently reach the target distance (while also employing more costly tools).

jonnyawsom3: Good point, leave it as it was or change it to "resulting in more detail or smaller size"?

monad0:
```suggestion
        "    Default: 7. Higher numbers allow more computation, generally producing smaller files at same quality. For lossy output, higher effort should more accurately reach the target quality.",
```

monad0: Main thing bugging me about my suggestion is the word 'files' which may not always be semantically accurate, but I'm out of time right now and think it communicates the point to the average user.

jonnyawsom3: I think that'll do, this _is_ for cjxl itself rather than the library after all, so it'll probably always be files anyway.

jonnyawsom3:
Oh, right. The suggestion has added @monad0 as a contributor, so they'll need to sign the CLA too now I suppose

jonnyawsom3:
Everything except the Authors check should pass now, but that's due to it associating me with the GitHub noreply instead of my actual email, so can likely be ignored and merged if all looks good

mo271:
rebased and fixed authors

mo271 (approved):

mo271:
@jonsneyers please take a look, since you have just rewritten most of this...

tools/cjxl_main.cc
```diff
@@ -144,7 +144,9 @@ struct CompressArgs {
     cmdline->AddOptionValue(
         'e', "effort", "EFFORT",
         "Encoder effort setting. Range: 1 .. 9.\n"
-        "    Default: 7. Higher number is more effort (slower).",
+        "Default: 7. Higher numbers allow more computation, generally "
```
jonsneyers, line 147: Please check indentation and line breaks. Aiming to stay below 100 characters width and indentation consistent (the first line of the description automatically get indented 4 spaces so any additional lines also need 4 spaces).

jonnyawsom3: I know the lines are still long, but should be within what you asked. Also fixed the group ordering descriptions so they matched all the other formatting

tools/cjxl_main.cc
```diff
@@ -144,7 +144,10 @@ struct CompressArgs {
     cmdline->AddOptionValue(
         'e', "effort", "EFFORT",
         "Encoder effort setting. Range: 1 .. 9.\n"
-        "    Default: 7. Higher number is more effort (slower).",
+        "    Default: 7. Higher numbers allow more computation "
+        "at the expense of time.\n"
+        "    For lossless, generally it will produce smaller files at the same quality.\n"
```
jonsneyers, line 149: This sentence is weird. Just drop the "at the same quality"?

jonnyawsom3: Yeah, I did think it was a bit redundant, removing it now

jonsneyers (approved):

jonnyawsom3:
It _should_ be ready now, thanks for all the help on this, I know it's taken far longer than it should have

jonsneyers (approved):

tools/cjxl_main.cc
monad0: @jonsneyers @jonnyawsom3 The original suggestion https://github.com/libjxl/libjxl/pull/2552#discussion_r1231633026 described the behavior irrelevant of the mode. Indeed, qualifying it here as a feature of lossless makes the phrase redundant, while the original sentence was more informative.

Similarly, the suggestion that allowing "more computation" comes "at the expense of time" is redundant (when true). The original suggestion rather highlighted the computation/size tradeoff.

PR #2555, opened 2023-06-12
Add PSNR target parameter to jpegli encoder.

szabadka:


mo271 (approved):

jonsneyers:
Could you explain here or in a code comment what this does at a high level? If I understand correctly, when optimizing for PSNR you don't need adaptive quantization and you basically do a binary search to find out what amount of global quantization to use to get to a given target PSNR?

Shouldn't it in principle be possible to one-shot it instead of searching? After all, PSNR should be more or less the same thing as the average quantization error, so the same quantization constants should lead to more or less the same PSNR (modulo some image content dependency).

szabadka:
This code is not optimizing for PSNR, it simply selects a distance value that would result in a given PSNR, while doing everything it would do otherwise, including adaptive quantization if enabled. You are right that disabling adaptive quantization leads to a better file size for the same PSNR value. It did this mostly to be able to compare against sjpeg codec that has a similar option.

jonsneyers:
I see, thanks for the clarification.

Issue #1450, opened 2022-05-26
libjxl uses abort()


jpcha2:
**Describe the bug**
libjxl calls abort(). Since it is intended as a library to be used in other projects, it shouldn't do this. 

**To Reproduce**
Basically any flow, including the loading of a corrupted file or API misuse, that results in a call to one of the following:
- jxl::Abort()
- JxlThreadParallelRunner::Abort()

**Expected behavior**
Calling abort() within a library is a bug because the library has no way of knowing what the calling application is doing. If the calling application is a process that is either statically linked to the library, a process that is dynamically linked to it or a process that calls a DLL that loads the library in any fashion, that abort() call will lead to the entire application suddenly crashing, sometimes for apparently no reason or a very trivial one. 

Consider, for instance, a case where libjxl is being used to store and load small images used as, say, employee mugshots in a listing in an app that does something completely unrelated to graphics. The mugshots are an unimportant feature and an invalid thumbnail file should ideally just result in a blank square, but the current implementation will crash the entire app as a result of the abort() call. The HR app for payroll crashing because the new intern's photo was causing one of the many image-file loading libraries to intentionally abort() the program cannot be described as a good design. 

In my view, there is no error that can be encountered within a library that justifies aborting the calling process. 

There are two ways to address this:
- Use the approach taken in butteraugli.cc, which is to wrap the fault-handler that calls abort() in an "enable_checks" switch
- Implement a proper fault-handling scheme that results in the relevant error being sent back to the calling application via the return value of the currently executing libjxl function.

**Screenshots**
n/a

**Environment**
 - OS: Windows
 - Compiler version: MSVC 2017 (ver 14.16)]
 - CPU type: x86_64
 - cjxl/djxl version string: n/a

**Additional context**

rickbrew:
I second this. I'd love to use libjxl to add JPEG XL support to Paint.NET by way of including @0xC0000054 's [plugin](https://github.com/0xC0000054/pdn-jpegxl), but I can't let a failure in loading a file cause the app to crash. `abort()` is a banned method.

jonsneyers:
Should be fixed in v0.11.
Please reopen if this is still an issue.

Issue #762, opened 2021-10-22
better handling of memory allocation failures


novomesk:
Hello,

especially on 32bit systems, there is very easy to run out of available memory (address space is limited).

It is enough to have few large images with float precision and together with other objects/libraries the address space is exhausted.

I can easily get following assertion when decoding another big image:
`./lib/jxl/image.cc:94: JXL_CHECK: bytes_.get()`

It is from:
```
    bytes_ = AllocateArray(bytes_per_row_ * ysize);
    JXL_CHECK(bytes_.get());
```

Unfortunately it aborts whole process/application.

I'd prefer if decoding just returns with an error without closing the whole application.

PR #2557, opened 2023-06-12
return JXL_ENC_ERR_OOM instead of aborting when out of memory

jonsneyers:
See #762 and #1450 : currently when allocations fail (out of memory), libjxl will just abort. For a command line tool like cjxl that is fine, but when integrated in a bigger application (when not taking specific sandboxing precautions), this will bring down the entire application which of course is not so nice. In particular, when an artist is working on a large image, and trying to save the image causes the application to crash, it could be very frustrating...

This PR is a first step in the direction of trying to fix that, starting with the encoder.

I probably missed many spots, but basically the goal is to make sure that `std::bad_alloc` gets thrown whenever something running out of memory and then it gets caught at the API interface (in `encode.cc`).

Probably there are other places where allocs are happening, but the main allocations are done in `CacheAligned::Allocate` (e.g. to allocate image buffers) and when using things like std::vector (which already does throw that exception).

Probably there are other places where the exception needs to be caught, but these are the main ones: when passing frames to the encoder (since it then makes a copy of it, allocating a big buffer) and when doing an actual frame encode (since that allocates all kinds of auxiliary data structures).

When running cjxl with `ulimit` to reduce the available memory to something small, I now get nice errors instead of aborts when trying to encode images that are too large, e.g.

```
JPEG XL encoder v0.9.0 7b50c120 [SSE4,SSSE3,SSE2]
Read 3072x2048 image, 7607002 bytes, 75.7 MP/s
Encoding [VarDCT, d1.000, effort: 7], 
./lib/jxl/encode.cc:1912: Not enough memory to allocate input buffer
JxlEncoderAddImageFrame() failed.
EncodeImageJXL() failed.
```

if the image is too large to copy the input to an internal buffer, and something like this:

```
JPEG XL encoder v0.9.0 7b50c120 [SSE4,SSSE3,SSE2]
Read 1361x907 image, 7406580 bytes, 665.2 MP/s
Encoding [VarDCT, d1.000, effort: 7], 
./lib/jxl/encode.cc:581: Out of memory during frame encoding
JxlEncoderProcessOutput failed.
EncodeImageJXL() failed.
```

if the OOM happens somewhere during the encoding.

Further work is needed to make sure we're not missing anything (since uncaught exceptions will still cause an abort), and to also do this for decoding. And then I guess we also need to look at all places where we do an explicit abort and replace those with returning error instead.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -104,7 +104,6 @@ else ()
   else()  # WIN32
     list(APPEND JPEGXL_INTERNAL_FLAGS
       -fsized-deallocation
-      -fno-exceptions
 
       # Language flags
       -fmath-errno
@@ -113,7 +112,6 @@ else ()
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       list(APPEND JPEGXL_INTERNAL_FLAGS
         -fnew-alignment=8
-        -fno-cxx-exceptions
         -fno-slp-vectorize
         -fno-vectorize
 
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -96,7 +96,6 @@ typedef enum {
   JXL_ENC_ERR_GENERIC = 1,
 
   /** Out of memory
-   *  TODO(jon): actually catch this and return this error
    */
   JXL_ENC_ERR_OOM = 2,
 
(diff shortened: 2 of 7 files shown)


jonsneyers:
So after some discussion we decided not to use this try/catch approach after all, mostly because there are build processes that don't support exceptions (in particular Chrome) so it would not be a full solution.

Instead we'll do the code plumbing to make allocation failures return an error instead of aborting or throwing uncaught exceptions. It will be quite some work to pass the status around in all functions that allocate stuff, but it will be the best solution. I'm keeping this PR open for now in case someone needs an encoder right now that will more gracefully handle OOM, but we decided not to go this route so a different PR will replace this one at some point.

PR #2559, opened 2023-06-12
Release v0.8.2

mo271:


veluca93 (dismissed):

mo271:
The CIFuzz / fuzzing fails because of https://github.com/libjxl/libjxl/pull/2427 and relatedly https://github.com/google/oss-fuzz/pull/10171.

We can't just cherry-pick those fixes to make the test pass, since they would require other jpegli-changes.
However, I think this workflow is not blocker. What do you think @veluca93 and @szabadka?

szabadka:
I don't think it is a blocker. Ideally the code that builds the fuzzers should also be in the jxl repo and the ossfuzz should only reference one script that does everything, but it is probably too late to change that for this release.

veluca93:
> I don't think it is a blocker. Ideally the code that builds the fuzzers should also be in the jxl repo and the ossfuzz should only reference one script that does everything, but it is probably too late to change that for this release.

+1, just submit :P

veluca93 (dismissed):

mo271:
ok, reverting to only having one cherry-pick, because I realized that fixing the MSYS2 / mingw32 tests would require at least 4 more cherry-picks...
I think it is cleaner to only have the more relevant fix here.

PR #2560, opened 2023-06-13
allow using cjxl/djxl with stdin/stdout

jonsneyers:
Replaces https://github.com/libjxl/libjxl/pull/1958

As is the convention in many other tools, the filename `-` is treated as stdin/stdout. So `cjxl - - < input.png > output.jxl` will work. For djxl, `djxl - - < input.jxl > output.png` does not work since djxl needs to know what format to output.

So this adds syntax similar to what ImageMagick's tools do: besides communicating the output format via the filename extension (e.g. `output.png`) you can also communicate it with a prefix (e.g. `jpg:output`). This allows you to do `djxl - png:- < input.jxl > output.png`.

In the case where multiple output files would be generated (e.g. when decoding an animation to ppm, where it generates one output file per frame), it will simply concatenate everything to stdout. Tools like ffmpeg do take concatenated frames as input, so that seems reasonable (unfortunately timing info will be lost but that's also the case when decoding to separate files).

### DIFF
--- a/tools/djxl_main.cc
+++ b/tools/djxl_main.cc
@@ -215,6 +215,7 @@ bool WriteOptionalOutput(const std::string& filename,
 std::string Filename(const std::string& base, const std::string& extension,
                      int layer_index, int frame_index, int num_layers,
                      int num_frames) {
+  if (base == "-") return "-";
   auto digits = [](int n) { return 1 + static_cast<int>(std::log10(n)); };
   std::string out = base;
   if (num_frames > 1) {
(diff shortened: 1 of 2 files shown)


mo271 (approved):
LGTM

dalbani:
Excellent, thanks a million @jonsneyers for taking it over and implementing it! 👍

eustas:
Does it work with filenames like "`C:\images\photo.jpg`"?

eustas:
In next PR I'm going to swap the order of check - first period, then colon

PR #2562, opened 2023-06-13
Implement a StatusOr type.

veluca93:
This can be used i.e. to implement static methods that construct an Image3F with proper allocation failure handling.

### DIFF
--- a/lib/jxl/base/status.h
+++ b/lib/jxl/base/status.h
@@ -13,6 +13,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include <type_traits>
+#include <utility>
+
 #include "lib/jxl/base/compiler_specific.h"
 #include "lib/jxl/base/sanitizer_definitions.h"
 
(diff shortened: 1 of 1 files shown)


lib/jxl/base/status.h
```diff
@@ -321,6 +324,84 @@ inline JXL_FORMAT(2, 3) Status
...
+template <typename T>
+class JXL_MUST_USE_RESULT StatusOr {
+  union Storage {
+    char dummy_;
+    T data_;
+  } storage_;
+
+  static_assert(!std::is_convertible<StatusCode, T>::value &&
+                    !std::is_convertible<T, StatusCode>::value,
+                "You cannot make a StatusOr with a type convertible from or to "
```
jonsneyers, line 336:
```suggestion
                "You can only make a StatusOr with a type convertible from or to "
```

veluca93: I think your suggestion changes the meaning of what the error says :-)

jonsneyers: Oh, I misunderstood, I thought the template param was to eventually use it for both jxl::Status and C API error codes, but the template param is for the type that may fail to construct due to OOM. Nevermind!

lib/jxl/base/status.h
```diff
@@ -321,6 +324,84 @@ inline JXL_FORMAT(2, 3) Status
...
+    }
+    code_ = other.code_;
+  }
+
+  StatusOr operator=(StatusOr&& other) noexcept {
+    if (ok() && other.ok()) {
+      storage_.data_ = std::move(other.storage_.data_);
+    } else if (other.ok()) {
+      new (&storage_.data_) T(std::move(other.storage_.data_));
+    }
```
danakj, line 370: Need to `else if (ok()) storage_.data_.~T();`

lib/jxl/base/status.h
```diff
@@ -321,6 +324,84 @@ inline JXL_FORMAT(2, 3) Status
...
+
+  StatusOr(StatusOr&& other) noexcept {
+    if (other.ok()) {
+      new (&storage_.data_) T(std::move(other.storage_.data_));
+    }
+    code_ = other.code_;
+  }
+
+  StatusOr operator=(StatusOr&& other) noexcept {
+    if (ok() && other.ok()) {
```
danakj, line 366: Maybe if this == &other then do nothing? Otherwise you require T::operator=(T&&) to handle aliasing itself.

lib/jxl/base/status.h
```diff
@@ -321,6 +324,84 @@ inline JXL_FORMAT(2, 3) Status
   return status;
 }
 
+template <typename T>
+class JXL_MUST_USE_RESULT StatusOr {
+  union Storage {
```
danakj, line 329: Do you not need to provide an empty ctor and dtor here?

PR #2569, opened 2023-06-14
fix --already_downsampled

jonsneyers:
The cjxl option `--already_downsampled` was broken: it would produce an image with the non-upsampled size, and then pass it a buffer with that size, which is too large since the encoder is expecting to get an already-downsampled frame. So it would just cause an error:

```
./lib/jxl/enc_external_image.cc:84: JXL_FAILURE: Buffer size is too large
./lib/jxl/enc_external_image.cc:133: JXL_RETURN_IF_ERROR code=1: ConvertFromExternal(bytes, xsize, ysize, bits_per_sample, format, c, pool, &color.Plane(c))
./lib/jxl/encode.cc:1896: Invalid input buffer
JxlEncoderAddImageFrame() failed.
EncodeImageJXL() failed.
```

This fixes that by setting the nominal image dimensions correctly in the JxlBasicInfo (to the upsampled size), so the expected frame dimensions will correspond to the input.

### DIFF
--- a/lib/extras/enc/jxl.cc
+++ b/lib/extras/enc/jxl.cc
@@ -88,6 +88,8 @@ bool EncodeImageJXL(const JXLCompressParams& params, const PackedPixelFile& ppf,
   } else {
     size_t num_alpha_channels = 0;  // Adjusted below.
     JxlBasicInfo basic_info = ppf.info;
+    basic_info.xsize *= params.already_downsampled;
+    basic_info.ysize *= params.already_downsampled;
     if (basic_info.alpha_bits > 0) num_alpha_channels = 1;
     if (params.intensity_target > 0) {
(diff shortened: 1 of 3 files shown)


mo271:
Should we add a test here?

jonsneyers:
> Should we add a test here?

I dunno, libjxl itself was ok but the way cjxl was using it was wrong. It would probably be good in general to have some cjxl/djxl roundtrip tests to check if everything works, but we currently don't have that. Something like "toolchain fuzzing": running cjxl with random combinations of options, checking if it produces something and djxl decodes it, and then computing some metric to see if the decoded image is within expectations (depending on options like distance and resampling). The --already_downsampled option is a bit special though because it creates a jxl with dimensions that are different from the input dimensions.

mo271 (approved):

PR #2570, opened 2023-06-14
fix RoundtripLosslessAnimation test

mo271:


mo271:
see, https://github.com/libjxl/libjxl/actions/runs/5256709528/jobs/9498417869
Was this caused by https://github.com/libjxl/libjxl/pull/2523, @jonsneyers ?

mo271:
I don't understand why it only fails in the Bazel test

jonsneyers:
> see, https://github.com/libjxl/libjxl/actions/runs/5256709528/jobs/9498417869 Was this caused by #2523, @jonsneyers ?

I haven't checked but my guess would be yes. Lossless 
will be different when palette can be done (mostly better, I hope), so it could be this is an example of that. I wonder why it didn't fail earlier though — is this a test that is not in the PR CI?

mo271:
I think it is in the PR CI when you add https://github.com/libjxl/libjxl/labels/CI%3Afull

jonsneyers (approved):

PR #2571, opened 2023-06-15
add a way to do custom upsampling

jonsneyers:
Currently the default upsampling weights are always used by the libjxl encoder. This adds a function to use custom upsampling weights. This is in particular useful when encoding pixel art with JXL_ENC_FRAME_SETTING_ALREADY_DOWNSAMPLED, since the default upsampling is great for photographic images but it's not what you typically want for pixel art.

For example, for this original image:
mario

encoding with `cjxl -d 0 --already_downsampled --resampling 8` produces this:

mario png default jxl

while when setting the upsampling mode to nearest neighbor (`--upsampling_mode 0`), it will look like this:

mario png NN jxl

Just for fun I also added another upsampling mode which I called "pixel dots" and it looks like this:

mario png funky2 jxl

For truly customized upsampling we would need a function that allows you to set the actual weights as opposed to picking from some presets, but that would be pretty inconvenient to use. This is of course less flexible, but we can always add more presets later if we want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `JXL_ENC_FRAME_SETTING_JPEG_KEEP_EXIF`
   - `JXL_ENC_FRAME_SETTING_JPEG_KEEP_XMP`
   - `JXL_ENC_FRAME_SETTING_JPEG_KEEP_JUMBF`
+ - encoder API: new function `JxlEncoderSetUpsamplingMode` to change the upsampling
+   method, e.g. to use nearest-neighbor upsampling for pixel art
  - cjxl can now be used to explicitly add/update/strip Exif/XMP/JUMBF metadata using
    the decoder-hints syntax, e.g. `cjxl input.ppm -x exif=input.exif output.jxl`
  - djxl can now be used to extract Exif/XMP/JUMBF metadata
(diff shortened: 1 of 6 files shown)


AlifianK:
I'm sorry if this isn't the right place, but is the ```resampling``` definition in ```--help``` intended to be the same as ```ec_resampling```? Both are defined like this
```
--resampling=-1|1|2|4|8
    Resampling for extra channels. Default of -1 applies resampling only for low quality. Value 1 does no downsampling (1x1), 2 does 2x2 downsampling, 4 is for 4x4 downsampling, and 8 for 8x8 downsampling.
--ec_resampling=-1|1|2|4|8
    Resampling for extra channels. Default of -1 applies resampling only for low quality. Value 1 does no downsampling (1x1), 2 does 2x2 downsampling, 4 is for 4x4 downsampling, and 8 for 8x8 downsampling.
```

mo271:
> I'm sorry if this isn't the right place, but is the `resampling` definition in `--help` intended to be the same as `ec_resampling`? Both are defined like this
> 
> ```
> --resampling=-1|1|2|4|8
>     Resampling for extra channels. Default of -1 applies resampling only for low quality. Value 1 does no downsampling (1x1), 2 does 2x2 downsampling, 4 is for 4x4 downsampling, and 8 for 8x8 downsampling.
> --ec_resampling=-1|1|2|4|8
>     Resampling for extra channels. Default of -1 applies resampling only for low quality. Value 1 does no downsampling (1x1), 2 does 2x2 downsampling, 4 is for 4x4 downsampling, and 8 for 8x8 downsampling.
> ```

this discussion is best done here: https://github.com/libjxl/libjxl/pull/2552

jonsneyers:
> Consider adding a changelog entry for this api change

done

mo271 (approved):

PR #2575, opened 2023-06-16
libjpegli is ready for system install now

malaterre:


szabadka (approved):

rouault:
I got hit by that just today where it crashes my tests due to the mix between the libjpeg from my system and the one from libjxl: https://github.com/OSGeo/gdal/actions/runs/5289282972/jobs/9572015255?pr=7959 . I wouldn't expect libjxl to build a libjpeg.so by default and install it.   JPEGXL_ENABLE_JPEGLI_LIBJPEG=NO would probably be a safer default

szabadka:
In the meantime I will add a separate CMAKE flag JPEGXL_INSTALL_JPEGLI_LIBJPEG that is false by default.

PR #2576, opened 2023-06-16
remove Butteraugli from the libjxl API

jonsneyers:
Currently we have butteraugli as part of the libjxl API, which doesn't really make sense — libjxl does use butteraugli internally (for e8+ vardct encoding) but it doesn't really make sense to expose that functionality as part of the libjxl API. If we make it part of the 1.0 API, then we're basically forcing everyone who wants to make an API/ABI-compatible libjxl to implement a function for computing butteraugli scores, which I don't think is a good idea.

So for now I propose to just remove those API functions from the libjxl API now we still can.

It would probably be useful to at some point have a `libbutteraugli` or maybe `libjxl_metrics` (which could include also ssimulacra2 and any other metrics we feel like having there, e.g. we could also have psnr and y(uv)-(m)ssim). Then this could be an optional dependency for libjxl (only needed if you want to do high-effort encoding that is optimizing for a metric), with maybe a runtime option to select which metric to use for optimization. Also benchmark_xl could then use that library, as well as the command line tools for butteraugli and ssimulacra2.


mo271:
Should this be mentioned in the changelog?

jonsneyers:
It probably should — I don't know to what extent we "advertised" the Butteraugli api to be part of libjxl and I don't think anyone is using libjxl in that way, but it doesn't hurt to mention it.

mo271 (approved):

PR #2578, opened 2023-06-16
Don't install jpegli system-wide by default.

szabadka:


jonsneyers (approved):

lib/jpegli.cmake
```diff
@@ -96,10 +96,13 @@ set_target_properties(jpeg PROPERTIES
 set_property(TARGET jpeg APPEND_STRING PROPERTY
   LINK_FLAGS " -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/jpegli/jpeg.version.${JPEGLI_LIBJPEG_LIBRARY_SOVERSION}")
 
-install(TARGETS jpeg
-  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(TARGETS jpeg RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
kmilos, line 99: This will always install the DLL (i.e. shared object) on Windows, should't it be gated by `if (JPEGXL_INSTALL_JPEGLI_LIBJPEG)` as well?

szabadka: I changed it like you suggested, now all jpegli installs are behind the if.

mo271 (approved):

PR #2582, opened 2023-06-19
Only one file_io

jonsneyers:
This is just a bit of code cleanup.

We currently have a `lib/jxl/base/file_io.h` and a `tools/file_io.h` and they basically duplicate the same code (a function to read a file and one to write a file) with some minor differences like the order and types of the arguments.

The only code that uses file io in `lib/jxl/` are some of the tests; libjxl itself doesn't do any file io. So it's better to just remove the `lib/jxl/base/` version of the file io code and keep only the one in `tools/`.


mo271 (dismissed):

jonsneyers:
Moving `file_io.h` to `lib/extras/` instead of `lib/jxl/base` and `tools/`, since that seems to be the most logical place for it after all...

mo271 (approved):
I think this is indeed a better place; lgtm

PR #2584, opened 2023-06-19
Harden current directory is a Git repository

dpino:
I've pulled libjxl as tarball inside a project which is managed by Git. When executing `deps.sh` the script detects `libjxl` is a Git repository, but because it's inside a parent Git repository. However, the provision of the submodules fails (tries to provision the submodules of the parent Git repository, which are none).

To correctly provision `libjxl` submodules the script should not only check libjxl is a Git repository but its toplevel directory is the absolute path of libjxl.


mo271:
Thanks for the pull request; please add yourself to the AUTHORS file...

jonsneyers (approved):

Issue #2581, opened 2023-06-18
Losslessly recompressing JXL turns enum tags into ICC profiles


Traneptora:
**Describe the bug**
~~Consider the conformance sample `animation_icos4d.jxl`. (Here, for convenience: [https://buzo.us/H.jxl](https://buzo.us/H.jxl)). 
Attempting to losslessly recompress it using `cjxl -d 0` appears to be taking the 8-bit pixel data from the original and encoding it losslessly as 32-bit float, which inflates the file size pretty dramatically.~~ Fixed by #2591 although the next issue remains, at the moment: 

It also seems to be embedding an ICC profile rather than tagging the new file with an sRGB enum, despite the original being tagged with an sRGB enum.

**To Reproduce**
```
$ cjxl -d 0 animation_icos4d.jxl test.jxl
$ jxlinfo -v animation_icos4.jxl
$ jxlinfo -v test.jxl
```

**Expected behavior**
I expected cjxl to preserve the tagged enum space rather than synthesize an ICC profile and then embed the synthesized profile.

**Log**
[https://buzo.us/D.log](https://buzo.us/D.log)

**Environment**
 - OS: Arch Linux x86_64
 - Compiler version: GCC 13
 - cjxl/djxl version string: JPEG XL encoder v0.9.0 00935aaa [AVX2]

jonsneyers:
Ugh, why does github auto-close the issue if I referenced it saying my PR "partially fixes" this issue.

The bitdepth thing should be fixed now, but it will still convert an enum colorspace to a synthesized icc profile...

Traneptora:
Should I open a separate issue for that?

Traneptora:
Updated title and body to reflect the current issue.

kmilos:
> Ugh, why does github auto-close the issue if I referenced it saying my PR "partially fixes" this issue.

@jonsneyers Because of the [keyword "fixes"](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) (and I don't know why they even say "may close" on the side when it seem to do it always when linked). Been there, using "partially addresses" now 😉 (But then it doesn't create the link between the issue and the PR apart from the mentions...)

PR #2591, opened 2023-06-21
preserve bitdepth when recompressing jxl

jonsneyers:
Partially fixes https://github.com/libjxl/libjxl/issues/2581

(it will still convert enum colorspace to icc, but at least the bit depth is preserved with this change)

On the API side, we currently don't allow JXL_BIT_DEPTH_FROM_CODESTREAM in combination with float buffers (we return "not implemented"), but I don't think there's a reason to disallow that. I think we can just define it to not have any impact on the interpretation of float buffers (they always use nominal range 0..1) and call it a day.

### DIFF
--- a/lib/extras/dec/jxl.h
+++ b/lib/extras/dec/jxl.h
@@ -55,7 +55,7 @@ struct JXLDecompressParams {
   bool unpremultiply_alpha = false;
 
   // Controls the effective bit depth of the output pixels.
-  JxlBitDepth output_bitdepth = {JXL_BIT_DEPTH_FROM_PIXEL_FORMAT, 0, 0};
+  JxlBitDepth output_bitdepth = {JXL_BIT_DEPTH_FROM_CODESTREAM, 0, 0};
 };
 
 bool DecodeImageJXL(const uint8_t* bytes, size_t bytes_size,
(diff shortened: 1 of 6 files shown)


lib/extras/dec/jxl.h
```diff
@@ -55,7 +55,7 @@ struct JXLDecompressParams {
   bool unpremultiply_alpha = false;
 
   // Controls the effective bit depth of the output pixels.
-  JxlBitDepth output_bitdepth = {JXL_BIT_DEPTH_FROM_PIXEL_FORMAT, 0, 0};
+  JxlBitDepth output_bitdepth = {JXL_BIT_DEPTH_FROM_CODESTREAM, 0, 0};
```
szabadka, line 58: this would break decoding 12-bit JXLs to PNG, see PR #2619

szabadka: it seems that we always rescale to full 8 or 16 bit in png writer, and that causes the small error in 12-bit pgm -> jxl -> png roundtrip, since it is not possible to exactly represent 12-bit images in 16-bit png, I think it is ok

szabadka (approved):

lib/extras/dec/jxl.h
jonsneyers: In principle we could use the `sBIT` chunk to indicate that it's actually a 12-bit png and make it roundtrip exactly, even if we do need to pad things to 16-bit. But anyway, that's kind of orthogonal to this PR :)

PR #2430, opened 2023-04-26
Migrate the GDK plugin from skcms to lcms2

sboukortt:

PR #2592, opened 2023-06-21
Do not require skcms when plugins are enabled

novomesk:


mo271 (approved):
Thanks!
Makes sense after https://github.com/libjxl/libjxl/pull/2430

mo271 (approved):
Thanks!
Makes sense after https://github.com/libjxl/libjxl/pull/2430

PR #2593, opened 2023-06-22
organize cjxl/djxl help a bit

jonsneyers:
Currently the help screens for cjxl/djxl are one long list and making the help more verbose shows more options in random places, which makes it hard to distinguish the basic options from the more advanced ones.

This reorders the options and adds some structure to it.

For `cjxl`, I made five tiers of options (so you need four `-v` to see them all). The first tier is always shown in the help screen and each additional `-v` 'unlocks' another tier.

- Basic options
- Advanced options
- Expert options
- Options for experimentation / benchmarking
- Modular mode options

Those last two tiers should not really be useful for normal users, they're intended more for encoder development than for actual use.

The full `cjxl` help looks like this then:

```
$ cjxl -v -v -v -v
JPEG XL encoder v0.9.0 87202f32 [SSE4,SSSE3,SSE2]
Usage: tools/cjxl INPUT OUTPUT [OPTIONS...]
 INPUT
    the input can be PNG, APNG, GIF, JPEG, EXR, PPM, PFM, PAM, PGX, or JXL
 OUTPUT
    the compressed JXL output file

Basic options:
 -d DISTANCE, --distance=DISTANCE
    Target visual distance in JND units, lower = higher quality.
    0.0 = mathematically lossless. Default for already-lossy input (JPEG/GIF).
    1.0 = visually lossless. Default for other input.
    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0. Mutually exclusive with --quality.
 -q QUALITY, --quality=QUALITY
    Quality setting, higher value = higher quality. This is internally mapped to --distance.
    100 = mathematically lossless. 90 = visually lossless.
    Quality values roughly match libjpeg quality.
    Recommended range: 68 .. 96. Allowed range: 0 .. 100. Mutually exclusive with --distance.
 -e EFFORT, --effort=EFFORT
    Encoder effort setting. Range: 1 .. 9.
    Default: 7. Higher numbers allow more computation at the expense of time.
    For lossless, generally it will produce smaller files.
    For lossy, higher effort should more accurately reach the target quality.
 -V, --version
    Print encoder library version number and exit.
 --quiet
    Be more silent
 -v, --verbose
    Verbose output; can be repeated and also applies to help (!).

Advanced options:
 -a A_DISTANCE, --alpha_distance=A_DISTANCE
    Target visual distance for the alpha channel, lower = higher quality.
    0.0 = mathematically lossless. 1.0 = visually lossless.
    Default is to use the same value as for the color image.
    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0.
 -p, --progressive
    Enable (more) progressive/responsive decoding.
 --group_order=0|1
    Order in which 256x256 groups are stored in the codestream for progressive rendering.
    0 = scanline order, 1 = center-first order. Default: 0.
 --container=0|1
    0 = Do not encode using container format (strip Exif/XMP/JPEG bitstream reconstruction data).
    1 = Force using container format. Default: use only if needed.
 --compress_boxes=0|1
    Disable/enable Brotli compression for metadata boxes. Default is 1 (enabled).
 --brotli_effort=B_EFFORT
    Brotli effort setting. Range: 0 .. 11.
    Default: 9. Higher number is more effort (slower).
 -m 0|1, --modular=0|1
    Use modular mode (default = encoder chooses, 0 = enforce VarDCT, 1 = enforce modular mode).
 -j 0|1, --lossless_jpeg=0|1
    If the input is JPEG, losslessly transcode JPEG, rather than using reencode pixels.
 --num_threads=N
    Number of worker threads (-1 == use machine default, 0 == do not use multithreading).
 --photon_noise_iso=ISO_FILM_SPEED
    Adds noise to the image emulating photographic film or sensor noise.
    Higher number = grainier image, e.g. 100 gives a low amount of noise,
    3200 gives a lot of noise. Default is 0.
 --intensity_target=N
    Upper bound on the intensity level present in the image, in nits.
    Default is 0, which means 'choose a sensible default value based on the color encoding.
 -x key=value, --dec-hints=key=value
    This is useful for 'raw' formats like PPM that do not contain colorspace information.
    The key 'color_space' indicates an enumerated ColorEncoding, for example:
      -x color_space=RGB_D65_SRG_Per_SRG is sRGB with perceptual rendering intent
      -x color_space=RGB_D65_202_Rel_PeQ is Rec.2100 PQ with relative rendering intent
    The key 'icc_pathname' refers to a binary file containing an ICC profile.

Expert options:
 --jpeg_store_metadata=0|1
    If --lossless_jpeg=1, store JPEG reconstruction metadata in the JPEG XL container.
    This allows reconstruction of the JPEG codestream. Default: 1.
 --codestream_level=K
    The codestream level. Either `-1`, `5` or `10`.
 --faster_decoding=0|1|2|3|4
    0 = default, higher values improve decode speed at the expense of quality or density.
 --premultiply=-1|0|1
    Force premultiplied (associated) alpha.
 --keep_invisible=0|1
    disable/enable preserving color of invisible pixels (default: 1 if lossless, 0 if lossy).
 --center_x=-1..XSIZE
    Determines the horizontal position of center for the center-first group order.
    Default -1 means 'middle of the image', values [0..xsize) set this to a particular coordinate.
 --center_y=-1..YSIZE
    Determines the vertical position of center for the center-first group order.
    Default -1 means 'middle of the image', values [0..ysize) set this to a particular coordinate.
 --progressive_ac
    Use the progressive mode for AC.
 --qprogressive_ac
    Use the progressive mode for AC with shift quantization.
 --progressive_dc=num_dc_frames
    Progressive-DC setting. Valid values are: -1, 0, 1, 2.
 --resampling=-1|1|2|4|8
    Resampling for color channels. Default of -1 applies resampling only for very low quality.
    1 = downsampling (1x1), 2 = 2x2 downsampling, 4 = 4x4 downsampling, 8 = 8x8 downsampling.
 --ec_resampling=-1|1|2|4|8
    Resampling for extra channels. Same as --resampling but for extra channels like alpha.
 --already_downsampled
    Do not downsample before encoding, but still signal that the decoder should upsample.
 --epf=-1|0|1|2|3
    Edge preserving filter level, 0-3. Default -1 means encoder chooses, 0-3 set a strength.
 --gaborish=0|1
    Force disable/enable the gaborish filter. Default is 'encoder chooses'
 --override_bitdepth=BITDEPTH
    Default is zero (use the input image bit depth); if nonzero, override the bit depth

Options for experimentation / benchmarking:
 --jpeg_reconstruction_cfl=0|1
    Enable/disable chroma-from-luma (CFL) for lossless JPEG reconstruction.
 --num_reps=N
    How many times to compress. (For benchmarking).
 --disable_output
    No output file will be written (for benchmarking)
 --dots=0|1
    Force disable/enable dots generation. (not provided = default, 0 = disable, 1 = enable).
 --patches=0|1
    Force disable/enable patches generation. (not provided = default, 0 = disable, 1 = enable).
 --frame_indexing=INDICES
    INDICES is of the form '^(0*|1[01]*)'. The i-th position indicates whether the
    i-th frame will be indexed in the frame index box.
 --allow_expert_options
    Allow specifying advanced options; this allows setting effort to 10, for
    somewhat better lossless compression at the cost of a massive speed hit.

Modular mode options:
 -I PERCENT, --iterations=PERCENT
    Percentage of pixels used to learn MA trees. Higher values use
    more encoder memory and can result in better compression. Default of -1 means
    the encoder chooses. Zero means no MA trees are used.
 -C K, --modular_colorspace=K
    Color transform: -1 = default (try several per group, depending
    on effort), 0 = RGB (none), 1-41 = fixed RCT (6 = YCoCg).
 -g K, --modular_group_size=K
    Group size: -1 = default (let the encoder choose),
    0 = 128x128, 1 = 256x256, 2 = 512x512, 3 = 1024x1024.
 -P K, --modular_predictor=K
    Predictor(s) to use: 0=zero, 1=left, 2=top, 3=avg0, 4=select,
    5=gradient, 6=weighted, 7=topright, 8=topleft, 9=leftleft, 10=avg1, 11=avg2, 12=avg3,
    13=toptop predictive average, 14=mix 5 and 6, 15=mix everything.
    Default is 14 at effort < 9 and 15 at effort 9.
 -E K, --modular_nb_prev_channels=K
    Number of extra (previous-channel) MA tree properties to use.
 --modular_palette_colors=K
    Use palette if number of colors is smaller than or equal to this.
 --modular_lossy_palette
    Use delta palette in a lossy way; it is recommended to also
    set --modular_palette_colors=0 with this option to use the default palette only.
 -X PERCENT, --pre-compact=PERCENT
    Use global channel palette if the number of sample values is smaller
    than this percentage of the nominal range. 
 -Y PERCENT, --post-compact=PERCENT
    Use local (per-group) channel palette if the number of sample values is
    smaller than this percentage of the nominal range.
 -R K, --responsive=K
    Do the Squeeze transform, 0=false, 1=true (default: 1 if lossy, 0 if lossless)
 -h, --help
    Prints this help message.
```

For `djxl`, I reduced it to three tiers of options:

- Basic options
- Advanced options
- Options for experimentation / benchmarking

That last tier should not really be useful for normal users.

The full `djxl` help screen looks like this then:

```
$ djxl -v -v 
JPEG XL decoder v0.9.0 87202f32 [SSE4,SSSE3,SSE2]
Usage: tools/djxl INPUT OUTPUT [OPTIONS...]
 INPUT
    The compressed input file (JXL). Use '-' for input from stdin.
 OUTPUT
    The output can be PNG, APNG, JPEG, EXR, PPM, PFM, or PAM. Use '-' for output to stdout.
    The output format is selected based on the extension or a prefix (e.g. 'png:-')

Basic options:
 -V, --version
    Print version number and exit.
 --quiet
    Silence output (except for errors).
 -v, --verbose
    Verbose output; can be repeated and also applies to help (!).

Advanced options:
 --num_threads=N
    Number of worker threads (-1 == use machine default, 0 == do not use multithreading).
 --bits_per_sample=N
    Sets the output bit depth. The value 0 (default for PNM) means the original (input) bit depth.
    The value -1 (default for other codecs) means it depends on the output format capabilities
    and the input bit depth (e.g. decoding a 12-bit image to PNG will produce a 16-bit PNG).
 --display_nits=N
    If set to a non-zero value, tone maps the image the given peak display luminance.
 --color_space=COLORSPACE_DESC
    Sets the desired output color space of the image. For example:
      --color_space=RGB_D65_SRG_Per_SRG is sRGB with perceptual rendering intent
      --color_space=RGB_D65_202_Rel_PeQ is Rec.2100 PQ with relative rendering intent
 -s 1|2|4|8, --downsampling=1|2|4|8
    If the input JXL stream is contains hints for target downsampling ratios,
    only decode what is needed to produce an image intended for this downsampling ratio.
 --allow_partial_files
    Allow decoding of truncated files.
 -j, --pixels_to_jpeg
    By default, if the input JXL is a recompressed JPEG file, djxl reconstructs that JPEG file.
    This flag causes the decoder to instead decode to pixels and encode a new (lossy) JPEG.
 -q N, --jpeg_quality=N
    Sets the JPEG output quality, default is 95. Setting this option implies --pixels_to_jpeg.

Options for experimentation / benchmarking:
 --num_reps=N
    Sets the number of times to decompress the image. Useful for benchmarking. Default is 1.
 --disable_output
    No output file will be written (for benchmarking)
 --norender_spotcolors
    Disables rendering of spot colors.
 --preview_out=FILENAME
    If specified, writes the preview image to this file.
 --icc_out=FILENAME
    If specified, writes the ICC profile of the decoded image to this file.
 --orig_icc_out=FILENAME
    If specified, writes the ICC profile of the original image to this file
    This can be different from the ICC profile of the decoded image if --color_space was specified.
 --metadata_out=FILENAME
    If specified, writes metadata info to a JSON file. Used by the conformance test script
 --print_read_bytes
    Print total number of decoded bytes.
 -h, --help
    Prints this help message.
```

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -1028,8 +1028,13 @@ int main(int argc, char** argv) {
     }
   }
   if (!args.quiet) {
-    cmdline.VerbosePrintf(0, "Compressed to %" PRIuS " bytes ",
-                          compressed.size());
+    if (compressed.size() < 100000) {
+      cmdline.VerbosePrintf(0, "Compressed to %" PRIuS " bytes ",
+                            compressed.size());
+    } else {
+      cmdline.VerbosePrintf(0, "Compressed to %.1f kB ",
+                            compressed.size() * 0.001);
+    }
     // For lossless jpeg-reconstruction, we don't print some stats, since we
     // don't have easy access to the image dimensions.
     if (args.container == jxl::Override::kOn) {
@@ -1040,9 +1045,7 @@ int main(int argc, char** argv) {
           static_cast<double>(compressed.size() * jxl::kBitsPerByte) / pixels;
       cmdline.VerbosePrintf(0, "(%.3f bpp%s).\n", bpp / ppf.frames.size(),
                             ppf.frames.size() == 1 ? "" : "/frame");
-      if (cmdline.verbosity > 0) {
-        JXL_CHECK(stats.Print(num_worker_threads));
-      }
+      JXL_CHECK(stats.Print(num_worker_threads));
     } else {
       cmdline.VerbosePrintf(0, "\n");
     }
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -94,21 +94,8 @@ struct CompressArgs {
                                  "the compressed JXL output file", &file_out);
 
     // Flags.
-    // TODO(lode): also add options to add exif/xmp/other metadata in the
-    // container.
-    cmdline->AddOptionValue(
-        '\0', "container", "0|1",
-        "0 = Do not encode using container format (strip "
-        "Exif/XMP/JPEG bitstream reconstruction data).\n"
-        "    1 = Force using container format. Default: use only if needed.",
-        &container, &ParseOverride, 1);
 
-    cmdline->AddOptionValue(
-        '\0', "jpeg_store_metadata", "0|1",
-        ("If --lossless_jpeg=1, store JPEG reconstruction "
-         "metadata in the JPEG XL container.\n"
-         "    This allows reconstruction of the JPEG codestream. Default: 1."),
-        &jpeg_store_metadata, &ParseUnsigned, 2);
+    cmdline->AddHelpText("\nBasic options:", 0);
 
     // Target distance/size/bpp
     opt_distance_id = cmdline->AddOptionValue(
@@ -120,15 +107,6 @@ struct CompressArgs {
         "    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0. "
         "Mutually exclusive with --quality.",
         &distance, &ParseFloat);
-
-    opt_alpha_distance_id = cmdline->AddOptionValue(
-        'a', "alpha_distance", "A_DISTANCE",
-        "Target visual distance for the alpha channel, lower = higher "
-        "quality.\n"
-        "    0.0 = mathematically lossless. 1.0 = visually lossless.\n"
-        "    Default is to use the same value as for the color image.\n"
-        "    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0.",
-        &alpha_distance, &ParseFloat, 1);
 
     // High-level options
     opt_quality_id = cmdline->AddOptionValue(
@@ -149,7 +127,48 @@ struct CompressArgs {
         "    For lossless, generally it will produce smaller files.\n"
         "    For lossy, higher effort should more accurately reach "
         "the target quality.",
-        &effort, &ParseUnsigned, -1);
+        &effort, &ParseUnsigned);
+
+    cmdline->AddOptionFlag('V', "version",
+                           "Print encoder library version number and exit.",
+                           &version, &SetBooleanTrue);
+    cmdline->AddOptionFlag('\0', "quiet", "Be more silent", &quiet,
+                           &SetBooleanTrue);
+    cmdline->AddOptionFlag('v', "verbose",
+                           "Verbose output; can be repeated and also applies "
+                           "to help (!).",
+                           &verbose, &SetBooleanTrue);
+
+    cmdline->AddHelpText("\nAdvanced options:", 1);
+
+    opt_alpha_distance_id = cmdline->AddOptionValue(
+        'a', "alpha_distance", "A_DISTANCE",
+        "Target visual distance for the alpha channel, lower = higher "
+        "quality.\n"
+        "    0.0 = mathematically lossless. 1.0 = visually lossless.\n"
+        "    Default is to use the same value as for the color image.\n"
+        "    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0.",
+        &alpha_distance, &ParseFloat, 1);
+
+    cmdline->AddOptionFlag('p', "progressive",
+                           "Enable (more) progressive/responsive decoding.",
+                           &progressive, &SetBooleanTrue, 1);
+
+    cmdline->AddOptionValue(
+        '\0', "group_order", "0|1",
+        "Order in which 256x256 groups are stored "
+        "in the codestream for progressive rendering.\n"
+        "    0 = scanline order, 1 = center-first order. Default: 0.",
+        &group_order, &ParseOverride, 1);
+
+    // TODO(lode): also add options to add exif/xmp/other metadata in the
+    // container.
+    cmdline->AddOptionValue(
+        '\0', "container", "0|1",
+        "0 = Do not encode using container format (strip "
+        "Exif/XMP/JPEG bitstream reconstruction data).\n"
+        "    1 = Force using container format. Default: use only if needed.",
+        &container, &ParseOverride, 1);
 
     cmdline->AddOptionValue('\0', "compress_boxes", "0|1",
                             "Disable/enable Brotli compression for metadata "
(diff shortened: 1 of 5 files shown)


mo271:
Might it be a good idea to get some spacing before 
 ```
 -h, --help
    Prints this help message.
  ```
  or put it somewhere further at the top?
  Also I think it would be good to mention somewhere when adding the `-v`  stabilizes (after 2 or 4 times). I think it is a bit silly for users to have to manually try out if `'-v -v -v -v' yields the same output as `-v -v -v -v -v `, and we should just tell them. Is there currently a difference between '-v' and '-v -v' for anything other than the help?
  
  Other than that LGTM

jonsneyers:
Updated it a bit. Now there's a newline before the `-h` text, and the text itself has been updated. When all options were shown, it now says that explicitly:

```
 -h, --help
    Prints this help message. All options are shown above.
```

while otherwise it says:

```
 -h, --help
    Prints this help message. Add -v to see more options.
```

Also the actual verbosity (when doing something other than showing help) has been made to depend on `-v`:

```
$ cjxl t.png t.jxl 
JPEG XL encoder v0.9.0 a25a6d9c [SSE4,SSSE3,SSE2]
Encoding [VarDCT, d1.000, effort: 7]
Compressed to 1104609 bytes (1.685 bpp).
$ djxl t.jxl t.png 
JPEG XL decoder v0.9.0 a25a6d9c [SSE4,SSSE3,SSE2]
Decoded to pixels.
```
versus
```
$ cjxl t.png t.jxl -v
JPEG XL encoder v0.9.0 a25a6d9c [SSE4,SSSE3,SSE2]
Read 2048x2560 image, 8117032 bytes, 110.4 MP/s
Encoding [VarDCT, d1.000, effort: 7]
Compressed to 1104212 bytes (1.685 bpp).
2048 x 2560, 4.03 MP/s [4.03, 4.03], 1 reps, 4 threads.
$ djxl t.jxl t.png -v
JPEG XL decoder v0.9.0 a25a6d9c [SSE4,SSSE3,SSE2]
Read 1104212 compressed bytes.
Decoded to pixels.
Wrote output to t.png
2048 x 2560, 70.47 MP/s [70.47, 70.47], 1 reps, 4 threads.
```

jonnyawsom3:
Part of me wonders if having the MP/s in the default verbosity of cjxl and djxl could be better. Then the average user knows the size, the efficiency and the speed of the operation at a glance, instead of just turning the effort up not realising the difference it makes.

On a side note, the MP/s has a minimum of 0.01Mp/s. Might be worth adding an extra decimal or two with the extra verbosity option for using effort 9/10

mo271:
> ```
>  -h, --help
>     Prints this help message. All options are shown above.
> ```
> 
> 
>     
>   
> 
> while otherwise it says:
> 
> 
> ```
>  -h, --help
>     Prints this help message. Add -v to see more options.
> ```

This still leaves the user guessing how many times `-v` can be added. Perhaps writing 
```
>     Prints this help message. Add -v (up to a total number of 4 times) to see more options.
```
or something like this would be easier

jonnyawsom3:
Maybe
`-h, --help
     Prints this help message. Add -v multiple times to see more options. (Basic, Advanced, Expert, Benchmarking, Modular Settings)`

monad0:
Why would you ever want to see more than the essential, but fewer than all options? I can't imagine seriously investigating the tool and saying "I'll just add two `-v`s and see what I get." One essential and one verbose level is quite justified: for a quick start relying on defaults, or for getting your hands dirty. With the improved readability, you can just skip over sections you're not interested in while having confidence you haven't missed anything.

If we must, something like https://github.com/libjxl/libjxl/pull/2593#issuecomment-1602523050 is what I imagined as an attempt at efficient communication, but sparing the user needing to count the number of sections.
```
-h, --help
     Prints this help message. Add multiple -v flags to include these sections:
     0 Basic, 1 Advanced, 2 Expert, 3 Benchmarking, 4 Modular Settings
```
Still, the granularity is awkward. It's not intuitive why I should want advanced options and not expert, or how many `-v`s I'll need next time when I want to reference a particular option.

mo271 (approved):
Clearly already an improvement over what we have now.

schrmh:
"Expert options:"
"--allow_expert_options"
The naming here is a bit unfortunate.

jonsneyers:
> By the way, the usage is including the path:
> 
> `Usage: tools/djxl INPUT OUTPUT [OPTIONS...]`
> 
> For mpv, ffmpeg, etc. no matter how you call them (`/usr/bin/mpv` or just `mpv`), they don't include the path:
> 
> `Usage: mpv [options] [url|path/]filename` `usage: ffmpeg [options] [[infile options] -i infile]... {[outfile options] outfile}...`
> 
> Is this intentional?

Both ways of doing it (showing a fixed name vs showing the actual `argv[0]` name) seem to be used. For example this is what libjpeg-turbo's `cjpeg` help looks like:

```
$ /usr/bin/cjpeg --help
usage: /usr/bin/cjpeg [switches] [inputfile]
Switches (names may be abbreviated):
  -quality N[,...]   Compression quality (0..100; 5-95 is most useful range,
                     default is 75)
  -grayscale     Create monochrome JPEG file
  -rgb           Create RGB JPEG file
  -optimize      Optimize Huffman table (smaller file, but slow compression)
  -progressive   Create progressive JPEG file
  -targa         Input file is Targa format (usually not needed)
Switches for advanced users:
  -arithmetic    Use arithmetic coding
  -dct int       Use accurate integer DCT method (default)
  -dct fast      Use less accurate integer DCT method [legacy feature]
  -dct float     Use floating-point DCT method [legacy feature]
  -icc FILE      Embed ICC profile contained in FILE
  -restart N     Set restart interval in rows, or in blocks with B
  -smooth N      Smooth dithered input (N=1..100 is strength)
  -maxmemory N   Maximum memory to use (in kbytes)
  -outfile name  Specify name for output file
  -memdst        Compress to memory instead of file (useful for benchmarking)
  -report        Report compression progress
  -strict        Treat all warnings as fatal
  -verbose  or  -debug   Emit debug output
  -version       Print version information and exit
Switches for wizards:
  -baseline      Force baseline quantization tables
  -qtables FILE  Use quantization tables given in FILE
  -qslots N[,...]    Set component quantization tables
  -sample HxV[,...]  Set component sampling factors
  -scans FILE    Create multi-scan JPEG per script FILE
```

I don't have a strong preference but I think using the `argv[0]` name like we do has some minor advantages, e.g. if you have multiple versions in different locations, or if you want to rename the binary for some reason (say you want to call it `png2jxl` instead of `cjxl`).


> Why would you ever want to see more than the essential, but fewer than all options?

Long help output that requires you to use `| less` or scroll up is not very convenient imo, especially if the options you're most likely to need are closer to the top. Some of the options like `--disable_output` and `--num_reps` are just not useful at all for people who just want to use the tools. Splitting it up in multiple tiers helps to avoid polluting the screen with output that is likely not useful anyway.

For `cjxl`, the tiers are something like this:

- Basic: the essentials (quality and effort). These options are fundamental, since encoding is all about the trade-off between quality, compression and speed. So this is about the core functionality of the tool.
- Advanced: additional options that you don't really need to set, i.e. the defaults should be fine for 95% of use cases, but you still sometimes might need them. This is still about functionality: what kind of bitstream to produce, how many threads to use, etc.
- Expert: these options are not really about functionality anymore but more about influencing internal encoder choices (though some are arguably a bit of both). For 99.9% of use cases, there is no need whatsoever to use these options.
- Experimentation / benchmarking: these options are not really meant for users but rather for encoder devs. These options could be removed or work differently in the future. They might evolve into something that will become an Advanced or Expert option (e.g. frame indexing, but then it will be something more user-friendly then than what it is now)
- Modular mode options: this is not about functionality at all and only about influencing internal encoder choices (specifically for modular mode). Not meant for users, almost impossible to know what these options even mean if you don't have a good understanding of how the format works. The goal is that 'effort' is the only thing the user has to set and the encoder then makes choices that make sense.

For `djxl` there are only 3 tiers:

- Basic: the basic usage is just `djxl INPUT OUTPUT` and besides knowing what output formats are available and how to specify them, there isn't really anything else you need for most use cases.
- Advanced: these are basically all the options meant for users.
- Experimentation / benchmarking: these options are not useful for users. They are only really there for benchmarking, conformance testing, and experimentation.

So basically `cjxl` / `djxl` gives a quick start guide, and `cjxl -v` / `djxl -v` gives the more detailed options. Any additional `-v` are things you very likely don't need, and that would e.g. very likely not be exposed or configurable in applications that use libjxl (say in a "save as jxl" dialog box in Gimp or Photoshop).

PR #2594, opened 2023-06-22
add lcms2 test to workflow

mo271:


lib/jxl.cmake
```diff
@@ -303,7 +307,7 @@ endif()  # BUILD_SHARED_LIBS
 set(JPEGXL_LIBRARY_REQUIRES
     "libhwy libbrotlienc libbrotlidec")
 if(NOT JPEGXL_ENABLE_SKCMS)
-  set(JPEGXL_LIBRARY_REQUIRES "${JPEGXL_LIBRARY_REQUIRES} lcms2")
+  set(JPEGXL_LIBRARY_REQUIRES "${JPEGXL_LIBRARY_REQUIRES} lcms::lcms2")
```
sboukortt, line 310:
```suggestion
  set(JPEGXL_LIBRARY_REQUIRES "${JPEGXL_LIBRARY_REQUIRES} lcms2")
```

This variable is used for `pkg-config`, where lcms is always `lcms2`.

sboukortt (changes requested):

examples/CMakeLists.txt
```diff
@@ -50,7 +50,13 @@ set_target_properties(StaticJxlThreads PROPERTIES
 add_executable(decode_oneshot_static decode_oneshot.cc)
 target_link_libraries(decode_oneshot_static
   -static StaticJxl StaticJxlThreads)
+if (NOT JPEGXL_ENABLE_SKCMS)
+  target_link_libraries(decode_oneshot_static -static lcms2)
```
sboukortt, line 54: This `-static` should be unnecessary since it’s already above. It doesn’t specifically affect the linking of lcms2 as far as CMake is concerned, it just adds `-static` to the linker flags.

third_party/lcms2.cmake
```diff
@@ -41,6 +41,7 @@ add_library(lcms2 STATIC EXCLUDE_FROM_ALL
   lcms/src/cmsxform.c
   lcms/src/lcms2_internal.h
 )
+
```
mo271, line 44:
```suggestion
```

mo271:
these kind of test were already added in https://github.com/libjxl/libjxl/pull/2916

Issue #2433, opened 2023-04-26
0.8.1: test suite is failing in ~`7% of the test units


kloczek:
**Describe the bug**

Looks lik esomething is wrong with test suite or I'm doing something wrongl;y.
On test suite exetion I have I have ~17% of failing units.

**To Reproduce**

cmake settings (`cmake -L` output):
<details>

```console
-- Cache values
ASCIIDOC:FILEPATH=/usr/bin/a2x
BASH_PROGRAM:FILEPATH=/usr/bin/bash
BUILD_SHARED_LIBS:BOOL=ON
BUILD_TESTING:BOOL=ON
CMAKE_BUILD_TYPE:STRING=RelWithDebInfo
CMAKE_INSTALL_PREFIX:PATH=/usr
GMOCK_INCLUDE_DIR:PATH=/usr/include
GTest_DIR:PATH=/usr/lib64/cmake/GTest
JPEGXL_BUNDLE_LIBPNG:BOOL=OFF
JPEGXL_BUNDLE_SKCMS:BOOL=ON
JPEGXL_DEP_LICENSE_DIR:STRING=
JPEGXL_ENABLE_BENCHMARK:BOOL=OFF
JPEGXL_ENABLE_BOXES:BOOL=ON
JPEGXL_ENABLE_COVERAGE:BOOL=OFF
JPEGXL_ENABLE_DEVTOOLS:BOOL=ON
JPEGXL_ENABLE_DOXYGEN:BOOL=ON
JPEGXL_ENABLE_EXAMPLES:BOOL=ON
JPEGXL_ENABLE_FUZZERS:BOOL=
JPEGXL_ENABLE_JNI:BOOL=ON
JPEGXL_ENABLE_JPEGLI:BOOL=ON
JPEGXL_ENABLE_JPEGLI_LIBJPEG:BOOL=ON
JPEGXL_ENABLE_MANPAGES:BOOL=ON
JPEGXL_ENABLE_OPENEXR:BOOL=ON
JPEGXL_ENABLE_PLUGINS:BOOL=OFF
JPEGXL_ENABLE_PROFILER:BOOL=OFF
JPEGXL_ENABLE_SIZELESS_VECTORS:BOOL=OFF
JPEGXL_ENABLE_SJPEG:BOOL=OFF
JPEGXL_ENABLE_SKCMS:BOOL=OFF
JPEGXL_ENABLE_TCMALLOC:BOOL=ON
[...]

PR #2596, opened 2023-06-22
Make sure to read the rendering intent before the CICP tag

sboukortt:
The code that reads the CICP tag exits the function if it succeeds, but it should not skip reading the rendering intent, so make sure to have already done it by that point.


mo271 (approved):

mo271:
at least a partial fix for https://github.com/libjxl/libjxl/issues/2433

PR #2603, opened 2023-06-26
Make extras io-agnostic

eustas:


jonsneyers (approved):

jonsneyers:
Thanks, this makes sense: leave the file io stuff to `tools/` and have the functions in `lib/` work with memory buffers only...

PR #2604, opened 2023-06-26
Use shared extras codec library for cjxl/djxl if available.

szabadka:


szabadka:
I just noticed that we build this shared library, but never use it and I wanted to see if we can actually use it. I would also be fine with just always statically link it to djxl/cjxl and not use it for anything else.

jonsneyers:
I am in favor of dynamically linking it, it's not just used by cjxl/djxl but by basically everything in `tools` (e.g. the metrics, the hdr tools, etc). Their binary sizes could probably be reduced significantly by factoring out the codec io to a shared library.

szabadka:
> I am in favor of dynamically linking it, it's not just used by cjxl/djxl but by basically everything in `tools` (e.g. the metrics, the hdr tools, etc). Their binary sizes could probably be reduced significantly by factoring out the codec io to a shared library.

That would be the long-term plan, yes, but since the other tools still use CodecInOut, we just statically link the jxl and jxl_extras libraries to them.

jonsneyers (approved):

PR #2611, opened 2023-06-27
Factor out the libjpeg-decoding parts of test_utils.cc

szabadka:
This is preparation for using jpegli's own jpeg headers for compiling the jpegli-static library.


mo271 (approved):
great!

PR #2612, opened 2023-06-28
simplify/fix local max MA tree size calculation

jonsneyers:
Probably doesn't make any difference on realistic examples, and it's a bit nitpicky, but in principle: if `size_t` is small (it can be only uint32) and the size of a local group is very large (e.g. because the image is a multispectral one with lots of channels), presumably the computation of the bound to use for the MA tree size could return failure while the bitstream was actually valid.

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -515,8 +515,7 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   const ANSCode *code = &code_storage;
   const std::vector<uint8_t> *context_map = &context_map_storage;
   if (!header.use_global_tree) {
-    size_t max_tree_size = 1024;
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/encoding/encoding.cc
```diff
@@ -516,23 +516,20 @@ Status ModularDecode(BitReader *br, Image &image, GroupHeader &header,
   const std::vector<uint8_t> *context_map = &context_map_storage;
   if (!header.use_global_tree) {
     size_t max_tree_size = 1024;
```
szabadka, line 518: this could be uint64_t and then we don't even need if (max_tree_size < pixels ...., and at the end we could have the std::min as before

jonsneyers: Yes, that also works and is even simpler.

szabadka (approved):

PR #2620, opened 2023-06-29
conditionally compile code for standalone-fjxl

jonsneyers:
The code for using `enc_fast_lossless` in a standalone way to produce a complete jxl file does not have to be in libjxl. This is the high-level function that produces a full file, and also the code that produces an image header (which is unused in libjxl).
Also removing `JxlFastLosslessMaxRequiredOutput()` since it's not really needed imo.

Slightly reduces `libjxl.so` binary size.

### DIFF
--- a/lib/jxl/enc_fast_lossless.cc
+++ b/lib/jxl/enc_fast_lossless.cc
@@ -202,6 +202,11 @@ size_t JxlFastLosslessOutputSize(const JxlFastLosslessFrameState* frame) {
   return frame->header.bytes_written + total_size_groups;
 }
 
+size_t JxlFastLosslessMaxRequiredOutput(
+    const JxlFastLosslessFrameState* frame) {
+  return JxlFastLosslessOutputSize(frame) + 32;
+}
+
 void JxlFastLosslessPrepareHeader(JxlFastLosslessFrameState* frame,
                                   int add_image_header, int is_last) {
   BitWriter* output = &frame->header;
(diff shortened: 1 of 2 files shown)


veluca93:
> Also removing JxlFastLosslessMaxRequiredOutput() since it's not really needed imo.

I disagree on that :) It is useful for people that might want to control how the allocation happens (i.e. mmapped files)

jonsneyers:
> > Also removing JxlFastLosslessMaxRequiredOutput() since it's not really needed imo.
> 
> I disagree on that :) It is useful for people that might want to control how the allocation happens (i.e. mmapped files)

Yes, but it's basically the same thing as `JxlFastLosslessOutputSize()`. Do we really need both?

veluca93:
> > > Also removing JxlFastLosslessMaxRequiredOutput() since it's not really needed imo.
> > 
> > 
> > I disagree on that :) It is useful for people that might want to control how the allocation happens (i.e. mmapped files)
> 
> Yes, but it's basically the same thing as `JxlFastLosslessOutputSize()`. Do we really need both?

Yes, and the comment explains why. That +32 is an implementation detail that might change in the future.

jonsneyers:
> > > > Also removing JxlFastLosslessMaxRequiredOutput() since it's not really needed imo.
> > > 
> > > 
> > > I disagree on that :) It is useful for people that might want to control how the allocation happens (i.e. mmapped files)
> > 
> > 
> > Yes, but it's basically the same thing as `JxlFastLosslessOutputSize()`. Do we really need both?
> 
> Yes, and the comment explains why. That +32 is an implementation detail that might change in the future.

In libjxl we only use `JxlFastLosslessOutputSize()`, and this implementation detail (minimum available output size of 32) is already mentioned in the comment for `JxlFastLosslessWriteOutput()` and has to be taken into account explicitly anyway by applications that want to use the low-level fjxl api (like we do in `encode.cc`)...

veluca93:
That 32 is not necessarily the same 32, and nothing says that these numbers
will not change in the future.

jonsneyers:
OK, so I'll keep it. But if libjxl isn't using that function, is that a bug in how libjxl is using it?

veluca93:
Not necessarily - IIRC libjxl only produces output in chunks from fjxl.

veluca93 (approved):

jonsneyers:
(this one needs a re-approve, had a silly typo in the one that was approved)

mo271 (approved):

PR #2621, opened 2023-06-29
add --noise to cjxl and make the noise test actually test noise

jonsneyers:
We have an option to enable adaptive noise synthesis in the encode API, but it doesn't exist in `cjxl`. Even though it's still experimental, it's nice to expose it.

Also updating the test for noise to actually test noise by picking a different image. The image that was used was too clean to trigger the noise detection, so it ended up early-aborting and not signaling any noise on that image (which caused most of the code in `enc_noise.cc` to not be reached). This other image does lead to some noise being generated.

### DIFF
--- a/tools/cjxl_main.cc
+++ b/tools/cjxl_main.cc
@@ -322,6 +322,12 @@ struct CompressArgs {
 
     cmdline->AddHelpText("\nOptions for experimentation / benchmarking:", 3);
 
+    cmdline->AddOptionValue('\0', "noise", "0|1",
+                            "Force disable/enable adaptive noise generation "
+                            "(experimental). Default "
+                            "is 'encoder chooses'",
+                            &noise, &ParseOverride, 3);
+
     cmdline->AddOptionValue(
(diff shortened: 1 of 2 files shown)


tools/cjxl_main.cc
```diff
@@ -322,6 +322,12 @@ struct CompressArgs {
 
     cmdline->AddHelpText("\nOptions for experimentation / benchmarking:", 3);
 
+    cmdline->AddOptionValue('\0', "noise", "0|1",
+                            "Force disable/enable adaptive noise generation "
+                            "(experimental). Default "
+                            "is 'encoder chooses'",
```
sboukortt, line 328: Isn’t the default “no”?

jonsneyers: Yes, the current encoder choice is always "no" :).

In the API the default is also described as "encoder chooses". At some point the encoder choice might become something a bit more interesting than just "always choose no".

sboukortt: Eh, fair, I suppose 😁

sboukortt (approved):

PR #2623, opened 2023-06-29
put code for ac strategy debug images behind a compile flag

jonsneyers:
There is quite a bit of code in `enc_ac_strategy.cc` that is just there to produce a nice debug image that shows the ac strategy (dct block type selection). This functionality is not exposed via the library api, but it can still be accessed via `benchmark_xl`. Probably better not to compile this into `libjxl.so` — if we want to use this debugging aid, we can easily enough set `JXL_DEBUG_AC_STRATEGY=1` (just like there's e.g. `JXL_DEBUG_DOT_DETECT`).

### DIFF
--- a/lib/jxl/enc_ac_strategy.cc
+++ b/lib/jxl/enc_ac_strategy.cc
@@ -44,12 +44,18 @@
 // sensitive to some kind of degradation. Unfortunately image quality
 // is still more of an art than science.
 
+// Set JXL_DEBUG_AC_STRATEGY to 1 to enable debugging.
+#ifndef JXL_DEBUG_AC_STRATEGY
+#define JXL_DEBUG_AC_STRATEGY 0
+#endif
+
 // This must come before the begin/end_target, but HWY_ONCE is only true
 // after that, so use an "include guard".
 #ifndef LIB_JXL_ENC_AC_STRATEGY_
 #define LIB_JXL_ENC_AC_STRATEGY_
 // Parameters of the heuristic are marked with a OPTIMIZE comment.
 namespace jxl {
+namespace {
 
 // Debugging utilities.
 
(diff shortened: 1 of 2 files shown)


lib/jxl/enc_ac_strategy.cc
```diff
@@ -44,6 +44,13 @@
 // sensitive to some kind of degradation. Unfortunately image quality
 // is still more of an art than science.
 
+// Set JXL_DEBUG_AC_STRATEGY to 1 to enable debugging.
+#ifndef JXL_DEBUG_AC_STRATEGY
+#define JXL_DEBUG_AC_STRATEGY 0
+#endif
+
+#if JXL_DEBUG_AC_STRATEGY
```
veluca93, line 52: Let's not use `#if` for this purpose

jonsneyers: What is the cleaner way to make sure these debug functions (`TypeColor()`, `TypeMask()`, `DumpAcStrategy()`) do not end up getting compiled? Define them anyway, put the call to `DumpAcStrategy()` behind an `if (constexpr) {}` and hope the compiler will remove the dead code?

veluca93: yes, the compiler is usually pretty good at doing that (at least if the functions are in an anonymous namespace), and that ensure that the function will still compile when needed.

jonsneyers: Makes sense, updated it.
These functions are in `namespace jxl`, should that be updated to an anonymous one?

jonsneyers: I don't really see why that debug code was in `namespace jxl` in the first place, so just turned it into an anonymous namespace.

veluca93 (approved):

lib/jxl/enc_ac_strategy.cc
jonsneyers: Oops, the `namespace jxl` was there for a reason after all, made it a nested anonymous namespace.

veluca93 (approved):

PR #2626, opened 2023-06-30
Various cleanups

jonsneyers:
Various cleanups all aimed at reducing the amount of unnecessary strings and debug code in libjxl.so:

1. JXL_ABORT() is often used in unreachable code paths that could only become reachable if the code would change (e.g. if the encoder would use a new blendmode for patches, or some programming error happened that caused a presumably unreachable path to become reachable). However, the compiler cannot know these are unreachable so even release builds still contain the code to print the abort message and to do the actual abort. This replaces these aborts with JXL_NOT_REACHABLE() which is a new macro that keeps the old behavior for debug builds (print message and abort), but in release builds it simply signals JXL_UNREACHABLE to inform the compiler that this is in fact an unreachable path.

2. Butteraugli had some code for generic convolutions that aren't actually used (and there was a printf warning). Replaced it with JXL_NOT_REACHABLE().

3. Removed deprecated dead code from `butteraugli.*`

4. Adaptive quantization debug stuff (dumping debug images and printf'ing stuff) ended up getting compiled in even in release builds. Putting it behind a compile flag `JXL_DEBUG_ADAPTIVE_QUANTIZATION`.

5. There are a bunch of printf strings in enc_aux_out that also end up in libjxl.so. Putting these behind `#if JXL_DEBUG_V_LEVEL > 0`.

### DIFF
--- a/lib/jxl/butteraugli/butteraugli.h
+++ b/lib/jxl/butteraugli/butteraugli.h
@@ -43,49 +43,6 @@ struct ButteraugliParams {
   float intensity_target = 80.0f;
 };
 
-// ButteraugliInterface defines the public interface for butteraugli.
-//
-// It calculates the difference between rgb0 and rgb1.
-//
-// rgb0 and rgb1 contain the images. rgb0[c][px] and rgb1[c][px] contains
-// the red image for c == 0, green for c == 1, blue for c == 2. Location index
-// px is calculated as y * xsize + x.
-//
-// Value of pixels of images rgb0 and rgb1 need to be represented as raw
-// intensity. Most image formats store gamma corrected intensity in pixel
-// values. This gamma correction has to be removed, by applying the following
-// function to values in the 0-1 range:
-// butteraugli_val = pow(input_val, gamma);
-// A typical value of gamma is 2.2. It is usually stored in the image header.
-// Take care not to confuse that value with its inverse. The gamma value should
-// be always greater than one.
-// Butteraugli does not work as intended if the caller does not perform
-// gamma correction.
-//
-// hf_asymmetry is a multiplier for penalizing new HF artifacts more than
-// blurring away features (1.0 -> neutral).
-//
-// diffmap will contain an image of the size xsize * ysize, containing
-// localized differences for values px (indexed with the px the same as rgb0
-// and rgb1). diffvalue will give a global score of similarity.
-//
-// A diffvalue smaller than kButteraugliGood indicates that images can be
-// observed as the same image.
-// diffvalue larger than kButteraugliBad indicates that a difference between
-// the images can be observed.
-// A diffvalue between kButteraugliGood and kButteraugliBad indicates that
-// a subtle difference can be observed between the images.
-//
-// Returns true on success.
-bool ButteraugliInterface(const Image3F &rgb0, const Image3F &rgb1,
-                          const ButteraugliParams &params, ImageF &diffmap,
-                          double &diffvalue);
-
-// Deprecated (calls the previous function)
-bool ButteraugliInterface(const Image3F &rgb0, const Image3F &rgb1,
-                          float hf_asymmetry, float xmul, ImageF &diffmap,
-                          double &diffvalue);
-
 // Converts the butteraugli score into fuzzy class values that are continuous
 // at the class boundary. The class boundary location is based on human
 // raters, but the slope is arbitrary. Particularly, it does not reflect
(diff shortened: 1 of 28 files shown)


veluca93:
Why do we need a new macro JXL_NOT_REACHABLE? Couldn't we just use
JXL_UNREACHABLE?

jonsneyers:
You mean change JXL_UNREACHABLE to take a message argument and also abort in debug builds?
Yes, that could also be done. I kept the existing macro like it was because perhaps there are cases where you just want to signal to the compiler that a path is unreachable, without writing a specific debug message and aborting.

veluca93:
> You mean change JXL_UNREACHABLE to take a message argument and also abort in debug builds? Yes, that could also be done. I kept the existing macro like it was because perhaps there are cases where you just want to signal to the compiler that a path is unreachable, without writing a specific debug message and aborting.

Considering actually reaching a `__builtin_unreachable()` statement is undefined behaviour, I cannot imagine a single valid reason not to want to debug-message+abort just before that.

veluca93:
I think it would cause less future confusion if the macros were called JXL_UNREACHABLE and JXL_UNREACHABLE_BUILTIN instead.

jonsneyers:
That makes sense. Renamed the macros accordingly.

Also made it so the debug messages of JXL_CHECK (which can be quite long in some cases) do not end up in release builds.

lib/jxl/butteraugli/butteraugli.h
```diff
@@ -43,49 +43,6 @@ struct ButteraugliParams {
...
-//
-// A diffvalue smaller than kButteraugliGood indicates that images can be
-// observed as the same image.
-// diffvalue larger than kButteraugliBad indicates that a difference between
-// the images can be observed.
-// A diffvalue between kButteraugliGood and kButteraugliBad indicates that
-// a subtle difference can be observed between the images.
-//
-// Returns true on success.
-bool ButteraugliInterface(const Image3F &rgb0, const Image3F &rgb1,
```
veluca93, line 80: We have some internal users for this API, please leave it around.

lib/jxl/enc_adaptive_quantization.cc
```diff
@@ -877,15 +883,15 @@ void FindBestQuantization(const ImageBundle& linear, const Image3F& opsin,
...
   }
   for (int i = 0; i < iters + 1; ++i) {
-    if (FLAGS_dump_quant_state) {
-      printf("\nQuantization field:\n");
-      for (size_t y = 0; y < quant_field.ysize(); ++y) {
-        for (size_t x = 0; x < quant_field.xsize(); ++x) {
-          printf(" %.5f", quant_field.Row(y)[x]);
-        }
-        printf("\n");
+#if JXL_DEBUG_ADAPTIVE_QUANTIZATION
```
veluca93, line 886: Let's avoid `#if` to skip compiling code and use `if (some_constexpr_boolean)` instead.

lib/jxl/enc_aux_out.cc
```diff
@@ -100,6 +102,7 @@ void AuxOut::Assimilate(const AuxOut& victim) {
 }
 
 void AuxOut::Print(size_t num_inputs) const {
+#if JXL_DEBUG_V_LEVEL > 0
```
veluca93, line 105: Ideally, let's do the same thing here (`if (JXL_DEBUG_V_LEVEL > 0)`)

lib/jxl/butteraugli/butteraugli.h
jonsneyers: OK, but probably it shouldn't be in `lib/jxl/` if it isn't used by libjxl (or anything in this repo, even).
Does the deprecated function below also need to be kept or can that one be removed?

veluca93: Let's leave it around for now, and discuss further with Jyrki.

jonsneyers:
Also reverted the manual unrolling in jpeg reconstruction from https://github.com/libjxl/libjxl/pull/2534 (see https://github.com/libjxl/libjxl/pull/2534#discussion_r1221643355), since it was adding ~100kb of binary size for a relatively small speedup in a relatively niche code path (reconstructing non-progressive jpegs).

lib/jxl/enc_aux_out.cc
jonsneyers: done

lib/jxl/enc_adaptive_quantization.cc
jonsneyers: ok — there are probably quite a few other spots where we were already doing that, feel free to update those too :)

lib/jxl/butteraugli/butteraugli.h
jonsneyers: ok, kept butteraugli as it was except for the not-implemented-kernel-size thing.

veluca93:
> Also reverted the manual unrolling in jpeg reconstruction from #2534 (see [#2534 (comment)](https://github.com/libjxl/libjxl/pull/2534#discussion_r1221643355)), since it was adding ~100kb of binary size for a relatively small speedup in a relatively niche code path (reconstructing non-progressive jpegs).

Can you split that in another pull request, and sent that to @szabadka ?

jonsneyers:
> > Also reverted the manual unrolling in jpeg reconstruction from #2534 (see [#2534 (comment)](https://github.com/libjxl/libjxl/pull/2534#discussion_r1221643355)), since it was adding ~100kb of binary size for a relatively small speedup in a relatively niche code path (reconstructing non-progressive jpegs).
> 
> Can you split that in another pull request, and sent that to @szabadka ?

sure

veluca93 (approved):

PR #2627, opened 2023-07-01
disable SSSE3 and AVX512 in default builds

jonsneyers:
According to https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam (see "Other Settings"), 99.18% of desktop hardware has SSE4 (and 100% has SSE2), so it is a bit unnecessary to have the highway target for SSSE3 enabled since it will only bring a small speedup to < 1% of people while the binary size bloat is for everyone.

Also in this survey we see 90% has AVX2 but less than 10% has AVX512. While of course AVX512 does give additional speedups over AVX2, the question is whether it is worth the binary size bloat (especially because there are two variants of this, the regular one and the Zen4 one). Also I'm assuming the steam survey will contain relatively more beefy gaming machines so it's likely an overestimate of the actual support for AVX512 in the consumer market. In the server market, AVX512 probably does have wider support, but imo we shouldn't aim our default build choices on that — these use cases can make optimized custom builds.

This PR disables the SSSE3, AVX512 and AVX512_ZEN4 targets by default. There are new cmake flags JPEGXL_ENABLE_AVX512 and JPEGXL_ENABLE_AVX512_ZEN4 to enable one or both of the AVX512 flavors, but it's opt-in now. For SSSE3, I didn't bother to add a cmake flag to keep it since the number of affected people is small and is only going to shrink even further.

So the 90% of people with AVX2 will get that, and of the remaining 10% who don't have AVX2, 90% has SSE4 and will get that, and the remaining 10% (1% of the total) will still get SSE2.

This makes the stripped x64 binary size of the default release build of `libjxl.so` (i.e. what will end up going into packages in linux distributions etc) go down from 4.26 MB to 2.83 MB (-33%).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,10 @@ set(JPEGXL_FORCE_NEON false CACHE BOOL
     "Set flags to enable NEON in arm if not enabled by your toolchain.")
 set(JPEGXL_TEST_TOOLS false CACHE BOOL
     "Run scripts that test the encoding / decoding tools.")
+set(JPEGXL_ENABLE_AVX512 false CACHE BOOL
+    "Build with AVX512 support (faster on CPUs that support it, but larger binary size).")
+set(JPEGXL_ENABLE_AVX512_ZEN4 false CACHE BOOL
+    "Build with Zen4-optimized AVX512 support (faster on CPUs that support it, but larger binary size).")
 
 # Force system dependencies.
 set(JPEGXL_FORCE_SYSTEM_BROTLI false CACHE BOOL
@@ -205,6 +209,20 @@ endif()
 
 message(STATUS
     "Compiled IDs C:${CMAKE_C_COMPILER_ID}, C++:${CMAKE_CXX_COMPILER_ID}")
+
+# Always disable SSSE3 since it is rare to have SSSE3 but not SSE4
+set(HWY_DISABLED_TARGETS "HWY_SSSE3")
+if (NOT JPEGXL_ENABLE_AVX512)
+  message(STATUS "Disabled AVX512 (set JPEGXL_ENABLE_AVX512 to enable it)")
+  set(HWY_DISABLED_TARGETS "${HWY_DISABLED_TARGETS}|HWY_AVX3")
+  add_definitions(-DFJXL_ENABLE_AVX512=0)
+endif()
+if (NOT JPEGXL_ENABLE_AVX512_ZEN4)
+  message(STATUS "Disabled AVX512_ZEN4 (set JPEGXL_ENABLE_AVX512_ZEN4 to enable it)")
+  set(HWY_DISABLED_TARGETS "${HWY_DISABLED_TARGETS}|HWY_AVX3_ZEN4")
+endif()
+
+
 
 # CMAKE_EXPORT_COMPILE_COMMANDS is used to generate the compilation database
 # used by clang-tidy.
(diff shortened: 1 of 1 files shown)


veluca93:
Do you have numbers for the speedup of AVX512 over AVX2?

jonsneyers:
I will do some speed measurements on Monday. I cannot test AVX512_Zen4 vs AVX512 though.

(also worth noting: fewer targets to build for also makes compilation faster, including in the CI)

veluca93:
> I will do some speed measurements on Monday. I cannot test AVX512_Zen4 vs AVX512 though.
> 
> (also worth noting: fewer targets to build for also makes compilation faster, including in the CI)

That's fair :) I don't think there will be a huge difference between the two.

jonsneyers:
With AVX512:
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d1           5760   370608    0.5147333   2.973  54.017   1.34597058  86.70561393   0.53598375  0.275888701782      0
jxl:d2           5760   239632    0.3328222   3.013  46.844   2.32679752  81.07308832   0.80932365  0.269360897092      0
jxl:d3           5760   169373    0.2352403   2.801  41.338   3.21663737  75.09191040   1.07429072  0.252716448356      0
jxl:d1:3         5760   409832    0.5692111  22.734  55.766   1.39670677  87.51147695   0.54275955  0.308944764944      0
jxl:d1:4         5760   415743    0.5774208  20.528  53.878   1.39571948  87.79915471   0.53795090  0.310624056258      0
Aggregate:       5760   303212    0.4211283   6.512  50.057   1.81392751  83.48616026   0.67104116  0.282594418823      0
```

Without AVX512 (but still AVX2):
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2        pnorm       BPP*pnorm   Bugs
-------------------------------------------------------------------------------------------------------------------------
jxl:d1           5760   370608    0.5147333   2.685  51.919   1.34597439  86.68902590   0.53598435  0.275889009746      0
jxl:d2           5760   239632    0.3328222   2.721  45.963   2.32680049  81.10690949   0.80932500  0.269361345161      0
jxl:d3           5760   169373    0.2352403   2.550  41.262   3.21663307  75.09711463   1.07429226  0.252716810071      0
jxl:d1:3         5760   409832    0.5692111  23.197  56.041   1.39670165  87.46158626   0.54275988  0.308944955020      0
jxl:d1:4         5760   415743    0.5774208  20.793  54.603   1.39571568  87.77881145   0.53795124  0.310624251221      0
Aggregate:       5760   303212    0.4211283   6.176  49.637   1.81392620  83.47769612   0.67104189  0.282594727072      0
```

(these are single-core timings on a Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz)

So there clearly is some benefit of AVX512 over AVX2, but the difference is not huge, and the impact is mostly on encode speed, not so much on decode speed.

jonsneyers:
So I guess the question is if this 5-10% encode speedup for < 10% of users is worth the binary bloat for 100% of users. I don't have a strong opinion on it. I feel like for a default build it's not really worth the binary bloat, and in any case having a cmake flag for it (whether it's default-on or default-off) is convenient imo for those who disagree with whatever conclusion we arrive at.

veluca93:
How many repetitions did you use? (encode_reps and decode_reps)

jonsneyers:
This was 10 encode reps, 30 decode reps. I can do a more accurate timing if you want.

CMakeLists.txt
```diff
@@ -206,6 +210,20 @@ endif()
 message(STATUS
     "Compiled IDs C:${CMAKE_C_COMPILER_ID}, C++:${CMAKE_CXX_COMPILER_ID}")
 
+# Always disable SSSE3 since it is rare to have SSSE3 but not SSE4
+set(HWY_DISABLED_TARGETS "HWY_SSSE3")
```
veluca93, line 214: Doesn't this override user settings?

veluca93:
> This was 10 encode reps, 30 decode reps. I can do a more accurate timing if you want.

No, this should be fine. The effect of AVX512 on encode is more than I expected...

CMakeLists.txt
jonsneyers: I guess it does, but we are already doing that anyway in our current build script: https://github.com/libjxl/libjxl/blob/main/CMakeLists.txt#L304

veluca93 (approved):

PR #2632, opened 2023-07-04
start streaming encoding

mo271:


lib/include/jxl/encode.h
```diff
@@ -630,6 +636,45 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * @param frame_settings
+ * @param pixel_format for pixels. Object owned by the caller and its contents
+ * are copied internally.
+ * @param x horizontal position of the top-left corner of the processed group.
+ * Must be divisible by 1024.
+ * @param y vertical postition of the top-left corner of the processed group.
+ * Must be divisible by 1024.
+ * @return JXL_EXPORT
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderChunkedImageFrameAddPart(
```
jonsneyers, line 674: Why offsets divisible by 1024? Shouldn't that be either 256 (or rather kGroupDim) or 2048, depending on the granularity we want for this?
Also: how does the actual input buffer and its dimensions get passed?

lib/include/jxl/encode.h
```diff
@@ -335,6 +335,12 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_JPEG_COMPRESS_BOXES = 33,
 
+  /** Control what kind of buffering is used, when using chunked image frames.
+   * 0 = can buffer anything, 1 = can buffer intermediate data,
+   * 2 = can buffer the output, 3 = minimize buffer usage
+   */
+  JXL_ENC_FRAME_SETTING_BUFFERING = 34,
```
jonsneyers, line 342: It's not clear to me what the option values mean, and what the implications of this are on how the chunked encode api can/should be used and/or how it affects compression.

lib/include/jxl/encode.h
```diff
@@ -630,6 +636,45 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+/**
+ * TODO(firsching): add documentation
+ *
+ * @param frame_settings
+ * @param pixel_format for pixels. Object owned by the caller and its contents
+ * are copied internally.
+ * @param x horizontal position of the top-left corner of the processed group.
+ * Must be divisible by 1024.
+ * @param y vertical postition of the top-left corner of the processed group.
+ * Must be divisible by 1024.
```
mo271, line 671:
```suggestion
 * Must be divisible by 2048.
 * @param y vertical position of the top-left corner of the processed group.
 * Must be divisible by 2048.
```

lib/include/jxl/encode.h
```diff
@@ -630,6 +636,45 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * @return JXL_EXPORT
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderChunkedImageFrameAddPart(
+    const JxlEncoderFrameSettings* frame_settings,
+    const JxlPixelFormat* pixel_format, size_t x, size_t y);
```
mo271, line 676:
```suggestion
JXL_EXPORT JxlEncoderStatus JxlEncoderChunkedImageFrameAddPart(
    const JxlEncoderFrameSettings* frame_settings, size_t x, size_t y,
    const JxlPixelFormat* pixel_format, 
    const void* input_data, 
    size_t input_size);
```

lib/include/jxl/encode.h
```diff
@@ -335,6 +335,12 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_JPEG_COMPRESS_BOXES = 33,
 
+  /** Control what kind of buffering is used, when using chunked image frames.
+   * 0 = can buffer anything, 1 = can buffer intermediate data,
+   * 2 = can buffer the output, 3 = minimize buffer usage
```
mo271, line 340:
```suggestion
   * 0 = buffers everything, basically the same as non-streamed code path (mainly for testing)
   * 1 = can buffer internal data (the tokens)
   * 2 = can buffer the output
   * 3 = minimize buffer usage: streamed input and streamed output, writing TOC last 
         (will not work with progressive)
    When the image dimensions is smaller than 2048 x 2048 all the options are the same
   Using 1, 2 or 3 can result increasingly in less compression density.
```

lib/include/jxl/encode.h
mo271: yes, 2048
Added the input buffer argument...

lib/include/jxl/encode.h
mo271: Added more explanation

lib/include/jxl/encode.h
```diff
@@ -335,6 +335,19 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_JPEG_COMPRESS_BOXES = 33,
 
+  /** Control what kind of buffering is used, when using chunked image frames.
+   * 0 = buffers everything, basically the same as non-streamed code path
+   (mainly for testing)
+   * 1 = can buffer internal data (the tokens)
+   * 2 = can buffer the output
+   * 3 = minimize buffer usage: streamed input and streamed output, writing TOC
+   last (will not work with progressive)
```
jonsneyers, line 344: How will writing the TOC last work? The output writing will be random access?
It's not really "streamed output" if there's going to be seeking involved...

lib/include/jxl/encode.h
```diff
@@ -335,6 +335,19 @@ typedef enum {
...
+
+   When the image dimensions is smaller than 2048 x 2048 all the options are the
+   same Using 1, 2 or 3 can result increasingly in less compression density.
```
jonsneyers, line 347:
```suggestion
   same. Using 1, 2 or 3 can result increasingly in less compression density.
```

lib/include/jxl/encode.h
```diff
@@ -630,6 +643,48 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ *
+ * @param frame_settings
+ * @param x horizontal position of the top-left corner of the processed group.
+ * Must be divisible by 2048.
+ * @param y vertical position of the top-left corner of the processed group.
+ * Must be divisible by 2048.
+ * @param pixel_format for pixels. Object owned by the caller and its contents
+ * are copied internally.
+ * @param input_data the input buffer.
+ * @param input_size size of the input data in bytes.
```
jonsneyers, line 680: How will we know the width and height of the input data?
The alignment from `pixel_format` can be used to get the stride between rows right, but we still need to know how large the chunk is.

Perhaps for simplicity we can say that a chunk needs to be exactly one 2048x2048 macrogroup, possibly smaller at the right and bottom side but always exactly one DC-group, basically.

veluca93: That's the idea, yeah

lib/include/jxl/encode.h
mo271: Yes.
Should we call it "buffered" instead?

jonsneyers: Or chunky/chunked?

lib/include/jxl/encode.h
mo271, line 680:
```suggestion
/**
 * TODO(firsching): add documentation
 * We process exactly one 2048x2048 DC-group.
 *
 * @param frame_settings
 * @param x horizontal position of the top-left corner of the processed group.
 * Must be divisible by 2048.
 * @param y vertical position of the top-left corner of the processed group.
 * Must be divisible by 2048.
 * @param pixel_format for pixels. Object owned by the caller and its contents
 * are copied internally.
 * @param input_data the input buffer.
 * @param input_size size of the input data in bytes.
```

mo271: I added a line.

jonsneyers (approved):

PR #2637, opened 2023-07-06
preparation for distinguishing bad input from bad api usage

jonsneyers:
Currently there is no distinction in the decoder API between errors caused by bad API usage (e.g. calling the functions in the wrong order, or with arguments that make no sense) and errors caused by bad input (corrupt/invalid/unsupported bitstreams).

Since that is probably a useful distinction to make (the former is an application bug while the latter is something that can happen at runtime and needs to be handled by the application), as a start I'm adding a macro JXL_INPUT_ERROR in addition to the JXL_API_ERROR macro that already existed, and changed all instances where the error is not an API usage error to use this macro instead. For now they both still do the same thing (return JXL_DEC_ERROR) but this will make it easier if we later want to have some more fine-grained error mechanism.

Flyby: the box size overflow check was not quite correct (the box header was counted twice), so fixing/simplifying that a bit.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -46,14 +46,6 @@ bool OutOfBounds(size_t a, size_t b, size_t size) {
   return false;
 }
 
-bool SumOverflows(size_t a, size_t b, size_t c) {
-  size_t sum = a + b;
-  if (sum < b) return true;
-  sum += c;
-  if (sum < c) return true;
-  return false;
-}
-
 JXL_INLINE size_t InitialBasicInfoSizeHint() {
   // Amount of bytes before the start of the codestream in the container format,
   // assuming that the codestream is the first box after the signature and
@@ -85,11 +77,21 @@ JXL_INLINE size_t InitialBasicInfoSizeHint() {
    JXL_DEC_ERROR)
 #endif  // JXL_CRASH_ON_ERROR
 
+// Error caused by bad input (invalid file) rather than incorrect API usage.
+// For now there is no way to distinguish these two types of errors yet.
+#define JXL_INPUT_ERROR(format, ...) JXL_API_ERROR(format, ##__VA_ARGS__)
+
 JxlDecoderStatus ConvertStatus(JxlDecoderStatus status) { return status; }
 
 JxlDecoderStatus ConvertStatus(jxl::Status status) {
   return status ? JXL_DEC_SUCCESS : JXL_DEC_ERROR;
 }
+
+#define JXL_API_RETURN_IF_ERROR(expr)               \
+  {                                                 \
+    JxlDecoderStatus status_ = ConvertStatus(expr); \
+    if (status_ != JXL_DEC_SUCCESS) return status_; \
+  }
 
 JxlSignature ReadSignature(const uint8_t* buf, size_t len, size_t* pos) {
   if (*pos >= len) return JXL_SIG_NOT_ENOUGH_BYTES;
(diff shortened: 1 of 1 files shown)


lib/jxl/decode.cc
```diff
@@ -1269,11 +1265,11 @@ JxlDecoderStatus JxlDecoderProcessCodestream(JxlDecoder* dec) {
         // No overflow, checked in CheckSizeLimit.
         size_t num_pixels = frame_dim.xsize * frame_dim.ysize;
         if (dec->used_cpu_base + num_pixels < dec->used_cpu_base) {
-          return JXL_API_ERROR("used too much CPU");
+          return JXL_INPUT_ERROR("used too much CPU");
```
szabadka, line 1268: I think something like "image too large" would be better here.

lib/jxl/decode.cc
```diff
@@ -2047,18 +2043,18 @@ JxlDecoderStatus JxlDecoderProcessInput(JxlDecoder* dec) {
   JxlDecoderStatus status = HandleBoxes(dec);
 
   if (status == JXL_DEC_NEED_MORE_INPUT && dec->input_closed) {
-    return JXL_API_ERROR("missing input");
+    return JXL_INPUT_ERROR("missing input");
```
szabadka, line 2046: this should be more generic, something like "premature end of input"

lib/jxl/decode.cc
```diff
@@ -2270,7 +2266,8 @@ JxlDecoderStatus JxlDecoderGetColorAsICCProfile(const JxlDecoder* dec,
   JxlDecoderStatus status =
       JxlDecoderGetICCProfileSize(dec, target, &wanted_size);
   if (status != JXL_DEC_SUCCESS) return status;
-  if (size < wanted_size) return JXL_API_ERROR("ICC profile output too small");
+  if (size < wanted_size)
+    return JXL_INPUT_ERROR("ICC profile output too small");
```
szabadka, line 2270: This is actually an API error, we get here if the buffer provided in the API call was too small.

lib/jxl/decode.cc
```diff
@@ -1269,11 +1265,11 @@ JxlDecoderStatus JxlDecoderProcessCodestream(JxlDecoder* dec) {
         // No overflow, checked in CheckSizeLimit.
         size_t num_pixels = frame_dim.xsize * frame_dim.ysize;
         if (dec->used_cpu_base + num_pixels < dec->used_cpu_base) {
-          return JXL_API_ERROR("used too much CPU");
+          return JXL_INPUT_ERROR("used too much CPU");
         }
         dec->used_cpu_base += num_pixels;
         if (dec->used_cpu_base > dec->cpu_limit_base) {
-          return JXL_API_ERROR("used too much CPU");
+          return JXL_INPUT_ERROR("used too much CPU");
```
szabadka, line 1272: "image too large"

lib/jxl/decode.cc
```diff
@@ -2303,7 +2300,7 @@ JxlDecoderStatus PrepareSizeCheck(const JxlDecoder* dec,
   *bits = BitsPerChannel(format->data_type);
 
   if (*bits == 0) {
-    return JXL_API_ERROR("Invalid/unsupported data type");
+    JXL_API_ERROR("Invalid/unsupported data type");
```
szabadka, line 2303: why remove the return?

jonsneyers: oops, putting it back

lib/jxl/decode.cc
jonsneyers: right, changing it

szabadka (approved):

PR #2639, opened 2023-07-06
Transform copy of libjpeg-turbo to module

eustas:


lib/jpegli.cmake
```diff
@@ -12,6 +12,8 @@ set(JPEGLI_INTERNAL_LIBS
   ${ATOMICS_LIBRARIES}
 )
 
+set(BITS_IN_JSAMPLE 8)
```
szabadka, line 15: we should also set  C_ARITH_CODING_SUPPORTED, and  D_ARITH_CODING_SUPPORTED to 0 and MEM_SRCDST_SUPPORTED to 1 so that clients that build against these headers know that we don't support arith coding.

eustas: Sure, done.

szabadka: ah, I misunderstood #cmakedefine it seems only MEM_SRCDST_SUPPORTED needs to be set to 1, since setting the var to zero or not setting it is the same (I think?)

eustas: ¯\_(ツ)_/¯ cmake is generally user-unfriendly =)
Will remove that line soon.

szabadka (approved):

PR #2652, opened 2023-07-11
Use only the public API in jxl benchmark codec.

szabadka:
The functionality regarding debug images and stats is almost the same, for the rest I added TODOs.

Removed the unused or publicly not exposed fields from the CompressParams struct.


mo271 (approved):

jyrkialakuijala:
could it be that this removed the functionality of --more_columns

I like to use it in quality optimizations -- it helps me to understand what is going on

szabadka:
The block size distribution was moved to the - - print_more_stats flag,
does that work?

PR #2674, opened 2023-07-20
Fix overflow when calculating spline area estimate.

eustas:
Drive-by: avoid `abs(int_x) < 0` situations

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -7,15 +7,13 @@
 
 #include <algorithm>
 #include <cmath>
+#include <limits>
 
-#include "lib/jxl/ans_params.h"
 #include "lib/jxl/base/printf_macros.h"
 #include "lib/jxl/base/status.h"
 #include "lib/jxl/chroma_from_luma.h"
 #include "lib/jxl/common.h"
 #include "lib/jxl/dct_scales.h"
-#include "lib/jxl/entropy_coder.h"
-#include "lib/jxl/opsin_params.h"
 
 #undef HWY_TARGET_INCLUDE
 #define HWY_TARGET_INCLUDE "lib/jxl/splines.cc"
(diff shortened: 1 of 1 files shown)


lib/jxl/splines.cc
```diff
@@ -379,7 +377,10 @@ QuantizedSpline::QuantizedSpline(const Spline& original,
   }
 
   const auto to_int = [](float v) -> int {
-    return static_cast<int>(roundf(v));
+    // Maximal int representable with float.
+    constexpr float kMax = std::numeric_limits<int>::max() - 127;
+    constexpr float kMin = -kMax;
+    return static_cast<int>(roundf(std::min(kMax, std::max(kMin, v))));
```
sboukortt, line 383: How about `Clamp1`?

```suggestion
    return static_cast<int>(roundf(Clamp1(v, kMin, kMax)));
```

lib/jxl/splines.cc
```diff
@@ -460,9 +469,12 @@ Status QuantizedSpline::Dequantize(const Spline::Point& starting_point,
   // it gives an indication of the influence of the color values on the area
   // that will need to be rendered.
   uint64_t logcolor = std::max(
-      uint64_t(1),
-      static_cast<uint64_t>(CeilLog2Nonzero(
-          uint64_t(1) + std::max(color[1], std::max(color[0], color[2])))));
+      kOne, static_cast<uint64_t>(CeilLog2Nonzero(
+                kOne + std::max(color[1], std::max(color[0], color[2])))));
```
sboukortt, line 473: How about:

```suggestion
                kOne + std::max({color[0], color[1], color[2]}))));
```

(since C++11)

lib/jxl/splines.cc
```diff
@@ -519,9 +529,13 @@ Status QuantizedSpline::Decode(const std::vector<uint8_t>& context_map,
   }
 
   const auto decode_dct = [decoder, br, &context_map](int dct[32]) -> Status {
+    constexpr int kWeirdNumber = std::numeric_limits<int>::min();
     for (int i = 0; i < 32; ++i) {
       dct[i] =
           UnpackSigned(decoder->ReadHybridUint(kDCTContext, br, context_map));
+      if (dct[i] == kWeirdNumber) {
```
sboukortt, line 536: Is this a recurrent value we have seen?

sboukortt (commented):
Is it on purpose that we are not using the `std::` qualifier for math functions (`round`, `ceil`…)?

lib/jxl/splines.cc
eustas: Applied.

eustas:
> Is it on purpose that we are not using the `std::` qualifier for math functions (`round`, `ceil`…)?

Don't know. Splines code was somewhat on itself from the beginning. Perhaps it requires some thorough review and combing.

lib/jxl/splines.cc
eustas: Done

lib/jxl/splines.cc
eustas: Thats the one for which std::abs(x) < 0

eustas:
PTAL

sboukortt (approved):

PR #2679, opened 2023-07-24
Fix UB in JPEG bit writer

eustas:


lib/jxl/jpeg/dec_jpeg_data_writer.cc
```diff
@@ -197,10 +202,10 @@ static JXL_INLINE void WriteSymbol(int symbol, HuffmanCodeTable* table,
     // "health" of JpegBitWriter.
     // Not checking this can cause bad output without error on corrupt input,
     // but that's OK. The check causes a noticeable slowdown.
-#if false
-    bw->healthy &= table->depth[symbol] != 0;
-#endif
-    WriteBits(bw, table->depth[symbol], table->code[symbol]);
+    int len = table->depth[symbol];
+    // Only when `len` is 0 litmus becomes negative; that property is sticky.
+    bw->litmus |= len - 1;
```
jonsneyers, line 207: What's the slowdown caused by updating the `litmus` variable?

How likely is it that this bitwriter health check is actually useful? In what circumstances could it trigger an error?

jonsneyers: Yes, the UB needs to be fixed. Question is if the litmus is needed - if there is no way it can get triggered, it's just unnecessary.

Regarding performance: something like `djxl some_large_image.jpg.jxl reconstructed.jpg --num_reps 30 --num_threads 0` ?

eustas: Redesigned to low-overhead; countered overhead with tiny improvements. PTAL

jonsneyers (approved):

jonsneyers:
Looks like this commit broke something. Can you check?

eustas:
Sure. Will investigate ASAP

PR #2694, opened 2023-07-31
roll highway to 1.0.5

mo271:


mo271:
Seems like we shouldn't use this version, because of the failing gcc8 tests. Those are still happening at the current highway head

Issue #2661, opened 2023-07-13
several memory bugs in cjpegli and djxl


kdsjZh:
**Describe the bug**
Hello, I'm testing my new fuzzer and find several memory issues in cjpegli and djxl.

**Environment**
 - OS: Ubuntu 20.04.5 LTS
 - Compiler version: gcc 9.4.0
 - CPU type: Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
 - cjxl/djxl version string: latest commit eda5654fd1c51c7285e1b1472161bb9e560e919f

**To Reproduce**

```
export CC="gcc -fsanitize=address -g"
export CXX="g++ -fsanitize=address -g"
mkdir build && pushd build && \
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=no -DJPEGXL_ENABLE_FUZZERS=no -DBUILD_TESTING=OFF .. && cmake --build . -- -j$(nproc) && \
popd

```

### BUG0

```
./tools/djxl /poc/djxl/poc0.JxlDecoderGetBasicInfo test.png
JPEG XL decoder v0.9.0 eda5654f [AVX2,SSE4,SSE2]
=================================================================
==3917716==ERROR: AddressSanitizer: stack-use-after-scope on address 0x5654f01d1540 at pc 0x5654ef256f4a bp 0x7ffcce6b9ec0 sp 0x7ffcce6b9eb0
READ of size 4 at 0x5654f01d1540 thread T0
    #0 0x5654ef256f49 in FixedAspectRatios /benchmark/libjxl/lib/jxl/headers.cc:37
    #1 0x5654ef256f49 in jxl::SizeHeader::xsize() const /benchmark/libjxl/lib/jxl/headers.cc:53
    #2 0x5654ef256f49 in jxl::SizeHeader::xsize() const /benchmark/libjxl/lib/jxl/headers.cc:51
    #3 0x5654ef1f6a66 in JxlDecoderGetBasicInfo /benchmark/libjxl/lib/jxl/decode.cc:2083
    #4 0x5654f0036cae in jxl::extras::DecodeImageJXL(unsigned char const*, unsigned long, jxl::extras::JXLDecompressParams const&, unsigned long*, jxl::extras::PackedPixelFile*, std::vector<unsigned char, std::allocator<unsigned char> >*) /benchmark/libjxl/lib/extras/dec/jxl.cc:284
[...]

PR #2699, opened 2023-08-02
fix overflow in corrupt exif offset

jonsneyers:
I think this fixes the djxl fuzzerbug in https://github.com/libjxl/libjxl/issues/2661 (though the asan trace suggests the issue is elsewhere). In any case this fixes _something_ in that example.

Since `offset` was uint32_t, the test `exif.size() < 12 + offset + 2` was not quite robust: in that fuzzer example, `offset` has the value 4294967287, so the right hand side evaluates to something that overflows (so the result is not 4294967301 but that number mod 2^32, which is 5). Doing the arithmetic in uint64_t avoids that problem.

We didn't catch this overflow before because this code is not called by libjxl (or djxl_fuzzer) but only by the `extras` encoders used by djxl.

### DIFF
--- a/lib/extras/exif.cc
+++ b/lib/extras/exif.cc
@@ -23,7 +23,7 @@ void ResetExifOrientation(std::vector<uint8_t>& exif) {
     return;  // not a valid tiff header
   }
   t += 4;
-  uint32_t offset = (bigendian ? LoadBE32(t) : LoadLE32(t));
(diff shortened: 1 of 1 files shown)


mo271 (approved):

jonsneyers:
`CI:full` will fail until https://github.com/libjxl/libjxl/pull/2696 is merged and this is rebased to that...

PR #2700, opened 2023-08-02
encoder output streaming

mo271:


lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+    JxlEncoder* enc, struct JxlEncoderOutputProcessor output_processor);
+
+/**
+ * Flushes any buffered input in the encoder, ensuring that all available input
+ * data has been processed and written to the output.
+ *
+ * @param enc encoder object.
+ * @return JXL_ENC_SUCCESS on success, JXL_ENC_ERROR on error.
+ */
+JXL_EXPORT JxlEncoderStatus JxlEncoderFlushInput(JxlEncoder* enc);
```
veluca93, line 740: We should mention how this function relates to ProcessOutput.

Also, we should mention that calling CloseInput should happen before the last call to FlushInput.

lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * will be interpreted as asking the output writing to stop. In such a case,
+   * the library will return an error. The client is expected to set the size of
+   * the returned buffer based on the suggested `size` when this function is
+   * called.
+   *
+   *
+   * @param opaque user supplied parameters to the callback
+   * @param size points to a suggested buffer size when called; must be set to
+   * the size of the returned buffer once the function returns.
+   * @return a pointer to the acquired buffer or NULL to indicate an error or
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * stop condition.
+   */
+  void* (*get_buffer)(void* opaque, size_t* size);
+
+  /**
+   * Required.
+   * Notifies the library that the current buffer's data has been written and
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+  void (*release_buffer)(void* opaque, size_t written_bytes);
+
+  /**
+   * Optional, can be NULL
+   * Seeks to a specific position in the output. This function is optional and
+   * can be set to NULL if the output doesn't support seeking. Can only be done
+   * when there is no buffer.
+   *
+   * @param opaque user supplied parameters to the callback
+   * @param opaque user supplied parameters to the callback
```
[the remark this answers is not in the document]
mo271: :)

lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * @param written_bytes the number of bytes written to the buffer.
+   */
+  void (*release_buffer)(void* opaque, size_t written_bytes);
+
+  /**
+   * Optional, can be NULL
+   * Seeks to a specific position in the output. This function is optional and
+   * can be set to NULL if the output doesn't support seeking. Can only be done
+   * when there is no buffer.
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
mo271: mentioned it

lib/include/jxl/encode.h
```diff
@@ -645,18 +645,99 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * for the encoder. It provides mechanisms to handle buffering, writing,
+ * seeking, and watermarking during the encoding process.
+ *
+ * At a high level, the processor can be in one of two states:
+ * - With an active buffer: meaning it has acquired a buffer using
+ *   `get_buffer` and can write encoded data to it.
+ * - Without an active buffer: which means no data can be written until
+ *   another buffer is acquired after releasing the current active buffer.
  *
+ * It is not allowed to acquire more than one buffer at a given time.
```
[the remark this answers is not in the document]
mo271: Tried to describe it. I altered the current second phrase and didn't skip it completely..

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
- * TODO(firsching): add documentation
+ * At a high level, the processor can be in one of two states:
+ * - With an active buffer: This indicates that a buffer has been acquired using
+ *   `get_buffer` and encoded data can be written to it.
+ * - Without an active buffer: In this state, no data can be written. A new
+ * buffer must be acquired after releasing any previously active buffer.
+ *
+ * It is not allowed to acquire more than one buffer at a given time.
+ *
+ * The JxlEncoder OuputProcessor interacts with `position` and `watermark`,
```
jonsneyers, line 685:
```suggestion
 * The JxlEncoder OutputProcessor interacts with `position` and `watermark`,
```

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+  /**
+   * Required.
+   * Sets a watermark on the output data, at a specific position.
+   * Seeking will never request a position before the watermark.
+   *
+   * @param opaque user supplied parameters to the callback
+   * @param watermark_position the position, in bytes, where the watermark
+   * should be set.
+   */
+  void (*set_watermark)(void* opaque, uint64_t watermark_position);
```
jonsneyers, line 757: 'Watermark' is not a good name for this imo. Maybe something like committed_position would be clearer.

mo271: This is of course not the meaning of watermark as in "watermark in a paper", but rather marking how high the water has risen.

jonsneyers: Yes, but that's not the most common meaning of that word so it's a bit confusing. Maybe something like 'gauge position' would be better to avoid that confusion, but imo it's even better to use a name that clarifies _what_ is being marked/gauged here. Something like 'committed position', 'confirmed position' or 'finalized position'.

szabadka: I also think that e.g. finalized_position would be a better name.

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * position.
+   *
+   * @param opaque user supplied parameters to the callback
+   * @param position the position to seek to, in bytes.
+   */
+  void (*seek)(void* opaque, uint64_t position);
+
+  /**
+   * Required.
+   * Sets a watermark on the output data, at a specific position.
```
szabadka, line 750: it will only be called when there is no active buffer, right? if yes, this should be documented here

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+  /**
+   * Required.
+   * An opaque pointer that the client can use to store custom data.
+   * This data will be passed to the associated callback functions.
+   */
+  void* opaque;
+
+  /**
+   * Required.
+   * Acquires a buffer into which the library will write the output data.
```
szabadka, line 708: Acquires a buffer at the current position ...

lib/jxl/encode_internal.h
```diff
@@ -161,38 +168,157 @@ struct JxlEncoderQueuedInput {
...
+ private:
+  void ReleaseBuffer(size_t bytes_used);
+
+  // Tries to write all the bytes up to the watermark position.
+  void FlushOutput();
+
+  bool AppendBufferToExternalProcessor(void* data, size_t count);
+
+  struct InternalBuffer {
+    // Bytes in the range `[consumed_bytes, written_bytes)` need to be flushed
```
szabadka, line 235: what is consumed_bytes?

lib/jxl/encode_internal.h
```diff
@@ -161,38 +168,157 @@ struct JxlEncoderQueuedInput {
...
+  // Invariant: `internal_buffers_` does not contain chunks that are entirely
+  // below the output position.
+  std::map<size_t, InternalBuffer> internal_buffers_;
+
+  uint8_t** next_out_ = nullptr;
+  size_t* avail_out_ = nullptr;
+  size_t position_ = 0;
+  size_t watermark_position_ = 0;
+  // Either the position of the `external_output_processor_` or the position
+  // `next_out_` points to.
```
szabadka, line 251: what is the difference between position_ and output_position_ ?

lib/include/jxl/encode.h
mo271: renamed to `finalized_position` and similar accordingly

lib/jxl/encode_test.cc
```diff
@@ -1404,3 +1407,188 @@ TEST(EncodeTest, JXL_TRANSCODE_JPEG_TEST(JPEGFrameTest)) {
...
+  size_t ysize = 259;
+  jxl::test::TestImage image;
+  image.SetDimensions(xsize, ysize)
+      .SetDataType(JXL_TYPE_UINT8)
+      .SetChannels(3)
+      .SetAllBitDepths(p.use_container() ? 16 : 8);
+  image.AddFrame().RandomFill();
+  const auto& frame = image.ppf().frames[0].color;
+  JxlBasicInfo basic_info;
+  jxl::test::JxlBasicInfoSetFromPixelFormat(&basic_info, &frame.format);
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode_test.cc
```diff
@@ -1404,3 +1407,188 @@ TEST(EncodeTest, JXL_TRANSCODE_JPEG_TEST(JPEGFrameTest)) {
...
+    }
+  };
+
+  {
+    JxlEncoderPtr enc = JxlEncoderMake(nullptr);
+    ASSERT_NE(nullptr, enc.get());
+    configure_encoder(enc.get());
+    uint8_t* next_out = compressed.data();
+    size_t avail_out = compressed.size();
+    JxlEncoderCloseFrames(enc.get());
```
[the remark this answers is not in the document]
mo271: moved `JxlEncoderCloseInput` from below, so that `JxlEncoderCloseFrames` is not necessary anymore...

lib/jxl/encode_test.cc
```diff
@@ -1404,3 +1407,188 @@ TEST(EncodeTest, JXL_TRANSCODE_JPEG_TEST(JPEGFrameTest)) {
...
+    ProcessEncoder(enc.get(), compressed, next_out, avail_out);
+  }
+
+  {
+    JxlEncoderPtr enc = JxlEncoderMake(nullptr);
+    ASSERT_NE(nullptr, enc.get());
+    JxlStreamingAdapter streaming_adapter(enc.get(), p.return_large_buffers(),
+                                          p.can_seek());
+    configure_encoder(enc.get());
+    JxlEncoderCloseInput(enc.get());
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode.cc
```diff
@@ -2050,36 +2375,46 @@ void JxlEncoderCloseInput(JxlEncoder* enc) {
...
+  }
+  return JXL_ENC_SUCCESS;
+}
+
+JXL_EXPORT JxlEncoderStatus JxlEncoderSetOutputProcessor(
+    JxlEncoder* enc, JxlEncoderOutputProcessor output_processor) {
+  if (enc->output_processor.HasAvailOut()) {
+    return JXL_API_ERROR(
+        enc, JXL_ENC_ERR_API_USAGE,
+        "Set an output processor when some output was already produced");
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
- * @param pixel_format for pixels. Object owned by the caller and its contents
- * are copied internally.
- * @param input_data the input buffer.
- * @param input_size size of the input data in bytes.
- * @return JXL_EXPORT
+ * Flushes any buffered input in the encoder, ensuring that all available input
+ * data has been processed and written to the output.
+ *
+ * This function can only be used after @ref JxlEncoderSetOutputProcessor.
+ * Before making the last call to @ref JxlEncoderFlushInput, users should call
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
- * TODO(firsching): add documentation
+ * At a high level, the processor can be in one of two states:
+ * - With an active buffer: This indicates that a buffer has been acquired using
+ *   `get_buffer` and encoded data can be written to it.
+ * - Without an active buffer: In this state, no data can be written. A new
+ * buffer must be acquired after releasing any previously active buffer.
+ *
+ * It is not allowed to acquire more than one buffer at a given time.
+ *
+ * The JxlEncoder OutputProcessor interacts with `position` and `watermark`,
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+ * which have the following meaning.
+ *
+ * - position: Represents the current position, in bytes, within the output
+ * stream where the encoded data will be written next. This position moves
+ * forward as data is written, and can also be adjusted through the optional
```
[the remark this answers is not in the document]
mo271: done

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
-
-/**
- * TODO(firsching): add documentation
+ * At a high level, the processor can be in one of two states:
+ * - With an active buffer: This indicates that a buffer has been acquired using
+ *   `get_buffer` and encoded data can be written to it.
+ * - Without an active buffer: In this state, no data can be written. A new
+ * buffer must be acquired after releasing any previously active buffer.
+ *
+ * It is not allowed to acquire more than one buffer at a given time.
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode_internal.h
mo271: If we support seeking, `output_position_ == position_`.
Otherwise `output_position` is smaller or equal to `position` and points to the beginning of the current buffer, and `position_` where we write next, I think 
@veluca93

lib/jxl/encode_internal.h
mo271: should be output_position, I suppose?! @veluca93

lib/include/jxl/encode.h
mo271: done

lib/jxl/encode_internal.h
veluca93: kinda, it should be output_position - start_of_the_buffer

lib/jxl/encode_internal.h
veluca93: position_ = where the next GetBuffer call will write bytes to
output_position_ = the position for either the user-given output_processor, or `next_out`.

lib/jxl/encode.cc
```diff
@@ -2050,36 +2375,46 @@ void JxlEncoderCloseInput(JxlEncoder* enc) {
   JxlEncoderCloseFrames(enc);
   JxlEncoderCloseBoxes(enc);
 }
+
+JXL_EXPORT JxlEncoderStatus JxlEncoderFlushInput(JxlEncoder* enc) {
+  while (!enc->input_queue.empty()) {
```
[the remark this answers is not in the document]
mo271: I guess we should. How to do this best? Can we initialize `JxlEncoderOutputProcessor`, so that it contains null pointers before `JxlEncoderSetOutputProcessor` was called or should we add something to the `JxlEncoder` struct? Like a bool 'using_output_processor`?

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,123 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * when there is no buffer. Cannot be used to seek before the watermark
+   * position.
+   *
+   * @param opaque user supplied parameters to the callback
+   * @param position the position to seek to, in bytes.
+   */
+  void (*seek)(void* opaque, uint64_t position);
+
+  /**
+   * Required.
```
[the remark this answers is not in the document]
mo271: good questions, should we make it optional, @veluca93?

lib/include/jxl/encode.h
mo271: I think yes

lib/jxl/encode.cc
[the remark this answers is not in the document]
veluca93: We can just check if `external_output_processor_` is null or not.

lib/include/jxl/encode.h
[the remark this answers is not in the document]
veluca93: Making it optional complicates the API for no reason IMO.

lib/jxl/encode_internal.h
mo271: done

lib/jxl/encode_internal.h
mo271: added comments...

lib/include/jxl/encode.h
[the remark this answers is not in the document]
mo271: ok, let's keep it required then...

lib/jxl/encode.cc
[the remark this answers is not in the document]
mo271: Ok, added a `OutputProcessorSet` with that check

lib/include/jxl/encode.h
mo271: Added a sentence to document it.

lib/jxl/encode.cc
```diff
@@ -29,43 +33,364 @@
...
+  return true;
+}
+
+void JxlEncoderOutputProcessorWrapper::ReleaseBuffer(size_t bytes_used) {
+  JXL_ASSERT(has_buffer_);
+  has_buffer_ = false;
+  auto it = internal_buffers_.find(position_);
+  if (it->second.owned_data.empty() && external_output_processor_) {
+    external_output_processor_->release_buffer(
+        external_output_processor_->opaque, bytes_used);
```
szabadka, line 185: shouldn't output_position_ be incremented by bytes_used here? otherwise it will not hold that output_position_ is the position of the external_output_processor_

lib/jxl/encode_internal.h
```diff
@@ -161,38 +168,162 @@ struct JxlEncoderQueuedInput {
...
+
+  // Invariant: `internal_buffers_` does not contain chunks that are entirely
+  // below the output position.
+  std::map<size_t, InternalBuffer> internal_buffers_;
+
+  uint8_t** next_out_ = nullptr;
+  size_t* avail_out_ = nullptr;
+  // Where the next GetBuffer call will write bytes to.
+  size_t position_ = 0;
+  // The position for either the user-given output_processor, or next_out.
```
szabadka, line 253: This does not seem to be true, isn't it simply the position of the last SetWatermark() call?

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,128 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
- * TODO(firsching): add documentation
+ * At a high level, the processor can be in one of two states:
+ * - With an active buffer: This indicates that a buffer has been acquired using
+ *   `get_buffer` and encoded data can be written to it.
+ * - Without an active buffer: In this state, no data can be written. A new
+ * buffer must be acquired after releasing any previously active buffer.
+ *
+ * The library will not acquire more than one buffer at a given time.
+ *
+ * The state of the processor includes with `position` and `finalized
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode.cc
```diff
@@ -29,43 +33,364 @@
...
+  }
+  if (bytes_used == 0) {
+    internal_buffers_.erase(it);
+    return;
+  }
+  JXL_ASSERT(it != internal_buffers_.end());
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode_internal.h
mo271: Changed it and also renamed `SetWatermark` to `SetFinalizedPosition`

lib/jxl/encode.cc
mo271: thanks makes sense. Changed it.

lib/include/jxl/encode.h
```diff
@@ -669,47 +669,128 @@ JXL_EXPORT JxlEncoderStatus JxlEncoderAddImageFrame(
...
+   * Acquires a buffer at the current position into which the library will write
+   * the output data.
+   *
+   * If the `size` argument points to 0 and the returned value is NULL, this
+   * will be interpreted as asking the output writing to stop. In such a case,
+   * the library will return an error. The client is expected to set the size of
+   * the returned buffer based on the suggested `size` when this function is
+   * called.
+   *
+   *
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode.cc
mo271: ... and now tests fail
Will take a look tomorrow

lib/jxl/encode_internal.h
szabadka: so now this comment should be updated because output_position_ is not the position of the external output processor, since it is not incremented after a release_buffer call

lib/jxl/encode.cc
```diff
@@ -29,43 +33,364 @@
...
-namespace jxl {}  // namespace jxl
+jxl::StatusOr<JxlOutputProcessorBuffer>
+JxlEncoderOutputProcessorWrapper::GetBuffer(size_t min_size,
+                                            size_t requested_size) {
+  JXL_ASSERT(min_size > 0);
+  JXL_ASSERT(!has_buffer_);
+  if (stop_requested_) return jxl::StatusCode::kNotEnoughBytes;
+  requested_size = std::max(min_size, requested_size);
+
+  // If we support seeking, output_position_ == position_.
```
szabadka, line 98: could you add an assert here to verify this?

lib/jxl/encode.cc
mo271: Removed the increment again for now.

So in the case where we don't support seeking, `outout_position_` can be less than `position_`, since in that case only `position_` gets increased. Will in this case `output_position_` not be the position of the `external_output_processor_`?

szabadka: No, the position of the output processor is always incremented after a release_buffer, that is part of its definition.

" position: Represents the current position, in bytes, within the output
 * stream where the encoded data will be written next. This position moves
 * forward with each `release_buffer` call as data is written, and can also be
 * adjusted through the optional seek callback, if provided."

lib/jxl/encode.cc
mo271: I added an assert. It was failing and stopped failing when output_position_ be incremented by bytes_used as suggested above.

lib/jxl/encode_internal.h
mo271: Still not sure if the correct thing is to increase output_position_ by bytes_used

lib/jxl/encode.cc
mo271: Ok

In the 6 failing tests, the  "streaming_adapter.CheckFinalWatermarkPosition();" now fails. I suppose this means that at some place we are not calling `JxlEncoderOutputProcessorWrapper::SetFinalizedPosition` or somehting is wrong in FlushOutput, at some point we should call `output_processor.set_finalized_position` and don't.

wdyt, @veluca93 ?
`

mo271: restructed it a bit, should work now.

lib/jxl/encode.cc
```diff
@@ -29,43 +33,392 @@
...
+    // anymore.
+    if (it->second.owned_data.empty()) {
+      external_output_processor_->release_buffer(
+          external_output_processor_->opaque, bytes_used);
+      // If we don't support seeking, this implies we will never modify again
+      // the bytes that were written so far. Advance the finalized position and
+      // flush the output to clean up the internal buffers.
+      if (!external_output_processor_->seek) {
+        finalized_position_ += bytes_used;
+        FlushOutput();
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode_internal.h
```diff
@@ -161,38 +168,162 @@ struct JxlEncoderQueuedInput {
...
+
+  // Caller can never overwrite a previously-written buffer. Asking for a buffer
+  // with `min_size` such that `position + min_size` overlaps with a
+  // previously-written buffer is invalid.
+  jxl::StatusOr<JxlOutputProcessorBuffer> GetBuffer(size_t min_size,
+                                                    size_t requested_size = 0);
+
+  void Seek(size_t pos);
+
+  void SetFinalizedPosition(size_t pos);
```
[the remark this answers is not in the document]
mo271: done

lib/jxl/encode.cc
```diff
@@ -29,43 +33,392 @@
...
+      if (size == 0 || user_buffer == nullptr) {
+        stop_requested_ = true;
+        return jxl::StatusCode::kNotEnoughBytes;
+      }
+      if (size < min_size + additional_size) {
+        external_output_processor_->release_buffer(
+            external_output_processor_->opaque, 0);
+      } else {
+        internal_buffers_.emplace(position_, InternalBuffer());
+        this->has_buffer_ = true;
```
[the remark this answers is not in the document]
mo271: yeah, done

lib/jxl/encode.cc
```diff
@@ -29,43 +33,391 @@
...
+    if (additional_size == 0) {
+      size_t size = requested_size;
+      uint8_t* user_buffer =
+          static_cast<uint8_t*>(external_output_processor_->get_buffer(
+              external_output_processor_->opaque, &size));
+      if (size == 0 || user_buffer == nullptr) {
+        stop_requested_ = true;
+        return jxl::StatusCode::kNotEnoughBytes;
+      }
+      if (size < min_size + additional_size) {
```
szabadka, line 119: no additional_size needed here

lib/jxl/encode.cc
```diff
@@ -29,43 +33,391 @@
...
+        external_output_processor_->release_buffer(
+            external_output_processor_->opaque, 0);
+      } else {
+        internal_buffers_.emplace(position_, InternalBuffer());
+        has_buffer_ = true;
+        return JxlOutputProcessorBuffer(user_buffer + additional_size,
```
szabadka, line 125: no additional_size needed here

szabadka (approved):

lib/jxl/encode.cc
mo271: done, and also one more... (where it was `- additional_size`)

szabadka (approved):

PR #2708, opened 2023-08-03
Fix using uninitialized value (formally even out-of-range array element)

eustas:


jonsneyers (dismissed):

jonsneyers:
Good catch, oops. How come this didn't trigger *san failures? Or even failures in the regular builds, since if it's really using uninitialized values, it should lead to incorrect or failing decodes, right? Just luck in always getting uninitialized values that happen to be zero?

eustas:
Internal build caught that on various tests under MSAN. Not sure why GitHub actions are passing...

PR #2714, opened 2023-08-07
Add "use before define" safeguard for huffman tables in JPEG stream

eustas:


lib/jxl/jpeg/jpeg_data.cc
```diff
@@ -370,6 +370,38 @@ Status JPEGData::VisitFields(Visitor* visitor) {
     }
   }
 
+  {
+    size_t dht_index = 0;
+    size_t scan_index = 0;
```
szabadka, line 375: dht_index and scan_index don't seem to be changed in the loop

eustas: Whoa! Fixed.

szabadka (approved):

PR #2716, opened 2023-08-08
[dec_group] Optimizing with SIMD in dec_group

gz83:
Introduce `emmintrin.h` in dec_group to optimize `Transpose8x8InPlace` and `Status LoadBlock` using SIMD.

At present, only the x86 architecture and the situation that does not support SIMD are processed, and the situation of non-X86 architecture has not been processed for the time being.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,6 +37,7 @@ Fred Brennan <copypaste@kittens.ph>
 gi-man
 Gilles Devillers (GilDev) <gildev@gmail.com>
 Heiko Becker <heirecka@exherbo.org>
+Ho Cheung <uioptt24@gmail.com>
 Jim Robinson <jimbo2150@gmail.com>
 Jon Sneyers <jon@cloudinary.com>
 Jonathan Brown (Jonnyawsom3) <jonathanbr30@gmail.com>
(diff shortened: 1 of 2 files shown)


jonsneyers:
Could you try doing this with highway instead? That would be more consistent with the rest of the project, and it would make non-x86 sims work too.

gz83:
@veluca93 PTAL thank you

This is the first time I contribute code to libjxl. I don't know how to do a performance comparison test (that is, a comparison between using SIMD and not using SIMD), please give me some guidance.

veluca93:
> @veluca93 PTAL thank you
> 
> This is the first time I contribute code to libjxl. I don't know how to do a performance comparison test (that is, a comparison between using SIMD and not using SIMD), please give me some guidance.

The problem is that the code as you have written it is not correct. Implementing transpose with SIMD is somewhat painful, you can perhaps reuse the code from https://github.com/libjxl/libjxl/blob/main/lib/jxl/transpose-inl.h#L78 (in general, https://github.com/libjxl/libjxl/blob/main/lib/jxl/transpose-inl.h unfortunately only implements non-in-place transposition, so reusing it - which would also be an option - would require changing the usage of this function to write to a different array on the stack).

mo271:
please feel free to re-open, especially when getting this to work with highway

PR #2717, opened 2023-08-08
Factor out jxl_cms as cmake library

mo271:
This allows us to use GetJxlCms from the decoder


bobfriesenhahn:
While this change set has not yet been merged, I discovered (while doing oss-fuzz builds) that libjxl.pc already mentions libjxl_cms as a dependency.  This causes builds with pkg-config --static to fail.

mo271:
> While this change set has not yet been merged, I discovered (while doing oss-fuzz builds) that libjxl.pc already mentions libjxl_cms as a dependency. This causes builds with pkg-config --static to fail.

Thanks for the heads up. There is already such a dependency, it was implemented in a series of pull request like #2884 #2855 #2838 #2782 #2830.
Just forgot to close this one here...

PR #2732, opened 2023-08-15
remove paths-ignore

mo271:
This removes the paths-ignores for ichange to .md and AUTHORS. This allows us to have required workflows. The ci is faster now and also we don't have many such pull requests.


mo271:
helps with #2730

szabadka (approved):

PR #2753, opened 2023-08-26
Fix NEON build issue: type marker instance creation

sshockwave:
Build error
```
/libjxl/lib/jxl/fast_dct-inl.h:162:64: error: expected '(' for function-style cast or type construction
      N * M * 2 + 3 * std::max(N, M) * MaxLanes(HWY_FULL(float)));
```
due to `HWY_FULL(float)` (aka `ScalableTag<float>`) being a type.

Related: 4b8aad231bc7857158be8e4bf554a805596b1916

### DIFF
--- a/lib/extras/metrics.cc
+++ b/lib/extras/metrics.cc
@@ -41,8 +41,8 @@ double ComputeDistanceP(const ImageF& distmap, const ButteraugliParams& params,
 #else
     using T = float;
 #endif
-    const HWY_FULL(T) d;
-    constexpr size_t N = MaxLanes(HWY_FULL(T)());
+    constexpr HWY_FULL(T) d;
+    constexpr size_t N = MaxLanes(d);
     // Manually aligned storage to avoid asan crash on clang-7 due to
     // unaligned spill.
     HWY_ALIGN T sum_totals0[N] = {0};
--- a/lib/extras/metrics.cc
+++ b/lib/extras/metrics.cc
@@ -41,7 +41,7 @@ double ComputeDistanceP(const ImageF& distmap, const ButteraugliParams& params,
 #else
     using T = float;
 #endif
-    constexpr HWY_FULL(T) d;
+    const HWY_FULL(T) d;
     constexpr size_t N = MaxLanes(d);
     // Manually aligned storage to avoid asan crash on clang-7 due to
     // unaligned spill.
--- a/lib/jxl/fast_dct-inl.h
+++ b/lib/jxl/fast_dct-inl.h
@@ -159,8 +159,8 @@ HWY_NOINLINE void TestFastIDCT() {
   int16_t* idct = idct_mem.get();
 
   constexpr HWY_FULL(float) df;
-  auto scratch_space_mem =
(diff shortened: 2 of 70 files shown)


mo271:
Thanks for the fix!
Please add your name to the AUTHORS file.

sshockwave:
@mo271 Done!

mo271:
@sshockwave Thanks!
Our check still does not pass, because the name used as the author in the commit, see 
https://github.com/libjxl/libjxl/commit/9b03f99e199735ecf129e68f45c76c8e01420eb5.patch
does not match the name in the AUTHORS file...

Could you either edit the AUTHORS file or change the author of the commits?

sshockwave:
@mo271 I see, I just updated the AUTHORS list.

eustas:
Better fix would be to extract `const HWY_FULL(float) df;`.

sshockwave:
@eustas Why create a dummy variable? Could you explain more, please? Thanks!

eustas:
Mostly for readability. Out of 132 places where HWY_FULL is used only 7 are not variable declaration. Likely we should convert those 7 as well.

sshockwave:
Understood. I'll make the change.

mo271 (approved):

mo271:
After merging #2755, we have a mix of `const` and `consexpr`. Should it make it more consistent, @eustas?

sshockwave:
Seems that `const HWY_FULL(T) d;` is automatically `constexpr`. I'm OK with both.

lib/extras/metrics.cc
```diff
@@ -41,7 +41,7 @@ double ComputeDistanceP(const ImageF& distmap, const ButteraugliParams& params,
 #else
     using T = float;
 #endif
-    const HWY_FULL(T) d;
+    constexpr HWY_FULL(T) d;
```
mo271, line 44:
```suggestion
    const HWY_FULL(T) d;
```

lib/jxl/fast_dct-inl.h
```diff
@@ -158,8 +158,9 @@ HWY_NOINLINE void TestFastIDCT() {
   auto idct_mem = hwy::AllocateAligned<int16_t>(N * M);
   int16_t* idct = idct_mem.get();
 
+  constexpr HWY_FULL(float) df;
```
mo271, line 161:
```suggestion
  const HWY_FULL(float) df;
```

lib/jxl/gauss_blur.cc
```diff
@@ -381,7 +381,7 @@ void FastGaussianVertical(const hwy::AlignedUniquePtr<RecursiveGaussian>& rg,
                           ImageF* JXL_RESTRICT out) {
   JXL_CHECK(SameSize(in, *out));
 
-  const HWY_FULL(float) df;
+  constexpr HWY_FULL(float) df;
```
mo271, line 384:
```suggestion
  const HWY_FULL(float) df;
```

mo271:
ok, great! Then let's make it `const` in all places...

mo271 (approved):

mo271:
Thanks for the contribution!

PR #2755, opened 2023-08-28
Fix WASM build

eustas:
Drive-by: avoid nameless HWY_FULL(T) instantiations


mo271 (approved):

mo271:
maybe we can only fix the WASM build here and not the drive-by after merging #2753 first.

PR #2762, opened 2023-08-30
don't special case intensity target for pq

mo271:
Benchmarking with
```
build/lib/jxl_gbench --benchmark_filter=".*PQ.*"
```
showed that TF_PQ() was not made significantly slower by this change.

Drive-by: add `static_assert` when evaluating polynomials

### DIFF
--- a/lib/jxl/dec_xyb.cc
+++ b/lib/jxl/dec_xyb.cc
@@ -303,14 +303,11 @@ Status OutputEncodingInfo::SetColorEncoding(const ColorEncoding& c_desired) {
 
   // The internal XYB color space uses absolute luminance, so we scale back the
   // opsin inverse matrix to relative luminance where 1.0 corresponds to the
-  // original intensity target, or to absolute luminance for PQ, where 1.0
-  // corresponds to 10000 nits.
+  // original intensity target.
   if (xyb_encoded) {
-    float intensity_target =
-        (c_desired.tf.IsPQ() ? 10000 : orig_intensity_target);
     InitSIMDInverseMatrix(inverse_matrix, opsin_params.inverse_opsin_matrix,
-                          intensity_target);
-    all_default_opsin = (std::abs(intensity_target - 255.0) <= 0.1f &&
+                          orig_intensity_target);
+    all_default_opsin = (std::abs(orig_intensity_target - 255.0) <= 0.1f &&
                          inverse_matrix_is_default);
   }
 
--- a/lib/jxl/rational_polynomial-inl.h
+++ b/lib/jxl/rational_polynomial-inl.h
@@ -87,6 +87,9 @@ HWY_INLINE HWY_MAYBE_UNUSED V EvalRationalPolynomial(const D d, const V x,
   if (kDegP >= 7) yp = MulAdd(yp, x, LoadDup128(d, p + ((kDegP - 7) * 4)));
   if (kDegQ >= 7) yq = MulAdd(yq, x, LoadDup128(d, q + ((kDegQ - 7) * 4)));
 
+  static_assert(kDegP < 8, "Polynomial degree is too high");
+  static_assert(kDegQ < 8, "Polynomial degree is too high");
+
   return FastDivision<T, V>()(yp, yq);
 }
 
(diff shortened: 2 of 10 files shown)


lib/jxl/dec_xyb.cc
```diff
@@ -303,14 +303,11 @@ Status OutputEncodingInfo::SetColorEncoding(const ColorEncoding& c_desired) {
 
   // The internal XYB color space uses absolute luminance, so we scale back the
   // opsin inverse matrix to relative luminance where 1.0 corresponds to the
-  // original intensity target, or to absolute luminance for PQ, where 1.0
-  // corresponds to 10000 nits.
+  // original intensity target.
   if (xyb_encoded) {
-    float intensity_target =
-        (c_desired.tf.IsPQ() ? 10000 : orig_intensity_target);
```
sboukortt, line 310: Aah, the heart of the issue, isn’t it. 😁

lib/jxl/rational_polynomial-inl.h
```diff
@@ -87,6 +87,9 @@ HWY_INLINE HWY_MAYBE_UNUSED V EvalRationalPolynomial(const D d, const V x,
   if (kDegP >= 7) yp = MulAdd(yp, x, LoadDup128(d, p + ((kDegP - 7) * 4)));
   if (kDegQ >= 7) yq = MulAdd(yq, x, LoadDup128(d, q + ((kDegQ - 7) * 4)));
 
+  static_assert(kDegP < 8, "Polynomial degree is too high");
+  static_assert(kDegQ < 8, "Polynomial degree is too high");
```
sboukortt, line 91: This seems unrelated but not a bad idea, I suppose.

lib/jxl/transfer_functions-inl.h
```diff
@@ -165,6 +165,13 @@ class TF_709 {
 // Perceptual Quantization
 class TF_PQ {
  public:
+  explicit TF_PQ(float display_intensity_target) {
+    display_scaling_factor_to_10000_nits_ =
+        display_intensity_target * (1.0f / 10000.0f);
+    display_scaling_factor_from_10000_nits_ =
+        1.0f / display_scaling_factor_to_10000_nits_;
+  }
```
sboukortt, line 173: These two fields could be made const by initialising them as:

```suggestion
  explicit TF_PQ(float display_intensity_target)
      : display_scaling_factor_to_10000_nits_(display_intensity_target *
                                              (1.0f / 10000.0f)),
        display_scaling_factor_from_10000_nits_(
            1.0f / display_scaling_factor_to_10000_nits_) {}
```

(Also, for the latter, if we are going to compute a division anyway then perhaps we might as well make it `10000.f / display_intensity_target`.)

lib/jxl/rational_polynomial-inl.h
veluca93: yup, it is just a drive-by caused by me looking at that function (when we were still considering pushing the multiplies in the rational polynomial)

lib/jxl/transfer_functions-inl.h
veluca93: Wouldn't making the const break the assignment operator?

Not sure if it necessarily matters, but it might.

lib/jxl/rational_polynomial-inl.h
mo271: yes, added remark about this drive-by to commit message

sboukortt (approved):

lib/jxl/transfer_functions-inl.h
```diff
@@ -165,7 +165,7 @@ class TF_709 {
 // Perceptual Quantization
 class TF_PQ {
  public:
-  explicit TF_PQ(float display_intensity_target)
+  explicit TF_PQ(float display_intensity_target = 255.0)
```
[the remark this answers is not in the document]
mo271: ... it didn't work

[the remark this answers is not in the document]
sboukortt: Is it because of the `explicit`? Because then we could have the explicit one without a default value for the argument, and then add a no-parameter constructor like this:

```
TF_PQ() : TF_PQ(kDefaultIntensityTarget) {}
```

veluca93:
How sure are we that this is not triggering because we do something not OK
with a TF_PQ array/pointer in the tonemapping file?

sboukortt (approved):

PR #2766, opened 2023-08-31
reduce libjxl_dec binary size

jonsneyers:
The code for separable 5x5 and 7x7 convolutions as well as Gaussian blurring is not used anywhere in the decoder, only in the encoder. This PR adjusts the build scripts accordingly so the decoder-only library libjxl_dec doesn't include unnecessary code.

On my x64 linux build, after stripping debug symbols, this reduces `libjxl_dec.so` from 1456 KB to 1251 KB.


eustas:
Let's just rename files. That will be less surprising for future developers... I'll prepare a PR

eustas:
Landed this piecewise. Now we can dismiss this PR.

PR #2771, opened 2023-09-04
Remove forced `-pthread` compilation for Emscripten builds

aryanpingle:


veluca93:
Could you remove instead of just commenting the relevant lines?

aryanpingle:
@mo271 thanks for that! I'm a tad busy rn but I'll add context to the PR as soon as I can.

eustas (approved):

eustas:
WASM CI workflow gets broken by this change.

eustas:
In what scenario we need to avoid `-ptheread`?

aryanpingle:
Hey @eustas. Not all browsers/versions support wasm threads, so web apps using the JXL wasm (like Squoosh) would prefer to have a single-threaded encoder-decoder. [This commit](https://github.com/libjxl/libjxl/commit/ef97fe07656e385bbd1af3801635b66bdbd413f2#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR237-R242) changed the JXL behaviour so that every build uses `-pthread`, regardless of whether the user invokes Emscripten with `-pthread` or not.

Single-threading worked fine AFAIR after removing the lines that force the pthread option, and multi-threading works if you supply it. Giving the user the option to choose their build type would be more convenient I think.

eustas:
Going to add a build option then: library and "binary" options should agree (i.e. both have `-pthreads` or none of them).

aryanpingle:
Yep, that sounds great

eustas:
Should be fixed with #3135

PR #2792, opened 2023-09-13
parameterize box test

mo271:
This is factored out, in order to make #2791 smaller and easier to review.
The diff only looks large, because we remove the  outer loop
`for (int compress_box = 0; compress_box <= 1; ++compress_box) {...}`

### DIFF
--- a/lib/jxl/encode_test.cc
+++ b/lib/jxl/encode_test.cc
@@ -8,8 +8,10 @@
 #include <jxl/encode.h>
 #include <jxl/encode_cxx.h>
 
+#include <cstddef>
 #include <cstdio>
 #include <ostream>
+#include <string>
 #include <vector>
 
 #include "jxl/types.h"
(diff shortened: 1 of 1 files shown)


lib/jxl/encode_test.cc
```diff
@@ -1338,11 +1340,20 @@ TEST(EncodeTest, JXL_BOXES_TEST(BoxTest)) {
...
-  }
 }
 
+std::string nameBoxTest(
+    const ::testing::TestParamInfo<std::tuple<bool>>& info) {
+    return (std::get<0>(info.param) ? "C" : "Unc") + std::string("ompressed");
+}
+
+INSTANTIATE_TEST_SUITE_P(EncodeBoxParamsTest, EncodeBoxTest,
+                         ::testing::Combine(::testing::Values(false, true)),
```
mo271, line 1354: of course it is a bit silly to use a 1-tuple `std::tuple<bool>>`, we are planning to make this a two-tuple soon....

mo271: making it a two-tuple, because apparently older version of gtest don't support `Combine()` with only one argument, causing the debian test to fail...

szabadka (approved):

PR #2793, opened 2023-09-14
adapeted SetPreferredColorProfileTest

mo271:
part of a bug report to show what goes wrong with primaries in `SetPreferredColorProfileTest`


mo271:
TODO: make the tests pass, even with `177`. This will need some work on taking primaries/white points into account.

PR #2798, opened 2023-09-18
relax distance check

mo271:


mo271:
https://github.com/libjxl/libjxl/actions/runs/6197279671/job/16825533521

PR #2825, opened 2023-09-26
fix e8+ VarDCT with noise enabled

jonsneyers:
At effort >= 8 in VarDCT mode, i.e. when adaptive quantization iterations are being done, if noise is enabled then this causes trouble:

- noise generation is not properly initialized or something in this code path (decoding-while-encoding), causing nondeterministic results (!) and probably also wildly wrong noise (haven't checked but this is what I would expect)
- BA scores will be lower because of the noise (possibly wildly lower if the noise is wildly wrong), causing adaptive quantization to bump up the quality a lot (so the file size will be a lot larger with --noise=1 than with --noise=0)

This simply disables noise rendering when doing decoding-while-encoding, which makes the resulting main image data identical whether noise is enabled or not, the only difference being the signaling of noise or not — just like it is at lower effort settings too.

### DIFF
--- a/lib/jxl/dec_cache.cc
+++ b/lib/jxl/dec_cache.cc
@@ -28,7 +28,7 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
                                            PipelineOptions options) {
   const FrameHeader& frame_header = shared->frame_header;
   size_t num_c = 3 + frame_header.nonserialized_metadata->m.num_extra_channels;
-  if ((frame_header.flags & FrameHeader::kNoise) != 0) {
+  if (options.render_noise && (frame_header.flags & FrameHeader::kNoise) != 0) {
     num_c += 3;
   }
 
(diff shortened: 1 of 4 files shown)


lib/jxl/dec_cache.cc
```diff
@@ -28,7 +28,7 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
                                            PipelineOptions options) {
   const FrameHeader& frame_header = shared->frame_header;
   size_t num_c = 3 + frame_header.nonserialized_metadata->m.num_extra_channels;
-  if ((frame_header.flags & FrameHeader::kNoise) != 0) {
+  if (options.render_noise && (frame_header.flags & FrameHeader::kNoise) != 0) {
```
sboukortt, line 31: Not super important but I think I would tend to swap the conditions, as it saves a check most of the time (when `options.render_noise` is true but there isn’t any).

sboukortt (approved):

sboukortt:
> This feels like a bit of a hack - should we instead try to figure out why noise generation is not properly initialized?

It would be good for our intellectual curiosity but we would probably still not use noise in the adaptive quantization loop anyway, would we?

veluca93:
> > This feels like a bit of a hack - should we instead try to figure out why noise generation is not properly initialized?
> 
> It would be good for our intellectual curiosity but we would probably still not use noise in the adaptive quantization loop anyway, would we?

I am not entirely sure... I can see noise being helpful in the AQ loop at low enough distances, but perhaps not.

lib/jxl/dec_cache.cc
jonsneyers: I think the difference between both variants is too small to justify running the CI another time :)

PR #2834, opened 2023-09-29
add some hardening compiler flags

jonsneyers:
Adds some compiler flags (as described on https://wiki.debian.org/Hardening) that are supposed to improve security without hurting performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,6 +369,10 @@ else ()
 
     # Global warning settings.
     -Wall
+
+    # Hardening options
+    -D_FORTIFY_SOURCE=2
+    -fstack-protector-strong
   )
 
   if (JPEGXL_WARNINGS_AS_ERRORS)


veluca93:
The first three flags are already enabled.

The last two, I am not sure whether they are a good idea. I am not certain they don't affect ABI, I would like to see some validation on the fact that they don't hurt performance, and they don't seem entirely portable.

jonsneyers:
Regarding performance, I couldn't measure a difference before/after (tried encode and decode with vardct and modular), but this might depend on the platform/compiler/image type/encode settings. I only did a quick benchmark_xl check though, maybe a more careful measurement could show a (small) difference.

I don't know about ABI and portability impact.

jonsneyers:
Looks like this causes trouble in the wasm build... drop the flags in the wasm build?

jonsneyers:
OK, I am closing this PR then. I added the above link and a link to this discussion to https://www.bestpractices.dev/en/projects/7845?criteria_level=2#security

PR #2849, opened 2023-10-06
Fix building the comparison tool

sboukortt:


sboukortt:
Oops, one sec, this is not enough.

sboukortt:
There.

veluca93 (approved):

veluca93:
Now the question is: why did the CI not catch the issue?

PR #2876, opened 2023-10-13
Add cheat-sheet how to cross-compile in docker

eustas:


jonsneyers:
Isn't it a bit strange to put a .md file with documentation in `.github/workflows/`?
Wouldn't it be better to put this in `doc/`?

eustas:
The reason is that this cheat-sheet closely relates to what workflows does. But I'm fine moving it to docs. One moment.

.github/workflows/build_test_cross.md
```diff
@@ -0,0 +1,44 @@
+=== Running in docker ===
+
```
jonsneyers, line 2: Maybe add one or two sentences to give some context on what this snippet is doing, where you'll find the resulting binaries, etc.

jonsneyers (dismissed):

mo271 (approved):

mo271 (approved):

PR #2881, opened 2023-10-16
Fix lcms2 build

sboukortt:


lib/jxl/cms/jxl_cms.cc
```diff
@@ -49,7 +49,7 @@
 namespace jxl {
 namespace {
 
-using jxl::cms::ColorEncoding;
+using ::jxl::cms::ColorEncoding;
```
sboukortt, line 52: (Not a build fix, just a drive-by https://abseil.io/tips/119 fix)

eustas (dismissed):
Thanks

mo271 (approved):

PR #2905, opened 2023-10-24
Set PKGCONFIG_TARGET_LIBS/INCLUDES for libjxl_cms.pc

BtbN:
Without this, the .pc file ends up with those variables empty, breaking the build in hilarious ways as the empty -L / -I eat up the next argument.

This section is straight up copied from the two sibling .cmake files.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -69,6 +69,7 @@ Stephan T. Lavavej <stl@nuwen.net>
 StepSecurity Bot <bot@stepsecurity.io>
 Sylvestre Ledru <sylvestre@debian.org>
 Thomas Bonfort <thomas.bonfort@airbus.com>
+Timo Rothenpieler <timo@rothenpieler.org>
 tmkk <tmkkmac@gmail.com>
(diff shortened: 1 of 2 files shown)


mo271 (approved):

mo271:
Thanks!

mo271:
Please add yourself to the AUTHORS file (I can also do it for you, if that is more convenient...)

mo271:
Thanks for attempting to fix the AUTHORS check! They still fail, because a different name/email address are used in the commits, see here:

https://github.com/libjxl/libjxl/commit/2b680c3a7c2b05db2e322a2900beb226b272d866.patch
https://github.com/libjxl/libjxl/commit/0fedc58f36c1e09b55c644d7c85354e135727fc2.patch

eustas:
I believe it is better to move this piece to `lib/CMakeLists.txt` and remove copy in `lib/jxl.cmake`, so that all `lib/` artifacts are affected.

eustas:
Should be fixed by #2964

PR #2907, opened 2023-10-24
do not ignore avif result

eustas:


mo271 (approved):

mo271:
Are we sure `avifEncoderSetCodecSpecificOption` might return not return `void` under some circumstances?

I'm getting:
```
[...]benchmark/benchmark_codec_avif.cc:258:9: error: cannot initialize a variable of type 'avifResult' with an rvalue of type 'void'
        JXL_RETURN_IF_AVIF_ERROR(avifEncoderSetCodecSpecificOption(
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]/benchmark/benchmark_codec_avif.cc:20:16: note: expanded from macro 'JXL_RETURN_IF_AVIF_ERROR'
    avifResult jxl_return_if_avif_error_result = (result);                     \
               ^                                 ~~~~~~~~
1 error generated.
```

PR #2927, opened 2023-11-02
implement cms api

mo271:


lib/jxl/render_pipeline/stage_from_linear.cc
```diff
@@ -4,6 +4,8 @@
 // license that can be found in the LICENSE file.
 
 #include "lib/jxl/render_pipeline/stage_from_linear.h"
+#include <cstdio>
+#include "lib/jxl/color_encoding_internal.h"
```
mo271, line 8:
```suggestion
```

lib/jxl/dec_cache.cc
```diff
@@ -195,23 +196,57 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
...
+          output_encoding_info.orig_color_encoding.IsGray());
+    } else {
+      linear_color_encoding = output_encoding_info.linear_color_encoding;
+    }
+
     auto tone_mapping_stage = GetToneMappingStage(output_encoding_info);
     if (tone_mapping_stage) {
       if (!linear) {
         auto to_linear_stage = GetToLinearStage(output_encoding_info);
         if (!to_linear_stage) {
```
sboukortt, line 211: I think maybe I would swap this condition around (`if (!cond) { a } else { b }` -> `if (cond) { b } else { a }`).

lib/jxl/dec_xyb.cc
```diff
@@ -250,6 +251,8 @@ Status OutputEncodingInfo::MaybeSetColorEncoding(
 
 Status OutputEncodingInfo::SetColorEncoding(const ColorEncoding& c_desired) {
   color_encoding = c_desired;
+  linear_color_encoding = color_encoding;
+  linear_color_encoding.Tf().SetTransferFunction(TransferFunction::kLinear);
```
sboukortt, line 255: Is this missing the call to `CreateICC()`?

lib/jxl/decode.cc
```diff
@@ -2653,28 +2662,63 @@ JxlDecoderStatus JxlDecoderGetFrameName(const JxlDecoder* dec, char* name,
 
 JxlDecoderStatus JxlDecoderSetPreferredColorProfile(
     JxlDecoder* dec, const JxlColorEncoding* color_encoding) {
+  return JxlDecoderSetOutputColorProfile(dec, color_encoding, nullptr, 0);
```
sboukortt, line 2665: Parameter comments for the last two arguments would be helpful.

lib/jxl/decode_test.cc
```diff
@@ -1864,12 +1895,41 @@ TEST(DecodeTest, SetPreferredColorProfileTestFromGray) {
...
+}
+
+class DecodeAllEncodingsVariantsTest
+    : public ::testing::TestWithParam<
+          std::tuple<jxl::test::ColorEncodingDescriptor, bool, bool>> {};
+JXL_GTEST_INSTANTIATE_TEST_SUITE_P(
+    DecodeAllEncodingsVariantsTestInstantiation, DecodeAllEncodingsVariantsTest,
+    ::testing::Combine(::testing::ValuesIn(jxl::test::AllEncodings()),
+                       ::testing::Values(false, true),
+                       ::testing::Values(false, true)),
```
sboukortt, line 1926:
```suggestion
                       ::testing::Bool(),
                       ::testing::Bool()),
```

lib/jxl/dec_xyb.cc
mo271: This call moved to here:

https://github.com/libjxl/libjxl/blob/6cc2150ecb9f3b219fc18bec592b6dacf34bc071/lib/jxl/dec_cache.cc#L242

because it was causing failures in the case where we don't set a CMS

lib/jxl/dec_cache.cc
mo271: done

lib/jxl/decode.cc
mo271: done

lib/include/jxl/decode.h
```diff
@@ -877,7 +877,8 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetOutputColorProfile(
  * @param cms structure representing a CMS implementation. See @ref
  * JxlCmsInterface for more details.
  */
-JXL_EXPORT void JxlDecoderSetCms(JxlDecoder* dec, JxlCmsInterface cms);
```
veluca93, line 880: What's the reason for this change?

What are the lifetime requirements for the cms parameter?

lib/jxl/dec_cache.cc
```diff
@@ -195,23 +196,57 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
       }
     }
 
+    ColorEncoding linear_color_encoding;
+    if (frame_header.color_transform == ColorTransform::kXYB) {
+      linear_color_encoding = output_encoding_info.linear_color_encoding;
+    } else {
+      linear_color_encoding = ColorEncoding::LinearSRGB(
```
veluca93, line 203: I admit I don't exactly remember the semantics of linear_color_encoding, but is that supposed to change primaries wrt the original ones? I'd guess not...

lib/jxl/dec_cache.cc
```diff
@@ -195,23 +196,57 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
...
-        builder.AddStage(std::move(to_linear_stage));
         linear = true;
       }
       builder.AddStage(std::move(tone_mapping_stage));
     }
 
     if (linear) {
-      builder.AddStage(GetFromLinearStage(output_encoding_info));
+      const size_t channels_src =
+          (output_encoding_info.linear_color_encoding.IsCMYK()
```
veluca93, line 229: ... For that matter, what's the linear_color_encoding supposed to be in the CMYK case?

lib/jxl/dec_cache.cc
```diff
@@ -195,23 +196,57 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
...
+               ? 4
+               : output_encoding_info.linear_color_encoding.Channels());
+      const size_t channels_dst =
+          output_encoding_info.color_encoding.Channels();
+      const bool not_mixing_color_and_grey =
+          (channels_src == channels_dst ||
+           (channels_src == 4 && channels_dst == 3));
+      if (output_encoding_info.color_encoding_is_original ||
```
veluca93, line 237: A comment that explains what is happening with this if could be useful

lib/jxl/decode.cc
```diff
@@ -2653,28 +2665,64 @@ JxlDecoderStatus JxlDecoderGetFrameName(const JxlDecoder* dec, char* name,
...
     return JXL_API_ERROR("too late to set the color encoding");
   }
-  if (dec->image_metadata.color_encoding.IsGray() &&
-      color_encoding->color_space != JXL_COLOR_SPACE_GRAY &&
-      dec->image_out_buffer_set && dec->image_out_format.num_channels < 3) {
-    return JXL_API_ERROR("Number of channels is too low for color output");
-  }
-  if (color_encoding->color_space == JXL_COLOR_SPACE_UNKNOWN) {
-    return JXL_API_ERROR("Unknown output colorspace");
+  // TODO(firsching): handle the non-icc case
```
veluca93, line 2687: Do we want to do this before tagging a release?

lib/jxl/decode_test.cc
```diff
@@ -1782,9 +1788,12 @@ void SetPreferredColorProfileTest(
       jxl::CreateTestJXLCodestream(jxl::Bytes(pixels.data(), pixels.size()),
                                    xsize, ysize, num_channels, params);
   auto all_encodings = jxl::test::AllEncodings();
-  all_encodings.push_back(
-      {jxl::ColorSpace::kXYB, jxl::WhitePoint::kD65, jxl::Primaries::kCustom,
-       jxl::TransferFunction::kUnknown, jxl::RenderingIntent::kPerceptual});
+  // TODO(firsching): understand why XYB does not work together with icc_dst.
```
veluca93, line 1791: That'd be nice to understand:)

lib/jxl/decode_test.cc
```diff
@@ -1864,12 +1896,40 @@ TEST(DecodeTest, SetPreferredColorProfileTestFromGray) {
...
+JXL_GTEST_INSTANTIATE_TEST_SUITE_P(
+    DecodeAllEncodingsVariantsTestInstantiation, DecodeAllEncodingsVariantsTest,
+    ::testing::Combine(::testing::ValuesIn(jxl::test::AllEncodings()),
+                       ::testing::Bool(), ::testing::Bool()),
+    DecodeAllEncodingsVariantsTestName);
+TEST_P(DecodeAllEncodingsVariantsTest, SetPreferredColorProfileTest) {
+  const auto& from = std::get<0>(GetParam());
+  bool icc_dst = std::get<1>(GetParam());
+  bool use_cms = std::get<2>(GetParam());
+  SetPreferredColorProfileTest(from, icc_dst, use_cms);
```
veluca93, line 1932: Do we have a test that checks directly that asking for the same output in a different space gives the same image?

lib/include/jxl/decode.h
mo271: I think this was due to some problem when cms-refactoring. 
Perhaps let's change it back, so we don't have to worry about the lifetime of `cms`. 
Plan to do something like this:
https://github.com/mo271/libjxl/compare/6a61ca734d29479b23aafeb208f6968ddc03988c..33984446603d920a9350f60d82bc91f65fb19010

lib/jxl/decode_test.cc
mo271: No, not yet, indirectly yes, because the test image we use is always the same.
Should we?

lib/include/jxl/decode.h
mo271: reverted

lib/jxl/decode.cc
mo271: I think we just forgot to delete this TODO and the non-icc is actually handled in the lines below.

lib/jxl/dec_cache.cc
mo271: no idea. @sboukortt ?

lib/jxl/dec_cache.cc
mo271: I added a comment.

lib/jxl/decode_test.cc
mo271: https://github.com/libjxl/libjxl/issues/2946

lib/jxl/dec_cache.cc
mo271: The `linear_color_encoding` is whatever is output from the from_xyb stage, also in the CMYK stage.
I added a comment to the new field `linear_color_encoding` to explain what it is.

lib/jxl/dec_cache.cc
veluca93: Yes, my point is that that linear colorspace will likely not have SRGB primaries if the data encoding did not have SRGB primaries...

lib/jxl/dec_cache.cc
```diff
@@ -195,23 +196,56 @@ Status PassesDecoderState::PreparePipeline(ImageBundle* decoded,
...
+          (output_encoding_info.orig_color_encoding.IsCMYK()
+               ? 4
+               : output_encoding_info.orig_color_encoding.Channels());
+      const size_t channels_dst =
+          output_encoding_info.color_encoding.Channels();
+      bool mixing_color_and_grey = (channels_dst != channels_src);
+      if ((output_encoding_info.color_encoding_is_original) ||
+          (!output_encoding_info.cms_set) || mixing_color_and_grey) {
+        // in those cases we only need a linear stage in other cases we attempt
+        // to obtain an cms stage
```
veluca93, line 238: This comment did not clarify to me why this condition is correct :)

mo271: I expanded the comment. 
The last condition was added when the `jxl_test` `RoundtripRGBToGray` was failing. 
If we don't have the first condition (`color_encoding_is_original`) everything should still work and indeed I just tried and all tests pass, but there is an superfluous cms_stage that is the identity.

veluca93: Ah, I see. Perhaps there should be a todo about the Gray condition then...

lib/jxl/dec_cache.cc
mo271: Is that true, @sboukortt? Does it have the primaries of the data encoding or always SRGB primaries?

veluca93: After some consideration, we figured out (probably...) that we don't need that variable anymore.

veluca93 (dismissed):

lib/jxl/dec_cache.cc
mo271: Yes, removed it. For completeness: `linear_color_encoding` should be in the primaries of the ouput color space.

lib/jxl/dec_cache.cc
mo271: added TODO

veluca93 (approved):

PR #2938, opened 2023-11-08
Fix wording

mo271:


eustas:
I believe "mastering" means more "crafting" here. Let's ask @sboukortt

veluca93:
I am also not sure that deviating from industry-standard terminology is beneficial here...

mo271:
I agree, let's close

PR #2942, opened 2023-11-13
plugins/pixbufloader-jxl: pass ICC profile instead of converting to sRGB

Traneptora:
The "icc-profile" gdk-pixbuf option accepts a base64-encoded string, so we can just pass that instead of relying on lcms2 to convert the pixels to sRGB. This saves most of the time on load.

### DIFF
--- a/plugins/gdk-pixbuf/pixbufloader-jxl.c
+++ b/plugins/gdk-pixbuf/pixbufloader-jxl.c
@@ -9,8 +9,6 @@
 #include <jxl/resizable_parallel_runner.h>
 #include <jxl/types.h>
 
-#include "lcms2.h"
-
 #define GDK_PIXBUF_ENABLE_BACKEND
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #undef GDK_PIXBUF_ENABLE_BACKEND
@@ -60,10 +58,7 @@ struct _GdkPixbufJxlAnimation {
   uint64_t tick_duration_us;
   uint64_t repetition_count;  // 0 = loop forever
 
-  gpointer icc_buff;
-  cmsContext context;
-  cmsHPROFILE profile, srgb;
-  cmsHTRANSFORM transform;
+  gchar *icc_base64;
 };
 
 #define GDK_TYPE_PIXBUF_JXL_ANIMATION (gdk_pixbuf_jxl_animation_get_type())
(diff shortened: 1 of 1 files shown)


plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -501,8 +471,10 @@ static gboolean load_increment(gpointer context, const guchar *buf, guint size,
                         "Failed to allocate output pixel buffer");
             return FALSE;
           }
+          gdk_pixbuf_set_option(frame.data, "icc-profile", decoder_state->icc_base64);
           decoder_state->pixel_format.align =
               gdk_pixbuf_get_rowstride(frame.data);
+          decoder_state->pixel_format.data_type = JXL_TYPE_UINT8;
```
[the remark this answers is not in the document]
Traneptora: GDK pixbuf only supports 8-bit. I think it might be worthwhile to request an HDR space and attach an HDR icc profile if the image is tagged as HDR (e.g. PQ)

Traneptora:
We request pixel data in the tagged space here, rather than just sRGB, so we can support HDR images on HDR screens.

Traneptora:
Change: Lint.

plugins/gdk-pixbuf/pixbufloader-jxl.c
```diff
@@ -349,7 +340,7 @@ static void draw_pixels(void *context, size_t x, size_t y, size_t num_pixels,
                 decoder_state->pixel_format.num_channels * x +
                 gdk_pixbuf_get_rowstride(output) * y;
 
-  cmsDoTransform(decoder_state->transform, pixels, dst, num_pixels);
+  memcpy(dst, pixels, num_pixels * decoder_state->pixel_format.num_channels);
```
mo271, line 343: It should be easier (and faster) to use the `JxlDecoderSetImageOutBuffer()` instead of using the callback with `JxlDecoderSetImageOutCallback()`.

Traneptora: Isn't this issue a bit off-topic though? i.e. should that behavior be changed in a separate, unrelated commit?

Traneptora:
Update: Rebase onto master.

plugins/gdk-pixbuf/pixbufloader-jxl.c
veluca93: If you have a strong opinion, this could be done in a separate commit, but since the simplification is possible because of the change of not doing the sRGB conversion, I would also be happy having it part of this one.

veluca93: SGTM

mo271: Let's submit this now and then change to not using the callback in a follow up pull request

mo271 (approved):

Issue #2940, opened 2023-11-10
condition always false


maddin200:
lib/jxl/render_pipeline/stage_tone_mapping.cc line 41
Code:
} else if (tf.IsHLG() && !tf.IsHLG()) { // <-- always false

sboukortt:
Thanks for reporting these!

PR #2947, opened 2023-11-15
Fix a few conditions

### LINKED ISSUE #2939: conditions always true
/lib/jxl/cms/color_management.cc line 800
Code:
  bool ok = (wx >= 0) || (wx <= 1) || (wy > 0) || (wy <= 1); // <-- both always true (maybe some && necessary)
  if (!ok) {

sboukortt:
Fixes #2939 and #2940.


veluca93:
Do we want tests that trigger the failure conditions?

veluca93 (approved):

sboukortt:
I’m not sure that I see a good way to write a test that wouldn’t be overly specific to the way that it currently behaves. (But then, maybe I’m making that a bigger deal than it would be.)

PR #2948, opened 2023-11-17
Implement chunked file input in cjxl.

szabadka:
This option works only for PPM and PGM for now, and it causes the encoder to not read the entire image file in memory but use the streaming input API of the encoder instead.

### DIFF
--- a/lib/extras/dec/pnm.cc
+++ b/lib/extras/dec/pnm.cc
@@ -19,17 +19,6 @@ namespace jxl {
 namespace extras {
 namespace {
 
-struct HeaderPNM {
-  size_t xsize;
-  size_t ysize;
-  bool is_gray;    // PGM
-  bool has_alpha;  // PAM
-  size_t bits_per_sample;
-  bool floating_point;
-  bool big_endian;
-  std::vector<JxlExtraChannelType> ec_types;  // PAM
-};
-
 class Parser {
  public:
   explicit Parser(const Span<const uint8_t> bytes)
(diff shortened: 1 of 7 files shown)


lib/extras/dec/pnm.cc
```diff
@@ -323,8 +312,77 @@ Span<const uint8_t> MakeSpan(const char* str) {
...
+  dec->data_start = pos - &buffer[0];
+
+  if (header.bits_per_sample == 0 || header.bits_per_sample > 16) {
+    return JXL_FAILURE("Invalid bits_per_sample");
+  }
+  if (header.has_alpha || !header.ec_types.empty() || header.floating_point) {
+    return JXL_FAILURE("Only PGM and PPM inputs are supported");
+  }
+
+  // PPM specify that in the raster, the sample values are "nonlinear" (BP.709,
```
veluca93, line 357: specifies?

lib/extras/dec/pnm.cc
```diff
@@ -323,8 +312,77 @@ Span<const uint8_t> MakeSpan(const char* str) {
   return Bytes(reinterpret_cast<const uint8_t*>(str), strlen(str));
 }
 
+void ReadLinePNM(void* opaque, size_t xpos, size_t ypos, size_t xsize,
```
veluca93, line 315: I am not sure I like the fact that the size of the buffer is inferred from xsize + opaque... What do you think about passing in the buffer size, and assert-failing if it does not match?

szabadka: Done.

lib/extras/dec/pnm.cc
szabadka: Done.

veluca93 (approved):

PR #2949, opened 2023-11-18
Implement streaming file output in cjxl.

szabadka:


tools/cjxl_main.cc
```diff
@@ -934,6 +938,62 @@ void ProcessFlags(const jxl::extras::Codec codec,
...
+    if (buffer_pos + *size > self->output.size()) {
+      self->output.resize(buffer_pos + *size, 0xDA);
+    }
+    *size = self->output.size() - buffer_pos;
+    return self->output.data() + buffer_pos;
+  }
+
+  static void ReleaseBuffer(void* opaque, size_t written_bytes) {
+    JxlOutputProcessor* self = reinterpret_cast<JxlOutputProcessor*>(opaque);
+    Seek(opaque, self->position + written_bytes);
```
veluca93, line 971: Why don't we just fwrite here and then resize the output vector to 0, getting rid of tracking finalized_position and making position match the file's position?

tools/scripts/roundtrip_test.sh
```diff
@@ -116,9 +116,9 @@ main() {
   roundtrip_test "jxl/flower/flower_small.rgb.png" "-e 1" 0.02
   roundtrip_test "jxl/flower/flower_small.rgb.png" "-e 1 -d 0.0" 0.0
   roundtrip_test "jxl/flower/flower_small.rgb.depth8.ppm" \
-		 "-e 1 --streaming_input" 0.02
+		 "-e 1 --streaming_input --streaming_output" 0.02
```
veluca93, line 119: Might make sense to test things separately

veluca93:
On a somewhat tangential note, given the number of times that we are doing gymnastics to pass C++ methods as callbacks for C code, what would you think of doing something like this? (with the template + macro magic in a shared header that is used in all the relevant places)

```
#include <stdio.h>
#include <utility>

namespace detail {

template <typename T> struct MethodToCCallbackHelper {};

template <typename T, typename R, typename... Args>
struct MethodToCCallbackHelper<R (T::*)(Args...)> {
  template <R (T::*method)(Args...)> static R Call(void *opaque, Args... args) {
    return (reinterpret_cast<T *>(opaque)->*method)(
        std::forward<Args>(args)...);
  }
};

} // namespace detail

#define METHOD_TO_C_CALLBACK(method)                                           \
  detail::MethodToCCallbackHelper<decltype(method)>::Call<method>

void f(void *opaque, void (*m)(void *opaque), int (*ma)(void *opaque, int i)) {
  printf("%d\n", ma(opaque, 1));
  m(opaque);
}

class X {
public:
  void Method() { fprintf(stderr, "METHOD\n"); }

  int MethodArg(int i) { return printf("%d\n", i); }
};

int main() {
  X x;
  f(&x, METHOD_TO_C_CALLBACK(&X::Method), METHOD_TO_C_CALLBACK(&X::MethodArg));
}
```

tools/cjxl_main.cc
szabadka: Thanks, that is indeed easier, I was trying to write the file sequentially for some reason, but with fseek we don't need to. I kept finalized position so that we know the file size at the end, but position we don't need since that is implicit in the output file.

tools/scripts/roundtrip_test.sh
szabadka: Done. And it caught a bug, which you probably noticed but it was probably the easiest way to point out :)

szabadka:
I was also thinking many times about the class method as callback, and wondered why there is no easy syntactic sugar to do this in C++. I would not object to this helper thing, but would not volunteer to do the change either ...

tools/cjxl_main.cc
veluca93: At the end, the finalized position is equal to the last byte written, so you could SEEK_END and ftell, but possibly keeping the finalized position is easier.

tools/scripts/roundtrip_test.sh
veluca93: I did not notice a bug :P what was it?

veluca93:
> I was also thinking many times about the class method as callback, and wondered why there is no easy syntactic sugar to do this in C++. I would not object to this helper thing, but would not volunteer to do the change either ...

I could do that, or maybe @sboukortt wants to give it a shot?

tools/scripts/roundtrip_test.sh
szabadka: JxlEncoderFlushInput was not called (and it called automatically for streaming input+output)

veluca93: Ah, indeed that would be a difference...

veluca93 (approved):

PR #2956, opened 2023-11-21
bump version to 0.10.0

mo271:


mo271:
made the base for v0.9.0 here: https://github.com/libjxl/libjxl/tree/v0.9.x

mo271:
will update the "update-version" script to take into account the quoted variables in the yaml file later.

eustas (approved):

mo271:
> will update the "update-version" script to take into account the quoted variables in the yaml file later.

done in #2972

PR #2958, opened 2023-11-22
Fix large box decoding.

szabadka:


veluca93 (approved):
Probably should also go in the change log - perhaps @mo271 can do that and also fix the test? :-)

mo271:
we probably also need to fix something in here: https://github.com/mo271/libjxl/blob/main/tools/box/box.cc

mo271:
> we probably also need to fix something in here: https://github.com/mo271/libjxl/blob/main/tools/box/box.cc

Done in https://github.com/libjxl/libjxl/pull/2974

PR #2960, opened 2023-11-22
Fix standalone fast lossless build.

### LINKED ISSUE #2959: Fix standalone fast lossless build.
PR #2957 breaks the standalone fast lossless build.

This should be fixed by removing all the parts that need the internal libjxl library with the FJXL_STANDALONE macro.

szabadka:
Fixes #2959

### DIFF
--- a/lib/jxl/enc_fast_lossless.h
+++ b/lib/jxl/enc_fast_lossless.h
@@ -5,7 +5,6 @@
 
 #ifndef LIB_JXL_ENC_FAST_LOSSLESS_H_
 #define LIB_JXL_ENC_FAST_LOSSLESS_H_
-#include <jxl/encode.h>
 #include <stdlib.h>
 
 // FJXL_STANDALONE=1 for a stand-alone jxl encoder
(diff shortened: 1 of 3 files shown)


lib/jxl/enc_fast_lossless.h
```diff
@@ -18,10 +17,32 @@
...
+  void* opaque;
+  const void* (*get_color_channel_data_at)(void* opaque, size_t xpos,
+                                           size_t ypos, size_t xsize,
+                                           size_t ysize, size_t* row_offset);
+  void (*release_buffer)(void* opaque, const void* buf);
+};
+// The standalone version does not use this struct, but we define it here so
+// that we don't have to clutter all the function signatures with defines.
+struct JxlEncoderOutputProcessorWrapper {
+  int palceholder;
```
veluca93, line 42: typo: placeholder

I assume this is not an empty struct b/c some ABI difference between C and C++? I'd probably rename it to `unused_` if so

veluca93 (dismissed):

lib/jxl/enc_fast_lossless.h
szabadka: Yes, because of the ABI difference. I renamed it to unused.

veluca93 (approved):

PR #2969, opened 2023-11-24
Improve streaming encoder testing and fix newly discovered bugs.

szabadka:


lib/BUILD
```diff
@@ -199,7 +199,10 @@ cc_library(
     name = "jpegxl",
     srcs = libjxl_dec_sources + libjxl_dec_box_sources + libjxl_dec_jpeg_sources + libjxl_enc_sources + libjxl_cms_sources,
     compatible_with = DEFAULT_COMPATIBILITY,
-    defines = ["JPEGXL_ENABLE_SKCMS=1"],
+    defines = [
+        "JPEGXL_ENABLE_SKCMS=1",
+        "FJXL_STANDALONE=0",
```
veluca93, line 204: Doesn't that get defined automatically? At least there is some logic for it somewhere (possibly the .cc file?)

lib/jxl/enc_fast_lossless.cc
```diff
@@ -715,6 +725,34 @@ void JxlFastLosslessPrepareHeader(JxlFastLosslessFrameState* frame,
   output->ZeroPadToByte();  // Groups are byte-aligned.
 }
 
+#if !FJXL_STANDALONE
+void JxlFastLosslessOutputAlignedSection(
```
veluca93, line 729: Doesn't WriteOutput already have a special case for this case? or am I missing something?

lib/BUILD
szabadka: Yes, there is, but apparently that does not work in bazel, and this seemed to be the easiest way to fix.

lib/jxl/enc_fast_lossless.cc
szabadka: The reason I did this was that WriteOutput always requested at least 32 bytes, and then the actual bytes written was less. This works if we write the output sequentially, but when we seek back and write the header, sometimes requesting at least 32 bytes is interpreted as trying to overwrite an already written output and we get an assert failure.

lib/BUILD
veluca93: That seems mildly suspicious, is perhaps some other define missing?

lib/jxl/enc_fast_lossless.cc
veluca93: Ah, that is indeed annoying... I wonder if there is a better solution than this, for example relaxing the WriteOutput requirements, but if nothing else comes to mind then that's fine with me.

lib/BUILD
szabadka: Defining the version macros works too.

lib/jxl/enc_fast_lossless.cc
szabadka: Not without making Write output more complex, and it is complex enough as it is. I would be in favor of handling the simple case with this straightforward function.

lib/BUILD
veluca93: Yeah, that seems better - although perhaps it fits best in a separate PR...

lib/jxl/enc_fast_lossless.cc
veluca93: Fair enough. Maybe we can just add a TODO for the future...

veluca93 (approved):

PR #2972, opened 2023-11-27
fix bump_version script

mo271:


mo271:
The dot inside the square bracket is treated as a literal, even when not escaped...

veluca93 (approved):

PR #2974, opened 2023-11-27
fix extended box in tools/box

mo271:


mo271:
maybe it is good to merge this even when we delete it afterwards (https://github.com/libjxl/libjxl/pull/2975) in the unlikely case this box file is ever revived...

veluca93 (approved):

PR #2975, opened 2023-11-27
remove box tool

mo271:


szabadka (commented):
box.h needs to be removed from the BUILD file to fix the bazel build

szabadka (approved):

Issue #2971, opened 2023-11-26
libjxl fails to AddImageFrame with large JxlPixelFormat.align


Traneptora:
**Describe the bug**
The actual size of the buffer passed to JxlEncoderAddImageFrame is ignored when bounds checking. A bound based on JxlPixelFormat.align will be calculated (correctly) but won't actually be compared to the buffer size passed to this function. Instead it'll be compared to the size `bytes_per_pixel * xsize * ysize`, and will then immediately fail if the align is significantly greater than 1.

**To Reproduce**
Encode an image frame that has JxlPixelFormat.align set to some large number, larger than the size of the row. For example, encode a 70x46 16-bit integer image, but set the linesize to something larger than 140. (FFmpeg does this when encoding the attached PFM, for example.) Although any API client call will fail regardless of what's in the pixel buffer so you can set up a primitive program to call libjxl and it will fail in the same way.

**Expected behavior**
JxlPixelFormat.align should do something. 

**Notable Info**
The culprit is right here: https://github.com/libjxl/libjxl/blob/bde01ad04255f41738d285d8422b69c35cb0d015/lib/jxl/encode.cc#L2270
This buffer is allocated as a `jxl::Bytes` and then its size is used, not the size of the buffer passed to `JxlEncoderAddImageFrame`.

The commit that introduced this regression is 6bd63f5880a36bc94172ff285064ebb1743b1964. 

**Environment**
 - OS: Reproduced on both Arch Linux x86_64, and Windows x86_64
 - Compiler version: reproduced with GCC and Clang

**Additional context**
I've attached a sample file to run through FFmpeg and generate the issue. It's the standard Rose image from ImageMagick, so no copyright issues there. Gzipped because pfm cannot be attached directly.

[in.pfm.gz](https://github.com/libjxl/libjxl/files/13465827/in.pfm.gz)

Credit to bonnibel on the JXL discord for reporting.

Traneptora:
Fixed by 308b5f1eed81becac506569080e4490cc486660c.

PR #2983, opened 2023-11-28
Use chunked frame adapter instead of image bundle in EncodeFrame.

szabadka:


lib/jxl/enc_external_image.cc
```diff
@@ -110,11 +92,84 @@ Status ConvertFromExternal(Span<const uint8_t> bytes, size_t xsize,
...
+Status ConvertFromExternal(const uint8_t* data, size_t size, size_t xsize,
+                           size_t ysize, size_t bits_per_sample,
+                           JxlPixelFormat format, size_t c, ThreadPool* pool,
+                           ImageF* channel) {
+  size_t bytes_per_channel = JxlDataTypeBytes(format.data_type);
+  size_t bytes_per_pixel = format.num_channels * bytes_per_channel;
+  const size_t last_row_size = xsize * bytes_per_pixel;
+  const size_t align = format.align;
+  const size_t row_size =
+      (align > 1 ? jxl::DivCeil(last_row_size, align) * align : last_row_size);
```
veluca93, line 150: Does this PR happen to also fix #2971?

lib/jxl/encode.cc
```diff
@@ -2258,64 +2242,37 @@ JxlEncoderStatus JxlEncoderAddImageFrameInternal(
...
   }
 
-  size_t row_offset;
-  const void* buffer = input.get_color_channel_data_at(
-      input.opaque, 0, 0, xsize, ysize, &row_offset);
-  if (!buffer) {
-    return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_GENERIC,
-                         "no buffer for color channels given");
+  if (!streaming) {
+    if (!frame_data.CopyBuffers()) {
```
veluca93, line 2246: I assume that this happens because we need to read the frame data after the point where it is guaranteed to be available anymore - in that case, could you add a comment? (and even not in that case...)

veluca93 (approved):

lib/jxl/encode.cc
szabadka: Done.

lib/jxl/enc_external_image.cc
szabadka: Hopefully yes, I made the "onegroup" mode in the streaming test use >row size align value to test this.

veluca93 (approved):

PR #2991, opened 2023-11-30
run authors check only on main

mo271:


mo271:
closing in favor of #2993

PR #2993, opened 2023-11-30
make authors check optional

mo271:


mo271:
The modification in the workflow still makes the authors check run, but it will continue-on-error, so will always be marked as green.

CONTRIBUTING.md
```diff
@@ -113,8 +115,8 @@ information on using pull requests.
 
   * Sign the CLA (only needed once per user, see above).
 
-  * AUTHORS: If this is your first contribution, add your name or your
-  company name to the [AUTHORS](AUTHORS) file for copyright tracking purposes.
+  * AUTHORS: If this is your first contribution, you can add your name
```
veluca93, line 118: Even if it is not the first contribution, no?

mo271: True, removed the condition

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,7 +7,7 @@
 ### Pull Request Checklist
 
 - [ ] **CLA Signed**: Have you signed the [Contributor License Agreement](https://code.google.com/legal/individual-cla-v1.0.html) (individual or corporate, as appropriate)? Only contributions from signed contributors can be accepted.
-- [ ] **Authors**: If this is your first contribution, have you added your name to the [AUTHORS](AUTHORS) file?
+- [ ] **Authors**: If this is your first contribution, have you considered adding your name to the [AUTHORS](AUTHORS) file?
```
veluca93, line 10: Here too :-)

mo271: done too

veluca93 (approved):

veluca93 (approved):

PR #2996, opened 2023-12-01
don't try to write files when --disable_output

### LINKED ISSUE #2994: djxl crashes with JXL_ABORT on any input with --disable_output
**Describe the bug**
djxl crashes when using --disable_output. It does not crash when writing to PNG, and libjxl works fine.

**To Reproduce**
Run djxl on any input JXL file with --disable_output. Preferably a debug build.

**Expected behavior**
Decoded to pixels, with output discarded.

**Log**
```
~/libjxl/conformance/testcases/bike :) $ djxl -v --disable_output input.jxl 
JPEG XL decoder v0.10.0 9b82fa10 [AVX2]
Read 388268 compressed bytes.
./lib/extras/packed_image.h:86: JXL_ABORT: Unhandled JxlDataType
Illegal instruction (core dumped)
```
This is reproducible with any JXL file, so no sample needed.

**Environment**
 - OS: x86_64 Arch Linux
 - Compiler version: GCC 13.2.1
 - cjxl/djxl version string: JPEG XL decoder v0.10.0 9b82fa10 [AVX2]

mo271:
fixes #2994

### DIFF
--- a/tools/djxl_main.cc
+++ b/tools/djxl_main.cc
@@ -552,7 +552,7 @@ int main(int argc, const char* argv[]) {
     }
     // When --disable_output was parsed, `filename_out` is empty and we don't
     // need to write files.
-    if (!filename_out.empty()) {
+    if (encoder) {
       if (args.alpha_blend) {
         float background[3];
         if (!ParseBackgroundColor(args.background_spec, background)) {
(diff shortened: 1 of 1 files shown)


tools/djxl_main.cc
```diff
@@ -550,57 +550,63 @@ int main(int argc, const char* argv[]) {
...
-    if (encoder && args.use_sjpeg) {
-      encoder->SetOption("jpeg_encoder", "sjpeg");
-    }
-    jxl::extras::EncodedImage encoded_image;
-    if (encoder) {
-      if (!args.quiet) cmdline.VerbosePrintf(2, "Encoding decoded image\n");
-      if (!encoder->Encode(ppf, &encoded_image)) {
-        fprintf(stderr, "Encode failed\n");
-        return EXIT_FAILURE;
+      if (encoder) {
```
szabadka, line 564: Now we can get rid of all the if (encoder ... parts, because we would have exited early in line 532 otherwise

mo271: done, and I used `encoder` instead of `!filename_out.empty()` in the main if, they should be the same value at that point.

PR #1421, opened 2022-05-12
Add Reference Links to Decoder Docs

thebombzen:
This PR will hyperlink functions included in the docs so you don't need to search them with Ctrl+F.

PR #3004, opened 2023-12-06
add more @ref to api documentation

mo271:


mo271:
> I’m not sure I see the motivation for this? When I build the documentation, most of these are already made into links automatically as far as I can tell.


Interesting! I can now also observe those cases that automatically turned into a link, but there's also many cases where it doesn't happen. Probably a good time to read the docs of readthedocs (or sphinx or whatever).

For convenience to compare:
[before changes in this pull request](https://libjxl.readthedocs.io/en/latest/)

[after changes in this pull request](https://libjxl--3004.org.readthedocs.build/en/3004/api_decoder.html)

There are also many occurence where links didn't work before and do now after this change, for example in encode.h.
Also all the changes where I added "::" are now getting shown as links, while they weren't before.


Perhaps @Traneptora knows what is going on, considering they made similar changes in 
 https://github.com/libjxl/libjxl/pull/1421...

lib/include/jxl/cms_interface.h
```diff
@@ -224,7 +225,7 @@ typedef void (*jpegxl_cms_destroy_func)(void*);
 typedef struct {
   /** CMS-specific data that will be passed to @ref set_fields_from_icc. */
   void* set_fields_data;
-  /** Populates a JxlColorEncoding from an ICC profile. */
+  /** Populates a 2ref JxlColorEncoding from an ICC profile. */
```
eustas, line 228:
```suggestion
  /** Populates a @ref JxlColorEncoding from an ICC profile. */
```

lib/include/jxl/decode.h
```diff
@@ -223,19 +223,19 @@ typedef enum {
...
-   * JXL_DEC_COLOR_ENCODING, and always earlier than any pixel data. While
-   * JPEG XL supports encoding a single frame as the composition of multiple
-   * internal sub-frames also called frames, this event is not indicated for the
-   * internal frames.
-   * In this case, @ref JxlDecoderReleaseInput will return all bytes from the
-   * end of the frame header (including ToC) as unprocessed.
+   * displayed frames, unless @ref JxlDecoderSetCoalescing was set to @ref
+   * JXL_FALSE "JXL_FALSE": in that case, the individual layers are returned,
+   * without blending. Note that even when coalescing is disabled, only frames
+   * of type kRegularFrame are returned; frames of type kReferenceOnly and
```
eustas, line 229: Somehow ref kConstants as well?

lib/include/jxl/decode.h
```diff
@@ -1151,13 +1152,13 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderExtraChannelBufferSize(
...
- * the @ref JXL_DEC_FRAME or @ref JXL_DEC_NEED_IMAGE_OUT_BUFFER event occurs,
+ * the ::JXL_DEC_FRAME or ::JXL_DEC_NEED_IMAGE_OUT_BUFFER event occurs,
  * and applies only for the current frame. The size of the buffer must be at
  * least as large as given by @ref JxlDecoderExtraChannelBufferSize. The buffer
- * follows the format described by JxlPixelFormat, but where num_channels is 1.
- * The buffer is owned by the caller. The amount of extra channels is given by
- * the num_extra_channels field in the associated JxlBasicInfo, and the
- * information of individual extra channels can be queried with @ref
+ * follows the format described by @ref JxlPixelFormat, but where num_channels
+ * is 1. The buffer is owned by the caller. The amount of extra channels is
```
eustas, line 1159: Is there a syntax for constants / literals?

eustas:
I'm fine with spicing documentation with to make things explicit (just a habit from Java times).

Traneptora:
>Perhaps @Traneptora knows what is going on, considering they made similar changes in
https://github.com/libjxl/libjxl/pull/1421...

I don't, I just made changes that weren't linked automatically in the build at the time. I couldn't tell you why some were automatically linked. I'd have to read the read the docs docs.

lib/include/jxl/decode.h
mo271: I could add @ref to all of the constants, where those defined are in some of the files where we build doxygen files. So it works for all, except `kRegularFrame` and `kReferenceOnly`, because those are defined in frame_header.h which is apparently not build in doxygen. 
(Should we move the definition of the enum FrameType to a more public place?)

lib/include/jxl/decode.h
mo271: Yes, I added single back tick in all places that seemed reasonable. Not entirely sure this is an improvement...

eustas (approved):

PR #3006, opened 2023-12-06
(TSAN) Do not compute mask1x1 for tile padding

eustas:
Drive-by: fix name shadowing


eustas:
"Fast benchmark release" have same compressed size as before. I'd guess pixels are not changed.

PR #3015, opened 2023-12-12
Reformat dot

eustas:


mo271:
still looks broken:
https://libjxl--3015.org.readthedocs.build/en/3015/api_color.html#_CPPv423jpegxl_cms_destroy_func

eustas:
That means, something is broken on rtd side. do we have any logs from them?
BTW: ee can host docs page on gh as well...

mo271:
fixed by #3028

PR #3019, opened 2023-12-12
Add a basic flake.nix file for development on NixOS.

veluca93:


sboukortt:
Is this missing the actual `shell.nix` or is there no need for that with flakes?

Incidentally, I also wrote one a while ago 😁 but it was still with LLVM 12 and Qt 5.

veluca93:
Ah yeah the PR title is out of date after flakeization :-)

sboukortt (approved):

PR #3020, opened 2023-12-13
mmap-backed implementation of PNM streaming input.

veluca93:
This PR recovers most of the low-core speed loss from using streaming input/output for fast lossless.

Memory usage may be increased, depending on OS choices. However, in case of out of memory, the OS should always unmap pages loaded because of this change.

```
Before:
7216 x 5412, geomean: 76.994 MP/s [76.29, 79.14], 20 reps, 2 threads.
3.67user 9.16system 0:10.18elapsed 126%CPU (0avgtext+0avgdata 8936maxresident)k
680inputs+223752outputs (4major+319857minor)pagefaults 0swaps

After:
7216 x 5412, geomean: 327.330 MP/s [296.45, 338.47], 20 reps, 2 threads.
3.28user 0.83system 0:02.47elapsed 166%CPU (0avgtext+0avgdata 123000maxresident)k
0inputs+111872outputs (10major+309603minor)pagefaults 0swaps

Non-streaming:
7216 x 5412, geomean: 369.804 MP/s [258.03, 371.21], 20 reps, 2 threads.
2.98user 0.69system 0:02.34elapsed 157%CPU (0avgtext+0avgdata 447432maxresident)k
784inputs+111872outputs (4major+517554minor)pagefaults 0swaps
```

### DIFF
--- a/lib/extras/dec/pnm.cc
+++ b/lib/extras/dec/pnm.cc
@@ -11,6 +11,7 @@
 #include <cmath>
 #include <mutex>
 
+#include "jxl/encode.h"
 #include "lib/extras/size_constraints.h"
 #include "lib/jxl/base/bits.h"
 #include "lib/jxl/base/compiler_specific.h"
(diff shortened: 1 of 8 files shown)


veluca93:
Note that this code will most certainly not work on Windows - if the approach seems sensible enough, I will add a Windows implementation too.

lib/extras/dec/pnm.cc
```diff
@@ -5,12 +5,16 @@
 
 #include "lib/extras/dec/pnm.h"
 
+#include <fcntl.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
```
szabadka, line 11: Maybe we can add some abstraction in another file that would handle OS specific stuff.

veluca93: Yeah, that's what I was thinking of doing - the only question is whether we should just error out on non-windows non-unix or if we should emulate it with fread...

szabadka: I think an error message saying that streaming input is not supported would be enough.

lib/extras/packed_image.h
```diff
@@ -225,50 +217,7 @@ class ChunkedPackedFrame {
   JxlPixelFormat format;
```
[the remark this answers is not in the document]
veluca93: They seem to be read by jxl.cc

szabadka (approved):

PR #3022, opened 2023-12-13
Replace CreateTableCurve with precalculated tables

eustas:
NB: tables are the same on SSE2/4, but one-off for one value on AVX2; so, fixing those will make output a little bit more stable across platforms


veluca93:
I think it might be best if @sboukortt gives this a look...

PR #3039, opened 2023-12-19
Remove sjpeg

sboukortt:


barracuda156:
Sorry, I am aware, what was wrong with sjpeg?

sboukortt:
Nothing “wrong” _per se_, but it’s not really clear that it warrants extra maintenance burden. However, the burden is also not that much, so we have decided to put off the removal for now.

PR #3045, opened 2023-12-20
Fix link

BtbN:
This fixes some issues I discovered when linking an external lcms2 library.
For one, the order of the dependency libraries are wrong. They need to come after the library that needs their symbols.

Furthermore, lcms2 can be built split into multiple libraries. The easiest way to deal with that is to simply require its pkg-config file as a dependency instead of manually linking to its lib.

### DIFF
--- a/lib/jxl_cms.cmake
+++ b/lib/jxl_cms.cmake
@@ -24,13 +24,14 @@ target_include_directories(jxl_cms PUBLIC
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>")
 
 set(JXL_CMS_PK_LIBS "")
+set(JPEGXL_CMS_LIBRARY_REQUIRES "")
 
 if (JPEGXL_ENABLE_SKCMS)
   target_link_libraries(jxl_cms PRIVATE skcms)
 else()
   target_link_libraries(jxl_cms PRIVATE lcms2)
   if (JPEGXL_FORCE_SYSTEM_LCMS2)
-    set(JXL_CMS_PK_LIBS "-llcms2")
+    set(JPEGXL_CMS_LIBRARY_REQUIRES "lcms2")
   endif()
 endif()
 
@@ -57,7 +58,6 @@ install(TARGETS jxl_cms
         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
-set(JPEGXL_CMS_LIBRARY_REQUIRES "")
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/jxl/libjxl_cms.pc.in"
                "libjxl_cms.pc" @ONLY)
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/libjxl_cms.pc"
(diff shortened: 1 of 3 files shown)


kmilos:
This will only look for lcms2 when static linking, don't you want it in any case? I.e. in `Requires` instead of `Requires.private`?

kmilos:
> This will only look for lcms2.pc when static linking, don't you want it in any case? I.e. in `Requires` instead of `Requires.private`? Like it was `-llcms2` in `Libs` instead of `Libs.private`? Or you're saying this should've been in `Libs.private` previously, clients of `libjxl_cms` don't access the lcms2 API directly anyway?

lib/jxl_cms.cmake
```diff
@@ -24,13 +24,14 @@ target_include_directories(jxl_cms PUBLIC
...
 set(JXL_CMS_PK_LIBS "")
+set(JPEGXL_CMS_LIBRARY_REQUIRES "")
 
 if (JPEGXL_ENABLE_SKCMS)
   target_link_libraries(jxl_cms PRIVATE skcms)
 else()
   target_link_libraries(jxl_cms PRIVATE lcms2)
   if (JPEGXL_FORCE_SYSTEM_LCMS2)
-    set(JXL_CMS_PK_LIBS "-llcms2")
+    set(JPEGXL_CMS_LIBRARY_REQUIRES "lcms2")
```
kmilos, line 34: I guess `JXL_CMS_PK_LIBS` could now be removed from here and the .pc.in template...

BtbN:
> This will only look for lcms2.pc when static linking, don't you want it in any case? I.e. in `Requires` instead of `Requires.private`? Like it was `-llcms2` in `Libs` instead of `Libs.private`? Or you're saying this should've been in `Libs.private` previously, clients of `libjxl_cms` don't access the lcms2 API directly anyway?

You want to link lcms2 if you are linking jxl as static library. If jxl is a shared library, lcms2 is either already built in (if it itself was static) or the jxl shared object will refer to its shared library all on its own.
And the usual place to put static-linking-only dependencies is in the `XYZ.private` variables, so that you don't end needlessly linking them when making a shared build.

It'd be in theory possible to alter the behaviour, depending on whether jxl is being build as a shared or static lib, but that seems like a bigger undertaking.

lib/jxl_cms.cmake
BtbN: I didn't remove it, since before this commit, JPEGXL_CMS_LIBRARY_REQUIRES was just sitting there unused.
So I figured they're kept around as easy to use future options.

Can easily remove it though, if preferred.

mo271:
is this fixed by https://github.com/libjxl/libjxl/pull/3061?

lib/jxl_cms.cmake
```diff
@@ -24,13 +24,14 @@ target_include_directories(jxl_cms PUBLIC
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>")
 
 set(JXL_CMS_PK_LIBS "")
```
mo271, line 26:
```suggestion
```

mo271 (approved):

veluca93 (dismissed):

mo271 (approved):

lib/jxl_cms.cmake
mo271: yeah, let's remove it.

mo271:
Thanks for the contribution, @BtbN !

PR #3046, opened 2023-12-20
fix RoundtripAnimationPatches

mo271:


mo271:
https://github.com/libjxl/libjxl/actions/runs/7273250477/job/19816876044

veluca93:
Unclear to me why it doesn't fail everywhere, but ok...

veluca93 (approved):

PR #3061, opened 2023-12-22
Fix pkgconfig for linking to an installed static library.

veluca93:
Drive-by: add some system deps to flake.nix

### DIFF
--- a/lib/jxl.cmake
+++ b/lib/jxl.cmake
@@ -257,6 +257,12 @@ install(TARGETS jxl
 set(JPEGXL_LIBRARY_REQUIRES
     "libhwy libbrotlienc libbrotlidec libjxl_cms")
 
+if (BUILD_SHARED_LIBS)
+  set(JPEGXL_REQUIRES_TYPE "Requires.private")
+else()
+  set(JPEGXL_REQUIRES_TYPE "Requires")
+endif()
+
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/jxl/libjxl.pc.in"
                "libjxl.pc" @ONLY)
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/libjxl.pc"
--- a/flake.nix
+++ b/flake.nix
@@ -26,6 +26,8 @@
               libclang.python
               libpng
               giflib
+              lcms2
+              brotli
             ];
             shellHook = ''
               export CC=clang
--- a/lib/jxl/libjxl.pc.in
+++ b/lib/jxl/libjxl.pc.in
@@ -6,7 +6,7 @@ includedir=@PKGCONFIG_TARGET_INCLUDES@
 Name: libjxl
 Description: Loads and saves JPEG XL files
 Version: @JPEGXL_LIBRARY_VERSION@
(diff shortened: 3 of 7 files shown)


mo271 (approved):

eustas (approved):

lib/jxl.cmake
```diff
@@ -257,6 +257,12 @@ install(TARGETS jxl
 set(JPEGXL_LIBRARY_REQUIRES
     "libhwy libbrotlienc libbrotlidec libjxl_cms")
 
+if (BUILD_SHARED_LIBS)
+  set(JPEGXL_REQUIRES_TYPE "Requires.private")
+else()
+  set(JPEGXL_REQUIRES_TYPE "Requires")
+endif()
+
```
kmilos, line 265: I could be wrong, but I think this paradigm (for all .pc files here) might not be fully correct?

From my understanding, you either rely on the depending library API directly, or you don't, regardless of linkage type.

In the former case, it goes into `Requires`, and in the latter it goes into `Requires.private`, again, regardless of linkage type. In the latter, indirect case, the dynamic linking is transitive anyway so it doesn't have to be in `Required`, but is necessary for static linkage only.

sboukortt: In our case, those libraries are conceptually `Requires.private`, but if we build a static `libjxl.a` instead of a dynamic `libjxl.so`, users do need them in their linking command.

kmilos: > In our case, those libraries are conceptually `Requires.private`, but if we build a static `libjxl.a` instead of a dynamic `libjxl.so`, users do need them in their linking command.

And they will be even if they're in `Requires.private`.

kmilos: There is also the scenario (e.g. common on MSYS2 and maybe some other distros), where you build both a static and shared lib, but ship only one .pc file.

sboukortt: > > In our case, those libraries are conceptually `Requires.private`, but if we build a static `libjxl.a` instead of a dynamic `libjxl.so`, users do need them in their linking command.
> 
> And they will be even if they're in `Requires.private`.

That doesn’t match our experience. (We wouldn’t have looked for a workaround like this one in the first place if we hadn’t encountered the issue.) pkg-config will add them if invoked as `pkg-config --static`, but apparently not if invoked normally and it just so happens that the only available `-ljxl` is a `.a`.

> There is also the scenario (e.g. common on MSYS2 and maybe some other distros), where you build both a static and shared lib, but ship only one .pc file.

That works with `Requires.private` because in that case, the static library is only ever used with `pkg-config --static`.

kmilos: > pkg-config will add them if invoked as `pkg-config --static`, but apparently not if invoked normally and it just so happens that the only available `-ljxl` is a `.a`.

Sure. But one could then argue that invoking `pkg-config` w/o `--static` when indenting to do static linking is somehow "not normal", no?

sboukortt: True, but the user may not necessarily intend to link statically specifically. They may just want to use whatever the installed libjxl happens to be, without committing to either one or knowing which one it is.

PR #3079, opened 2023-12-27
fix release contens permissions

mo271:


mo271:
The release workflow stopped uploading the artifacts after https://github.com/libjxl/libjxl/pull/2485, this should fix it.

veluca93 (approved):

PR #3088, opened 2024-01-02
refactor file extension handling

### LINKED ISSUE #3068: djxl: Full file path not working on Windows
djxl fails to write the output file if the output file name is specified using the full path.
```
d:\0>djxl test.jxl d:\test.png
JPEG XL decoder v0.10.0 ed80bfe [SSE4,SSSE3,SSE2]
can't decode to the file extension '.d'
```
However, it works if drive letter is not used.
```
d:\0>djxl test.jxl \test.png
JPEG XL decoder v0.10.0 ed80bfe [SSE4,SSSE3,SSE2]
Decoded to pixels.
512 x 512, 9.242 MP/s [9.24, 9.24], 1 reps, 4 threads.
```

mo271:
fixes #3068


lib/extras/dec/decode.cc
```diff
@@ -24,18 +24,8 @@ constexpr size_t kMinBytes = 9;
 
 void BasenameAndExtension(const std::string& path, std::string* filename,
```
szabadka, line 25: It seems that now in all code paths *filename = path, so I think we should simplify this function to something like ```std::string GetEtension(const std::string& path)```

lib/extras/dec/decode.cc
```diff
@@ -55,7 +45,7 @@ Codec CodecFromPath(std::string path, size_t* JXL_RESTRICT bits_per_sample,
   std::string ext;
   BasenameAndExtension(path, &base, &ext);
   if (filename) *filename = base;
```
szabadka, line 47: like wise in this function, we don't need the filename output any more.

lib/extras/dec/decode.cc
mo271: good idea, done

lib/extras/dec/decode.cc
mo271: also done

mo271:
adding @jonsneyers, because I think the "ext:name" syntax, which we remove here was added by him..

szabadka (approved):

PR #3098, opened 2024-01-04
Fix CheckOutOfBounds on 32-bit systems.

veluca93:


mo271:
I'm trying to fix the failing EncoderTest.CMYK here: #3099

sboukortt (approved):

Issue #3113, opened 2024-01-07
Two questions from non-programmers


DeepTulip:
1, Why does the v0.9.0 Window binary no longer contain the jxl.dll file? Compiling jxl.dll is very difficult for me, I have been using jxl.dll from releases in my scripting language.

2, JxlPixelFormat() seems to have been removed, so does the decoder now support outputting the BGRA format?

Traneptora:
Looks like the `jxl.dll` missing was just an oversight that got re-added. However, what do you mean by JxlPixelFormat has been removed? In FFmpeg we use this structure and we compile against 0.9.0.

DeepTulip:
To be precise, the JxlDecoderDefaultPixelFormat function was removed. Its documentation said it could define the output format, but it didn't actually support the BGRA format. I would like to know if its replacement supports the BGRA format.

Traneptora:
BGRA is still not supported by libjxl. RGB with alpha is only supported in the RGBA order.

fabiorug:
Open the issue.

DeepTulip:
Reopening this issue, is there a plan to support BGRA?
I'm very much looking forward to it.

I was introduced to both avif and jxl formats almost at the same time, but now I have 100K+ images on my computer that were converted to .avif and still haven't officially used the jxl format just because it doesn't support the BGRA format.

Traneptora:
Why does it matter that much? You can always convert from RGBA pixel data to BGRA with libswscale.

DeepTulip:
> Why does it matter that much? You can always convert from RGBA pixel data to BGRA with libswscale.

Because it's just "convertible", not optimal. I can convert even without libswscale.
And I think customizing the output format is a basic feature, without which jxl is an unfinished flawed product.

jonnyawsom3:
Closing in favour of #2519

PR #3117, opened 2024-01-08
Re-add jxl.dll asset

eustas:
fixes part of #3113

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -253,6 +253,14 @@ jobs:
             arch: '-A Win32'
           - triplet: x64-windows-static
             arch: '-A x64'
+          - triplet: x86-windows
+            arch: '-A Win32'
+            static: 'OFF'
+            tools: 'OFF'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -253,6 +253,10 @@ jobs:
             arch: '-A Win32'
           - triplet: x64-windows-static
             arch: '-A x64'
+          - triplet: x86-windows-dynamic
+            arch: '-A Win32'
+          - triplet: x64-windows-dynamic
```
mo271, line 258:
```suggestion
          - triplet: x86-windows
            arch: '-A Win32'
          - triplet: x64-windows
```
Does foo-windows is dynamic by default?

mo271: https://github.com/libjxl/libjxl/actions/runs/7445929299/job/20255177651?pr=3117

mo271 (dismissed):

mo271 (approved):

PR #3150, opened 2024-01-17
Add test for streaming and non-streaming pixel equivalence.

szabadka:
Also fix the streaming encoder tests to not set the streaming flag in the non-streaming mode.

### DIFF
--- a/lib/jxl/encode_test.cc
+++ b/lib/jxl/encode_test.cc
@@ -1774,7 +1774,7 @@ class EncoderStreamingTest : public testing::TestWithParam<StreamingTestParam> {
   static void SetupEncoder(JxlEncoderFrameSettings* frame_settings,
                            const StreamingTestParam& p,
                            const JxlBasicInfo& basic_info,
-                           size_t number_extra_channels) {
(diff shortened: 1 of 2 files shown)


lib/jxl/encode_test.cc
```diff
@@ -1901,7 +1906,7 @@ TEST_P(EncoderStreamingTest, OutputCallback) {
   }
 
   EXPECT_TRUE(SameDecodedPixels(compressed, streaming_compressed));
-  EXPECT_LE(streaming_compressed.size(), compressed.size() + 16);
+  EXPECT_LE(streaming_compressed.size(), compressed.size() * 1.02 + 1024);
```
veluca93, line 1909: is there some deeper reason behind this formula? I assume not, but...

veluca93 (approved):

lib/jxl/encode_test.cc
szabadka: No deep reason, and I realized that the multiplier is not needed since these images are all small.

veluca93 (approved):

fabiorug:
Yes

PR #3154, opened 2024-01-18
Disable streaming pixel equivalence test on non-x86-64 systems.

szabadka:


lib/jxl/jxl_test.cc
```diff
@@ -1635,6 +1635,8 @@ JXL_GTEST_INSTANTIATE_TEST_SUITE_P(
     JxlStreamingTest, JxlStreamingTest,
     testing::ValuesIn(StreamingTestParam::All()));
 
+// This is broken on mingw32, so we only enable it for x86_64 now.
+#if defined(__x86_64__)
```
veluca93, line 1639: Can we do this with `DISABLED_` instead?

Something like `#define SAME_PIXEL_TEST DISABLED_StreamingSamePixels` or thereabouts

szabadka: Done.

veluca93 (approved):

Issue #1794, opened 2022-09-28
Remove CodecInOut


mo271:
CodecInOut is only used in tests after the transition to using the API functions. Let's remove it and also use the API in the tests.

eustas:
Current state:
 - in `jxl/` used __only__ in tests
 - used in `extras/`
 - used in `tools/`

PR #3174, opened 2024-01-22
Use PPF instead of CodecInOut in the ImageCodec API.

szabadka:


veluca93 (approved):

mo271:
a step toward #1794...

lib/extras/enc/apng.cc
```diff
@@ -344,13 +344,11 @@ Status APNGEncoder::EncodePackedPixelFileToAPNG(
                  PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_BASE,
                  PNG_FILTER_TYPE_BASE);
     if (count == 0) {
-      if (!MaybeAddSRGB(ppf.color_encoding, png_ptr, info_ptr)) {
+      if (!ppf.icc.empty()) {
+        png_set_benign_errors(png_ptr, 1);
+        png_set_iCCP(png_ptr, info_ptr, "1", 0, ppf.icc.data(), ppf.icc.size());
+      } else if (!MaybeAddSRGB(ppf.color_encoding, png_ptr, info_ptr)) {
```
sboukortt, line 350: Out of curiosity, why was this changed? This means that we now never add cICP/cHRM/gAMA chunks anymore if there is an ICC profile.

szabadka: There were some ppfs, I think produced by the PNG ppf decoder that had an ICC profile, but color_encoding was set to sRGB, so the ICC profile was dropped.

Aren't cICP/cHRM/gAMA chunks redundant if we have an ICC profile?

sboukortt: If the color_encoding is indeed equivalent to sRGB, then it should be fine to have an sRGB chunk, and this precludes an ICC chunk (https://www.w3.org/TR/png/#11sRGB “It is recommended that the sRGB and iCCP chunks do not appear simultaneously in a PNG datastream.”).

In case we do write an ICC profile, the PNG specification [suggests](https://www.w3.org/TR/png/#11iCCP) that “A PNG encoder that writes the iCCP chunk is encouraged to also write gAMA chunks that approximate the ICC profile, to provide compatibility with applications that do not use the iCCP chunk.”

Regarding cICP: in situations where we would write one, there is probably the same CICP tag in the ICC profile that we write anyway, but we might as well cover all our bases.

PR #3196, opened 2024-01-25
Fix jxl_export CMake target to be a INTERFACE library

mgorny:
### Description

Fix the jxl_export target to be a INTERFACE library rather than an OBJECT library.  The latter type is only correct for libraries that actually include any source files, and therefore produce an actual object file.  Using it with jxl_export causes CMake to produce an invalid ninja file.  While ninja ignores the problem, samurai is more strict and rejects the file:

    samu: file is missing and not created by any action: 'lib/CMakeFiles/jxl_export.dir'

INTERFACE libraries "do not compile sources and do not produce library artifacts on disk", and therefore are more correct for a header-only target.  See:
https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

### Pull Request Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -69,6 +69,7 @@ Maarten DB <anonymous.maarten@gmail.com>
 Marcin Konicki <ahwayakchih@gmail.com>
 Martin Strunz
 Mathieu Malaterre <mathieu.malaterre@gmail.com>
+Michał Górny <mgorny@gentoo.org>
 Mikk Leini <mikk.leini@krakul.eu>
 Misaki Kasumi <misakikasumi@outlook.com>
 Moonchild Straver <moonchild@palemoon.org>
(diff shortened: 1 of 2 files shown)


eustas:
There is a comment right before that `add_library`:
```
# CMake does not allow generate_export_header for INTERFACE library, so we
# add this stub library just for file generation.
```

Also `jxl_export` does not contain any compilation units. Current solution is far from ideal. Will investigate soon if we can improve it (e.g. land this PR if the comment is no longer true for modern CMake).

eustas:
Alas, CMake build fails. Will check what else we can do.

eustas:
(Perhaps we will need to add Samurai to CI as well, to make sure we will not break it in the future)

mgorny:
> Alas, CMake build fails. Will check what else we can do.

Ah, sorry. I didn't notice because it was using `jxl_export.h` from the previously installed version.

> Alas, CMake build fails. Will check what else we can do.

Thanks a lot.

mgorny:
Thanks! I love the simplicity of this fix.

PR #3198, opened 2024-01-26
correct docstring for JxlDecoderGetBoxSizeRaw

### LINKED ISSUE #2022: JxlDecoderGetBoxSizeRaw documentation is incorrect
It actually returns the entire raw box size as the name suggests, including the 4 bytes of the box length, and the 4 bytes of the box type (at least using libjxl 0.7.0), vs. just the box content length as claimed.

So either the API documentation needs to be updated, or the implementation needs to change to match (breaking)...

mo271:
fixes #2022


lib/include/jxl/decode.h
```diff
@@ -1359,15 +1359,14 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderGetBoxType(JxlDecoder* dec,
...
- * bytes indicating "brob". This function gives the size of the data that will
- * be written in the output buffer when getting boxes in the default raw
- * compressed mode. When @ref JxlDecoderSetDecompressBoxes is enabled, the
- * return value of function does not change, and the decompressed size is not
- * known before it has already been decompressed and output.
+ * JXL_DEC_BOX event. This is the size of the contents, including the 4 bytes
+ * indicating the box size, and the 4 bytes indicating the type and optionally
+ * the 8 bytes indicating a the size (for extended boxes). This function gives
+ * the size of the data that will be written in the output buffer when getting
+ * boxes in the default raw compressed mode. When @ref
```
veluca93, line 1366: This seems false now, no?

mo271: yes, corrected it, please check

veluca93 (approved):

PR #3208, opened 2024-01-29
Fix OOB

eustas:


lib/jxl/enc_icc_codec.cc
```diff
@@ -226,6 +235,7 @@ Status PredictICC(const uint8_t* icc, size_t size, PaddedBytes* result) {
   // all point to the same one.
   Tag tag;
   size_t tagstart = 0, tagsize = 0, clutstart = 0;
+  bool tagsane = false;
```
veluca93, line 238: Can you move this definition to where it is initialized? This is not C89 :)

eustas: It it initialized under one if and used under the other (+ in the loop). So better to have it initialized.
I see it more as a derivative of tagSize. Will turn in to lambda(tagsize) for better readability

veluca93 (approved):

veluca93 (approved):

PR #3218, opened 2024-01-29
GDK Pixbuf Loader Fixes

### LINKED ISSUE #3144: 0.9.1 red-room.jxl fails to load with gdk-pixbuf plugin
**Describe the bug**
In jxl 0.8.2, the red-room.jxl test image is able to be loaded in EOG using the gdk-pixbuf plugin.

In jxl 0.9.1 the red-room jxl test image fails to load in EOG.

The test image can be downloaded here:
https://jpegxl.info/test-page/

**To Reproduce**
1. Compile the jxl 0.9.1 release with no patches
2. Download red-room.jxl from here: https://jpegxl.info/test-page/
3. Run `gdk-pixbuf-pixdata red-room.jxl test.pixdata`
4. It outputs `failed to load "red-room.jxl": Failed to load image “red-room.jxl”: JxlDecoderSetImageOutBuffer failed`

The error is due to here: https://github.com/libjxl/libjxl/blob/8e7dc5eff3d3dc92281ce4edbd5ea86e58c8aaf0/lib/jxl/decode.cc#L2443 as, gdb outputs:
```
(gdb) finish
Run till exit from #0  JxlDecoderImageOutBufferSize (
    dec=dec@entry=0x555555581750, format=format@entry=0x55555556dc50, 
    size=size@entry=0x7ffffffee0e0)
    at /builddir/libjxl-0.9.1/libjxl-0.9.1/lib/jxl/decode.cc:2402
JxlDecoderSetImageOutBuffer (dec=0x555555581750, format=0x55555556dc50, 
    buffer=0x7ffff44b0010, size=1351980)
    at /builddir/libjxl-0.9.1/libjxl-0.9.1/lib/jxl/decode.cc:2441
2441	in /builddir/libjxl-0.9.1/libjxl-0.9.1/lib/jxl/decode.cc
Value returned is $1 = JXL_DEC_SUCCESS
(gdb) print min_size
$2 = 1354416
(gdb) print size
$3 = 1351980
```

This became a problem in 0.9.0 due to https://github.com/libjxl/libjxl/commit/2ef999bbaa3e5d03b1894d224b6f44e756b6822e and reverting the commit does work around the issue.

Using `gdk_pixbuf_get_pixels_with_length()`, the length of pixel buffer is `1354413` which is also too small.

**Expected behavior**

The pixdata file is outputted with no errors.

This issue is specific to the gtk-pixbuf plugin.
This image loads fine with gimp and krita.
It also mostly works with djxl, `djxl -v red-room.jxl test.jpeg` outputs the file as expected however `djxl -v red-room.jxl --disable_output` aborts in packed_image.h on line 87. May not be related though.

**Environment**
 - OS: Void Linux
 - Compiler version: Tested with both gcc 12 and gcc 13
 - CPU type: x86_64
 - cjxl/djxl version string: djxl v0.9.1 0.9.1 [AVX2,SSE4,SSE2]

Traneptora:
**plugins/gdk-pixbuf: consider stride when passing buffer to libjxl**
    
    This change uses gdk_pixbuf_get_pixels_with_length to get the actual
    size of the buffer, with stride taken into account. The previous code
    just multiplies width by height and ignores stride.
    
    Signed-off-by: Leo Izen <leo.izen@gmail.com>

**lib/jxl/decode.cc: deduplicate and correct min_size calculation**
    
    Code to calculate the proper size of the buffer exists in three places
    in this function - one for the standard buffer, one for the extra
    channel buffer, and one for the preview buffer. However, only the code
    for the preview buffer correctly took into account the last row size
    possibly being slightly smaller than the remaining rows.
    
    This commit separates the code out into a separate function to remove
    duplicate code, and uses the algorithm from the prevew buffer which was
    the only correct one.
    
    Signed-off-by: Leo Izen <leo.izen@gmail.com>

Fixes #3144.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -2348,27 +2348,38 @@ JXL_EXPORT JxlDecoderStatus JxlDecoderSetCms(JxlDecoder* dec,
   return JXL_DEC_SUCCESS;
 }
 
-JXL_EXPORT JxlDecoderStatus JxlDecoderPreviewOutBufferSize(
-    const JxlDecoder* dec, const JxlPixelFormat* format, size_t* size) {
+static JxlDecoderStatus GetMinSize(const JxlDecoder* dec,
+                                   const JxlPixelFormat* format,
+                                   size_t num_channels, size_t* min_size,
+                                   bool preview) {
   size_t bits;
   JxlDecoderStatus status = PrepareSizeCheck(dec, format, &bits);
   if (status != JXL_DEC_SUCCESS) return status;
-  if (format->num_channels < 3 &&
-      !dec->image_metadata.color_encoding.IsGray()) {
-    return JXL_API_ERROR("Number of channels is too low for color output");
+  size_t xsize, ysize;
+  if (preview) {
+    xsize = dec->metadata.oriented_preview_xsize(dec->keep_orientation);
+    ysize = dec->metadata.oriented_preview_ysize(dec->keep_orientation);
+  } else {
+    GetCurrentDimensions(dec, xsize, ysize);
   }
-
-  size_t xsize = dec->metadata.oriented_preview_xsize(dec->keep_orientation);
-  size_t ysize = dec->metadata.oriented_preview_ysize(dec->keep_orientation);
-
+  if (num_channels == 0) num_channels = format->num_channels;
   size_t row_size =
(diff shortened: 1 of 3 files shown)


veluca93 (dismissed):

oreo639:
Thanks.
Testing this, it does appear to work for me.

mo271 (approved):

PR #3228, opened 2024-01-31
Enable streaming mode for modular mode. 

veluca93:
Different group sizes still TODO.

### DIFF
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -1647,11 +1647,12 @@ bool CanDoStreamingEncoding(const CompressParams& cparams,
   if (cparams.max_error_mode) {
     return false;
   }
-  if (cparams.modular_mode) {
+  if (cparams.modular_group_size_shift != 1 &&
+      cparams.modular_group_size_shift != -1) {
     return false;
   }
   if (!cparams.ModularPartIsLossless() || cparams.responsive > 0) {
-    if (metadata.m.num_extra_channels > 0) {
+    if (metadata.m.num_extra_channels > 0 || cparams.modular_mode) {
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


lib/jxl/enc_frame.cc
```diff
@@ -1647,7 +1647,8 @@ bool CanDoStreamingEncoding(const CompressParams& cparams,
   if (cparams.max_error_mode) {
     return false;
   }
-  if (cparams.modular_mode) {
+  if (cparams.modular_group_size_shift != 1 &&
+      cparams.modular_group_size_shift != -1) {
```
szabadka, line 1651: now if (metadata.m.num_extra_channels > 0) condition should be removed below, otherwise we would attempt to streaming encode modular, responsive images without alpha.

lib/jxl/jxl_test.cc
```diff
@@ -1699,5 +1699,26 @@ TEST(JxlTest, JXL_X86_64_TEST(StreamingSamePixelsAlpha)) {
   EXPECT_TRUE(jxl::test::SamePixels(ppf_out, ppf_out_streaming));
 }
 
+TEST(JxlTest, JXL_X86_64_TEST(StreamingSamePixelsLossless)) {
```
szabadka, line 1702: maybe it would be worth converting it to a parameterized test now and adding a few more modular settings like palette and modular color space and such

lib/jxl/enc_frame.cc
veluca93: Not quite - I believe it should be changed to `metadata.m.num_extra_channels > 0 || cparams.modular_mode`.

lib/jxl/jxl_test.cc
veluca93: Good point - will do that as a follow-up PR (after the group sizes also go in).

Ideally it should also include the lossy case and just test various different combinations...

szabadka (approved):

Issue #2289, opened 2023-03-11
djxl PNG Conversion Acting Up


gianni-rosato:
**Describe the bug**
When converting a particular JXL image to PNG, the PNG has noticeable color issues with patches of discoloration.

**To Reproduce**
Download [this](https://cdn.discordapp.com/attachments/794206170445119489/1083872278128885790/imagetest15-Q70.jxl)

djxl theimage.jxl out.png

Compare the png to the JXL.

**Expected behavior**
The PNG would be visually identical to the JXL, or at least close enough where I wouldn't see obvious discoloration or where the PNG is earning horrible ssimu2 scores.

**Screenshots**
If applicable, add screenshots or example input/output images to help explain your problem.

**Environment**
 - OS: Linux
 - Compiler version: gcc (GCC) 12.2.1 20230201
 - CPU type: x86_64
 - cjxl/djxl version string: djxl v0.9.0 13caf6f8 [AVX2,SSE4,SSSE3,SSE2]

**Additional context**
The problem persists with other versions of cjxl/djxl when using the source PNG (which is almost 100mb).

thebombzen:
Notably, the JXL file has an attached ICC profile and is XYB encoded. djxl is outputting a PNG in Linear Light rather than in sRGB.

gianni-rosato:
If you want the source PNG, let me know

novomesk:
It is the fact that djxl uses `RGB_D65_SRG_Rel_Lin` profile there.

If the JXL and decoded PNG look significantly visually different, that's a problem of non-implemented (or not working) color management support of the viewer application.

The functionality of djxl is not technically wrong.

Color profile used to decode can be changed:

`djxl --color_space=RGB_D65_SRG_Rel_SRG imagetest15-Q70.jxl out2.png`

gianni-rosato:
When inputting the weird PNG into ssimulacra2, it doesn't understand it & gives it a terrible score

kmilos:
Did you try the "keep original" color space option when exporting from darktable? The JPEG XL file will be considerably larger, but it'll preserve the profile. Otherwise, it discards it when encoding and only decodes to linear RGB (unless you're doing lossless, i.e. quality 100 as well when you get out what you put in).

jonsneyers:
There are three things that would fix this, and I think we need to do all of them:
1. Make cjxl recognize Display P3 and use the Enum representation for it rather than a generic icc profile. This would help for compression, and it circumvents the djxl bug.
[...]

PR #3233, opened 2024-02-01
Fix #2581 in a different way

### LINKED ISSUE #2581: Losslessly recompressing JXL turns enum tags into ICC profiles

sboukortt:
Always produce the ICC profile, but also indicate whether it’s the primary colorspace representation or not.

This lets `enc/jxl` ignore it if it’s not, and `enc/apng` include it for compatibility, helping with #2289.

### DIFF
--- a/lib/extras/packed_image.h
+++ b/lib/extras/packed_image.h
@@ -244,9 +244,21 @@ class PackedPixelFile {
   std::vector<PackedExtraChannel> extra_channels_info;
 
   // Color information of the decoded pixels.
-  // If the icc is empty, the JxlColorEncoding should be used instead.
-  std::vector<uint8_t> icc;
+  // `primary_color_representation` indicates whether `color_encoding` or `icc`
+  // is the “authoritative” encoding of the colorspace, as opposed to a fallback
+  // encoding. For example, if `color_encoding` is the primary one, as would
+  // occur when decoding a jxl file with such a representation, then `enc/jxl`
+  // will use it and ignore the ICC profile, whereas `enc/png` will include the
+  // ICC profile for compatibility.
+  // If `icc` is the primary representation, `enc/jxl` will preserve it when
+  // compressing losslessly, but *may* encode it as a color_encoding when
+  // compressing lossily.
+  enum {
+    kColorEncodingIsPrimary,
+    kIccIsPrimary
+  } primary_color_representation = kColorEncodingIsPrimary;
   JxlColorEncoding color_encoding = {};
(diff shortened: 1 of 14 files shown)


lib/extras/packed_image.h
```diff
@@ -244,9 +244,12 @@ class PackedPixelFile {
   std::vector<PackedExtraChannel> extra_channels_info;
 
   // Color information of the decoded pixels.
-  // If the icc is empty, the JxlColorEncoding should be used instead.
-  std::vector<uint8_t> icc;
+  enum {
+    kPreferColorEncoding,
+    kPreferIcc
+  } preferred_color_profile = kPreferColorEncoding;
```
jonsneyers, line 250: Could you add a comment here to explain what this toggle means / how it is used in enc/dec?

I'm not sure if `preferred_color_profile` is the best name for this variable, maybe something like `primary_colorspace_description` could make sense, with the other one being a 'secondary' colorspace description which is supposed to approximate the primary colorspace description, but it's the primary one that is authoritative and takes precedence, and the secondary one is only used to maximize compatibility and/or in cases where the format does not support signaling the primary one.

sboukortt: Thanks, something like this?

jonsneyers (approved):

lib/extras/packed_image.h
jonsneyers: LGTM!

PR #3254, opened 2024-02-06
Fix an incorrect check with kReferenceOnly non-patch non-crop frames.

veluca93:


lib/jxl/frame_header.cc
```diff
@@ -397,10 +397,14 @@ Status FrameHeader::VisitFields(Visitor* JXL_RESTRICT visitor) {
...
+      size_t ysize = custom_size_or_origin ? frame_size.ysize
+                                           : nonserialized_metadata->ysize();
       if (!save_before_color_transform &&
-          (frame_size.xsize < nonserialized_metadata->xsize() ||
-           frame_size.ysize < nonserialized_metadata->ysize() ||
-           frame_origin.x0 != 0 || frame_origin.y0 != 0)) {
+          (xsize < nonserialized_metadata->xsize() ||
+           ysize < nonserialized_metadata->ysize() || frame_origin.x0 != 0 ||
+           frame_origin.y0 != 0)) {
         return JXL_FAILURE(
```
jonsneyers, line 408: Shouldn't the error message below also use xsize/ysize?

veluca93: yup

jonsneyers (approved):

PR #3274, opened 2024-02-11
Fix the usage of fixed trees with streaming mode.

veluca93:
The total number of pixels was being incorrectly set to 0.


jonsneyers (dismissed):

jonsneyers:
What was the impact of setting it to zero?

veluca93:
> What was the impact of setting it to zero?

An assertion failure in debug builds, and either too big or too small trees for streaming mode.

jonsneyers (dismissed):

jonsneyers (approved):

PR #3201, opened 2024-01-26
Reject decoding to JPEG when DCT coefficients are invalid

eustas:

PR #3282, opened 2024-02-13
add minor versions to changelog

mo271:


eustas (approved):

eustas:
#3201? That is OOB

mo271:
Here I'm just copying what we had in the individual release note.

wwinniww:
Hello
Is the year for the last two entries correct? Should it be 2024 instead?

mo271:
> Hello Is the year for the last two entries correct? Should it be 2024 instead?

Yes, thanks, I will fix it...

PR #3283, opened 2024-02-14
(Just a test)

sboukortt:


sboukortt:
This has now fulfilled its purpose (to check whether #3289 was needed). Not sure whether we would like to keep `decode_oneshot` in C?

mo271:
> This has now fulfilled its purpose (to check whether #3289 was needed). Not sure whether we would like to keep `decode_oneshot` in C?

I have no strong opinion on that? What are the pros and cons here?

PR #3289, opened 2024-02-15
Include `-lm` and the C++ standard library in pkgconfig “Libs” as appropriate

sboukortt:
This ensures that if libjxl is built statically, C programs that use it will still link properly.


sboukortt:
(I mean, maybe not “ensures”, but at least removes a reason for it not to be the case.)

veluca93 (approved):

PR #3369, opened 2024-03-05
stop producing non-spec compliant files

jonsneyers:
This fixes the encoder to compute the `distance_multiplier` for lz77 correctly (i.e. not looking at meta-channels), while also avoiding to use it when the incorrect computation is different, so it does not produce files that cannot be decoded by a decoder that does the computation incorrectly (like the libjxl decoder at the moment).

At some point we can change this and turn it into an encoder option to allow producing bitstreams that cannot be decoded by non-spec-conforming decoders, but let's first make sure that such decoders are more ubiquitous :)

### DIFF
--- a/lib/jxl/modular/encoding/enc_encoding.cc
+++ b/lib/jxl/modular/encoding/enc_encoding.cc
@@ -642,16 +642,22 @@ Status ModularEncode(const Image &image, const ModularOptions &options,
   }
 
   size_t image_width = 0;
+  size_t image_width_bug = 0;
   size_t total_tokens = 0;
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

jonsneyers:
This doesn't fully fix things yet, I don't really understand why but it still produces files that don't decode.

jonsneyers:
Ah, this still causes it to produce bad lz77 distances since the encoder thinks there are no special distances while the decoder disagrees. The real fix is to make the encoder only use non-special distances while still making it act as if there are special distances. Ugh.

jonsneyers:
Closing. This is not a good fix.

PR #3381, opened 2024-03-06
Fix build for some compilers

eustas:


mo271 (approved):

GreenReaper:
This fixed my _other_ issue on FreeBSD with llvm11, so thanks for the patch.

PR #3419, opened 2024-03-13
extra channels for decoding to png

### LINKED ISSUE #3407: djxl does not output extra channels for png
**Describe the bug**
--output_extra_channels does not work and only a single PNG is written

**To Reproduce**
djxl suz-multichannel-800px.jxl --output_extra_channels multichannel.png

suz-multichannel-800px jxl
**Environment**
 - OS: Windows 10
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL decoder v0.10.2 e148959 [AVX2,SSE2]

mo271:
partially fixes #3407

### DIFF
--- a/lib/extras/enc/apng.cc
+++ b/lib/extras/enc/apng.cc
@@ -281,7 +281,13 @@ void MaybeAddCLLi(const JxlColorEncoding& c_enc, const float intensity_target,
 Status APNGEncoder::EncodePackedPixelFileToAPNG(
     const PackedPixelFile& ppf, ThreadPool* pool, std::vector<uint8_t>* bytes,
     bool encode_extra_channels, size_t extra_channel_index) const {
-  const auto& ec_info = ppf.extra_channels_info[extra_channel_index].ec_info;
+  JxlExtraChannelInfo ec_info{};
+  if (encode_extra_channels) {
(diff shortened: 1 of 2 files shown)


lib/extras/enc/apng.cc
```diff
@@ -440,6 +459,26 @@ Status APNGEncoder::EncodePackedPixelFileToAPNG(
   return true;
 }
 
+Status APNGEncoder::EncodeExtraChannelToPNG(const PackedImage& image,
+                                            size_t bits_per_sample,
+                                            std::vector<uint8_t>* bytes) const {
+  // Create a new PackedImage and move data from the original image
+  PackedImage moved_image(std::move(const_cast<PackedImage&>(image)));
```
szabadka, line 466: Since PNG encoding is used in a few other places than cjxl, it might have surprising effects (currently or in the future) if the const & ppf loses its extra channels, so it might be better to copy here. Or alternatively one could add extra arguments to EncodePackedPixelFileToAPNG to select which extra channel/frame to include in the output PNG.

mo271: I added extra arguments to `EncodePackedPixelFileToAPNG`.

szabadka (approved):

PR #3422, opened 2024-03-14
Corrected the checkout instruction in developing_in_github.md.

zond:
At least for me, "git clone git https://github.com/libjxl/libjxl --recursive" causes an error "fatal: repository git does not exist".

I believe the extra "git" is a typo.


mo271 (approved):

mo271:
Of course, thanks!

PR #3425, opened 2024-03-14
Added a tech report describing human evaluation of jpegli.

zond:
The tech report describes the study published at https://github.com/google-research/google-research/tree/master/mucped23.


zond:
I'll put this in https://github.com/google-research/google-research/tree/master/mucped23 instead.

Issue #3389, opened 2024-03-07
Should add STATIC property for jxl_extras_codec


LilyWangLL:
**Describe the bug**
`jxl_extras_codec` is a static library, but the codes [`add_library(jxl_extras_codec`](https://github.com/libjxl/libjxl/blob/491ee5ab4f17d35ec92a5e38ae2098683cf24ffa/lib/jxl_extras.cmake#L154) don't have `STATIC` property, so it will not create lib file when build dynamic, it just create dll file. This cause build tools failed, the [tools](https://github.com/libjxl/libjxl/blob/491ee5ab4f17d35ec92a5e38ae2098683cf24ffa/tools/CMakeLists.txt#L135-L139) need link library `jxl_extras_codec`.
**To Reproduce**
Open option `JPEGXL_ENABLE_TOOLS` and build `libjxl`.
**Expected behavior**
Build succeed.
**Screenshots**
If applicable, add screenshots or example input/output images to help explain your problem.

**Environment**
 - OS: [e.g. Windows]
 - Compiler version: [e.g. clang 11.0.1]
 - CPU type: [e.g. x86_64]
 - cjxl/djxl version string: [e.g. cjxl [v0.3.7 | SIMD supported: SSE4,Scalar]]

**Additional context**
The error log:
```
[...]

PR #3442, opened 2024-03-27
Update jxl_extras.cmake

ssh4net:
Fix for https://github.com/libjxl/libjxl/issues/3389
STATIC property for jxl_extras_codec

### Description

fix for jxl_extras.cmake from https://github.com/microsoft/vcpkg/pull/37211

### Pull Request Checklist

- [ ] **Code Style**: Have you ensured your code adheres to the project's coding style guidelines? You can use `./ci.sh lint` for automatic code formatting.


vt-idiot (approved):
Fixes compile errors on cjxl/djxl without forcing static compilation of everything else, works as intended now.

sboukortt (approved):
Thank you. I was somewhat hesitant to merge this without understanding why it was necessary, as I would have expected that:
```cmake
add_library(foo foo.cc)
add_executable(bar bar.cc)
target_link_libraries(bar foo)
```
would do the right thing even if `foo` was shared, but kkourin explained on Discord that:

> it doesn't work automatically for windows. cmake assumes theres a mylib.lib thunk lib generated for mylib, but if mylib doesn't export anything (either by putting dllexport in declarations or giving a .def file), the thunk will not be generated

With this newfound knowledge, I now feel more comfortable with merging this fix. If we decide later to make `jxl_extras` shared again, we can do it then, armed with this new information.

PR #3445, opened 2024-03-29
add shorthand colorspace descriptions

jonsneyers:
Adds a few special values for the colorspace description strings: `sRGB`, `DisplayP3`, `Rec2100PQ`, and `Rec2100HLG`. These are probably the most commonly used and relevant color spaces right now and in the foreseeable future.

This makes it somewhat more convenient to manually specify these colorspaces when e.g. using `cjxl` on a ppm or pfm input, that is, you can now also do `-x color_space=sRGB` instead of having to do `-x color_space=RGB_D65_SRG_Per_SRG` which I find hard to remember.

It also has the effect of making the description string a little shorter (and arguably easier to understand) in the corresponding synthesized ICC profiles.

### DIFF
--- a/lib/extras/dec/color_description.cc
+++ b/lib/extras/dec/color_description.cc
@@ -203,12 +203,38 @@ Status ParseTransferFunction(Tokenizer* tokenizer, JxlColorEncoding* c) {
 
 Status ParseDescription(const std::string& description, JxlColorEncoding* c) {
   *c = {};
-  Tokenizer tokenizer(&description, '_');
-  JXL_RETURN_IF_ERROR(ParseColorSpace(&tokenizer, c));
-  JXL_RETURN_IF_ERROR(ParseWhitePoint(&tokenizer, c));
-  JXL_RETURN_IF_ERROR(ParsePrimaries(&tokenizer, c));
-  JXL_RETURN_IF_ERROR(ParseRenderingIntent(&tokenizer, c));
-  JXL_RETURN_IF_ERROR(ParseTransferFunction(&tokenizer, c));
+  if (description == "sRGB") {
+    c->color_space = JXL_COLOR_SPACE_RGB;
+    c->white_point = JXL_WHITE_POINT_D65;
+    c->primaries = JXL_PRIMARIES_SRGB;
+    c->transfer_function = JXL_TRANSFER_FUNCTION_SRGB;
+    c->rendering_intent = JXL_RENDERING_INTENT_PERCEPTUAL;
+  } else if (description == "DisplayP3") {
+    c->color_space = JXL_COLOR_SPACE_RGB;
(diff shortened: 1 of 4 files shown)


sboukortt (approved):
Thanks!

kmilos:
> These are probably the most commonly used and relevant color spaces right now and in the foreseeable future.

P3 HLG (and P3 PQ?) options might be handy as well - I think this is what Adobe Camera Raw exports in AVIF, and what [Apple](https://developer.apple.com/documentation/coregraphics/cgcolorspace/3191857-displayp3_hlg) use as well.

jonsneyers:
> > These are probably the most commonly used and relevant color spaces right now and in the foreseeable future.
> 
> P3 HLG (and P3 PQ?) options might be handy as well - I think this is what Adobe Camera Raw exports in AVIF, and what [Apple](https://developer.apple.com/documentation/coregraphics/cgcolorspace/3191857-displayp3_hlg) use as well.

Adobe Camera Raw gives a choice between "HDR sRGB (Rec.709)", "HDR P3", and "HDR Rec.2020". It looks like these all use the PQ transfer function, and "HDR Rec.2020" corresponds to `Rec2100PQ`.
And yes, it looks like Apple uses P3 HLG.

Would it make sense to add `DisplayP3_HLG` and `DisplayP3_PQ`? (or without the underscore, or with a dash instead?) You can already use the longer description string `RGB_D65_DCI_Rel_HLG` / `RGB_D65_DCI_Rel_PeQ` for these though.

I want to keep the list of spaces with a shorthand name relatively short, because you could add spaces like Adobe98 and ProPhoto that are also still relatively common but at some point we're turning this into a colorspace name registry and that's not the intention :)

sboukortt:
I think I would have them without “Display”, because Display P3 is already “P3 primaries + sRGB white point and transfer function”, so replacing the transfer function again arguably makes it not Display P3 anymore.

kmilos:
> Would it make sense to add `DisplayP3_HLG` and `DisplayP3_PQ`? (or without the underscore, or with a dash instead?) You can already use the longer description string `RGB_D65_DCI_Rel_HLG` / `RGB_D65_DCI_Rel_PeQ` for these though.

I guess you're right - it's easier to add more shorthands later if really needed (i.e. frequently requested) than overthink it now...

jonsneyers:
> I think I would have them without “Display”, because Display P3 is already “P3 primaries + sRGB white point and transfer function”, so replacing the transfer function again arguably makes it not Display P3 anymore.

I agree, and also shorter is better. We could go both ways: Adobe uses the string "P3 PQ" (in the ICC description field) while Apple uses the string "displayP3_HLG". I would go for `P3PQ` and `P3HLG`, maybe. But for now, let's start with what's in this PR and more can be added later.

mo271 (approved):

Issue #3455, opened 2024-04-04
djxl silently failed to reconstruct some JPEG files


URenko:
**Describe the bug**
I use cjxl to recompress some JPEG files to JXL.
Then I use djxl to reconstruct them to JPEG, but it silently failed (no files were produced).
But I can successfully decode the JXL files into PNG files.

**To Reproduce**
JPEG file sample: [1c94fa30eea811e597bc25dc24f9c26c.zip](https://github.com/libjxl/libjxl/files/14865227/1c94fa30eea811e597bc25dc24f9c26c.zip)
```
$ cjxl 1c94fa30eea811e597bc25dc24f9c26c.jpg 1.jxl
JPEG XL encoder v0.10.2 e148959 [AVX2,SSE2]
Note: Implicit-default for JPEG is lossless-transcoding. To silence this message, set --lossless_jpeg=(1|0).
Encoding [JPEG, lossless transcode, effort: 7]
Compressed to 2746.1 kB including container
```
```
$ djxl 1.jxl 2.jpg
JPEG XL decoder v0.10.2 e148959 [AVX2,SSE2]
3677 x 5549,  282.590 MP/s [282.59, 282.59], , 1 reps, 16 threads.
```
The return code is 0, but nothing is produced.

```
$ djxl 1.jxl 2.png
JPEG XL decoder v0.10.2 e148959 [AVX2,SSE2]
Decoded to pixels.
3677 x 5549,  280.952 MP/s [280.95, 280.95], , 1 reps, 16 threads.
```
The content of 2.png is the same as the original JPEG.

**Environment**
 - OS: tested on Debian and Windows
 - Compiler version: all download from GitHub releases, so I guess clang
 - CPU type: x86_64
 - cjxl/djxl version string: see command line output above

mo271:
Thanks for the report, I can reproduce it at current head.

PR #3473, opened 2024-04-11
Return error in case of truncated streams for jpeg reconstruction.

szabadka:
If jpeg bytes were requested, we returned success even if the stream was truncated, without filling in the jpeg bytes. This can cause djxl to return successfully without writing any output file.


szabadka:
> This fixes #3455, right?

No, it just makes djxl retry in decode-to-pixels mode.

PR #3479, opened 2024-04-12
add tsan to ci

mo271:


mo271:
With CI:full tsan is about as slow as msan, but faster than debug...

jonsneyers (approved):

PR #3503, opened 2024-04-18
Support DQT markers after SOF marker in jpegli decoder.

### LINKED ISSUE #3486: Quantization table with index 0 not found
**Describe the bug**
Getting the error mentioned in title with the following image
[image.zip](https://github.com/libjxl/libjxl/files/14985844/image.zip)

**To Reproduce**
I get the error while reading the image through Qt's QImageReader when linking it with jpegli. But I guess it should be reproducible with the utilities from libjxl repo just as well?

**Expected behavior**
Image is decoded

**Environment**
 - OS: Linux
 - Compiler version: GCC 12
 - CPU type: x86_64
 - cjxl/djxl version string: I don't have cjxl/djxl... But I used v0.10.2 tag to build jpegli.

szabadka:
Fix suggested by @DCNick3 

Fixes #3486


jonsneyers:
What happens if the DQT marker is missing? Before, that error was caught too early (it would also trigger if the DQT marker was just later than SOF), but isn't it now not caught at all?

szabadka:
> What happens if the DQT marker is missing? Before, that error was caught too early (it would also trigger if the DQT marker was just later than SOF), but isn't it now not caught at all?

It is caught in PrepareForScan(), which is called from jpegli_start_decompress(). Do you think it is a problem that it is not caught in jpegli_read_header() ?

jonsneyers:
It is fine as long as it is caught somewhere.

PR #3507, opened 2024-04-18
Use YUV420 chroma subsampling in jpegli_set_defaults().

szabadka:
The default remains YUV444 in the cjpegli binary.

### DIFF
--- a/tools/scripts/jpegli_tools_test.sh
+++ b/tools/scripts/jpegli_tools_test.sh
@@ -163,12 +163,12 @@ djpeg_test() {
   local outfn="$(mktemp -p "${tmpdir}").pnm"
   LD_LIBRARY_PATH="${build_dir}/lib/jpegli:${LD_LIBRARY_PATH:-}" \
     djpeg -outfile "${outfn}" -colors 128 -onepass -dither fs "${jpgfn}"
-  verify_ssimulacra2 "${infn}" "${outfn}" 30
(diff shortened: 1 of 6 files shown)


tools/scripts/jpegli_tools_test.sh
```diff
@@ -185,28 +185,6 @@ djpeg_test() {
...
-  for scale in 1/4 3/8 1/2 5/8 9/8; do
-    local scalepct="$(python3 -c "print(100.0*${scale})")%"
-    local geometry=96x128+0+0
-    local outfn="$(mktemp -p "${tmpdir}").pnm"
-    LD_LIBRARY_PATH="${build_dir}/lib/jpegli:${LD_LIBRARY_PATH:-}" \
-      djpeg -outfile "${outfn}" -scale "${scale}" -crop "${geometry}" "${jpgfn}"
-    local outfn2="$(mktemp -p "${tmpdir}").pnm"
-    convert "${infn}" -scale "${scalepct}" -crop "${geometry}" "${outfn2}"
-    verify_ssimulacra2 "${outfn2}" "${outfn}" 80
-  done
```
mo271, line 209: was this dropped by accident?

szabadka: I was running the test locally and djpeg failed because apparently these flags were removed, they don't appear in --help or in the man page any more.

PR #3528, opened 2024-04-23
Fix compile error caused by moved rect.h

tirr-c:
### Description
`base/rect.h` was moved to `jxl/base/rect.h`, but the include was not updated in `tools/comparison_viewer/image_loading.cc`.


eustas (approved):

eustas:
Thanks.

PR #3531, opened 2024-04-25
Add an option for disabling perceptual optimizations.

veluca93:
There still likely is something off with the implementation, but this lays the groundwork.


lib/jxl/enc_heuristics.cc
```diff
@@ -201,18 +202,18 @@ Status FindBestDequantMatrices(const CompressParams& cparams,
   // TODO(veluca): heuristics for in-bitstream quant tables.
   *dequant_matrices = DequantMatrices();
   if (cparams.max_error_mode) {
```
szabadka, line 204: || cparamsdisable_percepeptual_optimizations ?

szabadka (approved):

PR #3552, opened 2024-05-02
add jhgm box and basic gain map support

mo271:


lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+#include "lib/jxl/memory_manager_internal.h"
+
+namespace {
+
+template <size_t N>
+void* FixedSizeMemoryManagerAlloc(void* opaque, size_t capacity);
+void FixedSizeMemoryManagerFree(void* opaque, void* pointer) {}
+
+template <size_t N>
+class FixedSizeMemoryManager {
```
veluca93, line 27: This will only handle one allocation, right? Perhaps we should remember that and return nullptr if trying to allocate again?

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+ public:
+  FixedSizeMemoryManager() = default;
+
+  JxlMemoryManager* memory_manager() { return &manager_; }
+
+  friend void* FixedSizeMemoryManagerAlloc<N>(void* opaque, size_t capacity);
```
veluca93, line 33: why is this not a static method?

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+                                 size_t* bundle_size) {
+  if (map_bundle == nullptr) return 0;
+
+  FixedSizeMemoryManager<sizeof(jxl::ColorEncoding)> memory_manager;
+  jxl::ColorEncoding internal_color_encoding;
+  jxl::BitWriter color_encoding_writer(memory_manager.memory_manager());
+  if (map_bundle->has_color_encoding) {
+    JXL_RETURN_IF_ERROR(
+        internal_color_encoding.FromExternal(map_bundle->color_encoding));
+    if (!jxl::Bundle::Write(internal_color_encoding, &color_encoding_writer,
```
veluca93, line 65: IIRC you mentioned that the function to get the size of a Bundle was broken?
If so, we should probably clean it up somehow.
If not, we should use it here ;-)

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+    cursor += 1;
+  }
+
+  uint16_t metadata_size_le = JXL_BSWAP16(map_bundle->gain_map_metadata_size);
+  if (cursor + 2 <= output_buffer_size) {
+    memcpy(output_buffer + cursor, &metadata_size_le, 2);
+    cursor += 2;
+  }
+
+  if (cursor + map_bundle->gain_map_metadata_size <= output_buffer_size) {
```
veluca93, line 118: Not fundamental in the encoder, but ideally those operations should check for overflow.

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+
+  color_encoding_writer.ZeroPadToByte();
+
+  size_t cursor = 0;
+  if (cursor + 1 <= output_buffer_size) {
+    memcpy(output_buffer + cursor, &jhgm_version, 1);
+    cursor += 1;
+  }
+
+  uint16_t metadata_size_le = JXL_BSWAP16(map_bundle->gain_map_metadata_size);
```
veluca93, line 112: Why not WriteU16LE & friends?

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+    return JXL_FALSE;
+  }
+
+  size_t cursor = 0;
+  // Read the version byte
+  map_bundle->jhgm_version = input_buffer[cursor];
+  cursor += 1;
+
+  // Read and swap gain_map_metadata_size
+  uint16_t gain_map_metadata_size_le;
```
veluca93, line 174: Same here with ReadLE16.

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+    memcpy(output_buffer + cursor, map_bundle->gain_map,
+           map_bundle->gain_map_size);
+    cursor += map_bundle->gain_map_size;
+  }
+
+  if (bytes_written != nullptr) *bytes_written = cursor;
+  return cursor == output_buffer_size ? JXL_TRUE : JXL_FALSE;
+}
+
+JXL_BOOL JxlGainMapReadBundle(JxlGainMapBundle* map_bundle,
```
veluca93, line 159: This function doesn't seem to check that reads are in-bounds, and it definitely should.

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+  uint32_t compressed_icc_size_le;
+  memcpy(&compressed_icc_size_le, input_buffer + cursor, 4);
+  uint32_t compressed_icc_size = JXL_BSWAP32(compressed_icc_size_le);
+  cursor += 4;
+  map_bundle->alt_icc_size = compressed_icc_size;
+  map_bundle->alt_icc = input_buffer + cursor;
+  cursor += compressed_icc_size;
+  // Remaining bytes are gain map
+  map_bundle->gain_map_size = input_buffer_size - cursor;
+  if (input_buffer_size < cursor + map_bundle->gain_map_size) return JXL_FALSE;
```
veluca93, line 211: I'm pretty sure this condition does nothing ;-)

lib/include/jxl/gain_map.h
```diff
@@ -0,0 +1,110 @@
...
+#endif
+
+/** Gain map bundle */
+typedef struct {
+  /** version number */
+  uint8_t jhgm_version;
+  /** size of the gain map metadata */
+  uint16_t gain_map_metadata_size;
+  /** pointer to binary blob of gain map metadata (ISO 21496-1) */
+  const uint8_t* gain_map_metadata;
```
veluca93, line 31: Something somewhere should clarify that these fields are not owned.

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,216 @@
...
+    return nullptr;
+  }
+  return manager->memory_;
+}
+
+}  // namespace
+
+JXL_BOOL JxlGainMapGetBundleSize(const JxlGainMapBundle* map_bundle,
+                                 size_t* bundle_size) {
+  if (map_bundle == nullptr) return 0;
```
veluca93, line 57: why 0?

lib/include/jxl/gain_map.h
```diff
@@ -0,0 +1,110 @@
...
+ * binary buffer. This function accounts for all the necessary space to
+ * serialize fields such as gain map metadata, color encoding, compressed ICC
+ * profile data, and the gain map itself.
+ *
+ * @param[in] map_bundle Pointer to the JxlGainMapBundle containing all
+ * necessary data to compute the size.
+ * @param[out] bundle_size The size in bytes required to serialize the bundle.
+ * @return Whether setting the size was successful.
+ */
+JXL_EXPORT JXL_BOOL JxlGainMapGetBundleSize(const JxlGainMapBundle* map_bundle,
```
veluca93, line 57: Do we expect this function to be able to fail in the future?

lib/extras/gain_map.cc
mo271: @sboukortt

lib/extras/gain_map.cc
mo271: @sboukortt

lib/extras/gain_map.cc
mo271: I think we only have the `MaxBits` function, which I think works, but I'm not sure that in this case will returns the actual number of Bits needed or an non-sharp upper bound. Should we use it nonetheless and then perhaps change the documentation and perhaps even the name to `JxlGainMapMaxBundleSize`?

veluca93: Very intuitively, the visitor to get the size appears to be called "CanEncode"...

lib/extras/gain_map.cc
mo271: I added the use of `SafeAdd`, which we have in common.h. 
Is this what you had in mind?

lib/extras/gain_map.cc
mo271: now returns `JXL_FALSE` instead.

lib/extras/gain_map.cc
mo271: Added ifs before every read

lib/extras/gain_map.cc
sboukortt: It’s meant to handle “reallocations”, for when `PaddedBytes` wants to increase its size. This is made possible by the fact that freeing is a no-op, and by the change from `memcpy` to `memmove` in `padded_bytes.h` (since the source and destination will be identical).

veluca93: I guess that's fine, I am not sure I like it too much but I don't see better alternatives...

lib/extras/gain_map.cc
mo271: now also actually returning an error

lib/extras/gain_map.cc
mo271: Now using this. Much better don't need the memory manager anymore in this function now..

lib/extras/gain_map.cc
mo271: now with the SafeAdd there is a possibilty of failure

lib/extras/gain_map.cc
mo271: done

lib/extras/gain_map.cc
mo271: done

lib/include/jxl/gain_map.h
mo271: It can already fail now, no?

lib/include/jxl/gain_map.h
mo271: added long docstring like we before struct like we have in cms_interface.h

lib/jxl/test_utils.cc
```diff
@@ -61,6 +61,76 @@ std::string GetTestDataPath(const std::string& filename) {
...
+      "\0\0\a\223\0\0\375\220\377\377\373\241\377\377\375\242\0\0\3\334\0\0\300"
+      "nXYZ \0\0\0\0\0\0o\240\0\08\365\0\0\3\220XYZ "
+      "\0\0\0\0\0\0$\237\0\0\17\204\0\0\266\304XYZ "
+      "\0\0\0\0\0\0b\227\0\0\267\207\0\0\30\331para\0\0\0\0\0\3\0\0\0\1\0\0\0\1"
+      "\0\0\0\0\0\0\0\1\0\0\0\0\0\0chrm\0\0\0\0\0\3\0\0\0\0\243\327\0\0T|"
+      "\0\0L\315\0\0\231\232\0\0&"
+      "g\0\0\17\\mluc\0\0\0\0\0\0\0\1\0\0\0\fenUS\0\0\0\f\0\0\0\34\0G\0o\0o\0g"
+      "\0l\0emluc\0\0\0\0\0\0\0\1\0\0\0\fenUS\0\0\0\26\0\0\0\34\0I\0m\0a\0g\0e"
+      "\0 \0c\0o\0d\0e\0c\0\0");
+  size_t profile_size = 896;
```
jonsneyers, line 97: Can we make this a file in the testdata repo instead of code?

lib/extras/enc/jxl.cc
```diff
@@ -306,10 +307,9 @@ bool EncodeImageJXL(const JXLCompressParams& params, const PackedPixelFile& ppf,
         const char* type;
         const std::vector<uint8_t>& bytes;
       } boxes[] = {
-          {"Exif", exif_with_offset},
-          {"xml ", ppf.metadata.xmp},
-          {"jumb", ppf.metadata.jumbf},
-          {"xml ", ppf.metadata.iptc},
+          {"Exif", exif_with_offset},   {"xml ", ppf.metadata.xmp},
+          {"jumb", ppf.metadata.jumbf}, {"xml ", ppf.metadata.iptc},
```
jonsneyers, line 311: Not really related to this pull request, but we should probably get rid of `ppf.metadata.iptc` — IPTC is a legacy metadata format that we don't really support; semantically it is a subset of XMP (and Exif, iirc), but syntactically it is different. Applications are supposed to use Exif or XMP instead. Just putting an IPTC blob in an `xml `box is wrong though. We should just remove it from our code.

lib/jxl/test_utils.cc
mo271: sure, but let's do that later when adding the ICC de/en-coding functions

lib/extras/gain_map.cc
jonsneyers: It feels a bit weird that this is defined in `gain_map.cc` and that this kind of functionality is only needed here.

lib/include/jxl/gain_map.h
```diff
@@ -0,0 +1,110 @@
...
+  /** pointer to binary blob of gain map metadata (ISO 21496-1) */
+  const uint8_t* gain_map_metadata;
+  /** indicate if it has a color encoding*/
+  bool has_color_encoding;
+  /** uncompressed color encoding */
+  JxlColorEncoding color_encoding;
+  /** size of the alt_icc profile  (compressed size) */
+  uint32_t alt_icc_size;
+  /** pointer to the compressed icc profile */
+  const uint8_t* alt_icc;
```
jonsneyers, line 57: This means applications have to implement their own ICC profile compression/decompression methods, since this is not something that is currently exposed in the libjxl API. That's not a good idea imo.

Either we should have an _uncompressed_ ICC profile here (and `JxlGainMapWriteBundle` compresses it before writing it, `JxlGainMapReadBundle` decompresses it after writing it), or we should expose functions to compress/decompress ICC profiles. I prefer the first option.

mo271: We are planning to expose the functions to compress/decompress the ICC profiles

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,269 @@
...
+
+  color_encoding_writer.ZeroPadToByte();
+
+  uint64_t cursor = 0;
+  uint64_t next_cursor = 0;
+
+  if (!jxl::SafeAdd(cursor, 1, next_cursor) ||
+      next_cursor > output_buffer_size) {
+    return JXL_FALSE;
+  }
```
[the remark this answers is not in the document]
sboukortt: Something like this?

lib/extras/gain_map.cc
sboukortt: Let’s have it here for now and factor it out if we eventually need it elsewhere?

jonsneyers (approved):

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,225 @@
...
+      return JXL_FALSE;
+    }
+  }
+
+  color_encoding_writer.ZeroPadToByte();
+
+  uint64_t cursor = 0;
+  uint64_t next_cursor = 0;
+
+#define ADVANCE_CURSOR(n)                        \
```
veluca93, line 103: I don't like the naming too much (it seems to imply that it advances the cursor by `n`, which it doesn't), but it's not a super strong opinion.

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,225 @@
...
+                              size_t* bytes_read) {
+  if (map_bundle == nullptr || input_buffer == nullptr ||
+      input_buffer_size == 0) {
+    return JXL_FALSE;
+  }
+
+  uint64_t cursor = 0;
+  uint64_t next_cursor = 0;
+
+#define ADVANCE_CURSOR(n)                        \
```
veluca93, line 162: same :-)

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,225 @@
...
+        next_cursor > input_buffer_size) {       \
+      return JXL_FALSE;                          \
+    }                                            \
+  } while (false)
+
+  // Read the version byte
+  ADVANCE_CURSOR(1);
+  map_bundle->jhgm_version = input_buffer[cursor];
+
+  // Read and swap gain_map_metadata_size
```
veluca93, line 175: Doesn't "swap" anymore ;-)

lib/extras/gain_map.cc
```diff
@@ -0,0 +1,225 @@
...
+  ADVANCE_CURSOR(4);
+  uint32_t compressed_icc_size = LoadBE32(input_buffer + cursor);
+
+  ADVANCE_CURSOR(compressed_icc_size);
+  map_bundle->alt_icc_size = compressed_icc_size;
+  map_bundle->alt_icc = input_buffer + cursor;
+
+  // Calculate remaining bytes for gain map
+  cursor = next_cursor;
+  map_bundle->gain_map_size = input_buffer_size - cursor;
```
veluca93, line 213: Maybe add a comment that this is guaranteed not to underflow, and why

veluca93 (approved):

lib/extras/gain_map.cc
mo271: done

lib/extras/gain_map.cc
mo271: how about "SAFE_CURSOR_UPDATE"?

lib/extras/gain_map.cc
mo271: same as above

lib/extras/gain_map.cc
mo271: done

veluca93 (approved):

PR #3566, opened 2024-05-08
fix arithmetic issue in dot detection

jonsneyers:
The bound was too tight, I encountered a case where it produced NaN output with that number equal to 1.13687e-12.

Haven't tested if this new bound is safe but it should at least be safer than the current one.

### DIFF
--- a/lib/jxl/enc_linalg.cc
+++ b/lib/jxl/enc_linalg.cc
@@ -17,7 +17,7 @@ void ConvertToDiagonal(const Matrix2x2& A, Vector2& diag, Matrix2x2& U) {
   JXL_ASSERT(std::abs(A[0][1] - A[1][0]) < 1e-15);
 #endif
 
-  if (std::abs(A[0][1]) < 1e-15) {
+  if (std::abs(A[0][1]) < 1e-10) {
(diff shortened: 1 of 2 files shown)


veluca93:
Can you add some tests? I.e. check a million or so random log-distributed values, and see if they trigger the issue.

veluca93 (approved):

jonsneyers:
I tried testing more random inputs but couldn't get one that triggers NaN even when trying lots of inputs.

The error bound that we had in the test was too tight though — I could find examples that get close to the new error bound.

The problematic behavior seems to be caused by taking the square root of something that can get slightly negative — but only for particular combinations of the two input values, which cannot be found easily by just using random values for both.

The new bailout condition specifically tests for `d < 0` too now.

I also added the problem input as a specific test case, and bumped up the amount of random testing to an amount that would find a counterexample for the previous error bound, but I kept it low enough so the test doesn't take much time; I tested with 1 billion inputs and couldn't find a counterexample for the new bound.

veluca93:
The test should use log-distributed values to have any chance of finding issues, I think :-)

PR #3581, opened 2024-05-14
support cmyk input for djpegli

### LINKED ISSUE #3474: djpegli fails to decode CMYK image
**Describe the bug**
`djpegli` fails to decode a CMYK JPG to PNG.

**To Reproduce**
1. Obtain a CMYK JPG, for example [this one](https://i.postimg.cc/ZTB4nKrW/crash.jpg) (from [this issue](https://gitlab.com/wg1/jpeg-xl/-/issues/122)).
2. `./djpegli image.jpg output.png`

**Actual behavior**
```
Read 2249602 compressed bytes.
jpegli decoding failed
```

**Expected behavior**
A PNG is produced.

**Environment**
* OS: Linux
* CPU type: x86_64
* [Binary used](https://artifacts.lucaversari.it/libjxl/libjxl/latest/jxl-linux-x86_64-static.zip) (`c04b3ba`)

mo271:
fixes #3474

### DIFF
--- a/lib/jpegli/color_transform.cc
+++ b/lib/jpegli/color_transform.cc
@@ -169,6 +169,33 @@ void CMYKToYCCK(float* row[kMaxComponents], size_t xsize) {
   RGBToYCbCr(row, xsize);
 }
 
+void CMYKToRGB(float* row[kMaxComponents], size_t xsize) {
+  const HWY_CAPPED(float, 8) df;
+  float* JXL_RESTRICT row_c = row[0];
+  float* JXL_RESTRICT row_m = row[1];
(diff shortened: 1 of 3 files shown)


lib/jpegli/color_transform.cc
```diff
@@ -169,6 +169,33 @@ void CMYKToYCCK(float* row[kMaxComponents], size_t xsize) {
   RGBToYCbCr(row, xsize);
 }
 
+void CMYKToRGB(float* row[kMaxComponents], size_t xsize) {
```
szabadka, line 172: This conversion should be handled on the application side, not in the library.

PR #3590, opened 2024-05-17
Updated CMake build to compile in C++17 mode if available

johnplatts:
Updated CMakeLists.txt to build libjxl in C++17 mode if supported by the C++ compiler.

For optimal dynamic dispatch with Google Highway, Clang 5 or later, GCC 7 or later, or MSVC 19.16 or later is needed, and all of these compilers have support for the C++17 standard.

Also added workarounds to suppress any -Wdeprecated-declarations warning that might be emitted by std::stable_sort with GCC or Clang in C++17 mode due to the libstdc++ implementation of std::stable_sort using std::get_temporary_buffer, which has been deprecated in C++17.


eustas (approved):
I'll also file an issue to get rid of stable_sort at some moment.

Thanks.

PR #3592, opened 2024-05-17
Fix MSAN findings

eustas:


eustas:
@szabadka friendly ping

PR #3613, opened 2024-05-29
Speedup debug tests

eustas:


eustas:
Now they run on CI ~15min instead of ~2h

mo271 (approved):

PR #3614, opened 2024-05-29
Speedup MSAN/TSAN tests; UB is checked on ASAN build

eustas:


eustas:
That shaves off ~45m of ~1h45 run

mo271 (approved):

PR #3622, opened 2024-06-02
improve lossless e2/e3 for > 10-bit images and make -E 1 do something when doing -e 2

jonsneyers:
The cutoffs used in the predefined trees for e2 (and the fallback e1 path) and e3 lossless encoding were designed with 8-bit images (possibly up to 10-bit) in mind. For higher bit depth images, compression improves when scaling up the cutoff values according to the bit depth.

Also: when explicitly enabling the previous channel extra properties (as set by the `-E` option in cjxl), we can use a previous channel  gradient difference property as context (based on the decoded value of the previous channel at the current pixel position), instead of the usual MA tree property we use in e2. This adds some encode/decode slowdown and is not necessarily better, but it can improve compression. This can be useful in particular for multi-spectral images with many correlated channels, like this example image with 21 channels with 16-bit data:

Before:
```
radiance-21.jxl
Encoding       kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
-----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:2         19902 396731430  159.4682655  16.820  31.253          nan 100.00000000  99.99   0.00000000  0.000000000000 159.468      0
jxl:d0:2:E1      19902 396731430  159.4682655  16.815  31.294          nan 100.00000000  99.99   0.00000000  0.000000000000 159.468      0
jxl:d0:3         19902 436822417  175.5830466   9.447  13.736          nan 100.00000000  99.99   0.00000000  0.000000000000 175.583      0
Aggregate:       19902 409668631  164.6684410  13.876  23.772   0.00000000 100.00000000  99.99   0.00000000  0.000000000000 164.668      0
```

After:
```
radiance-21.jxl
Encoding       kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
-----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:2         19902 387149978  155.6169510  11.994  14.058          nan 100.00000000  99.99   0.00000000  0.000000000000 155.617      0
jxl:d0:2:E1      19902 343799453  138.1919815  10.345  13.996          nan 100.00000000  99.99   0.00000000  0.000000000000 138.192      0
jxl:d0:3         19902 426401702  171.3943859   7.986  10.054          nan 100.00000000  99.99   0.00000000  0.000000000000 171.394      0
Aggregate:       19902 384297698  154.4704622   9.970  12.553   0.00000000 100.00000000  99.99   0.00000000  0.000000000000 154.470      0
```

Note: enc/dec speed does get worse, likely because before, the fixed MA tree ended up mostly collapsing when encoding 16-bit data with thresholds designed for 8-bit. The speeds are still quite reasonable for a 21-channel image.

### DIFF
--- a/lib/jxl/modular/encoding/encoding.h
+++ b/lib/jxl/modular/encoding/encoding.h
@@ -27,7 +27,7 @@ struct ANSCode;
 class BitReader;
 
 // Valid range of properties for using lookup tables instead of trees.
-constexpr int32_t kPropRangeFast = 512;
+constexpr int32_t kPropRangeFast = 512 << 6;
 
 struct GroupHeader : public Fields {
   GroupHeader();
--- a/lib/jxl/modular/encoding/encoding.h
+++ b/lib/jxl/modular/encoding/encoding.h
@@ -27,7 +27,7 @@ struct ANSCode;
 class BitReader;
 
 // Valid range of properties for using lookup tables instead of trees.
-constexpr int32_t kPropRangeFast = 512 << 6;
+constexpr int32_t kPropRangeFast = 512 << 4;
 
 struct GroupHeader : public Fields {
   GroupHeader();
@@ -59,15 +59,16 @@ FlatTree FilterTree(const Tree &global_tree,
                     size_t *num_props, bool *use_wp, bool *wp_only,
                     bool *gradient_only);
 
-template <typename T, bool HAS_MULTIPLIERS>
+template <typename T, bool HAS_OFFSETS, bool HAS_MULTIPLIERS>
 struct TreeLut {
   std::array<T, 2 * kPropRangeFast> context_lookup;
-  std::array<int8_t, 2 * kPropRangeFast> offsets;
+  std::array<int8_t, HAS_OFFSETS ? (2 * kPropRangeFast) : 0> offsets;
   std::array<int8_t, HAS_MULTIPLIERS ? (2 * kPropRangeFast) : 0> multipliers;
 };
 
-template <typename T, bool HAS_MULTIPLIERS>
-bool TreeToLookupTable(const FlatTree &tree, TreeLut<T, HAS_MULTIPLIERS> &lut) {
+template <typename T, bool HAS_OFFSETS, bool HAS_MULTIPLIERS>
+bool TreeToLookupTable(const FlatTree &tree,
+                       TreeLut<T, HAS_OFFSETS, HAS_MULTIPLIERS> &lut) {
   struct TreeRange {
     // Begin *excluded*, end *included*. This works best with > vs <= decision
     // nodes.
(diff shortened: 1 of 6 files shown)


veluca93:
What's the effect on enc/dec speed with "standard" images?

jonsneyers:
> What's the effect on enc/dec speed with "standard" images?

Up to 10-bit nothing at all changes, it's only for > 10-bit images that the cutoffs in the fixed tree change and that things become slower.

Looking into it, the issue was mostly that the lookup table specialization was not being used anymore when the cutoffs get scaled up, so the more generic branchy tree traversal ended up being used. Added a commit to bump up the limit for doing the lookup table thing, that seems to help to keep the speed closer to what it was:

```
radiance-21.jxl
Encoding       kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
-----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:2         19902 387149978  155.6169510  15.483  24.331          nan 100.00000000  99.99   0.00000000  0.000000000000 155.617      0
jxl:d0:2:E1      19902 343799453  138.1919815  10.005  13.660          nan 100.00000000  99.99   0.00000000  0.000000000000 138.192      0
jxl:d0:3         19902 426401702  171.3943859   9.354  14.675          nan 100.00000000  99.99   0.00000000  0.000000000000 171.394      0
Aggregate:       19902 384297698  154.4704622  11.316  16.959   0.00000000 100.00000000  99.99   0.00000000  0.000000000000 154.470      0
```

lib/jxl/modular/encoding/encoding.h
```diff
@@ -27,7 +27,7 @@ struct ANSCode;
 class BitReader;
 
 // Valid range of properties for using lookup tables instead of trees.
-constexpr int32_t kPropRangeFast = 512;
+constexpr int32_t kPropRangeFast = 512 << 6;
```
veluca93, line 30: This causes the table to occupy at least 32kB of stack size, which is probably too much.

jonsneyers: Agreed, it's too much. The table size currently is `6*kPropRangeFast` bytes, so that's even 196kB. It still worked reasonably well on my machine (I benchmarked reasonable speeds) but I assume it is not a good idea to have a table that large.

I'll cap the shifting to << 4, and also specialize the lookup table for the case of zero offsets and also adjust the encoder size specialization.

The table has then size `2*kPropRangeFast` instead of  `6*kPropRangeFast` for both encode and decode — in encode we had uint16 context_lookup and uint8 offsets (no multipliers), in decode we had uint8 context_lookup and uint8 offsets and multipliers.

Still it will be 16kB then instead of 3kB because `kPropRangeFast` is 16x as large as before. I'll benchmark this and put some results in the main PR discussion.

veluca93: Sorry, I somehow missed this.

The stack size issue is not a question of speed but of portability - I think 16kB should be fine though.

veluca93: We could also allocate a buffer per-group instead... Unclear if that'd be a significant improvement.

veluca93 (approved):

lib/jxl/modular/encoding/encoding.h
jonsneyers: Another approach could be to detect that the cutoffs were shifted (they all have N lsb zero bits) and shift the property values before using the lookup table. Adds a variable shift but reduces memory — possibly we can even get away with a smaller table in the 8-bit case too.

But for now let's just keep it simple and do it like this, unless there's an issue with it on some platform or something.

Issue #3594, opened 2024-05-18
Bundled SSIMULACRA2 detects differences in losslessly converted JXL


kylesacher:
**Describe the bug**
Converting a PNG to a JXL with `cjxl -d 0`, and then comparing the original and converted images with `ssimulacra2` results in a score < 100. 

**To Reproduce**
1. Use `cjxl -d 0` to convert a PNG to a JXL (e.g. `cjxl house.png house.jxl -d 0`)
2. Compare the original and converted image with `ssimulacra2` (e.g. `ssimulacra2 house.png house.jxl`)
3. Observe that the SSIMULACRA2 score is less than 100
4. Convert back to PNG with `djxl` (e.g. `djxl house.jxl house2.png`)
5. Compare the original PNG and new PNG with `ssimulacra2` (e.g. `ssimulacra2 house.png house2.png`)
6. Observe that the SSIMULACRA2 score is 100

**Expected behavior**
The SSIMULACRA2 score between the original PNG and converted JXL should be 100.

**Screenshots**
![house](https://github.com/libjxl/libjxl/assets/28654287/4a4be28d-bf89-40ed-8117-250ea3e22b4a)
![house jxl](https://github.com/libjxl/libjxl/assets/28654287/440bffdc-dcc0-46ce-ab42-3ce46b8687b3)
![house2](https://github.com/libjxl/libjxl/assets/28654287/53291008-e247-44b7-aa66-dd191594db20)

**Environment**
 - OS: Windows
 - Compiler version: N/A
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.10.2 e148959 [AVX2,SSE2], JPEG XL decoder v0.10.2 e148959 [AVX2,SSE2]

**Additional context**
- Only steps 1-3 are necessary to reproduce the problem. Steps 4-6 show that the data is not actually lost, despite the <100 score. 
- The attached `house.png`, `house.jxl.png`, and `house2.png` are the files used in the examples I've given in the steps to reproduce.
- The exact score I got from SSIMULACRA2 was 96.67210224.
- `cjxl`, `djxl`, and `ssimulacra2` all came from `jxl-x64-windows-static` in release 0.10.2.

jonsneyers:
Thanks for noticing this problem!

Looks like this is caused by slight differences in the way the conversion from int to float is implemented in `dec_modular.cc` (which is what is used when decoding a jxl) and in `enc_external_image.cc` (which is what is used when decoding a png). 

[...]

PR #3626, opened 2024-06-03
don't do a funky conversion from uint8 to float

jonsneyers:
When loading 8-bit images to float, we used to do `* 257 * (1.0f / 65535)` instead of just `* (1.0f / 255)`, which in floating point arithmetic is somehow not quite the same thing. For example, for the 8-bit value 254, the first formula produces the value 0.996078431606292724609375, while the second produces the value 0.996078491210937500000000, a difference of 5.96046e-08. This was basically a hack to ensure uint8 images decoded to uint16 buffers would still map all values to the same float32 values.

However, it only works for uint8 and uint16. For e.g. 10-bit or 12-bit images, this kind of trick doesn't work: while 65535 = 255*257, i.e. the uint16 maxval is an integer multiple of the uint8 maxval, so uint8 values can be represented exactly as uint16 values, but this is not the case for uint10 or uint12 values: there are no exactly corresponding uint16 values in the first place, so mapping them to the same float32 value would require a larger amount of cheating.

I assume that the main reason to do this hack was to allow benchmark_xl to decode jxl to uint16 and still get perfect metric scores when doing lossless encoding of 8-bit images (which would not work if uint8 values map to slightly different floats than the corresponding uint16 values). But then when decoding the jxl to a float buffer and comparing it to a funky-converted 8-bit png, as is done when you do `ssimulacra2 orig.png orig.png.jxl`, you get slight differences, see https://github.com/libjxl/libjxl/issues/3594.

Also when running benchmark_xl on 10-bit or 12-bit images, lossless jxl does not look lossless since it decodes the jxl to uint16. Decoding jxl to float in benchmark_xl and removing the funky slightly-incorrect uint8-to-float conversion solves these problems.

Before, on a 12-bit image:
```
507.ppm
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:2         1019  2045874   16.0536252  35.766 147.727   0.00227312  95.17732340 110.55   0.00118253  0.018983878811  16.054      0
Aggregate:       1019  2045874   16.0536252  35.766 147.727   0.00227312  95.17732340 110.55   0.00118253  0.018983878811  16.054      0
```

After:
```
507.ppm
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:2         1019  2045874   16.0536252  38.642 152.748          nan 100.00000000  99.99   0.00000000  0.000000000000  16.054      0
Aggregate:       1019  2045874   16.0536252  38.642 152.748   0.00000000 100.00000000  99.99   0.00000000  0.000000000000  16.054      0
```

Before:
```
$ cjxl -d 0 bees.png bees.png.jxl; ssimulacra2 bees.png bees.png.jxl
98.56508389
```

After:
```
$ cjxl -d 0 bees.png bees.png.jxl; ssimulacra2 bees.png bees.png.jxl
100.00000000
```

### DIFF
--- a/lib/jxl/base/float.h
+++ b/lib/jxl/base/float.h
@@ -62,9 +62,7 @@ static Status JXL_INLINE LoadFloatRow(const uint8_t* src, size_t count,
 
     case JXL_TYPE_UINT8:
       for (size_t i = 0; i < count; ++i) {
-        // Integer multiply uint8 value before scaling so that the UINT8 value
-        // and the corresponding UINT16 value convert to the same float
-        callback(i, (src[stride * i] * 257) * scale);
+        callback(i, src[stride * i] * scale);
       }
       return true;
 
--- a/lib/jxl/blending_test.cc
+++ b/lib/jxl/blending_test.cc
@@ -24,7 +24,7 @@ TEST(BlendingTest, Crops) {
   const std::vector<uint8_t> compressed =
       jxl::test::ReadTestData("jxl/blending/cropped_traffic_light.jxl");
   extras::JXLDecompressParams dparams;
-  dparams.accepted_formats = {{3, JXL_TYPE_UINT16, JXL_LITTLE_ENDIAN, 0}};
+  dparams.accepted_formats = {{3, JXL_TYPE_UINT8, JXL_LITTLE_ENDIAN, 0}};
   extras::PackedPixelFile decoded;
   ASSERT_TRUE(DecodeImageJXL(compressed.data(), compressed.size(), dparams,
                              /*decoded_bytes=*/nullptr, &decoded));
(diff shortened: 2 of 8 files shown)


jonsneyers:
This is undoing some of https://github.com/libjxl/libjxl/pull/3171 so pinging @szabadka in case I missed something.

jonsneyers:
> Just a piece of trivia: the 257 factor trick is also how Matlab does im2uint8() and im2uint16() conversion... [In Octave as well, only used between uint8/uint16](https://sourceforge.net/p/octave/image/ci/default/tree/inst/imcast.m#l103).

To be clear: there's nothing wrong with doing uint8 to uint16 conversion by multiplying by 257, that is mathematically correct. The problem is that converting uint8 to float by multiplying by 1.0f/255 does not produce the exact same result as first converting uint8 to uint16 and then converting that to float (by multiplying by 1.0f/65535). The difference is small, but not zero, so any test that expects a zero difference will fail.

szabadka (approved):

PR #3642, opened 2024-06-12
check multiple highway headers

### LINKED ISSUE #3641: FindHWY.cmake no longer works with the latest version of highway
**Describe the bug**
The `FindHWY.cmake` file uses the `hwy/highway.h` header file to determine the version when `pkgconfig` is not available but this moved to the file `hwy/base.h` and now the version cannot be found anymore.

**To Reproduce**
Build this project somewhere where `pkgconfig` is not available and the latest (`1.2.0`) highway library is installed on the system.

cc @jan-wassenberg

mo271:
attempts to fix #3641


dlemstra:
Should this part also be changed:

```cmake
find_path(HWY_INCLUDE_DIR
  NAMES hwy/highway.h
  HINTS ${PC_HWY_INCLUDEDIR} ${PC_HWY_INCLUDE_DIRS}
)
```

dlemstra:
Will test this later today.

dlemstra:
You should also do this:

```cmake
if(NOT HWY_VERSION_MAJOR STREQUAL "" AND NOT HWY_VERSION_MINOR STREQUAL "" AND NOT HWY_VERSION_PATCH STREQUAL "")
  set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
endif()
```

Otherwise the value will be overwritten from the other header file. But when I use `base-foo.h` and and `highway-foo` it will then not set `HWY_VERSION` and continue without any errors. That can be fixed with:

```cmake
if (HWY_INCLUDE_DIR AND NOT HWY_VERSION)
  set(HWY_VERSION "")
```

mo271:
> You should also do this:
> 
> ```cmake
> if(NOT HWY_VERSION_MAJOR STREQUAL "" AND NOT HWY_VERSION_MINOR STREQUAL "" AND NOT HWY_VERSION_PATCH STREQUAL "")
>   set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
> endif()
> ```
Make sense. I guess we could instead also break out of the foreach after setting this the first time


> 
> Otherwise the value will be overwritten from the other header file. But when I use `base-foo.h` and and `highway-foo` it will then not set `HWY_VERSION` and continue without any errors. That can be fixed with:
> 
> ```cmake
> if (HWY_INCLUDE_DIR AND NOT HWY_VERSION)
>   set(HWY_VERSION "")
> ```

I added this.
Thanks @dlemstra !

dlemstra:
Maybe a bit OCD but `if (NOT` looks better 😁

dlemstra (approved):
I can confirm that this fixes the issue.

cmake/FindHWY.cmake
```diff
@@ -20,20 +20,25 @@ find_library(HWY_LIBRARY
...
 
-    string(REGEX MATCH "#define +HWY_PATCH +([0-9]+)" _sink "${HWY_VERSION_CONTENT}")
-    set(HWY_VERSION_PATCH "${CMAKE_MATCH_1}")
-
-    set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
-  endif ()
+      string(REGEX MATCH "#define +HWY_PATCH +([0-9]+)" _sink "${HWY_VERSION_CONTENT}")
+      set(HWY_VERSION_PATCH "${CMAKE_MATCH_1}")
+      if (NOT HWY_VERSION_MAJOR STREQUAL "" AND NOT HWY_VERSION_MINOR STREQUAL "" AND NOT HWY_VERSION_PATCH STREQUAL "")
+        set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
```
sboukortt, line 38: How about `break()`ing out of the loop then?

sboukortt (approved):

cmake/FindHWY.cmake
mo271: yeah, let's do that, as suggested in this comment as well: https://github.com/libjxl/libjxl/pull/3642#issuecomment-2162809624

cmake/FindHWY.cmake
```diff
@@ -19,21 +19,30 @@ find_library(HWY_LIBRARY
...
 
-    string(REGEX MATCH "#define +HWY_PATCH +([0-9]+)" _sink "${HWY_VERSION_CONTENT}")
-    set(HWY_VERSION_PATCH "${CMAKE_MATCH_1}")
-
-    set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
-  endif ()
+      string(REGEX MATCH "#define +HWY_PATCH +([0-9]+)" _sink "${HWY_VERSION_CONTENT}")
+      set(HWY_VERSION_PATCH "${CMAKE_MATCH_1}")
+      set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
+      break()
```
dlemstra, line 39: The `break()` should only be done when `HWY_VERSION` is not `..`.

dlemstra (changes requested):

cmake/FindHWY.cmake
mo271: I kept the if now, and only added the `break`

cmake/FindHWY.cmake
mo271: done

cmake/FindHWY.cmake
```diff
@@ -19,21 +19,31 @@ find_library(HWY_LIBRARY
...
-  endif ()
+      string(REGEX MATCH "#define +HWY_PATCH +([0-9]+)" _sink "${HWY_VERSION_CONTENT}")
+      set(HWY_VERSION_PATCH "${CMAKE_MATCH_1}")
+      if (NOT HWY_VERSION_MAJOR STREQUAL "" AND NOT HWY_VERSION_MINOR STREQUAL "" AND NOT HWY_VERSION_PATCH STREQUAL "")
+        set(HWY_VERSION "${HWY_VERSION_MAJOR}.${HWY_VERSION_MINOR}.${HWY_VERSION_PATCH}")
+        break()
+      endif()
+    endif ()
+  endforeach ()
+  if(NOT HWY_VERSION)
```
dlemstra, line 44: `if (` ?

dlemstra (approved):

sboukortt (approved):

PR #3648, opened 2024-06-17
Allow to ask for memory with 0-side padding.

veluca93:
In the render pipeline, we promise 16 floats before the pointer passed to each stage to be readable.

Because of implementation details of the pre-memory-manager image allocation, that was always the case before. Once we started using the memory manager, this is no longer the case. This PR introduces the possibility to explicitly ask for such padding and uses it in the render pipeline.

### DIFF
--- a/lib/jxl/memory_manager_internal.cc
+++ b/lib/jxl/memory_manager_internal.cc
@@ -90,9 +90,9 @@ size_t BytesPerRow(const size_t xsize, const size_t sizeof_t) {
 }
 
 StatusOr<AlignedMemory> AlignedMemory::Create(JxlMemoryManager* memory_manager,
-                                              size_t size) {
-  size_t allocation_size = size + memory_manager_internal::kAlias;
-  if (size > allocation_size) {
+                                              size_t size, size_t pre_padding) {
+  size_t allocation_size = size + pre_padding + memory_manager_internal::kAlias;
+  if (size > allocation_size || size + pre_padding > allocation_size) {
     return JXL_FAILURE("Requested allocation is too large");
   }
   JXL_CHECK(memory_manager);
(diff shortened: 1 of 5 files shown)


eustas:
Could be simpler if we agree that pre_padding is always (A) `kAlignment/2` or (B) `kAlignment` (and thus make `pre_padding` a bool parameter).
In that case overflow test could be left intact; `pre_padding_groups` will be always zero (or 1, if (B) and pre_padding requested)...
pre_padding requirement could be checked with static assert in render_pipeline.

lib/jxl/memory_manager_internal.cc
```diff
@@ -113,15 +114,22 @@ AlignedMemory::AlignedMemory(JxlMemoryManager* memory_manager, void* allocation)
...
+
+  size_t pre_padding_groups = pre_padding / memory_manager_internal::kAlignment;
+  size_t pre_padding_offset =
+      pre_padding_groups * memory_manager_internal::kAlignment;
+
+  size_t offset =
+      memory_manager_internal::kAlignment * group + pre_padding_offset;
 
   // Actual allocation.
   uintptr_t address = reinterpret_cast<uintptr_t>(allocation);
```
szabadka, line 126: Wouldn't just adding pre_padding to address here and doing everything else the same be enough? Why do we need to manipulate offset at all?

veluca93: Specifically adding pre_padding might cause the resulting address to not be aligned, which is not great...

This way, we get at most one extra kAlignment block of additional padding. This version of the code also works if pre_padding is larger than kAlias (just changing the check at line 132 would work otherwise).

szabadka (approved):

eustas (approved):

PR #3652, opened 2024-06-19
More constrained, but easier to read pre_padding implementation.

eustas:


eustas:
It has no advantages, but rather requires much less thinking to prove that result is correct for all combination of {size, padding, group, allocation_offset}.

sboukortt:
Maybe this is an illusion but I feel I might have a better grasp on the previous (current `main`) implementation.

PR #3656, opened 2024-06-19
Propagate error from RunOnPool DataFn

eustas:


lib/extras/hlg.cc
```diff
@@ -26,27 +26,27 @@ Status HlgOOTF(ImageBundle* ib, const float gamma, ThreadPool* pool) {
...
+      if (std::isfinite(ratio)) {
+        red *= ratio;
+        green *= ratio;
+        blue *= ratio;
+      }
+    }
+    return true;
+  };
+
+  JXL_RETURN_IF_ERROR(RunOnPool(pool, 0, ib->ysize(), ThreadPool::NoInit,
```
veluca93, line 48: Why not just `return RunOnPool(...`?

lib/extras/tone_mapping.cc
```diff
@@ -44,23 +44,25 @@ Status ToneMapFrame(const std::pair<float, float> display_nits,
...
+      V green = Load(df, row_g + x);
+      V blue = Load(df, row_b + x);
+      tone_mapper.ToneMap(&red, &green, &blue);
+      Store(red, df, row_r + x);
+      Store(green, df, row_g + x);
+      Store(blue, df, row_b + x);
+    }
+    return true;
+  };
+  JXL_RETURN_IF_ERROR(RunOnPool(pool, 0, ib->ysize(), ThreadPool::NoInit,
```
veluca93, line 63: ditto

lib/extras/hlg.cc
eustas: It is rare in JXL code when we return status that we just got. So I did that for consistency.

veluca93 (approved):

PR #3657, opened 2024-06-20
Add a new DEC_BOX_COMPLETE event to handle complete boxes without having to wait for *either* DEC_SUCCESS or DEC_BOX

sboukortt:


examples/decode_exif_metadata.cc
```diff
@@ -72,7 +73,7 @@ bool DecodeJpegXlExif(const uint8_t* jxl, size_t size,
       exif->resize(exif->size() + kChunkSize);
       JxlDecoderSetBoxBuffer(dec.get(), exif->data() + output_pos,
                              exif->size() - output_pos);
-    } else if (status == JXL_DEC_SUCCESS) {
+    } else if (status == JXL_DEC_BOX_COMPLETE) {
```
veluca93, line 76: Doesn't this also need a change to the Subscribe call?

veluca93 (approved):

lib/include/jxl/decode.h
```diff
@@ -290,12 +290,12 @@ typedef enum {
...
    * The buffer set with @ref JxlDecoderSetBoxBuffer must be set again for each
    * next box to be obtained, or can be left unset to skip outputting this box.
-   * The output buffer contains the full box data when the next ::JXL_DEC_BOX
-   * event or ::JXL_DEC_SUCCESS occurs. ::JXL_DEC_BOX occurs for all
-   * boxes, including non-metadata boxes such as the signature box or codestream
-   * boxes. To check whether the box is a metadata type for respectively EXIF,
-   * XMP or JUMBF, use @ref JxlDecoderGetBoxType and check for types "Exif",
-   * "xml " and "jumb" respectively.
+   * The output buffer contains the full box data when the
+   * ::JXL_DEC_BOX_COMPLETE event occurs. ::JXL_DEC_BOX occurs for all boxes,
```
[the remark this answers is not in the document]
sboukortt: Strictly speaking, yes, as well as when any other event than `JXL_DEC_BOX_NEED_MORE_OUTPUT` (or an error) occurs, but is it really worth pointing out?

examples/decode_exif_metadata.cc
sboukortt: It does, and there is.

sboukortt:
It seems this actually breaks a bunch of tests so might need some more work after all.

lib/include/jxl/decode.h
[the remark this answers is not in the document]
veluca93: More precisely, this seems to imply that you need to subscribe to this event to know when the box is finished.

sboukortt:
Fixed. I was afraid of some deeper issue but it turned out to be a relatively trivial change.

lib/include/jxl/decode.h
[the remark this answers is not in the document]
sboukortt: Thanks, that’s a fair point. How is this?

lib/jxl/decode.cc
```diff
@@ -1662,7 +1667,7 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
             metadata.resize(metadata.size() * 2);
           } else if (box_result == JXL_DEC_NEED_MORE_INPUT) {
             break;  // box stage handling below will handle this instead
-          } else if (box_result == JXL_DEC_SUCCESS) {
+          } else if (box_result == JXL_DEC_BOX_COMPLETE) {
```
veluca93, line 1670: shouldn't this be an or?

sboukortt: As far as I can tell, it shouldn’t need to be (this specific `box_result` is the unchanged output of `JxlBoxContentDecoder::Process`, which I have changed to only ever return `BOX_COMPLETE`). But do you think it would make things clearer?

Should I revert that function to return `JXL_DEC_SUCCESS` again, and change the code above to turn it into `BOX_COMPLETE` if subscribed, rather than the other way around?

veluca93 (dismissed):

veluca93 (approved):

PR #3658, opened 2024-06-20
Try to fix enc_fast_lossless build (#7223)

eustas:
Drive-by: add standalone enc_fast_lossless target


CMakeLists.txt
```diff
@@ -204,7 +206,14 @@ if (NOT JPEGXL_ENABLE_AVX512_ZEN4)
...
 endif()
 
-
+check_c_source_compiles("#include <stdio.h>
+int main() {
+  __builtin_cpu_init();
+  printf(\"%d\", __builtin_cpu_supports(\"avx2\"));
+}" JXL_HAVE_BUILTIN_CPU_SUPPORTS)
+if(JXL_HAVE_BUILTIN_CPU_SUPPORTS)
+  #add_definitions(-DJXL_HAVE_BUILTIN_CPU_SUPPORTS)
```
veluca93, line 215: Intended?

lib/jxl/enc_fast_lossless.cc
```diff
@@ -19,18 +19,24 @@
 
 #if !FJXL_STANDALONE
 #include "lib/jxl/encode_internal.h"
-#endif
+#endif  // !FJXL_STANDALONE
+
+#if defined(JXL_HAVE_BUILTIN_CPU_SUPPORTS)
+#define FJXL_BUILTIN_CPU_SUPPORTS(F) __builtin_cpu_supports(F)
+#else  // JXL_HAVE_BUILTIN_CPU_SUPPORTS
+#define FJXL_BUILTIN_CPU_SUPPORTS(F) 1
```
veluca93, line 27: I am not sure I like this ;)

lib/jxl/enc_fast_lossless.cc
```diff
@@ -788,7 +800,7 @@ size_t JxlFastLosslessWriteOutput(JxlFastLosslessFrameState* frame,
                                          unsigned char*, uint64_t&) = nullptr;
 
 #if FJXL_ENABLE_AVX512
-  if (__builtin_cpu_supports("avx512vbmi2")) {
+  if (FJXL_BUILTIN_CPU_SUPPORTS(avx512vbmi2)) {
```
veluca93, line 803: Seems inconsistent with below

CMakeLists.txt
eustas: Ooops, was debugging both options. One sec.

lib/jxl/enc_fast_lossless.cc
eustas: Fixed

lib/jxl/enc_fast_lossless.cc
eustas: Unfortunately we have to avoid `__builtin_cpu_supports` if compiler-rt library does not provide `___cpu_model`.

We can borrow some code from HWY targets.cc... WDYT?

veluca93: That seems a lot better :-)

Also, I would probably flip the logic here - i.e. have a JXL_MISSING_BUILTIN_CPU_SUPPORTS macro, so that if you do nothing you get the standard builtin.

eustas: SGTM. Will do soon.

eustas: Done. Waiting for CI.

eustas:
Yup, why not.

veluca93 (dismissed):

eustas:
Fixed build, please re-stamp

veluca93 (approved):

PR #3660, opened 2024-06-24
allow passing custom primaries

jonsneyers:
When trying to pass custom primaries in a color space string, the current code always produces an error even if the primaries parsed correctly. This fixes that.


sboukortt:
Thanks for catching this. It seems that it has been that way for a while, more specifically since #495 (prior to that, the code returned before that `JXL_FAILURE` if setting the primaries succeeded).

sboukortt (approved):

Issue #3659, opened 2024-06-24
old encoded image jxl cannot decode by new version


doanquocnam:
**Describe the bug**
I have along time to use this image format. The version I started using for my application is v0.6.x.  So I had a problem, what I cant  resolve. Please help me to fix it.

A received an report that there was an image jxl cannot be decoded, so that I tried to decode by chrome old version (supported jxl), firefox trunk, ... so I put  to the sample decode_oneshot, it failed to decompress this file (what I upload for this report),
I tried all version 0.7.x, 0.8.x, ... 0.10.x .. and the main git version but the image wasnt able to decode.

I tried the version 0.6.x and the image could be decoded. So that please fix for this case. Thanks!
![image](https://github.com/libjxl/libjxl/assets/41658839/fd187718-821e-4f94-bc6b-69d6ff3a8f4d)

![c56f50ec47dfe481bdce jxl](https://github.com/libjxl/libjxl/assets/41658839/2e7d4168-09ad-4218-aa02-47e5f11d9295)

**Environment**
 - OS: [e.g. Windows, Android, Linux, iOS]
 - Compiler version: [e.g. clang 11.0.1]
 - CPU type: [e.g. x86_64, ARM64, ARM32]
 - cjxl/djxl version string: [e.g. cjxl [v0.3.7 | SIMD supported: SSE4,Scalar]]

mo271:
Bisecting suggests that this started happening at https://github.com/libjxl/libjxl/pull/1046. Any thoughts, @jonsneyers?

mo271:
Thanks for the report, @doanquocnam!

jonsneyers:
Could you share the jxl file that no longer decodes?

mo271:
@jonsneyers it is linked above

doanquocnam:
> Could you share the jxl file that no longer decodes?

Didnt you see that link of jxl image what I've uploaded.?
![image](https://github.com/libjxl/libjxl/assets/41658839/f78f39af-4baf-4d5b-bcbd-790c0af5aa14)

You can download it here: https://b-f67-zpg-r.zdn.vn/jxl/7220520585340013593/c56f50ec47dfe481bdce.jxl

jonsneyers:
Oh, interesting, on my browser (Thorium) I didn't see a broken image icon and a filename but an empty area the size of that image :)
(so it doesn't decode but it does know the image dimensions, it just doesn't paint any pixels)

Thanks for finding this regression!

PR #3662, opened 2024-06-24
fix bug in huffman-rle-gradient (e1) specialized decode path

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/3659

This was a regression introduced in https://github.com/libjxl/libjxl/pull/1046 when making a specialized decode path for RLE-only Huffman decoding as in bitstreams produced by lossless effort 1 (fast lossless).

The bug was that at the end of a decoded channel, there could still be a run going that goes into the next channel. The code assumed that there would be no runs crossing channel boundaries (i.e. run == 0 at the end of every channel), which is a valid assumption for bitstreams produced by fast_lossless, but it is not a valid assumption in general.

The cases where libjxl happens to produce RLE-only Huffman bitstreams outside the fast_lossless code path are pretty rare (otherwise we would have noticed this regression way earlier), but apparently it does happen, see https://github.com/libjxl/libjxl/issues/3659 for an example. Outside the fast_lossless encode path, there is no reason why runs would respect channel boundaries.

This fixes that bug and makes runs across channel boundaries work again, while preserving the specialized decode path (a simpler fix would be to just use the generic decode path).

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -146,7 +146,8 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
                                  const weighted::Header &wp_header,
                                  pixel_type chan, size_t group_id,
                                  TreeLut<uint8_t, false, false> &tree_lut,
-                                 Image *image) {
+                                 Image *image, uint32_t &fl_run,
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

mo271:
Should we perhaps mention this in the changelog (under "Fixed")

PR #3665, opened 2024-06-25
Parse cLLi chunk

ledoge:
### Description

Added code for parsing the cLLi chunk from HDR png files, so that MaxCLL metadata can be preserved when converting to jxl using `cjxl`. Previously, the value had to be specified manually with `--intensity_target`.

I tested the code as follows:
```sh
# generate HDR png file with some MaxCLL value
cjxl image.pfm image.jxl -x color_space=RGB_D65_202_Rel_PeQ --intensity_target=1234 && djxl image.jxl image.png
# encode to jxl
cjxl image.png image2.jxl
# confirm value of intensity_target
jxlinfo image2.jxl
[...]
intensity_target: 1234.000000 nits
[...]
```

### Pull Request Checklist


lib/extras/dec/apng.cc
```diff
@@ -257,6 +257,17 @@ Status DecodeChrmChunk(Bytes payload, JxlColorEncoding* color_encoding) {
   return true;
 }
 
+/** Extract information from 'cLLi' chunk. */
```
sboukortt, line 260:
```suggestion
/** Extracts information from 'cLLi' chunk. */
```

(https://google.github.io/styleguide/cppguide.html#Function_Comments

> Function comments should be written with an implied subject of _This function_ and should start with the verb phrase; for example, "Opens the file", rather than "Open the file".

)

lib/extras/dec/apng.cc
```diff
@@ -257,6 +257,17 @@ Status DecodeChrmChunk(Bytes payload, JxlColorEncoding* color_encoding) {
...
+Status DecodeClliChunk(Bytes payload, float* max_content_light_level) {
```
sboukortt, line 261: Perhaps the function could return a `StatusOr<float>`, and then line 1136 would be something like:

```c++
JXL_ASSIGN_OR_RETURN(ppf->info.intensity_target, DecodeClliChunk(payload));
```

sboukortt (changes requested):
Thanks! Just a couple of minor comments.

lib/extras/dec/apng.cc
ledoge: I'm not sure about this, since other chunks like `gAMA` and `cHRM` are also parsed with `JXL_RETURN_IF_ERROR`, so the current state seems better for code consistency.

sboukortt: I think this is mainly for historical reasons (we haven’t always had `StatusOr`), but we can also change it later.

sboukortt (approved):
Thanks again 😊

PR #3669, opened 2024-06-27
Add draft JNI adaptor for jpegli encoder

eustas:


szabadka (approved):

mo271:
This is causing failing coverage tests, for instance here:
https://github.com/libjxl/libjxl/actions/runs/10457103722/job/28960940495

PR #3698, opened 2024-07-04
Fix "space in script path" problem in all scripts.

eustas:
Scripts that were not prone to this problem were updated also, to make it more homogeneous.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -12,7 +12,8 @@ set -eu
 
 OS=`uname -s`
 
-MYDIR=$(dirname $(realpath "$0"))
+SELF=$(realpath "$0")
+MYDIR=$(dirname "${SELF}")
 
 ### Environment parameters:
 TEST_STACK_LIMIT="${TEST_STACK_LIMIT:-256}"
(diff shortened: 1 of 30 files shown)


szabadka (approved):

ci.sh
```diff
@@ -12,7 +12,8 @@ set -eu
 
 OS=`uname -s`
 
-MYDIR=$(dirname $(realpath "$0"))
+SELF=$(realpath "$0")
+MYDIR=$(dirname "${SELF}")
```
sboukortt, line 16: (No action required)

While it can look a bit surprising at first, it’s technically possible to nest double quotes:

```suggestion
MYDIR=$(dirname "$(realpath "$0")")
```

eustas: True. That works for `$()`, not sure about `${}`.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03

So I decided to write the way that requires least amount of prior knowledge to read.

TransRapid:
Still not fixed unfortuantely.

PR #3725, opened 2024-07-29
Fix incorrect encoding behaviour in some cases in Modular mode.

### LINKED ISSUE #3717: 2k²~ image encoded in lossless mode has color artifact above effort 4
**Describe the bug**
When using `cjxl` to encode a particular 2096x2088 PNG image to lossless JPEG XL, a reddish 48x40 rectangle appears at the bottom-right corner of the encoded JXL.

**To Reproduce**
Encode this image using `cjxl -d 0 -e 5`
</img>

**Expected behavior**
The encoded JXL should contain all pixels of the PNG unaltered.

**Screenshots**
This is a 512x512 crop of the bottom-right corner of the encoded JXL
image

**Environment**
 - OS: Windows 10
 - Compiler version: Unknown (downloaded from GitHub releases)
 - CPU type: x86_64
 - cjxl/djxl version string: `JPEG XL encoder v0.10.2 e148959 [AVX2,SSE2]`

**Additional context**
The image I had this issue on was the [cover art for femtanyl's "CHASER" EP](https://musicbrainz.org/release/ada3880b-7522-47cc-b526-81d2aadb5a4a/cover-art) at original size, which is **slightly graphic.**
This would be rather distracting, so the provided test image is that cover art with the graphic part blacked out. 
It exhibits the same problem as the source image.

veluca93:
Fixes #3717.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -1345,6 +1345,12 @@ Status ModularFrameEncoder::PrepareStreamParams(const Rect& rect,
                                                 const ModularStreamId& stream,
                                                 bool do_color, bool groupwise) {
   size_t stream_id = stream.ID(frame_dim_);
+  if (stream_id == 0 && frame_dim_.num_groups != 1) {
+    // If we have multiple groups, then the stream with ID 0 holds the full
+    // image and we do not want to apply transforms or in general change the
+    // pixel values.
+    return true;
+  }
   Image& full_image = stream_images_[0];
   JxlMemoryManager* memory_manager = full_image.memory_manager();
   const size_t xsize = rect.xsize();


jonsneyers (approved):

jonsneyers:
How did this one go undetected for so long?

Can we add some roundtrip test that would have caught this bug?

(in general, maybe not as part of regular CI, it wouldn't hurt to do some larger-scale roundtrip testing to check that lossless is actually lossless, regardless of the effort / number of threads / image dimensions / etc)

veluca93:
> How did this one go undetected for so long?
> 
> Can we add some roundtrip test that would have caught this bug?
> 
> (in general, maybe not as part of regular CI, it wouldn't hurt to do some larger-scale roundtrip testing to check that lossless is actually lossless, regardless of the effort / number of threads / image dimensions / etc)

The bug would have been caught by any thread-sanitizer run with an image of size least 2049x2049 and effort >=5, I think.

I'd have guessed this to be caused by streaming encoding, but the fact that 0.8.2 seems to show the issue confuses me...

PR #3750, opened 2024-08-13
tools/file_io.h: set binary mode for stdin and stdout on Windows

### LINKED ISSUE #542: Support operating on stdin and stdout in cjxl/djxl
For command line programs on *nix systems a common method of use is through redirecting the input and output of programs from/into other programs or files. This is done by redirecting standard input and standard output, and is a common way of doing multiple processing steps without having to write a bunch of temporary files. One of the many benefits of this is that it handles streaming data for situations when storage of the intermediate files (which are often in fully uncompressed formats) exceeds the storage of the host.

Currently cjxl and djxl only support reading and writing to files, which means that they cannot (easily) be integrated into this kind of workflow, and miss out in the described benefits.

There are several ways this could be implemented.
Some programs, such as `cjpeg` will automatically use stdin/stdout if no files are specified as arguments.
```sh
cjpeg -quality 60 <input.ppm >output.jpg
```
Other programs have an option that will cause them to read stdin or stdout.
```sh
cjxl --stdin output.jxl < input.png
cjxl --stdout input.png > output.jxl
cjxl --stdin --stdout < input.png  > output.jxl
```
Another common pattern is using `-` in place of the filename as an alias of stdin and stdout, but I can't remember how that works.

Something that needs to be considered is that cjxl and djxl currently have output to stdout, in the encode/decode status. When using stdin/stdout this output would have to be suppressed.

I think my preferred option would be the second one, because it makes things explicit, and avoids any potential issues around not being able to operate on files called `-`, but others may have better ideas as to how this should be implemented.

**Describe alternatives you've considered**
Currently it is possible through virtual files that are aliases of stdin and stdout, as below, but this is clunky.
```sh
cjxl /dev/stdin /dev/stdout <input.png >output.jxl
```

**Additional context**
Requested by [Sedna#4917 on the JXL discord](https://discord.com/channels/794206087879852103/803663417881395200/883852155952439316).

Traneptora:
Windows stdin and stdout default to text mode, but they can be changed to binary mode with _setmode from <io.h>. This is relevant because text mode enables LF <-> CRLF translation, and binary mode disables it, and it needs to be disabled for image reading/writing. This is technically also true on Unix-like operating systems, but text and binary do the same thing on macOS, Linux, etc.

Fixes #542

### DIFF
--- a/tools/file_io.h
+++ b/tools/file_io.h
@@ -19,6 +19,11 @@
 
 #include "lib/jxl/base/compiler_specific.h"
 
+#ifdef _WIN32
+#include <fcntl.h>
+#include <io.h>
+#endif
+
 namespace jpegxl {
 namespace tools {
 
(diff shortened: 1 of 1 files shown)


Traneptora:
It appears CI is failing after the commits from august 14th, even though it passed earlier.

mo271 (approved):

PR #3752, opened 2024-08-14
Add PPC to the CI

sboukortt:


mo271 (dismissed):

eustas:
FWIW: was successful running build/test under docker on ppc64le ubuntu; likely tests failures are because of glibc linking problems; some tests took 1300+ seconds on my PC.

PR #3770, opened 2024-08-27
Expose a wrapper function instead of Allotment

sboukortt:
This lets us detect and handle failure in initialisation without requiring that we make `Allotment` movable for use with `JXL_ASSIGN_OR_RETURN`.


lib/jxl/enc_bit_writer.h
sboukortt, line 1: The diff for this file is a bit misleading. What I did was move `Allotment` to the private section of `BitWriter`.

PR #3798, opened 2024-09-05
check realloc in basicinfo

### LINKED ISSUE #3788: check realloc
File:
/libjxl-main/tools/jxlinfo.c: line 389
Code:
  box_index += chunk_size - remaining;
  box_data = realloc(box_data, box_size); // <-- check realloc

mo271:
fixes #3788

### DIFF
--- a/tools/jxlinfo.c
+++ b/tools/jxlinfo.c
@@ -386,7 +386,14 @@ static int PrintBasicInfo(FILE* file, int verbose) {
       const size_t remaining = JxlDecoderReleaseBoxBuffer(dec);
       box_size += chunk_size;
       box_index += chunk_size - remaining;
-      box_data = realloc(box_data, box_size);
+      void* temp = realloc(box_data, box_size);
+      if (temp == NULL) {
+          free(box_data);
+          fprintf(stderr, "Memory reallocation failed\n");
+          break;
+      } else {
+          box_data = temp;
+      }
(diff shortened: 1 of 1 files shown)


tools/jxlinfo.c
```diff
@@ -386,7 +386,14 @@ static int PrintBasicInfo(FILE* file, int verbose) {
       const size_t remaining = JxlDecoderReleaseBoxBuffer(dec);
       box_size += chunk_size;
       box_index += chunk_size - remaining;
-      box_data = realloc(box_data, box_size);
+      void* temp = realloc(box_data, box_size);
+      if (temp == NULL) {
+        free(box_data);
```
sboukortt, line 391: I think I would feel slightly more comfortable with an additional:

    box_data = NULL;
    box_size = 0;
    box_index = 0;

sboukortt (changes requested):

tools/jxlinfo.c
```diff
@@ -386,7 +386,17 @@ static int PrintBasicInfo(FILE* file, int verbose) {
...
+        box_data = NULL;
+        box_size = 0;
+        box_index = 0;
+        fprintf(stderr, "Memory reallocation failed\n");
+        break;
+      } else {
+        box_data = temp;
```
sboukortt, line 398: Another comment I forgot to make: if the true branch is going to break anyway, we don’t need the `else`.

mo271: good point, done

sboukortt (approved):

PR #3819, opened 2024-09-13
Faster PNG compression

jonnyawsom3:
I did the thing.

Lowers the zlib compression level from the default 6 to 1, to lessen the impact of PNG writing on JXL decoding tests.
This results in a 20% density decrease and a 73% speed increase of djxl when outputting to PNGs.


sboukortt (approved):
Good idea, thanks!

PR #3824, opened 2024-09-14
allow setting modular 16 bit flag to true

jonsneyers:
Could be useful for testing what happens in various implementations, and also to possibly make slightly smaller jxl art by saving a few header bits (this flag has to be true for all_default)

(also technically if this flag is false then the image is not Level 5 so it requires a container and a `jxll` box, so having a way to make jxl art that is actually a valid bitstream is nice)


CrushedAsian255:
Maybe the name should be "16BitModular", as all the others seem to use PascalCase instead of snake_case.

CrushedAsian255:
Any progress on this?

veluca93 (approved):

PR #3826, opened 2024-09-14
cmake: Do not use -mrelax-all with clang on RISCV64

kraj:
This option causes clang 19+ to crash on riscv64

For RISC-V -mrelax-all causes the assembler stage to convert all conditional branches into

  bcc reg, reg, tmplabel
  jmp target
tmplabel:

Unconditional jmp can reach targets a lot further than a conditional branch. Unfortunately, jmp also has a limited range so sometimes needs to be converted to an indirect jump which requires a scratch register. We do all branch relaxation early for RISC-V during the compilation stage so we can allocate the scratch register.

If the assembler also does branch relaxation it will increase the size of the function and may cause unconditional jmps to go out of range. If this happens, linking will fail.

### Pull Request Checklist


mo271 (approved):

CONTRIBUTORS
```diff
@@ -10,6 +10,7 @@ Iulia-Maria Comșa
 Jan Wassenberg
 Jon Sneyers
 Jyrki Alakuijala
+Khem Raj
```
sboukortt, line 13: The appropriate file would be [AUTHORS](https://github.com/libjxl/libjxl/blob/main/AUTHORS).

sboukortt (changes requested):

CONTRIBUTORS
kraj: done

sboukortt (approved):

sboukortt:
Thanks!

PR #3831, opened 2024-09-16
binary releases for Ubuntu 24.04

mo271:


mo271:
seems like it worked: https://github.com/libjxl/libjxl/actions/runs/10878245339/artifacts/1936067225

PR #3833, opened 2024-09-16
detect broken ICC when encoding jpg

### LINKED ISSUE #3131: Failed to losslessly recompress and reconstruct some JPEG files
**Describe the bug**
Failed to losslessly recompress and reconstruct some JPEG files.

**To Reproduce**
JPEG file sample: [181.zip](https://github.com/libjxl/libjxl/files/13927234/181.zip)

Afer v0.9.0
```
$ cjxl --lossless_jpeg=1 181.jpg 181.jxl
JPEG XL encoder v0.9.1 b8ceae3 [AVX2,SSE4,SSSE3,SSE2]
Encoding [JPEG, lossless transcode, effort: 7]
Compressed to 58662 bytes including container

$ djxl 181.jxl 181_output.jpg
JPEG XL decoder v0.9.1 b8ceae3 [AVX2,SSE4,SSSE3,SSE2]
Failed to decode image
Warning: could not decode losslessly to JPEG. Retrying with --pixels_to_jpeg...
Failed to decode image
```

Before v0.9.0
```
$ cjxl --lossless_jpeg=1 181.jpg 181.jxl
JPEG XL encoder v0.8.2 954b460 [AVX2,SSE4,SSSE3,Unknown]
Read JPEG image with 88508 bytes.
JxlEncoderAddJPEGFrame() failed.
EncodeImageJXL() failed.
```

The behavior after v0.9.0 is particularly dangerous, as one might think that files successfully processed by cjxl can always be reconstructed by djxl, thereby deleting the original JPEG file and losing it forever.

**Environment**
 - OS: Windows, also tested on Debian
 - Compiler version: all download from GitHub releases, so I guess clang
 - CPU type: x86_64
 - cjxl/djxl version string: see command line output above

mo271:
Attempt to fix #3131

### DIFF
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -2101,7 +2101,7 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
             *io.Main().jpeg_data,
             &frame_settings->enc->metadata.m.color_encoding)) {
       return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_BAD_INPUT,
-                           "Error decoing the ICC profile");
+                           "Error decoding the ICC profile embedded in the input JPEG");
     }
     frame_settings->enc->color_encoding_set = true;
   }
--- a/lib/jxl/encode.cc
+++ b/lib/jxl/encode.cc
@@ -2097,8 +2097,12 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
   }
 
   if (!frame_settings->enc->color_encoding_set) {
-    SetColorEncodingFromJpegData(
-        *io.Main().jpeg_data, &frame_settings->enc->metadata.m.color_encoding);
+    if (!SetColorEncodingFromJpegData(
+            *io.Main().jpeg_data,
+            &frame_settings->enc->metadata.m.color_encoding)) {
+      return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_BAD_INPUT,
+                           "Error decoing the ICC profile");
+    }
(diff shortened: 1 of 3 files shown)


mo271:
Not quite sure that this is the way to go here

lib/jxl/encode.cc
```diff
@@ -2097,8 +2097,12 @@ JxlEncoderStatus JxlEncoderAddJPEGFrame(
   }
 
   if (!frame_settings->enc->color_encoding_set) {
-    SetColorEncodingFromJpegData(
-        *io.Main().jpeg_data, &frame_settings->enc->metadata.m.color_encoding);
+    if (!SetColorEncodingFromJpegData(
+            *io.Main().jpeg_data,
+            &frame_settings->enc->metadata.m.color_encoding)) {
+      return JXL_API_ERROR(frame_settings->enc, JXL_ENC_ERR_BAD_INPUT,
+                           "Error decoing the ICC profile");
```
jonsneyers, line 2104:
```suggestion
                           "Error decoding the ICC profile embedded in the input JPEG");
```

jonsneyers:
I guess the other option would be to allow creating recompressed jpegs with a broken ICC profile and fix the reconstruction process to make that work, but I don't think that's a good option. That would produce a jxl bitstream that is technically invalid, which I don't think we should do.

So I think just considering such JPEGs bad input, like it is done in this PR, is the right way to go.

sboukortt (approved):

sboukortt (approved):

Traneptora:
Does this still consider it bad input with `--strip`?

PR #3842, opened 2024-09-21
fix cmake script for a2x

zzzsyyy:
In NixOS, the shebang of the `a2x` script is `xxx/bash -e`, which doesn't match the current regex, causing the build to default to Python and fail. 

This PR updates the regex to correctly match such shebangs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -463,7 +463,7 @@ if(JPEGXL_ENABLE_MANPAGES)
   find_program(ASCIIDOC a2x)
   if(ASCIIDOC)
     file(STRINGS "${ASCIIDOC}" ASCIIDOC_SHEBANG LIMIT_COUNT 1)
-    if(ASCIIDOC_SHEBANG MATCHES "sh$" OR ASCIIDOC_SHEBANG MATCHES "libexec/bin/python$" OR MINGW)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -460,7 +460,7 @@ if(JPEGXL_ENABLE_MANPAGES)
   find_program(ASCIIDOC a2x)
   if(ASCIIDOC)
     file(STRINGS "${ASCIIDOC}" ASCIIDOC_SHEBANG LIMIT_COUNT 1)
-    if(ASCIIDOC_SHEBANG MATCHES "sh$" OR ASCIIDOC_SHEBANG MATCHES "libexec/bin/python$" OR MINGW)
+    if(ASCIIDOC_SHEBANG MATCHES "sh( -e$|$)" OR ASCIIDOC_SHEBANG MATCHES "libexec/bin/python$" OR MINGW)
```
sboukortt, line 463: Thanks. Would this also work?
```suggestion
    if(ASCIIDOC_SHEBANG MATCHES "sh( -e)?$" OR ASCIIDOC_SHEBANG MATCHES "libexec/bin/python$" OR MINGW)
```

zzzsyyy: this also works, thanks

eustas (approved):

PR #3850, opened 2024-09-24
ClangTidy: included header cmath is not used directly

mo271:


mo271:
@jmroot I'm getting this from ClangTidy, although #3828 makes sense, I guess.

jmroot:
Well, `loop_filter.cc` is definitely still calling `std::abs` on floating-point values, and the floating-point overrides for `std::abs` are definitely provided by `<cmath>` as per the standards. Does ClangTidy perhaps want the include to go in `loop_filter.cc` instead, since nothing in `loop_filter.h` itself uses it?

sboukortt (approved):

PR #3862, opened 2024-09-27
add ShrinkTo to ppf and remove some CodecInOut in passes_test

mo271:


mo271:
after this there are only 11 uses of `CodecInOut::ShrinkTo` left.

lib/extras/packed_image.h
```diff
@@ -149,6 +149,38 @@ class PackedImage {
...
+        malloc(std::max<size_t>(1, new_pixels_size)), free);
+
+    if (!new_pixels) {
+      // TODO(firsching): use specialized OOM error code
+      return JXL_FAILURE("Failed to allocate memory for shrunk image");
+    }
+
+    // Copy data row by row
+    for (size_t y = 0; y < new_ysize; ++y) {
+      memcpy(reinterpret_cast<uint8_t*>(new_pixels.get()) + y * new_stride,
```
szabadka, line 169: why do we need to reallocate + copy here? wouldn't leaving stride as is and just setting xsize/ysize to the new values work?

mo271: I guess yes, but I thought after we reallocate and copy here, less memory is used.

szabadka: In the CodecIO/ImageF case we only set the new size, and don't reallocate. I think the idea was that this operation should be cheap.

Since this is only used in tests, one more reason to just set the new size is to find any cases where we incorrectly assume that stride = xsize * pixel_stride_

eustas:
#4407

PR #3910, opened 2024-10-27
Fix e10 butteraugli iterations

Galaxy4594:
This bug explains why e10 is faster than e9. It's only doing 2 iterations instead of 4!


CrushedAsian255:
LGTM

CrushedAsian255:
ok now LGTM

eustas (approved):

Issue #3134, opened 2024-01-15
Encoding fails for 'Optimized' GIFs


jonnyawsom3:
**Describe the bug**
On some GIFs, cjxl fails with "Getting pixel data failed."
Using "Coalesce" on https://ezgif.com/optimize allows the same GIF to then be read by cjxl

**To Reproduce**
`cjxl -d 0 Optimized.gif Optimized.jxl`
`cjxl -d 0 Un-optimized.gif Un-optimized.jxl`

**Environment**
 - OS: Windows
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.9.1 b8ceae3 [AVX2,SSE4,SSSE3,SSE2]

Original GIF
![Optimized](https://github.com/libjxl/libjxl/assets/51539240/d5a56cb4-b9e4-47c8-a20f-53bcf1f5a098)
Un-optimized GIF
![Un-optimized](https://github.com/libjxl/libjxl/assets/51539240/e4d0698e-dcf9-42cb-b3b4-efd203e957a8)

Traneptora:
Here's the error message from a debug build:
```
$ cjxl -d 0 296631940-d5a56cb4-b9e4-47c8-a20f-53bcf1f5a098.gif test.jxl
JPEG XL encoder v0.10.0 875e72dd [AVX2]
./lib/extras/dec/gif.cc:284: JXL_FAILURE: GIF with dispose-to-0 is not supported for non-full or blended frames
./lib/extras/dec/decode.cc:141: JXL_FAILURE: Codecs failed to decode
Getting pixel data failed.
```

jonnyawsom3:
Ah, so it would need to use an additive(?) blend mode on those frames instead of erroring out

fabiorug:
> Ah, so it would need to use an additive(?) blend mode on those frames instead of erroring out. Although that sounds like it defaults to only accepting full frames with dispose 0 rather than the default GIF behaviour of overlapping

Not overlapping

Traneptora:
Yea, but since this is a not-yet-implemented feature rather than an actual defect, and not impossible to implement, in the short-term it would be nice if an error message was printed that explained that.

mo271:
I think this is a more general problem of how much error messages are printed in `release` mode versus `opt` and `debug`. 
Not sure if this is easy to fix.

Lunaphied:
[...]

PR #3918, opened 2024-11-04
warn user when GIF disposal mode is not supported

Amiralgaby:
### Description

With `cjxl`, if a GIF has dispose-to-0 mode, an error `Getting data error` is printed (#3134)

This pull request change 
```
$ cjxl -d 0 296631940-d5a56cb4-b9e4-47c8-a20f-53bcf1f5a098.gif test.jxl
JPEG XL encoder v0.12.0 adeeccc4 [SSE2]
Getting pixel data failed.
```
to
```
$ ./cjxl ~/Images/icons/296631940-d5a56cb4-b9e4-47c8-a20f-53bcf1f5a098.gif test.jxl -d 0
JPEG XL encoder v0.12.0 adeeccc4 [SSE2]
GIF with dispose-to-0 is not supported for non-full or blended frames
Getting pixel data failed.
```

### Pull Request Checklist

- [x] **CLA Signed**: Have you signed the [Contributor License Agreement](https://code.google.com/legal/individual-cla-v1.0.html)


jonsneyers:
If this happens in practice, then maybe we should get rid of that TODO and actually make it work, rather than just improving the error messaging...

@veluca93 do you remember what the complication was that caused this case not to be handled?

veluca93:
> If this happens in practice, then maybe we should get rid of that TODO and actually make it work, rather than just improving the error messaging...
> 
> @veluca93 do you remember what the complication was that caused this case not to be handled?

I don't think it would be *too* complicated, but I suspect it would need inserting dummy frames, which would be why we did not do it to begin with...

jonnyawsom3:
> I don't think it would be _too_ complicated, but I suspect it would need inserting dummy frames, which would be why we did not do it to begin with...

Could reference frames or patches be used instead? Should have minimal impact on density then, albeit with a more complex solution.

eustas:
Since support for that disposal method hasn't been added in 1+ year, let's land this at least

eustas (approved):

jonsneyers (approved):

PR #3940, opened 2024-11-19
Prevent squeeze to 0px dimension

eustas:


jonsneyers:
Changing the default Squeeze script seems risky, aren't there cases where this leads to jxl files no longer decoding?

eustas:
This has been found by `streaming_fuzzer`; it crashes, because channel is "moved" to groups, but later on it still formally eligible for decoding into it.

Will check if crashes cjxl as well.

Code change only affects 1xK and Kx1 images; if those do not encode normally, then the change is considered safe (as there are no such encoded images in the wild).

eustas:
Repro:
```
convert -size 1x1025 xc: narrow.png
./build/tools/cjxl -m 1 -g 3 ./narrow.png ./narrow.jxl
```

Result:
```
JPEG XL encoder v0.12.0 d28c7a24 [AVX2,SSE4,SSE2]
Encoding [Modular, d1.000, effort: 7]
lib/jxl/image.h:101: JXL_DASSERT: y < ysize_
Illegal instruction
```

Also with `-g 2`; but OK with `-g 1` and `-g 0`

Issue #3937, opened 2024-11-18
Modular encoder fails for tall/wide images


eustas:
```
convert -size 15x167941 xc: tall.png
# if fails, update /etc/ImageMagick-6/policy.xml

./build/tools/cjxl ./tall.png ./tall.jxl
# OK

./build/tools/cjxl -m 1 ./tall.png ./tall.jxl
# Fails
```

Error:
```
JPEG XL encoder v0.12.0 538c140d [AVX2,SSE4,SSE2]
Encoding [Modular, d1.000, effort: 7]
lib/jxl/modular/encoding/encoding.cc:523: JXL_FAILURE: Inconsistent transforms
lib/jxl/enc_modular.cc:1036: JXL_RETURN_IF_ERROR code=1: ValidateChannelDimensions(gi, stream_options_[0])
lib/jxl/enc_frame.cc:1612: JXL_RETURN_IF_ERROR code=1: enc_modular.ComputeEncodingData( frame_header, metadata->m, &color, extra_channels, group_rect, frame_dim, frame_area_rect, &enc_state, cms, pool, aux_out, cparams.modular_mode)
lib/jxl/enc_frame.cc:2125: JXL_RETURN_IF_ERROR code=1: ComputeEncodingData( cparams, frame_info, metadata, frame_data, jpeg_data.get(), 0, 0, frame_data.xsize, frame_data.ysize, cms, pool, frame_header, enc_modular, enc_state, &group_codes, aux_out)
lib/jxl/encode.cc:999: Failed to encode frame
JxlEncoderProcessOutput failed.
EncodeImageJXL() failed.
```

eustas:
Disabling this check causes considerable diff in "streaming_fuzzer" (-> failure).

eustas:
@jonsneyers could you take a look at `ValidateChannelDimensions` / `DefaultSqueezeParameters`. Likely one is over-restrictive or the other is not taking into account group dim...

jonnyawsom3:
Should be fixed by #3942 (I think).

eustas:
Re-open; a case found where `ValidateChannelDimensions` fails.

eustas:
```
Image size: 57601 X 1, d=0.000000
JXL_ENC_FRAME_SETTING_EFFORT = 8
JXL_ENC_FRAME_SETTING_DECODING_SPEED = 3
JXL_ENC_FRAME_SETTING_NOISE = -1
[...]

PR #3942, opened 2024-11-20
Make 0px channels eligible for "stop" check

eustas:
Before this PR 0px channels were "skipped" before the check. In some cases 1xK and Kx1 images are squeezed (for 420 subsampling) and thus 0xK / Kx0 channels appear.

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -1031,6 +1031,7 @@ Status ModularFrameEncoder::ComputeEncodingData(
       if (fc.w > frame_dim_.group_dim || fc.h > frame_dim_.group_dim) break;
     }
     for (; ch < full_image.channel.size(); ch++) {
+      // TODO(eustas): shrink / assign channel to keep size consistency
       full_image.channel[ch].plane = ImageI();
     }
   }
--- a/lib/jxl/modular/encoding/enc_encoding.cc
+++ b/lib/jxl/modular/encoding/enc_encoding.cc
@@ -617,13 +617,13 @@ Status ModularEncode(const Image &image, const ModularOptions &options,
           options.max_property_values);
     }
     for (size_t i = 0; i < nb_channels; i++) {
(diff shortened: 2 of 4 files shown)


jonsneyers:
Is this a spec change?

eustas:
Going to check.

eustas:
(sorry, only old spec is at hands reach now)

C.4.8: `The decoder shall decode a modular sub-bitstream (C.9). However, the decoder shall only decode the first nb_meta_channels channels and any further channels that have a width and height that are both at most 256.`

C.8.2: `In the partial image decoded from the GlobalModular and ModularDcGroup sections, the pixels in the remaining channel data that correspond to the group are decoded as another modular image sub-bitstream (C.9)... The number of channels and their dimensions are derived as follows: for every remaining channel in the partially decoded GlobalModular image (i.e. it is not a meta-channel, the channel dimensions exceed 256×256,...`

C.9.2: `The decoder shall then decode the data for each channel (in ascending order of index) as specified in C.9.3, skipping any channels whose width or height is zero. Finally the series of inverse transformations shall be applied as described in C.9.4.`

So, both C.4.8 and C.8.2 add "extra" condition on the top of C.9. C.9.2 gives a lower level hint - "no pixels -> skip".

But of course, that could be interpreted the other way round.

eustas:
@jonsneyers friendly ping

jonsneyers:
I mean: are there files produced by older cjxl/libjxl that will no longer decode if this gets merged, or will the new cjxl/libjxl produce files that will not decode by older djxl/libjxl?

eustas:
Current encoder and decoder fail when they meet specific input (very tall/wide image + modular + group-shift).
So, likely there are no such files in the wild.
New encoder will be able to produce files on which old decoder will fail.

Context: #3937

jonsneyers:
OK, let's just check if this is something that needs to be made more clear in the spec.

jonsneyers (approved):

PR #3988, opened 2024-12-02
Drop Symmetric5 overload

eustas:
It is used only once.

This PR is pre-work for removing convolve out of decoder sources


mo271 (dismissed):

eustas:
@mo271 please re-stamp

PR #4021, opened 2024-12-19
Update SKCMS

eustas:
It's structure has been changed - avx2 / avx512 specializations have been factored out


eustas:
It seems that SKCMS "dynamic dispatch" does not really work. Investigating

eustas:
Likely dynamic dispatch works. But depending on compiler flags HSW (or even Baseline) emits not allowed instructions.

PR #4060, opened 2025-01-13
Undisable SVE

eustas:


jonsneyers:
Please revert or make this conditional on the platform being something where it works.
On MacOS, I'm running into compile issues after this commit...

eustas:
Sure. I was under impression it was fixed on HWY side. Going to:
 - roll back
 - check if it could be fixed by rolling on HWY
 - roll-forward with condition for osx (or compilation check)

kleisauke:
Updating the highway submodule to https://github.com/google/highway/commit/1d5cc073467fe38c79de229797e90d07ea287cf7 ought to fix that.

eustas:
Interesting: 'macos-latest' git-hub ruuner is supposed to be M1 instance... and it did not see the problem.
Though, it uses AppleClang15.
Going to try with (now public preview) macos-15 runner.

PR #4062, opened 2025-01-13
add decode option to set leniency

### LINKED ISSUE #3712: Decode option for decoding corrupted bitstreams
Currently, if corruption is detected (typically by ANS end state not matching), the decoder refuses to decode the image.

It would be useful to have an option in the decoder to ignore corrupted sections and still return an image in this case. It makes sense to disable this option by default and preserve the current behavior, but it would be a useful option to have in order to get a basic form of error resilience / data recovery.

If the corruption only happens in AC groups, ignoring the corrupted groups (i.e. assuming all coefficients are zero) would allow still recovering a reasonable image (something similar to a progressive preview).

It might be a good idea to add this option to the API before v1.0.

Possibly this could be not just a boolean setting, but an integer with different levels of 'stubbornness': default 0 would be to give up on any error, value 1 could be "allow 1 corrupt AC group but not more", value 2 could be "allow any number of corrupt AC groups", value 3 could be "also allow corrupt DC groups" (e.g. setting the corrupt DC values to gray so you can still see any available high-frequency data in the group).

Setting this option to a nonzero value does not guarantee that you will get an image in all cases: for corruptions in the headers etc, the decoder will still refuse to decode the image.

jonsneyers:
This allows decoding (some) corrupt bitstreams. Fixes #3712.

Does not catch all kinds of corruptions yet, nor does it handle them in the best possible way, but it's a start.

The function `JxlDecoderSetLeniency()` is added, as well as the corresponding `djxl` option `--leniency=N`.
Default leniency is zero which does not change any behavior.

Leniency 1:
Corrupted HF sections are replaced with LF only (in case of VarDCT) or zeroes (in case of Modular).
This is a reasonable "graceful degradation": in case of lossy (VarDCT or Modular) it means the corrupt region becomes blurry; in case of lossless it means the corrupt region becomes black.

Leniency 2:
On top of leniency 1, also continue if LF sections are corrupt (corresponding to 2048x2048 pixels).
Corrupted LF sections are just turned entirely black.

This creates a basic mechanism for some amount of error recovery; instead of just returning an error when a bitstream is corrupt, you can now get a reasonable image.

TODO: better handle multi-pass and multi-frame.

Negative leniency (not implemented yet) can be useful if we want to have a way to be "stricter than default".

E.g. we currently don't check value ranges / integer overflows in places where checking would cause a slowdown (e.g. checking that modular data remains in int16 / int32 range, etc) so libjxl does decode some invalid codestreams without complaining. Having a way to validate codestreams would be nice, but it's probably best to keep that a non-default option since it will cause slowdown. Also I think we currently only check level 10 limits but not level 5 ones. Anyway: negative leniency is not implemented yet but it could be another use of this decode option besides positive leniency to handle corrupt bitstreams.

### DIFF
--- a/lib/include/jxl/decode.h
+++ b/lib/include/jxl/decode.h
@@ -474,6 +474,24 @@ JXL_EXPORT size_t JxlDecoderSizeHintBasicInfo(const JxlDecoder* dec);
 JXL_EXPORT JxlDecoderStatus JxlDecoderSubscribeEvents(JxlDecoder* dec,
                                                       int events_wanted);
 
+/** Configure how strict or lenient the decoder is when the input bitstream
+ * is technically non-conforming, truncated or corrupt.
+ * By default, the value is 0. Higher values mean more leniency, lower values
+ * mean stricter.
+ * Currently supported values are:
+ * -2: fully check bitstream conformance including value range (not implemented)
+ * -1: check signaled level limitations, error on corruption (not implemented)
+ * 0 (default): check Main profile Level 10 limitations, error on corruption
+ * 1: recover HF corruption, error on other corruption
+ * 2: recover all recoverable corruption, error on non-recoverable corruption
+ *
+ * @param dec decoder object
+ * @param leniency amount of leniency regarding invalid input
+ * @return ::JXL_DEC_SUCCESS if no error, ::JXL_DEC_ERROR otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetLeniency(JxlDecoder* dec,
+                                                  int leniency);
+
 /** Enables or disables preserving of as-in-bitstream pixeldata
  * orientation. Some images are encoded with an Orientation tag
  * indicating that the decoder must perform a rotation and/or
--- a/lib/jxl/dec_frame.cc
+++ b/lib/jxl/dec_frame.cc
@@ -317,21 +317,61 @@ Status FrameDecoder::ProcessDCGroup(size_t dc_group_id, BitReader* br) {
   const size_t gx = dc_group_id % frame_dim_.xsize_dc_groups;
   const size_t gy = dc_group_id / frame_dim_.xsize_dc_groups;
   const LoopFilter& lf = frame_header_.loop_filter;
+  Status status(true);
   if (frame_header_.encoding == FrameEncoding::kVarDCT &&
       !(frame_header_.flags & FrameHeader::kUseDcFrame)) {
-    JXL_RETURN_IF_ERROR(modular_frame_decoder_.DecodeVarDCTDC(
-        frame_header_, dc_group_id, br, dec_state_));
+    status = modular_frame_decoder_.DecodeVarDCTDC(frame_header_, dc_group_id,
+                                                   br, dec_state_);
+    if (!status) {
+      if (dec_state_->leniency > 1) {
+        fprintf(stderr,
+                "Error in VarDCT LF group %" PRIuS " at position (%" PRIuS
+                ",%" PRIuS ") group_dim: %" PRIuS "\n",
+                dc_group_id, gx * frame_dim_.dc_group_dim,
+                gy * frame_dim_.dc_group_dim, frame_dim_.dc_group_dim);
+      } else {
+        return JXL_FAILURE("Error in ProcessDCGroup/DecodeVarDCTDC");
+      }
+    }
   }
   const Rect mrect(gx * frame_dim_.dc_group_dim, gy * frame_dim_.dc_group_dim,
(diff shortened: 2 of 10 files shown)


jonnyawsom3:
Mentioning a 'negative leniency', perhaps that could have an equivalent in cjxl to make sure lossless files are actually lossless, as some have asked recently. Positive vales enabling invisible pixel zeroing to give it more than 1 use, ect.

jonsneyers:
> Mentioning a 'negative leniency', perhaps that could have an equivalent in cjxl to make sure lossless files are actually lossless, as some have asked recently. Positive vales enabling invisible pixel zeroing to give it more than 1 use, ect.

This PR is a decode option, not an encode option.

There is already an encode option to select what to do with invisible pixels iirc. And lossless should always be lossless, if it's not that's a bug. We could add an explicit roundtrip test option to cjxl if that's a desirable feature, but I don't think there needs to be a libjxl-level feature for it.

lib/jxl/dec_frame.cc
```diff
@@ -317,21 +317,61 @@ Status FrameDecoder::ProcessDCGroup(size_t dc_group_id, BitReader* br) {
...
+      } else {
+        return JXL_FAILURE("Error in ProcessDCGroup/DecodeVarDCTDC");
+      }
+    }
   }
   const Rect mrect(gx * frame_dim_.dc_group_dim, gy * frame_dim_.dc_group_dim,
                    frame_dim_.dc_group_dim, frame_dim_.dc_group_dim);
-  JXL_RETURN_IF_ERROR(modular_frame_decoder_.DecodeGroup(
+  bool zerofill, allow_truncated;
+  zerofill = allow_truncated = (dec_state_->leniency > 1);
```
veluca93, line 340: I'd prefer
```
bool zerofill = dec_state_->leniency > 1;
bool allow_truncated = zerofill;
```

lib/jxl/dec_group.cc
```diff
@@ -805,10 +805,16 @@ Status DecodeGroup(const FrameHeader& frame_header,
...
       dec_state, first_pass));
 
-  JXL_RETURN_IF_ERROR(HWY_DYNAMIC_DISPATCH(DecodeGroupImpl)(
+  Status status = HWY_DYNAMIC_DISPATCH(DecodeGroupImpl)(
       frame_header, get_block.get(), group_dec_cache, dec_state, thread,
-      group_idx, render_pipeline_input, jpeg_data, draw));
-
+      group_idx, render_pipeline_input, jpeg_data, draw);
+  if (!status && dec_state->leniency > 0) {
+    printf("making it LF\n");
```
veluca93, line 812: :-)

lib/include/jxl/decode.h
```diff
@@ -474,6 +474,24 @@ JXL_EXPORT size_t JxlDecoderSizeHintBasicInfo(const JxlDecoder* dec);
...
+ * -1: check signaled level limitations, error on corruption (not implemented)
+ * 0 (default): check Main profile Level 10 limitations, error on corruption
+ * 1: recover HF corruption, error on other corruption
+ * 2: recover all recoverable corruption, error on non-recoverable corruption
+ *
+ * @param dec decoder object
+ * @param leniency amount of leniency regarding invalid input
+ * @return ::JXL_DEC_SUCCESS if no error, ::JXL_DEC_ERROR otherwise.
+ */
+JXL_EXPORT JxlDecoderStatus JxlDecoderSetLeniency(JxlDecoder* dec,
```
veluca93, line 492: I'm not sure I like the name... I think we'll discuss this in the next team meeting and get back to you :-)

jonsneyers: Could also call it SetStrictness and reverse the scale...

lib/jxl/dec_frame.cc
jonsneyers: Sure, that's more readable indeed.

lib/jxl/dec_group.cc
jonsneyers: Oops, that should have been a comment, not a printf :)

jonnyawsom3:
Was testing it out just now and discovered `--leniency=2` causes squeeze encoded modular files to break
Test

jonsneyers:
> Was testing it out just now and discovered `--leniency=2` causes squeeze encoded modular files to break ![Test]

Oops! Thanks for spotting this! Silly mistake, this should fix it.
The image you get is the result of accidentally zeroing modular LF groups (while global LF is still fine and so is the HF), while they should only be zeroed if they are corrupted. That messes up the mid-frequencies, which actually produces a surprisingly reasonable image still, I was expecting something worse :)

PR #4077, opened 2025-01-21
Flatten Separable5

eustas:
Move "controller" into separable5
Fix mirroring.
5x less codegen, likely smaller stack frames.

Row mirroring got a bit bigger (but, hopefully correct; see #4076)


mo271 (approved):
nice fix

PR #4079, opened 2025-01-22
fix JXL_DEBUG_V_LEVEL build

mo271:


mo271:
I don't think it is worth the effort to add a test for that

eustas (approved):

PR #4084, opened 2025-01-23
Fix to build static linked binaries on macOS

lisanet:
This PR enables building statically linked binaries on macOS.

When attempting to build statically linked binaries with Apple's C++ compiler, the -static flag causes an error due to the absence of static versions of libc++ and libSystem. 

omiting this flag, will build everything statically linked, with these two system libraries remaining dynamically linked—which is absolutely fine.

Additionally, the changes include copying pnglibconf.h.prebuild to libpngconf.h to resolve build dependencies when compiling lib/extras/dec/apng.cc or lib/extras/enc/apng.cc


eustas:
Thanks

PR #4099, opened 2025-01-30
Remove unnecessary space

alantudyk:


tools/speed_stats.cc
```diff
@@ -51,7 +51,7 @@ bool SpeedStats::GetSummary(SpeedStats::Summary* s) {
 
     s->central_tendency = pow(product, 1.0 / (elapsed_.size() - 1));
     s->variability = 0.0;
-    s->type = " geomean:";
+    s->type = "geomean:";
```
eustas, line 54: this seems to be inconsistent with other stats, e.g. `second`, `median`.

alantudyk: > inconsistent

I don't care. And I won't fix it.

PR #4130, opened 2025-02-25
slight improvement in jpeg recompression

jonsneyers:
Changes the AC context modeling used when recompressing JPEGs.
Previous model: each DC  component was split into up to 3 equal buckets, and then luma AC used 3x3=9 contexts based on chroma DC while chroma AC used 3 contexts (shared between Cb and Cr) based on luma DC.
New model: only care about luma DC, split it in up to 4 buckets, and use up to 4 contexts per component (not shared between Cb and Cr).

This results in 0.1 - 0.2% better compression overall, depending on quality and image size. For larger images at higher quality, the improvement can be up  to 0.3 - 0.4 %; for smaller images the improvement is smaller (0.05 % or so).

Also the improvement seems to be a bit larger for jpegli encoded images than for libjpeg-turbo encoded images, which is nice because the amount of saving from recompression is in general lower for (already well-optimized) jpegli images than for (not very optimized) libjpeg-turbo images.

Your mileage may vary on individual images, I only looked at corpus totals.

### Detailed results:

corpus  | jpeg encoder | q setting | chroma | jpeg corpus size | jxl before | jxl after | improvement | saving before | saving after
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
imagecompression.info | libjpeg-turbo | q30 | 4:4:4 | 11027077 | 7511255 | 7504575 | 0.09% | 31.88% | 31.94%
  | libjpeg-turbo | q50 | 4:4:4 | 15787104 | 11846444 | 11809831 | 0.31% | 24.96% | 25.19%
  | libjpeg-turbo | q60 | 4:4:4 | 18540959 | 14242903 | 14199379 | 0.31% | 23.18% | 23.42%
  | libjpeg-turbo | q70 | 4:4:4 | 22798622 | 17870310 | 17817181 | 0.30% | 21.62% | 21.85%
  | libjpeg-turbo | q80 | 4:4:4 | 30277359 | 24083456 | 23988861 | 0.39% | 20.46% | 20.77%
  | libjpeg-turbo | q90 | 4:4:4 | 49169245 | 39177048 | 39045153 | 0.34% | 20.32% | 20.59%
  | libjpeg-turbo | q92 | 4:4:4 | 56139556 | 44712883 | 44574988 | 0.31% | 20.35% | 20.60%
  | libjpeg-turbo | q95 | 4:4:4 | 78390474 | 61858309 | 61709516 | 0.24% | 21.09% | 21.28%
[...]

PR #4132, opened 2025-02-27
More precise histogram

Melirius:
### Description

The remark in the normalize histogram code by @robryk suggested that rescaling targets can provide a better histogram, and this is indeed the case - many test cases show file size reduction with application of that rescaling.

Additionally `SmallestIncrement` is simplified for the invocation with `count > 0` (that is the only case used), and `target` was chosen to be float instead of int. The latter is connected to the problem in current implementation that arizes on near-flat histograms with `inc = 1` due to the fact that then `inc / 2 = 0`.

To illustrate the problem, take a histogram of 2048 values of 2 and 2048 values of 0 with `inc = 1`, then in the current realization `target = 2/0` correspondingly, and for all 2 values `target >= counts[n] + inc / 2` is true, therefore first method will provide histogram of 2047 values 3 and the first value negative, 3-2048, and only the second method produces the correct histogram.

Even in the case when the second method is not needed, the first one with `inc = 1` will systematically overincrease the relatively rare values frequency, that will lead to slightly worse compression.

In all the cases tested compression with the new normalization of histogram is better than with the original version, however, by a small margin of $(1\div4)\times10^{-3}$ %. The same reduction appears on the version of #4130.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/enc_ans.cc
+++ b/lib/jxl/enc_ans.cc
@@ -115,16 +115,18 @@ const uint8_t kLogCountSymbols[ANS_LOG_TAB_SIZE + 2] = {
 
 // Returns the difference between largest count that can be represented and is
 // smaller than "count" and smallest representable count larger than "count".
-int SmallestIncrement(uint32_t count, uint32_t shift) {
-  int bits = count == 0 ? -1 : FloorLog2Nonzero(count);
-  int drop_bits = bits - GetPopulationCountPrecision(bits, shift);
-  return drop_bits < 0 ? 1 : (1 << drop_bits);
+// Should be invoked with `count > 0` only
+uint32_t SmallestIncrement(uint32_t count, uint32_t shift) {
+  uint32_t bits = FloorLog2Nonzero(count);
+  uint32_t drop_bits = bits - GetPopulationCountPrecision(bits, shift);
+  return 1u << drop_bits;
 }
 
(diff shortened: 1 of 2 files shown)


lib/jxl/enc_ans.cc
```diff
@@ -115,24 +115,26 @@ const uint8_t kLogCountSymbols[ANS_LOG_TAB_SIZE + 2] = {
 
 // Returns the difference between largest count that can be represented and is
 // smaller than "count" and smallest representable count larger than "count".
-int SmallestIncrement(uint32_t count, uint32_t shift) {
-  int bits = count == 0 ? -1 : FloorLog2Nonzero(count);
-  int drop_bits = bits - GetPopulationCountPrecision(bits, shift);
-  return drop_bits < 0 ? 1 : (1 << drop_bits);
+// Should be invoked with `count > 0` only
```
jonsneyers, line 118: Better to make that a `JXL_DASSERT()` instead of a comment. Or rename the function to `SmallestIncrementNonzero` like `FloorLog2Nonzero`.

Melirius: Done both. BTW, do you know the reasoning for overestinmation of total_pixels value (https://github.com/libjxl/libjxl/pull/4135)?

Melirius (commented):
Done.

PR #4135, opened 2025-03-02
Draft: Fix for handling total_pixels

Melirius:
### Description

Handling of the number of pixels in `ModularEncode` functions is strange: being applied to a sequence of stream images like in `ModularFrameEncoder::ComputeTree` or to an individual image like in other cases, it overestimates the number of pixels processed due to special treatment of `*total_pixels == 0` case.

If initially `total_pixels = 0` then we have after the function `total_pixels = 2 * (number of pixels in the image)` increased by this part and `GatherTreeData` , in any other case we have `total_pixels += (number of pixels in the image)` increased by `GatherTreeData` (modulo edge cases).

It looks like that the intention was to not allow zero in the number of pixels as it crashes `LearnTree` function. However, it can be done in a more straightforward way without pixel number overestimation. 

In my testing the proposed change also leads to a slightly better compression, of the order of $6\times10^{-3}$ %. However, some of the tests are failing due to increase of the file size, so I mark this as draft before I can find out the cause.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/modular/encoding/enc_encoding.cc
+++ b/lib/jxl/modular/encoding/enc_encoding.cc
@@ -589,17 +589,6 @@ Status ModularEncode(const Image &image, const ModularOptions &options,
   TreeSamples tree_samples_storage;
   size_t total_pixels_storage = 0;
   if (!total_pixels) total_pixels = &total_pixels_storage;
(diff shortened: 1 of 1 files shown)


eustas:
Supplanted by #4209

PR #4111, opened 2025-02-14
Fix coverity defects

sudhackar:
### Description

A coverity scan is one of the steps in Main Inclusion Review [MIR](https://github.com/canonical/ubuntu-mir) needed to get any package in Ubuntu `main`. This PR fixes some trivial issues coverity identified. A private report can be sent on maintainer's request too

### Pull Request Checklist

- [x] **CLA Signed**: Have you signed the [Contributor License Agreement](https://code.google.com/legal/individual-cla-v1.0.html) (individual or corporate, as appropriate)? Only contributions from signed contributors can be accepted.
- [x] **Authors**: Have you considered adding your name to the [AUTHORS](AUTHORS) file?
- [x] **Code Style**: Have you ensured your code adheres to the project's coding style guidelines? You can use `./ci.sh lint` for automatic code formatting.

Please review the full [contributing guidelines](https://github.com/libjxl/libjxl/blob/main/CONTRIBUTING.md) for more details.

PR #4140, opened 2025-03-04
fix test that was broken by #4111

jonsneyers:
In #4111, changes to GetSomeTestImage() cause test images to change which means some of the thresholds have to be adjusted.


jonsneyers:
Already included in https://github.com/libjxl/libjxl/pull/4142

Issue #4143, opened 2025-03-06
`clamp` in alpha.cc has unclear meaning and usage


zond:
**Describe the bug**

The `clamp` parameter used by multiple functions in alpha.cc sometimes means "clamp the value", and sometimes means "don't clamp the value".

`float fga = clamp ? Clamp(fg.a[x]) : fg.a[x];`

vs

`float fa = clamp ? fga[x] : Clamp(fga[x]);`

This might or might not be a bug.

If it is a bug:
- The spec needs to be read to verify that the spec doesn't have the same bug.
- The conformance tests needs to be extended to capture it.

If it is not a bug:
- The parameter should (in the cases where it means "don't clamp the value") renamed to something that makes it more clear that this is what it does.

jonsneyers:
It was a bug, clamp always means "clamp the value". Should be fixed now.

PR #4144, opened 2025-03-06
Made the meaning of the clamp argument uniform.

zond:
See https://github.com/libjxl/libjxl/issues/4143

Tests still pass. Spec says "If patch[i].blending[j].clamp[c] is true, alpha values are clamped to the interval [0,1] before blending."


eustas:
Whoa. This code doesn't seem to be tested thorougly =(

I wonder, if `bga` needs to be clamped as well...

jonsneyers (approved):

PR #4147, opened 2025-03-09
Resampling tweaks and updated docs

jonnyawsom3:
### Description

This PR lowers the threshold for resampling to be enabled during lossy encoding, now using 2x2 resampling at distance 10 and up, with a tweaked equation to closer match the non-resampled bpp. It varies by image content, but on average ssimulacra2 scores were 10 points higher at a lower or equal bpp, with most score regressions looking better after visual inspection.

I encountered one image that was using 50% higher bpp with resampling than without, causing a jump in image quality at distance 10, but I think butteraugli was undershooting without due to the extremely low image quality otherwise. I tested 4x4 resampling at distance 25, and while promising, the results were far too blurry on anything below 4K, and results were highly inconsistent between images.

It also moves the extremely slow iterative downsampling algorithm to effort 11, speeding up the default effort 7 by 3-10x with a significant reduction in memory due to chunked encoding being re-enabled. Effort 10 should be around 2x faster from my testing too, hence choosing to move it to effort 11 since lossy did nothing different before.

Documentation has been updated and incorrect/outdated comments fixed too.

Overall, distances above 10 should have improved image quality and/or bpp, with distances above 20 having higher bpp to match no resampling and marginally better quality. Images were also tested with version 0.8 due to quality regressions since, but the relative improvements are consistent across encoder changes.

Current Distance 10 (SSIMULACRA2: 25.41)
image
New Resampled Distance 10 (SSIMULACRA2: 32.18)
image

### Pull Request Checklist

### DIFF
--- a/doc/encode_effort.md
+++ b/doc/encode_effort.md
@@ -26,18 +26,18 @@ The following table describes what the various effort settings do:
 | e6 | e5 + more RCTs and MA tree properties | e5 + error diffusion, full variable blocks heuristics |
 | e7 | e6 + more RCTs and MA tree properties | e6 + patches (including dots) |
 | e8 | e7 + more RCTs, MA tree properties, and Weighted predictor parameters | e7 + Butteraugli iterations for adaptive quantization |
-| e9 | e8 + more RCTs, MA tree properties, and Weighted predictor parameters | e8 + more Butteraugli iterations and disables chunked encoding |
-| e10 | e9 + global MA tree, try all predictors, and disables chunked encoding | e9 + more thorough adaptive quantization |
-| e11 | e10 + previous-channel MA tree properties, different group dimensions, and try multiple e10 configurations | N/A |
+| e9 | e8 + more RCTs, MA tree properties, and Weighted predictor parameters | e8 + more Butteraugli iterations |
+| e10 | e9 + global MA tree, try all predictors, and disables chunked encoding | e9 + more thorough adaptive quantization and disables chunked encoding |
+| e11 | e10 + previous-channel MA tree properties, different group dimensions, and try multiple e10 configurations | e10 + iterative downsampling for high distance values |
 
 For the entropy coding (context clustering, lz77 search, hybriduint configuration): slower/more exhaustive search as effort goes up.
 
 <u>Chunked encoding is also disabled under these circumstances:</u>
 * When the image is smaller than 2048x2048.
 * Lossless Jpeg transcoding.
-* VarDCT at distances ≥20.
 * Effort 7 VarDCT at distances ≥3.0.
 * Effort 8 VarDCT at distances >0.5.
+* Effort 11 VarDCT with --resampling 2.
 * Lossy Modular.
 * When using any of these flags:
   * `--patches=1`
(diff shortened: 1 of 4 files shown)


jonnyawsom3:
Converting to draft while we explore resolution thresholds and a more conservative distance threshold to avoid regressions.

jonnyawsom3:
Was renaming branches then realised it had closed the PR.

Firepal:
I tried this out with a very synthetic sample (basic octahedral panorama in Blender). There's an oddly desirable "sharpness"(?) with the v0.11.1 result, but it has artifacts on top of that "sharpness." I'd say PR is an improvement.

v0.11.1 -d 10 -e 10 = **34.6KiB**
octo2048_v0 11 1 jxl

PR -d 10 -e 10 = **48.1KiB**
octo2048_pr jxl

**Original PNG** for evaluation
octo2048

Might try a more organic sample.

jonnyawsom3:
The current issue is that photographic gets lower bpp, but synthetic is higher bpp, making the advantages questionable compared to using a lower distance setting.

Currently I erred on the side of caution, aiming for quality overshoot rather than undershoot, so you're more likely to get a larger file than a ruined image.

Firepal:
PR has less blockiness and better represents things like the wind turbines here. It also does away with how v0.11.1 tries to keep *some* edges sharp, like the sponsors on the car and the headlight covers. 

I'm on the fence about this... but d10 is a fairly large distance, so maybe it's fine.

v0.11.1 d10 e10 = **31.8KiB**
quattro png 0 11 1 jxl

PR d10 e10 = **33.7KiB**
quattro png pr jxl

Original
quattro

Source image: https://commons.wikimedia.org/wiki/File:Audi_Sport_Quattro_S1-E2_Sachs_Franken_Classic_2018_P5190483.jpg Reinhold Möller, CC BY-SA 4.0 via Wikimedia Commons

jonsneyers (dismissed):

jonnyawsom3:
Jyrki will want to check this thoroughly. It could do with more tuning but I'm unsure in what direction. Changing internal distance causes (more) under/overshoot depending on image, and also causes a sudden change in quality and density whatever distance threshold the resampling is enabled at.

Given the lack of tuning above distance 3 or so, I think it's probably a decent option to improve quality when people do comparisons with AVIF at low qualities.

jyrkialakuijala (dismissed):

jyrkialakuijala:
These kind of optimizations are complex compromises of decoding speed, encoding speed, quality between small sharp details vs. textures. Let's remain sensitive to the possible feedback from the field and stay open to adjusting these and other heuristics.

jonnyawsom3:
Decode speed is actually 40% faster, encode speed 30% faster (25% slower with patches disabled) and memory usage 50% lower, thanks to the image internally being half the resolution. Strangely, the current behaviour seems to favour high contrast areas.

Text, logos and sharp edges get almost perfectly preserved, while the rest of the image gets smoothed due to bitrate starving.
image
The PR more evenly distributes bits, causing even quality degradation. Metrics disagree, but visually it's like putting on glasses.
image

jonnyawsom3:
Was doing a few final checks and realised the slower downsampling method could never be used in lossy, also updated docs again to reflect it.

jonsneyers (dismissed):

jonsneyers (approved):

PR #4148, opened 2025-03-10
Remove useless asserts

Melirius:
### Description

Here the asserts are useless: two `size_t`  added cannot give you a negative number. And being applied to other cases of usage appears to be the same: `size_t` as large unsigned forces promotion of other number to unsigned too.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/base/rect.h
+++ b/lib/jxl/base/rect.h
@@ -74,31 +74,26 @@ class RectT {
 
   template <template <class> class P, typename V>
   V* Row(P<V>* image, size_t y) const {
-    JXL_DASSERT(y + y0_ >= 0);
     return image->Row(y + y0_) + x0_;
   }
 
   template <template <class> class P, typename V>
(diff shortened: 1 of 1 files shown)


eustas:
Overall, second expression (`static_cast<T>(y) + y0_ >= 0`) is more readable, since we overcome undesired "promotion" and show our desire to check that accessed row index is not negative. Additionally to this, we could add `static_assert(sizeof(T) >= sizeof(size_t), "blah-blah");` somewhere around.

Melirius:
Yeah, you're right, I exchanged types of y and y0_ in my mind somehow :). I agree that `static_cast<T>(y) + y0_ >= 0` is more expressive, maybe also ensure not going out of rectangle like `y < ysize_`?

eustas:
Extra check won't hurt, but I believe it is done inside `(Const)Row` already (see `VoidRow`).

Melirius:
JPEG reconstruction fails on this additional check, wow. Let me see why.

PR #4154, opened 2025-03-17
Fix ineffciency in quantize histogram

Melirius:
### Description

Histogram quantization function `QuantizeHistogram` has a deficiency that generates runs of false thresholds for zeros in histogram when `cumsum == threshold * sum / num_chunks`. As a result it can provide `thresholds` vectors much longer than supposed `num_chunks` , especially if histogram has a long zero-tail (can be verified by putting assert there on tests). Such vectors mess up with `property_mapping` providing not only much more buckets than intended, but also wrapping them out as `size_t mapped` is saved into `uint8_t`.

This PR fixes the problem. Some sizes in the tests get up as the number of buckets decreases to prescribed values.

The fix uncovered also a problem in the second `RebalanceHistogram` approach, that can fail on nearly-flat histograms with many sub-1 entries in targets (can be verified by `QuantWeightsTest.RAW` excluding changes in [lib/jxl/enc_ans.cc](https://github.com/libjxl/libjxl/compare/main...Melirius:Fix-ineffciency-in-QuantizeHistogram?expand=1#diff-eae1dcaff25053843308883e33aff691cb95c69e78af1909f1695a1cdd73d4fa)). Allowing also rounding down from targets, this case can be covered.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/enc_ans.cc
+++ b/lib/jxl/enc_ans.cc
@@ -160,6 +160,9 @@ bool RebalanceHistogram(const float* targets, int max_symbol,
       if (target >= count + 0.5f * inc && count + inc < table_size) {
         count += inc;
       }
+      if (target < count - 0.5f * inc && count - inc > 0) {
+        count -= inc;
+      }
       sum += count;
       counts[n] = count;
       const int count_log = FloorLog2Nonzero(count);
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -925,7 +925,7 @@ TEST(JxlTest, RoundtripAlphaPremultiplied) {
 
   std::vector<uint8_t> compressed;
   EXPECT_TRUE(test::EncodeFile(cparams, io.get(), &compressed));
-  EXPECT_LE(compressed.size(), 18000u);
+  EXPECT_LE(compressed.size(), 18313u);
 
   for (bool use_image_callback : {false, true}) {
     for (bool unpremul_alpha : {false, true}) {
@@ -1081,7 +1081,7 @@ JXL_SLOW_TEST(JxlTest, RoundtripLossless8) {
   dparams.accepted_formats.push_back(t.ppf().frames[0].color.format);
 
   PackedPixelFile ppf_out;
-  EXPECT_EQ(Roundtrip(t.ppf(), cparams, dparams, pool.get(), &ppf_out), 223044);
+  EXPECT_EQ(Roundtrip(t.ppf(), cparams, dparams, pool.get(), &ppf_out), 223047);
   EXPECT_EQ(ComputeDistance2(t.ppf(), ppf_out), 0.0);
 }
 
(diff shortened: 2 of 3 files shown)


jonsneyers:
There are merge conflicts that have to be resolved.

Melirius:
Final version to rerun CI.

Melirius:
There are three cases of test file size mismatch in CI: [i386 shard 1](https://github.com/libjxl/libjxl/actions/runs/14463843272/job/40563111743?pr=4154#step:9:1049) and two fails on low precision in JxlTest.RoundtripAnimationPatches: [Ubuntu Build release:arm64-lowprecision](https://github.com/libjxl/libjxl/actions/runs/14463843260/job/40561546228?pr=4154#step:18:16723) and [Ubuntu Build lowprecision](https://github.com/libjxl/libjxl/actions/runs/14463843260/job/40561543608?pr=4154#step:18:15288). I suspect different floating-point results. Will it worth the effort to dig further or just increase the tolerance of these 2 tests?

jonsneyers:
> There are three cases of test file size mismatch in CI: [i386 shard 1](https://github.com/libjxl/libjxl/actions/runs/14463843272/job/40563111743?pr=4154#step:9:1049) and two fails on low precision in JxlTest.RoundtripAnimationPatches: [Ubuntu Build release:arm64-lowprecision](https://github.com/libjxl/libjxl/actions/runs/14463843260/job/40561546228?pr=4154#step:18:16723) and [Ubuntu Build lowprecision](https://github.com/libjxl/libjxl/actions/runs/14463843260/job/40561543608?pr=4154#step:18:15288). I suspect different floating-point results. Will it worth the effort to dig further or just increase the tolerance of these 2 tests?

Just increase the tolerance.

Melirius:
Done.

jonsneyers:
Can this be done in a way that does not cause file sizes to go up?

Melirius:
> Can this be done in a way that does not cause file sizes to go up?

Unfortunately, not without further modifications. The effctively used number of buckets is reduced sometimes (as thresholds are defined on only 10% of samples, there is a good chance that we have pixels going into this produced "thresholds tail") and it increases the size. To fight it we need to increase entropy of buckets back, either by upping their number in the effort parameters, or maybe by some another scheme of threshold production. I'm investigating the possibilities now.

lib/jxl/pack_signed.h
```diff
@@ -29,7 +29,8 @@ constexpr uint32_t PackSigned(int32_t value)
 // convenience its type is `size_t` (i.e. `uint32_t` or `uint64_t`).
 constexpr int32_t UnpackSigned(size_t value)
     JXL_NO_SANITIZE("unsigned-integer-overflow") {
-  JXL_DASSERT((value & 0xFFFFFFFF) == value);  // no-op in 32-bit build
+  // TODO(Ivan): fails in C++11 mode, restore with a guard?
```
jonsneyers, line 32: When does this fail? This function is not supposed to be called with an input that exceeds uint32...

Melirius: Yes, it fails to compile in C++11 mode on Clang and GCC insisting that `__builtin_trap()` is not `constexpr`.

Melirius:
```
[build] In file included from /home/ivan/libjxl/lib/jxl/entropy_coder.cc:21:
[build] /home/ivan/libjxl/lib/jxl/pack_signed.h: In function ‘constexpr int32_t jxl::UnpackSigned(size_t)’:
[build] /home/ivan/libjxl/lib/jxl/pack_signed.h:34:1: error: body of ‘constexpr’ function ‘constexpr int32_t jxl::UnpackSigned(size_t)’ not a return-statement
[build]    34 | }
[build]       | ^
```

jonsneyers: Oh, I see.

jonsneyers (approved):

PR #4161, opened 2025-03-24
Introduce the new jxltran tool

LucasChollet:
### Description

This tool, as described in #871 will perform operations on existing jxl files, like changing the orientation or packing a codestream in a container.

---

I aimed small for this PR, but I plan to add more features to the tool. It will probably not be every single one listed in #871, but at least a few.

I tried to be the more idiomatic as possible to the `libjxl` style, but I probably did a few things wrong. I will happily make any requested change.

Oh, and I didn't add a test, but I can surely add one with some hints on how to do it.

### Pull Request Checklist

### DIFF
--- a/tools/jxltran.cc
+++ b/tools/jxltran.cc
@@ -1,0 +1,82 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include <string>
+#include <vector>
+
+#include "lib/include/jxl/decode.h"
+#include "tools/cmdline.h"
+#include "tools/file_io.h"
+
+namespace jpegxl {
+namespace tools {
+namespace {
+
+struct Args {
+  void AddCommandLineOptions(CommandLineParser* cmdline) {
+    cmdline->AddPositionalOption("INPUT", /* required = */ true,
+                                 "The JPEG XL input file.", &file_in);
+
+    cmdline->AddPositionalOption("OUTPUT", /* required = */ true,
+                                 "The JPEG XL output file.", &file_out);
+  }
+
+  const char* file_in = nullptr;
+  const char* file_out = nullptr;
+};
+
+}  // namespace
+
+int JxlTranMain(int argc, const char* argv[]) {
+  Args args;
+  CommandLineParser cmdline;
+  args.AddCommandLineOptions(&cmdline);
+
+  if (!cmdline.Parse(argc, const_cast<const char**>(argv))) {
+    // Parse already printed the actual error cause.
+    fprintf(stderr, "Use '%s -h' for more information.\n", argv[0]);
+    return EXIT_FAILURE;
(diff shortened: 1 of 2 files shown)


tools/jxltran.cc
```diff
@@ -0,0 +1,171 @@
...
+      fprintf(stderr, "Error, already provided all input\n");
+      return JXL_DEC_ERROR;
+    } else if (status == JXL_DEC_BOX) {
+      JxlBoxType type;
+      status = JxlDecoderGetBoxType(dec.get(), type, /*decompressed=*/ false);
+      if (status != JXL_DEC_SUCCESS) {
+        fprintf(stderr, "Error, failed to get box type\n");
+        return JXL_DEC_ERROR;
+      }
+      if (!memcmp(type, "jxlc", 4)) {
```
jonsneyers, line 80: There is also the possibility of `jxlp` boxes, which contain part of the codestream and their contents has to be concatenated to get the full codestream. These boxes start with a uint32 index which has the msb set to 1 if it's the last `jxlp` box; the other 31 bits are a counter that is supposed to increment by one, in the same order as the order in which the boxes appear (you can just ignore it).

tools/jxltran.cc
```diff
@@ -0,0 +1,171 @@
...
+                           &extract, &SetBooleanTrue);
+  }
+
+  const char* file_in = nullptr;
+  const char* file_out = nullptr;
+
+  bool extract = false;
+};
+
+JxlDecoderStatus extract_from_jxlc_box(
```
jonsneyers, line 40: It doesn't matter yet at this point since there is only one action anyway, but probably we shouldn't have one function per action and a set of mutually exclusive actions, but rather one big function that parses input and writes output, with some struct of options that describes what has to be done, i.e. what has to be copied and what has to be stripped at the file format level (and I suppose later we could also have modification operations that change stuff in the image or frame headers, for codestream level stuff like lossless cropping or changing the colorspace tag of an XYB image).

tools/jxltran.cc
LucasChollet: Do you want me to do it right now? I'm usually working by doing the "smallest step possible", and support for `jxlp` boxes was planned as the very next PR.
And if your answer is yes, should I add a new commit or fixup the existing one?

tools/jxltran.cc
LucasChollet: Sure, I should be able to do that today.

tools/jxltran.cc
jonsneyers: Can do it in a follow-up PR or in another commit in this one (they will be squashed when merging), both are fine.

LucasChollet: Will do in a follow-up then.

tools/jxltran.cc
LucasChollet: Done!

jonsneyers (approved):

LucasChollet:
Thanks!

PR #4168, opened 2025-03-29
tools/jxltran: warn if extracting raw codestream that is not level 5

Traneptora:
If the --extract option is used, the codestream is extracted without the container. However, raw JPEG XL codestreams cannot be higher than level 5, so we produce a warning if a jxll box was read first that indicates a higher level codestream is present.

### DIFF
--- a/tools/jxltran.cc
+++ b/tools/jxltran.cc
@@ -72,6 +72,16 @@ JxlDecoderStatus apply_file_format_options(
 
     JxlBoxType type;
 
+    uint32_t jxlc, jxll, jxlp;
+    // these aren't hardcoded because
+    // they're in native endian order
+    memcpy(&jxlc, "jxlc", sizeof(jxlc));
+    memcpy(&jxll, "jxll", sizeof(jxll));
+    memcpy(&jxlp, "jxlp", sizeof(jxlp));
+
+    uint32_t box = 0;
+    uint8_t level = 5;
+
     for (;;) {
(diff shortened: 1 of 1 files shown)


LucasChollet (commented):
Do we want to only show a warning and allow them? It means that an official tool will create non-conformant files.

jonnyawsom3:
IIRC libjxl doesn't currently check levels, which allows things within spec but not yet associated to a level, such as using over 256 extra channels

Traneptora:
> Do we want to only show a warning and allow them? It means that an official tool will create non-conformant files.

Perhaps? Better than silently doing the wrong thing with no warning. Maybe guard it behind a `--noncompliant` flag.

LucasChollet:
> Perhaps? Better than silently doing the wrong thing with no warning.

Sorry, I haven't been clear. I wanted to suggest that we should simply refuse to create them rather than emitting a warning. Obviously a warning is better than nothing, I like where this is going.

jonsneyers (dismissed):

Traneptora:
I have rebased main onto this PR, so it can be re-reviewed.

jonsneyers (approved):

Traneptora:
It looks like CI was failing because of `-Werror=maybe-uninitialized`. gcc doesn't know that `JXL_DEC_BOX` fires before `JXL_DEC_BOX_COMPLETE` so I initialized it to 0 and forcepushed a change.

jonsneyers (approved):

Issue #4187, opened 2025-04-13
Failure in benchmark


Melirius:
**Describe the bug**
Running on `main` among other output I got error that NC013_Evening_AdobeRGB_16bit.ppm is failed to decode.

**To Reproduce**
Steps to reproduce the behavior: run `./ci.sh benchmark` 

**Expected behavior**
No error.

**Screenshots**
Relevant log part:
```
./lib/extras/dec/color_description.cc:106: JXL_FAILURE: Invalid double: Ado
./lib/extras/dec/color_description.cc:165: JXL_RETURN_IF_ERROR code=1: ParseDouble(&xy_tokenizer, c->primaries_red_xy + 0)
./lib/extras/dec/color_description.cc:240: JXL_RETURN_IF_ERROR code=1: ParsePrimaries(&tokenizer, c)
./lib/extras/dec/color_hints.cc:37: JXL_FAILURE: Failed to apply color_space
./lib/extras/dec/color_hints.h:44: JXL_FAILURE: ColorHints::Foreach returned false
[...]

PR #4199, opened 2025-04-16
Adobe98

jonsneyers:
Should fix https://github.com/libjxl/libjxl/issues/4187

Adds `Ado` keywords in color description strings, as well as a shorthand "`Adobe98`".

Flyby: simplify the code for the shorthands, tweak jxlinfo layout for custom primaries.

### DIFF
--- a/lib/extras/dec/color_description.cc
+++ b/lib/extras/dec/color_description.cc
@@ -224,9 +224,9 @@ Status ParseTransferFunction(Tokenizer* tokenizer, JxlColorEncoding* c) {
 Status ParseDescription(const std::string& description, JxlColorEncoding* c) {
   *c = {};
   if (description == "sRGB") {
-    return ParseDescription("RGB_D65_SRG_Per_SRG", c);
+    return ParseDescription("RGB_D65_SRG_Rel_SRG", c);
   } else if (description == "DisplayP3") {
-    return ParseDescription("RGB_D65_DCI_Per_SRG", c);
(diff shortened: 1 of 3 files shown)


lib/extras/dec/color_description.cc
```diff
@@ -210,29 +224,15 @@ Status ParseTransferFunction(Tokenizer* tokenizer, JxlColorEncoding* c) {
...
-    c->rendering_intent = JXL_RENDERING_INTENT_PERCEPTUAL;
+    return ParseDescription("RGB_D65_SRG_Per_SRG", c);
   } else if (description == "DisplayP3") {
-    c->color_space = JXL_COLOR_SPACE_RGB;
-    c->white_point = JXL_WHITE_POINT_D65;
-    c->primaries = JXL_PRIMARIES_P3;
-    c->transfer_function = JXL_TRANSFER_FUNCTION_SRGB;
-    c->rendering_intent = JXL_RENDERING_INTENT_PERCEPTUAL;
+    return ParseDescription("RGB_D65_DCI_Per_SRG", c);
+  } else if (description == "Adobe98") {
```
sboukortt, line 230: Do you remember why we chose “perceptual” for these two and “relative” for the others? Maybe “relative” for all would make sense?

sboukortt (approved):

lib/extras/dec/color_description.cc
jonsneyers: No, I don't remember. Relative for all is fine with me. Updating.

sboukortt (approved):

PR #4202, opened 2025-04-18
tweaks to jpeg recompression

jonsneyers:
Similar to https://github.com/libjxl/libjxl/pull/4130

Changes the context model for JPEG recompression a bit.
Your mileage may vary: for some images this improves compression, for others it makes things worse.

On average, for the images on imagecompression.info encoded with libjpeg-turbo, compression improves by 0.3-0.4% (for both 4:4:4 and 4:2:0).
For the subset1 images encoded with libjpeg-turbo, on average compression improves by 0.1%.

For some random jpeg photos from my phone, some improve by 0.6%, but some get worse by up to 0.4%.


lib/jxl/enc_frame.cc
```diff
@@ -978,8 +975,10 @@ Status ComputeJPEGTranscodingData(const jpeg::JPEGData& jpeg_data,
           } else {
             idc = inputjpeg[base] + 1024 / qt_dc[c];
           }
-          dc_counts[c][std::min(static_cast<uint32_t>(idc + 1024),
-                                static_cast<uint32_t>(2047))]++;
+          if (c == 1) {
+            dc_counts[std::min(static_cast<uint32_t>(idc + 1024),
```
eustas, line 979: While on this: `std::min<uint32_t>` looks neater than double `static_cast`.

lib/jxl/enc_frame.cc
```diff
@@ -1023,41 +1022,51 @@ Status ComputeJPEGTranscodingData(const jpeg::JPEGData& jpeg_data,
...
-        cut = total_dc[i] * (dct[i].size() + 1) / (num_thresholds + 1);
-      }
+  }
+  // use more contexts for larger and higher quality images
+  int num_thresholds = CeilLog2Nonzero(total_dc[1]) -
+                       CeilLog2Nonzero(static_cast<unsigned>(
+                           qt[1] + qt[2] + qt[3] + qt[4] + qt[5])) -
+                       7;
+  // up to 8 buckets, based on luma only
+  num_thresholds = std::min(std::max(num_thresholds, 1), 7);
```
eustas, line 1035: Perhaps `jxl::Clamp1` is more readable

lib/jxl/enc_frame.cc
jonsneyers: updated

lib/jxl/enc_frame.cc
jonsneyers: updated

lib/jxl/jxl_test.cc
```diff
@@ -1475,7 +1475,7 @@ JXL_TRANSCODE_JPEG_TEST(JxlTest, RoundtripJpegRecompression444) {
   const std::vector<uint8_t> orig =
       ReadTestData("jxl/flower/flower.png.im_q85_444.jpg");
   // JPEG size is 696,659 bytes.
-  EXPECT_NEAR(RoundtripJpeg(orig, pool.get()), 568822u, 20);
+  EXPECT_NEAR(RoundtripJpeg(orig, pool.get()), 568879u, 20);
```
eustas, line 1478: Just a thought: 0.0035% threshold seems to be too tight. How about making it something like 0.05%

eustas (approved):

lib/jxl/jxl_test.cc
jonsneyers: yes, next time a test with such a small threshold fails, I'll fix it by bumping up the threshold instead...

PR #4210, opened 2025-04-25
output in DisplayP3, add option for producing ultrahdr_app input

jonsneyers:
Two changes to the `local_tone_map` tool:

1. Instead of producing SDR output in Rec2020 with the sRGB tf (which is too wide gamut for 8-bit formats, kind of defeating much the purpose of the tool), make the output DisplayP3 instead.
2. Add an optional argument to dump the HDR image in a format suitable for the libultrahdr example app `ultrahdr_app` (see https://github.com/google/libultrahdr). If a third filename is given on the command line, the input image will be converted to the P3 PQ space and stored as a raw little-endian rgba1010102 file which can be used as input for that tool. An example command line will also be provided, e.g. like this:

```
$ local_tone_map alps-sunset.png out_sdr.png out_hdr.raw
cjpegli out_sdr.png input_sdr.jpg
ultrahdr_app -m 0 -a 5 -t 2 -C 1 -w 840 -h 944 -c 1 -R 1 -s 1 -Q 95 -M 1 -L 2000 -p out_hdr.raw -i input_sdr.jpg -z output_ultrahdr.jpg
```

This makes it a little easier to produce these jpegs with gain maps, since `ultrahdr_app` does not exactly have the most user friendly interface...

Input file `alps-sunset.png`:
alps-sunset

Resulting output file `output_ultrahdr.jpg`:
output_ultrahdr

### DIFF
--- a/tools/hdr/local_tone_map.cc
+++ b/tools/hdr/local_tone_map.cc
@@ -19,6 +19,7 @@
 #include "lib/jxl/base/common.h"
 #include "lib/jxl/base/compiler_specific.h"
 #include "lib/jxl/base/data_parallel.h"
+#include "lib/jxl/base/printf_macros.h"
 #include "lib/jxl/base/rect.h"
 #include "lib/jxl/base/span.h"
 #include "lib/jxl/base/status.h"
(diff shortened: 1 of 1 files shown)


tools/hdr/local_tone_map.cc
```diff
@@ -284,12 +285,14 @@ Status ProcessFrame(CodecInOut* image, float preserve_saturation,
...
-  JXL_RETURN_IF_ERROR(rec2020_srgb.CreateICC());
+  ColorEncoding displayP3;
+  JXL_RETURN_IF_ERROR(displayP3.SetWhitePointType(WhitePoint::kD65));
+  JXL_RETURN_IF_ERROR(displayP3.SetPrimariesType(Primaries::kP3));
+  displayP3.Tf().SetTransferFunction(TransferFunction::kSRGB);
+  JXL_RETURN_IF_ERROR(displayP3.CreateICC());
   JXL_RETURN_IF_ERROR(
-      image->Main().TransformTo(rec2020_srgb, *JxlGetDefaultCms(), pool));
-  image->metadata.m.color_encoding = rec2020_srgb;
+      image->Main().TransformTo(displayP3, *JxlGetDefaultCms(), pool));
```
sboukortt, line 294: In a sense, there are now two successive gamut reductions (tone mapping result ->(`GamutMap(preserve_saturation)`)->Rec. 2020->(`TransformTo(displayP3)`)->P3), but maybe that’s not too much of a problem.

(This `TransformTo` was previously really just for the sRGB TRC.)

sboukortt (approved):
Neat, thanks 😁

PR #4220, opened 2025-04-29
try reducing workflow permissions from write to read

jonsneyers:
Not sure if these two workflows actually need `contents: write` permissions to do their thing.
If they do, then the permission should be at the lowest level where it is needed, not at the top-level like it is now.

See also: 
https://securityscorecards.dev/viewer/?uri=github.com/libjxl/libjxl
https://github.com/ossf/scorecard/blob/cd152cb6742c5b8f2f3d2b5193b41d9c50905198/docs/checks.md#token-permissions

### DIFF
--- a/.github/workflows/gitlab_mirror.yml
+++ b/.github/workflows/gitlab_mirror.yml
@@ -3,7 +3,7 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-# Workflow for building and running tests.
+# Workflow for mirroring the github libjxl repo to the wg1 gitlab repo
 
 name: Mirror to GitLab
 
@@ -21,8 +21,6 @@ permissions:
 
 jobs:
   mirror:
-    permissions:
-      contents: write  # for Git to git push
     if: github.repository_owner == 'libjxl'
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


eustas:
This PR reverts #3079 that partially reverts #2485.

I'll check if "contents" could depend on event type. If not, we would need to refactor workflow into parts: one for CI, another for actual release.

jonsneyers:
To make the scorecard thing happy, it probably suffices to put the write permissions at the job level (for the jobs that need it) instead of at the global level.

For the gitlab mirroring one, I don't quite understand why write permissions are needed at all — it writes to gitlab but not to github, so from the point of view of the github workflow it is read-only, right?

eustas:
The comment was:
> The release workflow stopped uploading the artifacts after https://github.com/libjxl/libjxl/pull/2485, this should fix it.

So it is about `Upload binaries to release` job. BTW, would be nice to get rid of `AButler/upload-release-assets` dependency in favor of `actions/upload-release-asset`.

eustas:
`actions/upload-release-asset` is not a drop-in replacement.

Let's do migration in separate PR. Since we are not releasing today we could land as is and I will fix it a bit later.

eustas (approved):

PR #4201, opened 2025-04-18
Rebalanced Faster Decoding and fixed Progressive for Lossless

jonnyawsom3:
### Description

This PR changes the encoding parameters for Lossless when `--faster_decoding` is specified, resulting in up to 80% smaller files and 25% faster decoding.

Progressive Lossless is also improved. Up to 40% smaller files and 20% faster decoding, with an additional 800% encode speed increase, thanks to enabling local MA trees.

These results change depending on encoding parameters, image content, CPU and decoder.
Only libjxl and jxl-oxide were tested, due to jxlatte being too unstable. Oxide proved to be considerably faster for Progressive Lossless and our new level 4 Faster Decoding, suggesting libjxl could be improved by examining it.

Tested on a Ryzen 5600H with a 10 MP image at effort 7.
BPP is now much more inline with the base effort 7 encode, following a logical progression of less density as you increase level.
| BPP | Main | PR |  Δ (%) |
|------|-------|----|--------|
| Default Lossless | 3.458  | 3.457  | N/A |
| Faster Decoding 1 | 15.760 | 3.497 | -77.81 |
| Faster Decoding 2 | 15.760 | 3.578 | -77.30 |
| Faster Decoding 3 | 5.755 | 4.172 | -27.51 |
| Faster Decoding 4 | 5.755  | 4.478 | -22.19 |
| Progressive | 7.948 | 5.047 | -36.50 |

Levels 1-4 now make sense for decoding speed as well, with 3 and 4 now faster than effort 1. Progressive also got a boost.
| Decode (MP/s) | Main | PR |  Δ (%) |
|------|-------|----|--------|
| Default Lossless | 43.01 | 43.01 | N/A |
| Faster Decoding 1 | 134.56 | 76.60  | -43.07 |
| Faster Decoding 2 | 134.56  | 84.31 | -37.34 |
| Faster Decoding 3 | 112.98 | 142.22 | +5.69 |
| Faster Decoding 4 | 112.98 | 169.22 | +25.76  |
| Progressive | 35.41 | 41.72 | +21.24 |

A bonus major speed up for progressive lossless encoding too, using Local MA trees but still buffering the full image.
| Encode (MP/s) | Main | PR |  Δ (%) |
|------|-------|----|--------|
| Progressive | 0.65 | 4.81 | +740 |

### Pull Request Checklist

[...]

PR #4232, opened 2025-05-06
Histogram retune

Galaxy4594:
A follow up of this PR #4154. It generally results in denser images. Fixes the regression I found [here](https://github.com/libjxl/libjxl/pull/4201#issuecomment-2849934762). Images are denser by about 0.1-0.5%. Faster decoding levels 1-3 benefit quite a bit from this change. I've seen progressive lossless be slightly less dense (<0.05%) as a trade-off.

Here are some of my results on 5828 x 4371 photo. All faster decoding & progressive runs are done with effort 7.
Settings|Main (BPP)|PR (BPP)|Δ (%)|
|--------|---------|--------|--------|
|Effort 3|7.867|7.867|=|
|Effort 5|7.829|7.795|-0.434|
|Effort 7|7.635|7.616|-0.249|
|Effort 9|7.589|7.589|=|
|Fd1|7.693|7.673|-0.260|
|Fd2|7.742|7.729|-0.168|
|Fd3|8.387|8.137|**-2.981**|
|Fd4|8.895|8.895|=|
|Progressive|8.058|8.058|=|


eustas (dismissed):

jonsneyers:
Doesn't this basically revert #4154 and reintroduce the potential runs of false thresholds in case `cumsum == threshold * sum / num_chunks` ? That was an unintentional possibility that would lead to better density (since more than `num_chunks` buckets were used in this case) at the cost of speed and memory.

I may be mistaken but I think the better approach here would be to bump up `max_property_values` in the code starting here: https://github.com/libjxl/libjxl/blob/67e5f5559f947a978319a1bdb6d5534e0afda098/lib/jxl/enc_modular.cc#L564

Pinging @Melirius and @veluca93 to check if my understanding is correct.

Galaxy4594:
Bumping up the max property values didn't really decrease file size as much as I hoped (and sometimes even made files bigger!), so I'm not sure if that is the answer to the regression I see. But I wondered if I could use these "unintentional false thresholds" to increase density with minimal speed and memory penalties, albeit even with the old version it's still very minor in the grand scheme of things. So in this new commit, I plugged in a new version that does something like that.

Now I'm not proficient C++ at all, so I used AI to assist me with my idea. So far this code compiles and images round-trip just fine on my system. But in advance, I'm sorry if I didn't catch any glaring issues with the code. Here is ChatGPT's rationale:
| Criterion | Pre-4154 | #4154 (present in main) | This PR |
| -- | -- | -- | -- |
Input safety | no num_chunks == 0 guard; no sum==0 | guarded against both | guarded against both|
Zero‐sum histograms | explodes into false thresholds | returns empty | returns empty|
Per‐bin cost | division + multiplication per check | multiplication + comparison | multiplication + comparison|
Exact quantiles | sometimes sprays many thresholds on ties | sprays until sum strictly exceeds | emits one threshold per quantile only|
Number of thresholds | up to histogram.size() | up to num_chunks (then pop_back) | exactly num_chunks−1|
Loop bounds | i + 1 < size | i < size then pop the extra | i < size && next_quantile < num_chunks (early out)|

Extended from my previous tests, it's still an improvement over main. Just lost a bit of density compared to the old version.
Settings|Main (BPP)|Old-PR (BPP)|New-PR (BPP)|
|--------|---------|--------|--------|
|Effort 5|7.829|7.795|7.822|
|Effort 7|7.635|7.616|7.629|
|Fd1|7.693|7.673|7.687|
|Fd2|7.742|7.729|7.738|
|Fd3|8.387|8.137|8.195|

jonsneyers:
Could you give this a try: https://github.com/libjxl/libjxl/pull/4236  and see how it does on your test image/corpus?

eustas:
```patch
diff --git a/lib/jxl/jxl_test.cc b/lib/jxl/jxl_test.cc
index e5923b24..84e8f1b0 100644
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -1085,7 +1085,7 @@ JXL_SLOW_TEST(JxlTest, RoundtripLossless8) {
   PackedPixelFile ppf_out;
   size_t compressed_size =
       Roundtrip(t.ppf(), cparams, dparams, pool.get(), &ppf_out);
-  EXPECT_EQ(compressed_size, 218429u);
+  EXPECT_NEAR(compressed_size, 217443u, 200);
   EXPECT_EQ(ComputeDistance2(t.ppf(), ppf_out), 0.0);
 }
 
@@ -1165,7 +1165,7 @@ TEST(JxlTest, RoundtripLossless8Alpha) {
 
   PackedPixelFile ppf_out;
   size_t compressed_size = Roundtrip(t.ppf(), cparams, dparams, pool, &ppf_out);
-  EXPECT_EQ(compressed_size, 246225u);
+  EXPECT_NEAR(compressed_size, 245224u, 200);
   EXPECT_EQ(ComputeDistance2(t.ppf(), ppf_out), 0.0);
   EXPECT_EQ(ppf_out.info.alpha_bits, 8u);
   EXPECT_TRUE(test::SameAlpha(t.ppf(), ppf_out));
@@ -1341,7 +1341,8 @@ TEST(JxlTest, RoundtripLossless8Gray) {
   dparams.accepted_formats.push_back(t.ppf().frames[0].color.format);
 
   PackedPixelFile ppf_out;
-  EXPECT_EQ(Roundtrip(t.ppf(), cparams, dparams, pool, &ppf_out), 92588u);
+  size_t compressed_size = Roundtrip(t.ppf(), cparams, dparams, pool, &ppf_out);
+  EXPECT_NEAR(compressed_size, 92504u, 100);
   EXPECT_EQ(ComputeDistance2(t.ppf(), ppf_out), 0.0);
   EXPECT_EQ(ppf_out.color_encoding.color_space, JXL_COLOR_SPACE_GRAY);
   EXPECT_EQ(ppf_out.info.bits_per_sample, 8u);
```

to re-baseline tests.

Galaxy4594:
Corrected the tests. I'll take another look at the code and redo my benchmarks to see if there is still a significant improvement.

eustas (dismissed):

Galaxy4594:
Reverted to the old algorithm again. The potential runs of false thresholds actually helps in compressing high-res images by quite a bit. It's interesting how a less correct algorithm produces smaller files.

However smaller images (less than 512x512) sometimes suffer a small density loss. Performance and memory used is about the same in my testing. 

Ideally it would be better to have an adaptive approach for finding the optimal histogram. Maybe it could switch algorithms based on resolution/image content?

Melirius:
@Galaxy4594 you can try splitting by Knuth algo from https://github.com/Melirius/libjxl/blob/a197e3845c32bca887af2306a933a95e426fe745/lib/jxl/enc_Knuth_partition.h. It should give the correct optimal splitting for given histogram, but it is O(n^2). Used in my optimizer exactly for that purpose: to bucket DC values for optimization optimally.

Galaxy4594:
O(n^2) might be too slow for effort 7, maybe effort 8 instead? However I think it would be a good idea to implement this idea in the `jxltran` tool here. It would be cool to losslessly recompress existing JXLs with better entropy encoding and palette sorting similar to `jpegtran` or `oxipng`.

Issue #4234, opened 2025-05-06
libjxl image cannot be decoded with jxl-oxide, `modular_16bit_buffers` seems to be incorrectly set


printfn:
**Describe the bug**
I wrote a small program that encodes a JXL image using libjxl (via [CImg](https://www.cimg.eu/)). While libjxl can successfully decode the image, jxl-oxide cannot, and it appears that the image is not encoded correctly.

Original jxl-oxide issue: https://github.com/tirr-c/jxl-oxide/issues/456

**To Reproduce**
I've made a minimal reproduction here: https://github.com/printfn/repro/tree/cimg-libjxl-spheres

1. (Optional) Run `make` and `./repro` to run the program, which encodes the `spheres.jxl` file (depending on your system you may need to adjust paths in the `Makefile`)
2. Decoding [`spheres.jxl`](https://github.com/printfn/repro/blob/cimg-libjxl-spheres/spheres.jxl) succeeds with `djxl`, but fails with `jxl-oxide`

The initial investigation at https://github.com/tirr-c/jxl-oxide/issues/456 seems to indicate that libjxl is incorrectly setting `modular_16bit_buffers` despite samples being outside the 16-bit range.

**Expected behavior**
It looks like there might be two issues:
* libjxl should have encoded the image such that it can be decoded by jxl-oxide
* libjxl should have refused to decode the image

**Screenshots**
N/A

**Environment**
 - OS: macOS
 - Compiler version: Apple clang version 17.0.0
 - CPU type: aarch64
 - cjxl/djxl version string: cjxl v0.11.1 0.11.1 [NEON]

I have also reproduced this on aarch64 Linux.

**Additional context**
The code in CImg.h is a bit strange in that it tries to encode losslessly by setting the distance to zero, but doesn't call `JxlEncoderSetFrameLossless`. But reading the documentation I don't think calling `JxlEncoderSetFrameLossless` is required; the bug seems to be unrelated to whether the image is lossless or not.

jonsneyers:
I guess this was caused by this commit: https://github.com/libjxl/libjxl/commit/117342ab6afaa94366a77b785fe6754a3aed09ac

That code should be made more careful, since with these XYB quant factors, modular_16bit_buffers is not guaranteed while the code is assuming it is (even assuming XYB only requires 12-bit, while in reality with these factors it can require up to uint16, at least for the Y component...)

[...]

PR #4238, opened 2025-05-08
ensure we stay within Level 5, also avoid "lossy d0"

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/4234

After all, it was not modular mode + XYB that caused this (as I first thought), but bad API usage that ended up doing VarDCT at d0. The current code would translate that to d0.001, but that distance is too low: the DC quantization factors will be huge and the result is that the DC goes outside the `int16` range, which means it does not conform to Level 5 and we are producing technically invalid bitstreams since we don't set `modular_16bit_buffers` to false as we should.

While we could in principle go to arbitrarily low distances and just go outside Level 5, there is no real practical use for that so for now let's ignore that option.
This sets the minimum distance to 0.05 (instead of 0.001), which should be a safe limit to stay within Level 5.

Additionally, when setting the distance to 0 in the API, JxlEncoderSetFrameLossless will be called, which will result in JXL_API_ERROR in the example of https://github.com/libjxl/libjxl/issues/4234 (which it is; when doing lossless you cannot use XYB). That should help catch these cases of bad API usage where people are accidentally doing lossy when they really want to do lossless.

### DIFF
--- a/lib/jxl/enc_cache.cc
+++ b/lib/jxl/enc_cache.cc
@@ -135,7 +135,7 @@ Status InitializePassesEncoder(const FrameHeader& frame_header,
           std::max(static_cast<int>(SpeedTier::kTortoise),
                    static_cast<int>(cparams.speed_tier) - 1));
       cparams.butteraugli_distance =
-          std::max(kMinButteraugliDistance,
+          std::max(kMinButteraugliDistance * 0.02f,
                    enc_state->cparams.butteraugli_distance * 0.02f);
     } else {
       cparams.max_error_mode = true;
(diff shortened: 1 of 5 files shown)


veluca93 (dismissed):

eustas (dismissed):

eustas:
```
[ RUN      ] JxlTest.RoundtripDots
ReadTestData D:/a/libjxl/libjxl/testdata/external/wesaturate/500px/cvo9xd_keong_macan_srgb8.png
Test data external/wesaturate/500px/cvo9xd_keong_macan_srgb8.png is 368318 bytes long.
lib\jxl\jxl_test.cc:1270: Failure
Expected: (_actual) <= (_expected), actual: 0.16014000773429871 vs 0.16
```
Under MSYS2/clang64

eustas (approved):

PR #4242, opened 2025-05-09
Fixed spline offset.

zond:
Subtracted row x0 from pixel to render to avoid rendering outside the row buffer.

Bug found in Rust translation jxl-rs, see https://github.com/libjxl/jxl-rs/pull/158.

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -79,7 +79,7 @@ float ContinuousIDCT(const Dct32& dct, const float t) {
 
 template <typename DF>
 void DrawSegment(DF df, const SplineSegment& segment, const bool add,
-                 const size_t y, const size_t x, float* JXL_RESTRICT rows[3]) {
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

veluca93 (approved):

eustas:
`rows[c] + x - row_x0 = {row_x - row_x0, row_y - row_x0, row_b - row_x0}[c] + x - row_x0 = {row_x, row_y, row_b}[c] + x - 2 * row_x0`
This duplicate deduction of `row_x0` does not sound right to me. Going to dig further. (Is codepath with nonzero row_x0 covered by tests?)

eustas:
In rust version `row_x0` is not subtracted from row address in `draw_segments` that is why it should be subtracted later. But here - I have some doubts.

zond:
Oh you are correct - the offset was already fixed in DrawSegments as you noted. I'll send a rollback.

eustas:
I'm preparing test for this.

PR #4247, opened 2025-05-12
Reverted the broken fix.

zond:
As @eustas noted, this was already fixed - it was just a case of bad code reading on my part.

See https://github.com/libjxl/libjxl/pull/4242

### DIFF
--- a/lib/jxl/splines.cc
+++ b/lib/jxl/splines.cc
@@ -79,7 +79,7 @@ float ContinuousIDCT(const Dct32& dct, const float t) {
 
 template <typename DF>
 void DrawSegment(DF df, const SplineSegment& segment, const bool add,
-                 const size_t y, const size_t x, const size_t row_x0, float* JXL_RESTRICT rows[3]) {
+                 const size_t y, const size_t x, float* JXL_RESTRICT rows[3]) {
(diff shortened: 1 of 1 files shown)


zond:
@eustas, @veluca93 

I'm still a bit worried about this.

DrawSegments looks like
```
void DrawSegments(float* JXL_RESTRICT row_x, float* JXL_RESTRICT row_y,
                  float* JXL_RESTRICT row_b, size_t y, size_t x0, size_t x1,
                  const bool add, const SplineSegment* segments,
                  const size_t* segment_indices,
                  const size_t* segment_y_start) {
  float* JXL_RESTRICT rows[3] = {row_x - x0, row_y - x0, row_b - x0};
  for (size_t i = segment_y_start[y]; i < segment_y_start[y + 1]; i++) {
    DrawSegment(segments[segment_indices[i]], add, y, x0, x1, rows);
  }
}

```

It does subtract x0 from all rows, but it still sends x0 as-is to DrawSegment - which will still do Load and Store to rows[c] + x0 (after some clamping to the spline segment width).

I suspect this should perhaps be changed to `DrawSegment(segments[segment_indices[i]], add, y, 0, x1-x0, rows);`

WDYT?

PR #4252, opened 2025-05-13
Refactor DrawSegment to avoid possibly temporarily invalid addresses

eustas:
Hopefully calculations are more clear now as well.


zond:
Looks much cleaner!

zond (approved):

Issue #4287, opened 2025-06-13
cjxl -q 100 -e 1 is lossy


y-guyon:
**Describe the bug**
Compressing and decompressing a PNG with cjxl and djxl changes the value of a pixel.

**To Reproduce**
```sh
git clone https://github.com/libjxl/libjxl.git # Fails at v0.11.1 and at head (0102eef)
cd libjxl
./deps.sh
cmake -S . -B build  -DBUILD_TESTING=OFF -DJPEGXL_ENABLE_BENCHMARK=OFF -DJPEGXL_ENABLE_EXAMPLES=OFF -DJPEGXL_ENABLE_OPENEXR=OFF  -DJPEGXL_ENABLE_DEVTOOLS=ON -DJPEGXL_ENABLE_JPEGLI=ON -DJPEGXL_ENABLE_JPEGLI_LIBJPEG=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++  -DBUILD_SHARED_LIBS=ON
cmake --build build --parallel
build/tools/cjxl emoji_u1f468_1f3fb_200d_1f373.png -q 100 -e 1 enc.jxl
build/tools/djxl enc.jxl dec.png
compare emoji_u1f468_1f3fb_200d_1f373.png dec.png diff.png
```

Source `emoji_u1f468_1f3fb_200d_1f373.png`:

![Image](https://github.com/user-attachments/assets/f9795bb8-80e3-4394-84c8-526411270f6e)

`diff.png`:

![Image](https://github.com/user-attachments/assets/cff6ec9b-d9ff-44e2-b2af-b177346034c9)

**Environment**
 - OS: linux
 - Compiler version: clang 19.1.7
 - CPU type: x64
 - cjxl/djxl version string: v0.12.0 0102eef5 [_AVX2_,SSE4,SSE2]

Galaxy4594:
I can replicate the issue, it only happens with effort 1, it seems to have been an issue since v0.8. So basically ever since fjxl has been a thing. I wonder if it's a gamma issue similar to this: https://github.com/libjxl/libjxl/issues/4139#issuecomment-2695180418
```
ssimulacra2.exe original.png effort1.jxl 91.04934214
ssimulacra2.exe original.png effort2.jxl 100
```

jonsneyers:
[...]

PR #4027, opened 2024-12-23
Fix bug in BitWriter::WriteMultiple.

veluca93:
Fixes #4026.

PR #4291, opened 2025-06-16
fix subtle bug in e1 palette detection

jonsneyers:
Fixes https://github.com/libjxl/libjxl/issues/4287

Pretty subtle bug that would cause it to become possible for a hash collision to be missed, which causes an incorrect encode down the road.


veluca93 (approved):

jonnyawsom3:
Shouldn't this, #4027 and #4281 (Probably more missed fixes too) also be mirrored to https://github.com/libjxl/libjxl-tiny? Or is that repo no longer maintained? In which case it should probably be stated in the README.


Edit: Disregard that, I thought fast lossless was originally in a separate repo.

Issue #4270, opened 2025-05-23
jxlinfo output may be confusing


gregbenz:
For some images, the data reported by jxlinfo may be a little confusing, particularly for occasional users of this handy tool. 

If you encode a 32-bit image to JXL in Photoshop with ProPhoto primaries, the following is reported by jxlinfo:

```
JPEG XL file format container (ISO/IEC 18181-2)
JPEG XL image, 1000x1250, lossy, 32-bit float (8 exponent bits) RGB+Alpha
Color space: RGB, Custom, white_point(x=0.345705,y=0.358541), Custom primaries: red(x=0.734700,y=0.265300),  green(x=0.159601,y=0.840398),  blue(x=0.036598,y=0.000100)Linear transfer function, rendering intent: Relative
```

The "color space:" line is a bit confusing here.

Suggestions for clarity:
- What does "Custom" refer to? I presume the white point - but it isn't clear, especially with the comma separating it from the other values. The following "custom primaries:" more clearly connects to the actual x,y data. If this is white point, probably better to show "Custom white point: x=0.345705,y=0.358541"
- Break the output to multiple lines (separate white point from primaries, etc)
- There is no break/space before "Linear transfer..."

jonnyawsom3:
https://github.com/libjxl/libjxl/pull/3445#issuecomment-2027073445
> I want to keep the list of spaces with a shorthand name relatively short, because you could add spaces like Adobe98 and ProPhoto that are also still relatively common but at some point we're turning this into a colorspace name registry and that's not the intention :)

Adobe98 was eventually added with https://github.com/libjxl/libjxl/pull/4199, so maybe ProPhoto can still be added too. I can't think of any other common profiles without shorthand names, so that would complete the set.

gregbenz:
@jonnyawsom3 Agreed, simple is all that is needed (Rec2020, P3, Adobe RGB, sRGB, ProPhoto is a solid list to support).

eustas:
Fixed formatting in #4300.

Will take a look if we could add primaries alias a wee later.

PR #4296, opened 2025-06-17
Colorspace names: add ProPhoto

jonsneyers:
See also https://github.com/libjxl/libjxl/issues/4270

This doesn't change jxlinfo output yet, but it does make synthesized ICC profiles have nicer names in the common cases (sRGB, Adobe98, ProPhoto, DisplayP3, Rec2100PQ, Rec2100HLG).

Also adds "ProPhoto" as a recognized keyword when specifying color spaces (e.g. in `cjxl`, `djxl`, `simplify_icc`)


sboukortt (dismissed):

jonsneyers:
Conformance is failing because the synthesized profiles no longer match (they get shorter names after this change). That's actually a bug in the conformance testing, it is too strict to expect synthesized profiles to have exact contents. This should fix that: https://github.com/libjxl/conformance/pull/42

(please retry running the checks after the conformance testing has been fixed)

jonsneyers:
For some reason, conformance is still failing for lossless_pfm but I cannot figure out what could be causing a change.

If someone feels like figuring out, feel free!

FooIbar:
Conformance will fail because it uses `conformance.py` from this repo instead of the conformance repo
https://github.com/libjxl/libjxl/blob/4867d7bb0853a557dbc53659c0485e6b7c856947/.github/workflows/conformance.yml#L148

FooIbar:
I found it was changed to use this repo's `conformance.py` in #1533, any particular reason for using that over the conformance repo one?

cc @szabadka

jonnyawsom3:
Superseded by #4824

PR #4298, opened 2025-06-19
Fix decoder: some valid streams were undecodable.

eustas:
`IsSingleValueAndAdvance` assumes that LZ77 does not "mix" images, i.e. tail of copy does not span over beginning of degenerate image.

### DIFF
--- a/lib/jxl/dec_ans.h
+++ b/lib/jxl/dec_ans.h
@@ -379,6 +379,8 @@ class ANSSymbolReader {
   bool IsSingleValueAndAdvance(size_t ctx, uint32_t* value, size_t count) {
     // TODO(veluca): No optimization for Huffman mode yet.
     if (use_prefix_code_) return false;
+    // TODO(eustas): Check if we could deal with copy tail as well.
+    if (num_to_copy_ != 0) return false;
     // TODO(eustas): propagate "degenerate_symbol" to simplify this method.
     const uint32_t res = state_ & (ANS_TAB_SIZE - 1u);
     const AliasTable::Entry* table = &alias_tables_[ctx << log_alpha_size_];


veluca93 (approved):

jonsneyers:
Wait, how are there valid bitstreams where `!use_prefix_code_ && num_to_copy_ != 0` ?

I would expect an LZ77 run to terminate at the end of an entropy-coded stream, and never continue into the next entropy-coded stream, especially not if that one has LZ77 disabled.

Maybe the spec needs to be clarified on this point. I'd prefer this to be invalid, and have the invariant that num_to_copy == 0 at the end of an entropy-coded stream. At the very least, this should be true at the end of every bitstream section because otherwise parallel decoding is impossible.

eustas:
The thing is - streams in question are produced by our encoder... Discovered by failing roundtrips.

eustas:
@jonsneyers aren't `use_prefix_code_` and `lz77.enabled` independent?

jonnyawsom3:
How rare is it, because I don't think we've had any issues opened about it, right? Might not be too late to fix spec and the encoder, while still leaving the fix in the decoder for the few that exist.

Unlike the special distance problem last year, these files already didn't decode, so the impact should've been noticed far sooner if it was a big issue.

eustas:
Nobody complained, because it is less likely to trigger with previous releases.

Recently heuristics that enable LZ77 have been tuned towards lower qualities, where degenerate planes are more likely.

jonnyawsom3:
I imagine my PR that enables LZ77 _at all_ for default lossless probably contributed too. Still seems like a good time to fix it encoder side, before the next release. I assume there's no downside to leaving this fix in for compatibility?

jonsneyers:
> @jonsneyers aren't `use_prefix_code_` and `lz77.enabled` independent?

Oh right, my bad. I must have seen too much Deflate, causing me to conflate lz77 and prefix coding :)

PR #4303, opened 2025-06-23
Comparison tool: don’t draw the left image if the cursor is to its left

sboukortt:
`drawPixmap` with an empty rect appears to draw the entire image. If the left and right image are the same size (as should usually be the case) then the right image would then cover it anyway, but this avoids the appearance of a glitch otherwise.


jonsneyers:
Shouldn't there be a symmetric condition too for drawing the right image if the cursor is to its right?

jonsneyers (approved):

sboukortt:
As far as I understand the situation, it’s problematic for the left image because the origin is `(0, 0)` (or to the left of that) so it takes everything, whereas for the right image, if the cursor is to the right then so is the origin, so taking everything to the right and bottom of that point is still nothing.

PR #4305, opened 2025-06-24
Replace ordered noise with blue noise

### LINKED ISSUE #3926: Use Blue Noise for dithering
#3090 (based on #1362) greatly improved perceptual quality of dark images, however at high zoom levels, zooming with NN upsampling or in images with smooth gradients, the crosshatching becomes obvious and even distracting on some images.

Adding a decode flag, or outright replacing the 8x8 ordered dithering with one of the fast solutions already in use for blue noise dithering https://psychopath.io/post/2022_07_24_owen_scrambling_based_dithered_blue_noise_sampling

Galaxy4594:
Addresses the repeating patterns seen when decoding to 8-bit with ordered noise.

This is visual example I made with the brightness cranked up to see the noise patterns better.
Noise_Comp
Edit: Newer comparison.

Fixes #3926

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -336,7 +336,7 @@ jobs:
 
     - name: Coverage upload to Codecov
       if: env.WILL_COV == 'true'
-      uses: codecov/codecov-action@fdcc8476540edceab3de004e990f80d881c6cc00 # v5.5.0
+      uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
       with:
         flags: unittests
         files: build/coverage.xml
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -60,7 +60,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@2d92b76c45b91eb80fc44c74ce3fce0ee94e8f9d # v3.29.5
+        uses: github/codeql-action/init@f1f6e5f6af878fb37288ce1c627459e94dbf7d01 # v3.29.5
         with:
           languages: ${{ matrix.language }}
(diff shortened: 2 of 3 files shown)


jonsneyers:
Seems like a good idea to me. Two things:

- Some of the tests are failing, probably just because they're comparing the new dithering to a reference decode with the old dithering and getting some differences.
- Using a 32x32 matrix might cause a bit of slowdown due to some extra pressure on the cpu cache, since it's 4 KiB instead of 0.5 KiB. Could you give some before/after results for `djxl input.jxl output.ppm --num_threads=0 --num_reps=100` to see if there's something noticeable?

Galaxy4594:
A very small but notable difference for me on my intel i5-12400, from 74.889 MP/s to 74.034 MP/s.

Speaking of the failing tests, what would be the best way to fix them?

jonsneyers:
> A very small but notable difference for me on my intel i5-12400, from 74.889 MP/s to 74.034 MP/s.
> 
> Speaking of the failing tests, what would be the best way to fix them?

Best to locally run the tests and check on a case-by-case basis if the failure indicates an actual problem or not.
I think there might be an actual problem related to the dithering weights being too high, making lossless uint8 no longer roundtrip. Once that's fixed, maybe there are no failing tests anymore.

eustas:
In my ideal world, dithering matrix should be generated at compile time. Most preferably if all values are `integer/(2**k)` (and no FPs are used in calculations before the final conversion)...

jonnyawsom3:
Theoretically, we could just call the photon noise in the decoder to get blue noise for 'free', set to a low level like 200.

Galaxy4594:
Okay I upped the precision from 4 decimal points to 5 and I constrained the values to [-0.49219, 0.49219]. Hopefully this fixes the failing tests.

Galaxy4594:
It's seems like @jonsneyers is on holiday. 😅

eustas:
@jonsneyers friendly ping

jonsneyers (approved):

PR #4306, opened 2025-06-24
Predictor Zero LZ77 Tweaks

jonnyawsom3:
A minor follow-up to #4225.

Adjusts behaviour when Predictor Zero is used, giving extra speed at low effort settings and removing a redundant faster decoding check. `cjxl -d 0 -e 2 -P 0` goes from `22.797 MP/s` to `27.256 MP/s` with no impact on density.

### DIFF
--- a/lib/jxl/enc_ans.cc
+++ b/lib/jxl/enc_ans.cc
@@ -1301,15 +1301,17 @@ HistogramParams HistogramParams::ForModular(
   if (cparams.decoding_speed_tier >= 2) {
     params.max_histograms = 12;
   }
-  if ((cparams.decoding_speed_tier >= 3 ||
-       cparams.options.predictor == Predictor::Zero) &&
-      cparams.modular_mode) {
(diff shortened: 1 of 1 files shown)


jonsneyers:
I guess this is only because `-d 0 -e 2 -P 0` ends up using the ClampedGradient predictor anyway, but adding the `-P 0` would make it try to do lz77 which is not very useful?

jonnyawsom3:
Exactly. File size was actually identical, suggesting it does the LZ77 but then doesn't save it. Likely because efforts 1-3 have some hardcoded parameters.

I also changed it so `-P 0` will use RLE for effort 5 too, as RLE 4 to 5 have a large density change (11.5 bpp to 9.3 bpp), but LZ77 5 to 6 had almost none (7.24 bpp to 7.2 bpp). Means the effort levels have better tradeoffs.

jonnyawsom3:
Needs to be added to the merge queue, if you don't mind

PR #4311, opened 2025-06-26
Improve speed stats reporting

eustas:
(especially for regular user case when reps==1)

Before: `5616 x 3744,  0.927 MP/s [0.93, 0.93], , 1 reps, 96 threads.`
After: `5616 x 3744, 0.934 MP/s, 96 threads.`

### DIFF
--- a/tools/speed_stats.cc
+++ b/tools/speed_stats.cc
@@ -38,7 +38,7 @@ bool SpeedStats::GetSummary(SpeedStats::Summary* s) {
   if (elapsed_.size() == 2) {
     s->central_tendency = elapsed_[1];
     s->variability = 0.0;
-    s->type = " second:";
+    s->type = "second: ";
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


jonnyawsom3:
I see there's an output for MB/s too, but I'm not aware of a way to display it. Maybe it could be shown with `-v`?

jonsneyers:
MB/s is a confusing unit for image compression: are you talking about megabytes of compressed bitstream? of uncompressed image buffers? if the latter, then in what pixel format?
E.g. do you count a 10-bit RGB image that is passed using 16-bit RGBA buffers (with a trivial/ignored alpha) as being 30 bits per pixel or 64 bits per pixel?

It does make sense in the case of lossless jpeg recompression though, so I think it's only printed in that case.

jonsneyers (approved):

jonnyawsom3:
Riiight, that would explain it. I was just curious if it was broken, since I've never seen the MB/s displayed but noticed the code here.

Issue #465, opened 2021-08-18
cjxl generates 16-bit float from 32-bit float exr input


xiota:
**Describe the bug**
`cjxl` generates 16-bit float from 32-bit float `exr` input.  `pfm`  files are not affected.

**To Reproduce**
* Convert image to 32-bit float in GIMP and export as an `exr` file.
* Confirm image is 32-bit float using `exrheader`.
* Use `cjxl` to compress file in both lossy and lossless modes.
* Use `jxlinfo` to examine `bits_per_sample` and `exponent_bits_per_sample`.
  - Result is (16, 5).  Expected is (32, 8). 

**Expected behavior**
`jxl` file should have the same pixel format as the original files (32-bit float in this case).

**Environment**
 - OS: Ubuntu 21.04; Linux 5.11.??
 - Compiler version: gcc 10.3.0
 - CPU type: x86_64
 - cjxl/djxl version string: 0.6.0 810ecc3

**Additional context**
It appears [`codec_exr.cc`](https://github.com/libjxl/libjxl/blob/7d6a7bd229db977bf94e1c1698cf07bf254ae2cc/lib/extras/codec_exr.cc#L32) has 16-bits hard coded for `exr` files, rather than detecting the number of bits.

Note: I'm just playing with `jxl` capabilities.  I don't have any need for 32-bit float files.

jonsneyers:
Looks like what is needed here is to fix `codec_exr.cc` to detect the correct float format and pass that to the encoder. It works for `pfm` so it should also be possible to get it to work for `exr`.

xiota:
I did some searching, and it looks like the OpenEXR API call that's being used supports only float16 RGB.  To get float32 or grayscale image data requires using a more complicated API call.

jonsneyers:
That's annoying. Well, I guess as a workaround you can use pfm?

xiota:
@jonsneyers  With pfm, there's the issue I opened then closed (#463).  Different programs treat the data differently.  GIMP treats it as linear data.  Imagemagick treats it as non-linear data.  Whether the ICC profile is used doesn't make any difference.

Maybe someone more familiar with the OpenEXR API will eventually fix this?

aras-p:
This should be fixed by #4312 that just got merged

Issue #1662, opened 2022-07-30
Problem with OpenEXR


Jamaika1:
cjxl.exe image_21447_24bit.exr image_21447_24bit_exr.jxl -v -m 1 -q 100 -e 5 -C 1 --num_threads=4
JPEG XL encoder v0.7.0 0.7.0-59f7d19 [Unknown]
cjxl_main.cc:676: JXL_FAILURE: Codecs failed to decode input.

mo271:
Currently OpenEXR is not supported, but we are planning on bringing this back.

jonnyawsom3:
What's the status of this after #4007?

MasterInQuestion:
Alike?
    [ https://github.com/libjxl/libjxl/issues/625#issuecomment-2460663870](https://github.com/libjxl/libjxl/issues/625#issuecomment-2460663870)
    (Pipe everything from capable decoders...)

PR #4312, opened 2025-06-26
cjxl: EXR improvements

aras-p:
### Description

Improvements in handling EXR files when reading them in `cjxl`:
- No longer makes color channels always be FP16 (even for EXR files where they were FP32), fixing #465
- Support EXR files where color channel names are not strictly just `R`, `G`, `B` (`A`). Multi-layer files could have names like `Composite.Combined.R` etc. Now finds the first EXR "layer" (channel prefix ending with `.`) that contains channels `R`, `G`, `B` and uses that for color channels. Uses alpha with the same prefix.
- If no "color channels" are found, put first EXR channel as "grayscale", this makes it support luminance-only files like [Garden.exr](https://openexr.com/en/latest/test_images/LuminanceChroma/Garden.html) from EXR test suite.
- Read all other channels as "extra channels" to be encoded.
- Additionally, this fixes `EncodeImageJXL` to actually support per-channel pixel formats; previously it was using the same base color pixel format for all the extra channels.

This makes `cjxl` be able to read [Blender40.exr](https://aras-p.info/files/exr_files/Blender40.exr) (135MB) or [Blender41small.exr](https://aras-p.info/files/exr_files/Blender41small.exr) (3.3MB) with many channels.

**Issues I haven't figured out yet**:
- ~~When mixed channel precisions are used (e.g. Blender41small.exr above), the FP32 channels seem to get encoded as garbage with `cjxl -d 0 -e 2`. The data that ends up in corresponding `PackedImage` object of that channel seems to be correct, so something further down in the whole compression pipeline goes wrong?~~ _Fixed in EncodeImageJXL_.
- Extra channel names do not come through to JXL file. I am setting them in `PackedExtraChannel` but maybe that is not used further down?

### Pull Request Checklist

### DIFF
--- a/lib/extras/dec/exr.cc
+++ b/lib/extras/dec/exr.cc
@@ -33,9 +33,10 @@ Status DecodeImageEXR(Span<const uint8_t> bytes, const ColorHints& color_hints,
 
 #else  // JPEGXL_ENABLE_EXR
 
+#include <ImfChannelList.h>
+#include <ImfFrameBuffer.h>
 #include <ImfIO.h>
-#include <ImfRgba.h>
-#include <ImfRgbaFile.h>
+#include <ImfInputFile.h>
 #include <ImfStandardAttributes.h>
 #include <OpenEXRConfig.h>
 #include <jxl/color_encoding.h>
@@ -45,6 +46,7 @@ Status DecodeImageEXR(Span<const uint8_t> bytes, const ColorHints& color_hints,
 #include <cstddef>
 #include <cstring>
 #include <memory>
+#include <set>
 #include <utility>
 #include <vector>
 
@@ -70,9 +72,6 @@ namespace OpenEXR = OPENEXR_IMF_NAMESPACE;
 // on macOS, where the definition for OpenEXR::Int64 was actually not equivalent
 // to uint64_t. This alternative should work in all cases.
 using ExrInt64 = decltype(std::declval<OpenEXR::IStream>().tellg());
-
-constexpr int kExrBitsPerSample = 16;
-constexpr int kExrAlphaBits = 16;
 
 class InMemoryIStream : public OpenEXR::IStream {
  public:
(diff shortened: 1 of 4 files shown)


jonsneyers:
It is very much possible that there are encoder bugs that cause channel names to get dropped somewhere or mixed-precision channels to get treated incorrectly. That wouldn't be surprising since this would be the first input format that supports channel names and mixed-precision channels :)

We can tackle those bugs in follow up pull requests. After this PR lands, it will be easier to reproduce such bugs.

lib/extras/dec/exr.cc
```diff
@@ -127,52 +126,108 @@ class InMemoryIStream : public OpenEXR::IStream {
...
+  OpenEXR::InputFile input(is);
 #endif
 
-  if ((input.channels() & OpenEXR::RgbaChannels::WRITE_RGB) !=
-      OpenEXR::RgbaChannels::WRITE_RGB) {
-    return JXL_FAILURE("only RGB OpenEXR files are supported");
+  const OpenEXR::Header& header = input.header();
+  const OpenEXR::ChannelList& channels = header.channels();
+
+  // we don't support subsampled or UINT channels yet
```
jonsneyers, line 179: Maybe add a `// TODO:` comment here, since some cases of subsampling could be handled by setting the channel shift; this only works if `ch.xSampling == ch.ySampling` ∈ {2, 4, 8} but that probably does cover most of the cases of subsampling that occur in practice.

lib/extras/dec/exr.cc
```diff
@@ -182,59 +237,129 @@ Status DecodeImageEXR(Span<const uint8_t> bytes, const ColorHints& color_hints,
...
+    const JxlPixelFormat ec_format{1, fp16 ? JXL_TYPE_FLOAT16 : JXL_TYPE_FLOAT, JXL_NATIVE_ENDIAN, 0};
+    JXL_ASSIGN_OR_RETURN(
+        PackedImage ec,
+        PackedImage::Create(imageWidth, imageHeight, ec_format));
+    frame.extra_channels.emplace_back(std::move(ec));
+
+    PackedExtraChannel pec = {};
+    pec.ec_info.bits_per_sample = fp16 ? 16 : 32;
+    pec.ec_info.exponent_bits_per_sample = fp16 ? 5 : 8;
+    pec.ec_info.type = JXL_CHANNEL_OPTIONAL;
```
jonsneyers, line 272: This is fine, but consider adding a `// TODO`  here to map recognized extra channels to other predefined channel types (e.g. depth or thermal) if there are sufficiently commonly used naming conventions for those in EXR (I don't know).

jonsneyers:
You'll need to run `clang-format --style=file -i` on the files you changed, otherwise the CI will fail on the `format` check.

jonsneyers (dismissed):

jonnyawsom3:
Maybe we could enable EXR support for Github releases in another PR too?
Not sure when or if it was enabled before, but might help #1662.
```
JPEG XL encoder v0.11.1 794a5dc [AVX2,SSE2]
Usage: cjxl INPUT OUTPUT [OPTIONS...]
 INPUT
    the input can be JXL, PPM, PNM, PFM, PAM, PGX, PNG, APNG, GIF, JPEG

aras-p:
Hmm the "msan" build failing is strange. The "asan" build seems to use the same OS, the same compiler and it pulls the same OpenEXR version, and that one is passing. Could it be that when msan is on, the std::string becomes a different type (like different template types etc.) and thus fails to find shared library imports that have std::string arguments?

jonsneyers:
@eustas do you have any idea why the msan build is not compiling? All other checks are passing...

eustas:
Going to look into now.

eustas:
Generally speaking, there might be 2 problems:
 - library version mismatch
 - order and list of included headers

Have had both before. Will play with this PR on my PC. If it reproes, will try to fix.

eustas:
The problem is:
 - compiler loos for: `Imf_3_1::ChannelList::findChannel(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const`
 - library provides: `Imf_3_1::ChannelList::findChannel(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const`

I.e. `std::__1` vs `std::__cxx11`. If `Imf_3_1::ChannelList::findChannel(char const*)` is a viable alternative (i.e. layer names does not contain `\0`), then I would recommend using it instead.

aras-p:
@eustas thanks! interesting that this only affects msan build though? like the asan build is using the same code and the same OpenEXR library, yet it passes. Anyway, I'll try to change this to not rely on STL string function overload.

eustas (approved):

jonsneyers (approved):

PR #4322, opened 2025-07-03
SIMDify EstimateCost

eustas:
While on release build it is ~+5% performance, on ASAN build it is ~30x speedup.

Why: std::min/max are tracked by ASAN as branch / decision point, and cost a lot...

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -42,6 +42,7 @@
 #include "lib/jxl/enc_cache.h"
 #include "lib/jxl/enc_fields.h"
 #include "lib/jxl/enc_gaborish.h"
+#include "lib/jxl/enc_modular_simd.h"
 #include "lib/jxl/enc_params.h"
 #include "lib/jxl/enc_patch_dictionary.h"
 #include "lib/jxl/enc_quant_weights.h"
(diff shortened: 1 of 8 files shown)


jonsneyers:
Have you checked how large the difference is in the result of EstimateCost between the two implementations?

lib/jxl/enc_modular.cc
```diff
@@ -273,46 +275,6 @@ float EstimateWPCost(const Image& img, size_t i) {
...
-      }
-    }
-    for (auto& h : histo) {
-      histo_cost += h.ShannonEntropy();
-      h.Clear();
-    }
-  }
-  return histo_cost + extra_bits;
-}
-
```
jonsneyers, line 315: Would it make sense to keep the old scalar implementation as a comment in the simd version? It is probably useful documentation.
Or at least add some high-level comment describing what it does (this was missing from the original code but becomes more important if the code becomes less readable). The function is basically estimating the cost based on the entropy of the residuals of an effort 2 style fixed prediction with ClampedGradient and a fixed context model based on a simple context property (one that doesn't actually exist but presumably is somewhat representative).

eustas: Sure, will re-add as scalar specialization.

eustas:
> Have you checked how large the difference is in the result of EstimateCost between the two implementations?

In tests we have 2 cases that trigger that function, each about dozen times. I've ported code to match exactly with 2 changes:
 - contexts are reverse-numbered (0-to-16, not 16-to-0); that __should not__ affect the cost, since all contexts are equal
 - (but __it does__, that is why) sum of histo costs is collected separately for whole and fraction parts; this avoids roundings done by float addition

Due to "reversed" order of histo summing results are different (in fractional part for our test cases).
Since I'm going to re-add scalar implementation I'll patch it with 2-nd change (cost cumulation), so that difference will be 0.

eustas:
PTAL. I have a change on top of this that makes encoding with`-m 1 -d 0 -e 8` 1.168x faster.

PR #4325, opened 2025-07-05
Image viewers: for jxl files, use Qt for color transforms instead of lcms2

sboukortt:
Simpler and much faster (an image that took over 11 seconds to load on my machine now takes 850ms).


sboukortt:
Or maybe not for now – it seems it’s not quite as accurate as lcms2, leading to noticeable inconsistencies in the comparison tool if one image is a JXL and the other isn’t. (Perhaps we can consider using Qt in both codepaths.)

PR #4331, opened 2025-07-09
Do not ignore rct_type=0

eustas:
This improves density in relevant cases


eustas:
NB: `do_transform` for rct_type=0 returns `false`. This is expected behaviour - we don't want it in encoded stream.

szabadka (approved):

PR #4336, opened 2025-07-10
Update predictor when setting ModularOptions::TreeMode::kNoWP

eustas:
Q: perhaps we need to update it conditionally

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -466,18 +466,25 @@ Status ModularFrameEncoder::Init(const FrameHeader& frame_header,
   }
 
   if (cparams_.ModularPartIsLossless()) {
+    const auto disable_wp = [this] () {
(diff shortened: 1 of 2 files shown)


jonnyawsom3 (changes requested):
We want to have (unfortunately) another nested if statement explicitly checking if the selected predictor is Weighted. Otherwise, effort 9 and 10 (Or -P 15 in cjxl) would skip trying any other predictors. Weighted is the most expensive by far, hence using kNoWP.

eustas:
> We want to have (unfortunately) another nested if statement explicitly checking if the selected predictor is Weighted. Otherwise, effort 9 and 10 (Or -P 15 in cjxl) would skip trying any other predictors. Weighted is the most expensive by far, hence using kNoWP.

Sure. That is why I left (Q) in description... Done.

szabadka (approved):

PR #4375, opened 2025-08-05
fix variable name typo

jonsneyers:
linter was complaining about this typo...


jonsneyers:
nevermind, was already flyby-fixed in another pull request

PR #4384, opened 2025-08-11
Faster upsampling

eustas:
2.5-4x faster upsampling


jonnyawsom3:
Good work. Should make `--resampling` properly scale decode speed with internal resolution.

jonnyawsom3:
Does this only apply to 8x upsampling? Would it be difficult to do 4x and 2x too? 8x is used for the LF but 2x is used at or above distance 10.

eustas:
It affects all 3 resampings (that is why speedup us a range, not a single number); though, I've tested it only on AVX2; most likely other vector engines will benefit. Not unlikely scalar too.

Issue #4389, opened 2025-08-12
libjxl decoder requires CMS set when pixels are requested in the tagged space with XYB disabled


Traneptora:
**Describe the bug**
If a JXL file is tagged with `uses_original_profile = true` (i.e. XYB is disabled) then calling `JxlDecoderSetPreferredColorProfile` with the exact same enum values returned by `JxlDecoderGetColorAsEncodedProfile` should succeed, because no CMS is required at all.

Instead, what happens is it fails and it dumps the following warning on debug mode.
```
./lib/jxl/decode.cc:2678: must set color management system via JxlDecoderSetCms
```

**To Reproduce**
Take any libjxl-based decoder (e.g. ffmpeg) and use it with libjxl 0.11.0, and it works as expected. Linking to git master however fails with the error as described above.

A code path we have in ffmpeg is the following, which is fairly common:

```c
    jret = JxlDecoderGetColorAsEncodedProfile(ctx->decoder, JXL_COLOR_PROFILE_TARGET_ORIGINAL, &jxl_color);
    if (jret == JXL_DEC_SUCCESS) {
        /* enum values describe the colors of this image */
        jret = JxlDecoderSetPreferredColorProfile(ctx->decoder, &jxl_color);
        if (jret == JXL_DEC_SUCCESS)
            jret = JxlDecoderGetColorAsEncodedProfile(ctx->decoder, JXL_COLOR_PROFILE_TARGET_DATA, &jxl_color);
        /* if we couldn't successfully request the pixel data space, we fall back on wide gamut */
        /* this code path is very unlikely to happen in practice */
        if (jret != JXL_DEC_SUCCESS)
            fallback = 1;
    } 
```

This should work even without setting a CMS engine. It does work in release 0.11.0, but it does not work in git master. I bisected and I believe the culprit is this commit: 81247d5cf29e472700d65ebef72d91edcde6a396

**Environment**
 - OS: Arch Linux
 - Compiler version: gcc 15.1.0
 - CPU type: x86_64
 - cjxl/djxl version string: djxl v0.12.0 095b8b2a [_AVX2_] {GNU 15.1.1}

PR #4399, opened 2025-08-18
lib/jxl/decode.cc: avoid erroring out if setting color encoding to same

Traneptora:
If the image is not XYB encoded and a library client sets the color encoding to the one tagged in the image, then CMS is not actually required, so we should not error out in this case.

Fixes: #4389.


eustas:
Thanks!

PR #4401, opened 2025-08-18
Undisable trixie

eustas:
Since Gimp3 has their own JXL plugin, we could just discontinue shipping ours (for distros that dropped Gimp2).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -120,10 +120,13 @@ jobs:
         - ubuntu:20.04
         - debian:bullseye
         - debian:bookworm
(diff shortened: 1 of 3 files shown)


mo271 (approved):

mo271 (commented):
https://github.com/libjxl/libjxl/blob/cfa5933bcd29b4094eedeb6aad4abfbaa70cc0db/doc/developing_in_debian.md?plain=1#L37 
should also remove the `libgimp2.0-dev`, I suppose

eustas:
Will do in separate PR. Because there are more updates need to be done.

PR #4442, opened 2025-09-11
Fix build on some systems

eustas:
"SYSTEM" includes get special treatment when Clang rewrites command line for AppleClang.
It seems that without that option all works fine on other systems.

Drive-by: fix specific warnings/errors.

### DIFF
--- a/lib/jxl/jxl_test.cc
+++ b/lib/jxl/jxl_test.cc
@@ -975,7 +975,7 @@ TEST(JxlTest, RoundtripAlphaResampling) {
   TestImage t;
   ASSERT_TRUE(t.DecodeFromBytes(orig));
   t.ClearMetadata();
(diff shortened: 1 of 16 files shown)


lib/jxl/jxl_test.cc
```diff
@@ -1626,74 +1626,359 @@ JXL_TRANSCODE_JPEG_TEST(JxlTest, RoundtripJpegRecompressionMetadata) {
...
+      0x00,  //
       // entropy coded data
-      0xfc, 0xaa, 0xaf,  //
+      0xfc,
+      0xaa,
+      0xaf,  //
       // EOI
-      0xff, 0xd9,  //
+      0xff,
+      0xd9,  //
```
sboukortt, line 1963: clang-format mishap?

sboukortt (approved):

sboukortt (approved):

PR #4446, opened 2025-09-16
Mingw build fix for lto

### LINKED ISSUE #4445: Compiling libjxl with clang and LTO is broken in mingw
I mainly compile libjxl using mingw64, and clang's lto seems to only work with the lld linker. However, sjpeg fails to link properly, meaning `djxl`, `benchmark_xl`, and `jpegli` all fail to compile by default. Only `cjxl` and `jxlinfo` can successfully compile. 

A short-term solution would be to disable sjpeg but it would be nice to know why it fails. I've tried compiling sjpeg by itself with clang+lto and it also seems to fail as well, implying the issue is sjpeg's fault and not libjxl's. Is it a build system problem or something deeper?

This is my working LTO build command:
`cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF -DJPEGXL_STATIC=ON -DJPEGXL_ENABLE_LTO=ON -DJPEGXL_ENABLE_SJPEG=OFF -DCMAKE_C_FLAGS=-fuse-ld=lld -DCMAKE_CXX_FLAGS=-fuse-ld=lld -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld .. && ninja`

Using the command above, enabling sjpeg spits out this error:
```
ld.lld: error: undefined symbol: std::__once_callable
>>> referenced by C:/msys64/home/user/libjxl/third_party/sjpeg/src/enc.cc
>>>               libsjpeg.a(enc.cc.obj)
```

OS: Windows, msys2-mingw64
Compiler: Clang 20.1.8

Galaxy4594:
Fixes #4445, although it doesn't really address the sjpeg issue. Also, I'm not sure if this happens outside of Windows, so the WIN32 check might be unnecessary.


eustas (commented):
IMHO, it would be better to fail rather than disable. It is easy to miss warning message.
Let's mention in message what extra flags are required for build and make it `FATAL`.

Galaxy4594:
Learning from https://github.com/libjxl/libjxl/issues/4445#issuecomment-3298123694. I added a check to determine if Clang and LTO are used, then making if `FATAL` if CLANG64 is not used.

Andarwinux:
> Learning from [#4445 (comment)](https://github.com/libjxl/libjxl/issues/4445#issuecomment-3298123694). I added a check to determine if Clang and LTO are used, then making if `FATAL` if CLANG64 is not used.

This is not acceptable, because it would break everything outside of msys2, such as llvm-mingw.

The real issue here is that gcc/libstdc++ in msys2 uses emutls, while LLVM currently cannot correctly handle emutls during LTO code generation.

Newer versions of gcc/libstdc++ have already implemented native tls, but msys2 has not switched to native tls yet, since that would completely change the ABI.

The correct approach is to detect whether the compiler uses native tls or emutls. Of course, this may be difficult to achieve. Given the current state of the mingw ecosystem, as a fallback it would be acceptable to simply detect the C++ STL implementation, since at least for libc++ there is no problem whether emutls or native tls is used.

Galaxy4594:
TIL, well that's annoying, I didn't know the MinGW ecosystem was so fractured, with many small crucial details between toolchains/versions. Unfortunately, I don't have the time to invest in fixing this since college classes have started for me. 😔

Does msys2 plan to fully switch native tls anytime soon? I really like using msys2, I'm curious about its future.

@eustas what would be the best way to approach this problem?

Andarwinux:
> Does msys2 plan to fully switch native tls anytime soon? I really like using msys2, I'm curious about its future.

I'm not sure. Because this would change the ABI, and they'd need to rebuild everything for that.

I recommend using clang64 and clangarm64, as they already uses native tls.

> TIL, well that's annoying, I didn't know the MinGW ecosystem was so fractured, with many small crucial details between toolchains/versions.

see https://www.msys2.org/docs/environments

I recommend to detect the STL implementation used by the compiler via `_LIBCPP_VERSION` (libc++) and `__GLIBCXX__` (libstdc++)
For the combination of MinGW + libstdc++ + clang + LTO, issue a warning that there may be potential [problem](https://github.com/llvm/llvm-project/issues/79715).

Issue #4447, opened 2025-09-17
Fast Lossless regression since v0.11


jonnyawsom3:
Encoding speed for `cjxl -d 0 -e 1` was already 2x slower than the standalone fast_lossless tool, but since v0.11 both have been a further 3x slower.

I realised when looking at old results in #1975, #2013 and https://www.lucaversari.it/FJXL_and_FPNGE.pdf, noticing significantly faster singlethreaded speeds than I could reach. Bisecting by Mario on Discord led to #3658 being the cause.

Related to #2268, building with Clang helps the problem somewhat, increasing encode speed by 6x but still over 2x slower than before.

Will reopen #2843 when fixed

Galaxy4594:
Here are my numbers. I'm on an i5-12400, and I compiled libjxl with clang64.
Comparing libjxl and fjxl-standalone, compiled with clang on Windows. MSVC libjxl performed around ~175 MP/s with 12 threads. So twice as slow as the worst clang build.
```
v0.8.4
╔═════════╤═════════════╤═══════════════╗
║ Threads │ cjxl        │ fast-lossless ║
╠═════════╪═════════════╪═══════════════╣
║ 1       │ 249.32 MP/s │ 295.357 MP/s  ║
╟─────────┼─────────────┼───────────────╢
║ 6       │ 694.60 MP/s │ 1027.186 MP/s ║
╟─────────┼─────────────┼───────────────╢
║ 12      │ 773.32 MP/s │ 1248.640 MP/s ║
╚═════════╧═════════════╧═══════════════╝

v0.10.4
╔═════════╤══════════════╤═══════════════╗
║ Threads │ cjxl         │ fast-lossless ║
╠═════════╪══════════════╪═══════════════╣
║ 1       │ 235.001 MP/s │ 274.598 MP/s  ║
╟─────────┼──────────────┼───────────────╢
║ 6       │ 606.156 MP/s │ 888.753 MP/s  ║
╟─────────┼──────────────┼───────────────╢
║ 12      │ 670.190 MP/s │ 1025.519 MP/s ║
╚═════════╧══════════════╧═══════════════╝

v0.11.1
╔═════════╤══════════════╤═══════════════╗
║ Threads │ cjxl         │ fast-lossless ║
╠═════════╪══════════════╪═══════════════╣
║ 1       │ 65.821 MP/s  │ 69.284 MP/s   ║
╟─────────┼──────────────┼───────────────╢
[...]

PR #4448, opened 2025-09-17
Restore fast lossless speed

sboukortt:


jonnyawsom3:
Couldn't we add a nested if to make sure AVX2 is present too? I doubt we'd ever have a CPU without it, but better safe than sorry

sboukortt:
Naïvely, I would almost ask the opposite question: do we need to check XCR0 at all in addition to the output of `cpuid`?

sboukortt:
Seems like it; sent #4449.

jonnyawsom3:
For Github issue linking, partially fixes #4447 (1-bit is still falling back to effort 2)

Edit: Perhaps not...

Issue #3881, opened 2024-10-07
Encoding float16 lossless appears to produce artifacts for specific values


Skielex:
**Describe the bug**
When using lossless float16 encoding, specific ranges of binary values appear to be corrupted. Specifically, it appears that all values with a binary representation corresponding to the uint16 ranges `[512:1023]`, `[31745:32767]`, `[33280:33793]`, and `[64513:65534]` are changed during an encode-decode cycle.

**To Reproduce**
1. Encode an image with any float16 value with a binary representation corresponding to the uint16 ranges `[512:1023]`, `[31745:32767]`, `[33280:33793]`, and `[64513:65534]`.
2. Decode the image.
3. Check that the image data has changed.

**Expected behavior**
Values should not change during a lossless encode-decode cycle. I've tested float32 encoding with all possible values and it works as expected.

**Screenshots**
Affected value ranges in yellow:
![image](https://github.com/user-attachments/assets/e1d1b3f0-a328-48fe-9bb0-0dbd12b89516)
Bit values (yellow = True) for all 65536 float16 values with affected ranges between red and blue lines.
![image](https://github.com/user-attachments/assets/80aea59d-fc2e-455c-aff0-443ec0e92493)

**Environment**
 - OS: Ubuntu WSL2 on Windows (+ Windows, AMD64 and ARM64, see https://github.com/cgohlke/imagecodecs/issues/114)
 - Compiler version: GCC 11 (I assume, it's a Python 3.10 extension)
 - CPU type: AMD Ryzen 5900X
 - libjxl version: 0.11.0 (according to [this](https://github.com/cgohlke/imagecodecs/blob/50bac419858de471b934943f949835aad18b5a96/imagecodecs/libjxl.pxd#L4))
 - cjxl/djxl version string: Cannot test with cjxl/djxl as they don't appear to support any formats that allow float16.

**Additional context**
I found this issue using the `imagecodes` Python package. Original issue is https://github.com/cgohlke/imagecodecs/issues/114. 

As mentioned in the issue linked above, I've not been able to test for the issue using cjxl/djxl or GIMP due to a lack of float16 support.

Skielex:
I don't think the issue is related to special float values, although they too are affected, for two reasons:
1. The first affected regions of values are `float16` values between 3.0517578125e-05 and 6.097555160522461e-05, which become values between 0.0 and 6.091594696044922e-05 after encode-decode.
[...]

PR #4461, opened 2025-09-26
Partial fix for roundtripping of float16 values (#3881)

sboukortt:


lib/jxl/base/float.h
eustas, line 21: Does this comment hold true?

sboukortt: At the very least, indeed not “for testing”. Might as well get rid of it altogether. Thanks.

eustas (approved):

PR #4491, opened 2025-10-24
Fix SSIMULACRA2 weight indexing for small images

nolmoonen:
### Description

This PR fixes what seems to be an indexing bug for small images. When calculating the average values, the loop breaks if the image has been downsampled to a point where either dimension is smaller than 8. This happens when at least one of the original dimensions is smaller than 256. If the loop breaks in this way, `Msssim::scales` has fewer values than `kNumScales`. Because the channel is the outer dimension of the loop that calculates the final weighted value and the index is incremented in the loop body, different weights are used for the same scale and channel than for large images. As far as I can tell from the README of https://github.com/cloudinary/ssimulacra2, no dataset with images that small are used in the training process, so this behavior is not taken into account when tuning.

### Pull Request Checklist


jonsneyers:
Good catch!

Probably ignoring the 1:32 scale is not the best way to handle small images; maybe it's better to relax that "min dimension 8" break condition in the loop instead (unless it's really needed because some of the operations assume such a minimum).

There will still be images too small so this fix is needed in any case, but I think it would be best to keep the contribution of the low freq scales for as long as possible instead of not computing it at all, which will probably lead to scores being a bit too high for small images where that happens.

eustas:
Perhaps CLI should emit warning and use tiled image (to fulfill expected minimum size)?

PR #4495, opened 2025-10-28
Fix low memory rendering pipleline

eustas:
Put together calculations of x0/xsize for ProcessRow

Drive-by: fix formatting


veluca93 (approved):

Jamaika1:
```
low_memory_render_pipeline.cc:609:17: warning: comparison of unsigned expression in '>= 0' is always true [-Wtype-limit]
  609 |   JXL_DASSERT(0 <= first_trailing_stage_);
```

PR #4502, opened 2025-10-31
Bump example project CMake version

jonnyawsom3:
Fixes a warning [reported on Discord](https://discord.com/channels/794206087879852103/848189884614705192/1431759854891958272). 

> When I install libjxl-0.11.1 on Gentoo, following message is displayed:
>  
> 
> - QA Notice: Compatibility w/ CMake < 3.16 will be removed in future ECM release.
> - If not fixed in upstream's code repository, we should make sure they are aware.
> - See also tracker bug #964407; check existing or file a new bug for this package.
> - The following CMakeLists.txt files are causing warnings:
> - examples/CMakeLists.txt:3.10
> - An upstreamable patch should take any resulting CMake policy changes
> - into account. See also:
> - https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html


novomesk:
Related to https://bugs.gentoo.org/964598

PR #4507, opened 2025-11-05
Workaround for MSAN bug

eustas:
Update expectation in RoundtripAnimationPatches to fit actual AVX3 score.


eustas:
Friendly ping

PR #4508, opened 2025-11-06
Do not depend on CPU lottery for testing AVX3

eustas:


jonsneyers:
what's [tools/scripts/test_cost-x64.zip](https://github.com/libjxl/libjxl/pull/4508/files#diff-cf485a3bf5cd6393e58c221c8cf9186864091272a48268a4d1acedc877d8fb94) ?

eustas:
After tests complete it contains how long each test run. Before running tests, if CMake have that on hands, it runs "heavy" tests first. On multicore systems that ensures faster test completion (not waiting for long running test being executed last and using just one core). In our case it helps shard balancing (both avoiding long tests being executed at the end, but also divide test set evenly).

Currently I'm running pipeline in "upload costs" mode. It uses only 1 shard, but produces costs files at the end. When it is done, I will switch back to normal mode and update costs files. The costs in current change-set is a stub made on my PC (without SDE emulation, so only loosely relevant to real pipeline costs).

eustas:
(5613/8168 tests completed @ 3h11m; hopefully won't time-out; if we are lucky each of 8 shards should complete in ~30min)

mo271 (approved):

PR #4514, opened 2025-11-10
Do not render noise on DC planes.

### LINKED ISSUE #4368: Bad behavior with lossy progressive and photon noise
There two bugs present when using photon noise combined with lossy progressive. First of all, photon noise seems be applied to the LF and the HF. This is causing a LF blurry noise texture to be present. This only happens with VarDCT when using `--photon_noise_iso` and `-p`. It also happens with `--photon_noise_iso` and `-progressive_dc`. None of the other progressive flags, `--progressive_ac` and `--qprogressive_ac` trigger this bug. 

noisecomp1

Another issue found is that for VarDCT, photon noise is much much stronger when used with progressive, likely meaning it's being applied multiple times over. A major condition to trigger this bug is that the image has to be larger than 2048x2048, implying chunked encoding must be at play here. For lossy modular, it's always stronger regardless if progressive is specified or not.

noisecomp2

eustas:
(partially) fix #4368


eustas:
@Galaxy4594 does if fix both problems or just LF noise?

jonnyawsom3:
Encoding now fails with `--photon_noise_iso` and `-p`

Galaxy4594:
Same here, it fails encoding most of the time.

eustas:
Sorry, will fix tomorrow morning.

mo271 (approved):

jonsneyers:
This fix was not correct, introduced a decoder bug.
If DC frames have noise, nothing in the spec allows decoders to not render that noise.

The correct fix would be for libjxl to stop adding noise to DC frames in the first place.

PR #4516, opened 2025-11-14
Change default color decoding from LinearSRGB to SRGB

### LINKED ISSUE #2147: Decoding from XYB with ICC always returns linear sRGB regardless of pixel format
**Describe the bug**
As stated in the [docs](https://libjxl.readthedocs.io/en/latest/api_common.html#_CPPv4N12JxlBasicInfo21uses_original_profileE):
`If the original profile is not used, the decoder only outputs the data as sRGB (linear if outputting to floating point, nonlinear with standard sRGB transfer function if outputting to unsigned integers) but will not convert it to to the original color profile.`
The decoder should returns with sRGB TRC when pixel format is uint, but this is not the case and instead it outputs as a linear TRC instead: `RGB_D65_SRG_rel_Lin`,

**To Reproduce**
Decode an uint JXL image that was converted from an color profile / ICC that's not currently supported by libjxl's enum. For example this image:
https://kampidh.com/jxlcoloranalysis/CC-proph-boost-xyb-icc-integer.jxl

**Expected behavior**
According to the docs, it should return as sRGB TRC, not linear.

**Environment**
 - OS: Windows 10
 - Compiler version: Clang 15
 - CPU type: x86_64
 - cjxl/djxl version string: 0.8.0

**Additional context**
This is possibly related to another reports that shows the image results were darker, since on non-color managed viewer / browser, linear TRC always looks dark.

### LINKED ISSUE #2289: djxl PNG Conversion Acting Up

### LINKED ISSUE #4307: PNG converted to JXL lossy becomes dark
**Describe the bug**
PNG -> JXL lossy conversion changes brightness

**To Reproduce**
* download this [sample](https://upload.wikimedia.org/wikipedia/commons/e/e5/%D0%94%D0%B5%D0%B2%D1%83%D1%88%D0%BA%D0%B0_%D1%81_%D0%A1%D0%B8%D1%80%D0%B5%D0%BD%D1%8C%D1%8E_%28%D0%98%D0%BD%D0%BD%D0%B0_%D0%A7%D0%B5%D1%80%D0%BD%D0%B5%D1%86%D0%BA%D0%B0%D1%8F%29.png)
* convert to jxl lossless (the image looks the same)
* convert to jxl quality=96 (the image becomes much darker)

**Expected behavior**
Image brightness should not change upon png->jxl conversion

**Screenshots**
Left half source/jxl_lossless, right half - jxl_q96
Image

**Environment**
 - OS: Windows
 - CPU type: x86_64
 - cjxl/djxl version string: cjxl 0.11.0 4df1e9e [AVX2,SSE2]

### LINKED ISSUE #4511: djxl crushes colors from certain profiles on lossy encodes
cases:  [crushed_decode.zip](https://github.com/user-attachments/files/23434306/crushed_decode.zip)

### Example 1
Image

```
JPEG XL file format container (ISO/IEC 18181-2)
Uncompressed Exif metadata: 318 bytes
Uncompressed xml  metadata: 15090 bytes
JPEG XL image, 599x360, lossy, 8-bit RGB+Alpha
584-byte ICC profile, CMM type: "ADBE", Color space: "RGB ", Rendering intent: 1
```

### Example 2
original PNG
Image

original PNG stripped
Image

cjxl 0.5.0 d1 JXL
Image

```
JPEG XL image, 303x416, lossy, 8-bit RGB
6636-byte ICC profile, CMM type: "appl", Color space: "RGB ", Rendering intent: 0
```

djxl 0.12.0 445e9098 PNG 
Image

djxl 0.12.0 445e9098 PNG stripped
Image

The behavior expresses first in djxl 0.7, which doesn't introduce mitigating dithering.

djxl 0.7.0 PNG
Image

djxl 0.6.1 PNG
Image

djxl 0.6.1 PNG stripped
Image

**Environment**
 - OS: Ubuntu
 - Compiler version: clang 18.1.3
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL decoder v0.12.0 445e9098 [\_AVX2\_,SSE4,SSE2] {Clang 18.1.3}

Galaxy4594:
Basically, it changes the default transfer curve from Linear to sRGB since that's what's usually expected of image viewers and is implied by the docs.

The dithering when decoding to 8-bit was masking the banding, but it's still quite visible since 8-bit linear is quite bad.
all-fixed-now

Fixes #2289
It also addresses these as well:
Fixes #2147
Fixes #4307
Fixes #4511

### DIFF
--- a/lib/jxl/dec_xyb.cc
+++ b/lib/jxl/dec_xyb.cc
@@ -140,7 +140,7 @@ bool CanOutputToColorEncoding(const ColorEncoding& c_desired) {
   return true;
 }
 
-Status OutputEncodingInfo::SetFromMetadata(const CodecMetadata& metadata, bool is_float_output) {
+Status OutputEncodingInfo::SetFromMetadata(const CodecMetadata& metadata) {
   orig_color_encoding = metadata.m.color_encoding;
   orig_intensity_target = metadata.m.IntensityTarget();
   desired_intensity_target = orig_intensity_target;
@@ -158,30 +158,9 @@ Status OutputEncodingInfo::SetFromMetadata(const CodecMetadata& metadata, bool i
             opsin_params.quant_biases);
   bool orig_ok = CanOutputToColorEncoding(orig_color_encoding);
   bool orig_grey = orig_color_encoding.IsGray();
-  ColorEncoding c_out;
-  if (!xyb_encoded || orig_ok) {
-      c_out = orig_color_encoding;
-  } else {
-      // If we are synthesizing the output encoding (XYB -> RGB):
-      c_out = ColorEncoding::SRGB(orig_grey);
-  }
-
-  // If the user requests FLOAT (is_float_output=true), we prefer Linear Light 
-  // to avoid unnecessary gamma correction steps for HDR/Compute.
-  // If the user requests UINT8/16 (is_float_output=false), we MUST use a non-linear 
-  // transfer function (like sRGB gamma) to avoid banding in shadows.
-  if (is_float_output) {
-      c_out.Tf().SetTransferFunction(TransferFunction::kLinear);
-  } else {
-      // Ensure we aren't using Linear for 8-bit. 
-      // If the original was already SRGB/709/PQ, keep it. 
-      // But if it was Linear, force it to sRGB Gamma.
-      if (c_out.Tf().IsLinear()) {
-          c_out.Tf().SetTransferFunction(TransferFunction::kSRGB);
-      }
-  }
-
-  return SetColorEncoding(c_out);
+  return SetColorEncoding(!xyb_encoded || orig_ok
+                              ? orig_color_encoding
+                              : ColorEncoding::SRGB(orig_grey));
 }
 
 Status OutputEncodingInfo::MaybeSetColorEncoding(
--- a/lib/jxl/dec_xyb.h
+++ b/lib/jxl/dec_xyb.h
@@ -68,7 +68,7 @@ struct OutputEncodingInfo {
   bool cms_set = false;
   JxlCmsInterface color_management_system;
 
-  Status SetFromMetadata(const CodecMetadata& metadata, bool is_float_output = true);
+  Status SetFromMetadata(const CodecMetadata& metadata);
   Status MaybeSetColorEncoding(const ColorEncoding& c_desired);
 
  private:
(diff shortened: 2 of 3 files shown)


eustas:
`DecodeTest.IccProfileTestXybEncoded` fails

Galaxy4594:
Tests should be fixed

goodusername123:
The docs *seem* to imply (unless I'm understanding wrong) that `LinearSRGB` be/is used when outputting to a floating point format and that `SRGB` be/is used when outputting to a integer format. however in practice it looks like the same thing ends up getting used for both, meaning that this PR doesn't solve the core issue and instead only shifts the problem around?

https://libjxl.readthedocs.io/en/stable/api_metadata.html#_CPPv4N12JxlBasicInfo21uses_original_profileE
> JXL_BOOL uses_original_profile
> 
> Whether the data in the codestream is encoded in the original color profile that is attached to the codestream metadata header, or is encoded in an internally supported absolute color space (which the decoder can always convert to linear or non-linear sRGB or to XYB). If the original profile is used, the decoder outputs pixel data in the color space matching that profile, but doesn’t convert it to any other color space. If the original profile is not used, the decoder only outputs the data as sRGB **(linear if outputting to floating point, nonlinear with standard sRGB transfer function if outputting to unsigned integers)** but will not convert it to to the original color profile. The decoder also does not convert to the target display color profile. To convert the pixel data produced by the decoder to the original color profile, one of the JxlDecoderGetColor* functions needs to be called with JXL_COLOR_PROFILE_TARGET_DATA to get the color profile of the decoder output, and then an external CMS can be used for conversion. Note that for lossy compression, this should be set to false for most use cases, and if needed, the image should be converted to the original color profile after decoding, as described above.

PR #4517, opened 2025-11-16
examples: require CMake 3.16

novomesk:
Fix for Gentoo QA Notice https://bugs.gentoo.org/964598

I think it is OK to require cmake 3.16 here in examples, because the root CMakeLists.txt from the repo requires 3.16 too.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -5,7 +5,7 @@
 
 # Example project using libjxl.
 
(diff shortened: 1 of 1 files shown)


jonnyawsom3:
I did it a few weeks ago in #4502, still not merged though

novomesk:
Sorry, I did not observe there is PR for this already.

PR #4533, opened 2025-12-06
Allow more Patch predictors

jonnyawsom3:
Previously Patches was hardcoded to use the Gradient predictor, however due to the nature of Patches, it's most likely to contain hard edges making Gradient a poor choice.

This PR enables all predictors except Weighted, effectively behaving like Faster Decoding level 1 to avoid a decode speed penalty.

Encode speed is bottlenecked by the Patch detection anyway, so this should have minimal impact while giving a density bump. Further improvements are likely possible too. 

I'm currently on my phone in London, so I'll add test results when available. Marking as Draft until then.


jonnyawsom3:
Trying all predictors added quite a slowdown, might need to wait until we add more specialised/cut down predictor sets (Eg, something between Gradient+Weighted and All)

Will continue experimenting in this PR when I have time though.

PR #4554, opened 2026-01-09
Fix unsigned underflow in box content decoder pointer adjustment

rootvector2:
## Summary

This PR fixes a memory-safety issue in the JPEG XL container box output path where an inconsistent box-position value could cause unsigned underflow and pointer misadjustment, leading to out-of-bounds reads when copying box payload data into an application-provided output buffer.

## Severity

S1 – out-of-bounds read on untrusted input.

## Affected Components

- Box output / container processing in the decoder
- `JxlBoxContentDecoder::Process` and the decoder box state machine

## Vulnerability Description

The decoder exposes an API for applications to stream container box payloads via `JXL_DEC_BOX` events and `JxlDecoderSetBoxBuffer`.

Internally, `JxlBoxContentDecoder` tracks its consumed position (`pos_`) and is called with a caller-provided `box_pos` indicating the current offset of `next_in` within the box. Previously, pointer adjustment was performed using `pos_ - box_pos` with unsigned arithmetic. If `box_pos` exceeded `pos_`, this subtraction underflowed to a large `size_t`, shifting the input pointer far beyond the provided buffer and enabling out-of-bounds reads.

This inconsistent state could arise if an application continued decoding past a box header without setting a box output buffer and later attempted to start output for that same box.

## Fix Description

- Treats decoding past a box header without setting an output buffer as an explicit opt-out for box output for that box.
- Prevents late buffer setup for a box that has already progressed without output.
- Adds validation in `JxlBoxContentDecoder::Process` to enforce invariants between `box_pos`, `pos_`, and `avail_in`, preventing unsigned underflow and unsafe pointer arithmetic even if invariants are violated.

## Before vs After

**Before:**  
Late `JxlDecoderSetBoxBuffer` could desynchronize internal state and trigger unsigned underflow, resulting in out-of-bounds reads.

**After:**  
Unsafe state transitions are rejected, and positional invariants are enforced to guarantee safe pointer adjustment.

## Testing

- Reviewed box state machine transitions to ensure normal streaming workflows remain unaffected.
- Existing box-related tests continue to pass.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -60,7 +60,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@1b168cd39490f61582a9beae412bb7057a6b2c4e # v3.29.5
+        uses: github/codeql-action/init@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v3.29.5
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
@@ -86,6 +86,6 @@ jobs:
             -DBUILD_TESTING=OFF
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@1b168cd39490f61582a9beae412bb7057a6b2c4e # v3.29.5
+        uses: github/codeql-action/analyze@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v3.29.5
         with:
           category: "/language:${{matrix.language}}"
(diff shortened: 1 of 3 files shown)


eustas:
What will happen if we return failure instead of un-setting `dec->box_event`?

rootvector2:
Returning an error instead of clearing dec->box_event would break valid and documented usage.

The API allows applications to subscribe to JXL_DEC_BOX and skip boxes simply by not calling JxlDecoderSetBoxBuffer. This is already used in the codebase and in tests where only selected box types are captured while others are ignored.

If we returned JXL_DEC_ERROR here then the decoder would fail whenever an application encounters a box it does not want to output.

Clearing box_event keeps the existing meaning that no buffer means skip this box while still preventing the unsafe case this patch fixes which is trying to start box output after decoding has already advanced through the box.

Returning an error would be too strict and would break compatibility while the current approach preserves behavior and fixes the vulnerability safely.

eustas (approved):

Issue #4471, opened 2025-10-06
cjxl fails on animated GIF images with the disposal method undefined


JacobDev1:
**Describe the bug**

`cjxl` fails to transcode animated GIF images with variable frame sizes and the disposal method undefined.

**Steps to reproduce**

Install ImageMagick and libjxl-tools, then run the following:

```bash
convert -delay 100 -loop 0 \
    -size 100x100 xc:red \
    -size 80x80 xc:blue \
    variable_frames_no_dp.gif
cjxl variable_frames_no_dp.gif variable_frames_no_dp.jxl
```

Output:

```bash
JPEG XL encoder v0.11.1 794a5dc [AVX2,SSE4,SSE2]
Getting pixel data failed.
```

Now, if we set `-dispose` to `None` (or `Background`), transcoding will succeed.

```bash
convert -delay 100 -loop 0 -dispose None \
    -size 100x100 xc:red \
    -size 80x80 xc:blue \
    variable_frames_dp.gif
cjxl variable_frames_dp.gif variable_frames_dp.jxl
```

Output:

```bash
JPEG XL encoder v0.11.1 794a5dc [AVX2,SSE4,SSE2]
Encoding [Modular, lossless, effort: 7]
Compressed to 70 bytes (0.028 bpp/frame).
100 x 100,  0.438 MP/s [0.44, 0.44], , 1 reps, 4 threads.
```

**Expected behavior**

Transcoding should be successful.

**Environment**
 - OS: Linux Mint 22.1 Xia
 - CPU type: x86_64
 - cjxl/djxl version string: cjxl v0.11.1 794a5dc \[AVX2,SSE4,SSE2\]
 - the official pre-builds were used from: https://github.com/libjxl/libjxl/releases/download/v0.11.1/jxl-linux-x86_64-static-v0.11.1.tar.gz
 - I also checked on Windows 10 x64 and `cjxl` behaves in the same way so it's not platform-dependent.

**Additional context**

This is a low priority issue so don't stress too much about it. You can fix it by treating the disposal method being `undefined` as `none`.

[images.zip](https://github.com/user-attachments/files/22723709/images.zip)

jonnyawsom3:
Possibly a duplicate of #3134

PR #4571, opened 2026-01-27
Enforce minimum GIF delay for parity with web browsers

FooIbar:
### Description

Many GIFs are mistakenly created with <= 10 ms delays but are meant to be played at 100 ms, this PR enforces a minimum GIF delay to match the behavior of web browsers.
See http://webkit.org/b/26455 for details.

### Pull Request Checklist

### DIFF
--- a/lib/extras/dec/gif.cc
+++ b/lib/extras/dec/gif.cc
@@ -213,6 +213,16 @@ Status DecodeImageGIF(Span<const uint8_t> bytes, const ColorHints& color_hints,
 
   bool replace = true;
   bool last_base_was_none = true;
+  bool follow_browser = true;
+  bool warn_gif_delay = true;
+  JXL_RETURN_IF_ERROR(color_hints.Foreach(
+      [&follow_browser](const std::string& key,
+                        const std::string& value) -> Status {
+        if (key == "gif_follow_browser") {
+          follow_browser = value != "0";
+        }
+        return true;
+      }));
   for (int i = 0; i < gif->ImageCount; ++i) {
     const SavedImage& image = gif->SavedImages[i];
     msan::UnpoisonMemory(image.RasterBits, sizeof(*image.RasterBits) *
@@ -285,7 +295,20 @@ Status DecodeImageGIF(Span<const uint8_t> bytes, const ColorHints& color_hints,
                         total_rect.xsize() == canvas.color.xsize &&
                         total_rect.ysize() == canvas.color.ysize;
     if (ppf->info.have_animation) {
-      frame->frame_info.duration = gcb.DelayTime;
+      // Enforce minimum GIF delay for parity with web browsers.
+      // See http://webkit.org/b/26455 for more information.
+      if (follow_browser && gcb.DelayTime <= 1) {
+        frame->frame_info.duration = 10;
+        if (warn_gif_delay) {
+          fprintf(stderr,
+                  "Warning: Frame delay is <= 10 ms; bumping to 100 ms to "
+                  "match browser behavior. Pass '-x gif_follow_browser=0' to "
+                  "keep the original delay.\n");
+          warn_gif_delay = false;
+        }
+      } else {
+        frame->frame_info.duration = gcb.DelayTime;
+      }
       frame->frame_info.layer_info.have_crop = static_cast<int>(!is_full_size);
       frame->frame_info.layer_info.crop_x0 = total_rect.x0();
       frame->frame_info.layer_info.crop_y0 = total_rect.y0();


goodusername123:
IMO this should really print a warning or similar somewhere (maybe just at the CLI would suffice?) Since this PR would mean that some GIFs are technically not being converted one-to-one anymore.

FooIbar:
> IMO this should really print a warning or similar somewhere (maybe just at the CLI would suffice?) Since this PR would mean that some GIFs are technically not being converted one-to-one anymore.

Well they were never 1:1.
`delay == 0` in GIF was converted to `duration == 0` in JXL, but they are not the same!
In JXL `duration == 0` means the frame won't be presented.

jonnyawsom3:
By that logic, making it only apply to 0 duration GIFs sounds safe to me

FooIbar:
10ms -> 100ms is the de facto standard, and users would expect converted JXLs to play at the same speed as the original GIFs in other software.
I think we need a strong reason to justify introducing a new behavior.

FooIbar:
@eustas PTAL, this is pretty trivial and I hope it can make it into v0.12.

eustas:
Hi. I'm almost fine (though changing existing behavior is a straight way to make harm to someone), but:
 - let's make that configurable; there is `ColorHints` passed; name is confusing, that is more like "decode hints"; we can check if instance contains specific key/value to return previous behavior; e.g. `gif_allow_one_tick_delays`
 - when new behavior takes action, emit (fprintf on stderr) a warning (and explanation how to avoid)

veluca93:
I would prefer to instead have an option to have a minimum delay.

FooIbar:
> Hi. I'm almost fine (though changing existing behavior is a straight way to make harm to someone), but:
> 
> * let's make that configurable; there is `ColorHints` passed; name is confusing, that is more like "decode hints"; we can check if instance contains specific key/value to return previous behavior; e.g. `gif_allow_one_tick_delays`
> * when new behavior takes action, emit (fprintf on stderr) a warning (and explanation how to avoid)

The current behavior for delay = 0 is a bug, so that needs to be changed regardless and what is considered as the previous behavior is unsure.
If `gif_allow_one_tick_delays` is set, should we set the duration to 100 ms or 10 ms when delay is 0?

> I would prefer to instead have an option to have a minimum delay.

I would agree with adding a minimum delay option if we simply clamp the values to that. However, it's a bit trickier here because we have different values for the threshold (10 ms) and the fallback (100 ms) due to historical reasons. Do we need to make both of them configurable?

veluca93:
> > Hi. I'm almost fine (though changing existing behavior is a straight way to make harm to someone), but:
> > 
> > * let's make that configurable; there is `ColorHints` passed; name is confusing, that is more like "decode hints"; we can check if instance contains specific key/value to return previous behavior; e.g. `gif_allow_one_tick_delays`
> > * when new behavior takes action, emit (fprintf on stderr) a warning (and explanation how to avoid)
> 
> The current behavior for delay = 0 is a bug, so that needs to be changed regardless and what is considered as the previous behavior is unsure. If `gif_allow_one_tick_delays` is set, should we set the duration to 100 ms or 10 ms when delay is 0?
> 
> > I would prefer to instead have an option to have a minimum delay.
> 
> I would agree with adding a minimum delay option if we simply clamp the values to that. However, it's a bit trickier here because we have different values for the threshold (10 ms) and the fallback (100 ms) due to historical reasons. Do we need to make both of them configurable?

I am also fine with a flag that basically says "follow browser implementations for GIF and not the GIF spec"

FooIbar:
> I am also fine with a flag that basically says "follow browser implementations for GIF and not the GIF spec"

To strictly follow the GIF spec, we need to merge the 0-delay frames to previous frames, and that would make it non-trivial.

veluca93:
> > I am also fine with a flag that basically says "follow browser implementations for GIF and not the GIF spec"
> 
> To strictly follow the GIF spec, we need to merge the 0-delay frames to previous frames, and that would make it non-trivial.

Isn't that equivalent to having them as JXL frames with 0 duration?

If it isn't, it should be, as otherwise "real" frames will not display correctly

FooIbar:
> > > I am also fine with a flag that basically says "follow browser implementations for GIF and not the GIF spec"
> > 
> > 
> > To strictly follow the GIF spec, we need to merge the 0-delay frames to previous frames, and that would make it non-trivial.
> 
> Isn't that equivalent to having them as JXL frames with 0 duration?
> 
> If it isn't, it should be, as otherwise "real" frames will not display correctly

Actually, you are right. I was wrong about it in https://github.com/libjxl/libjxl/pull/4571#issuecomment-3805863481, the test file I used didn't have them blended because of #4471.

FooIbar:
> Hi. I'm almost fine (though changing existing behavior is a straight way to make harm to someone), but:
> 
> * let's make that configurable; there is `ColorHints` passed; name is confusing, that is more like "decode hints"; we can check if instance contains specific key/value to return previous behavior; e.g. `gif_allow_one_tick_delays`
> * when new behavior takes action, emit (fprintf on stderr) a warning (and explanation how to avoid)

Done

eustas (approved):

PR #4611, opened 2026-02-10
Wire up buffering levels

jonnyawsom3:
Previously `JXL_ENC_FRAME_SETTING_BUFFERING` would always fall back to level 1, causing large speed discrepancies between 1080p and 1440p images, due to the 2048 x 2048 threshold.

This PR allows selecting buffering via cjxl and makes level 2 buffer all images larger than a single group, giving a 8-10x encode speed improvement and 2x memory reduction for 1080p images in some quick testing.

Currently marked as a Draft because I might try to integrate #4580 at the same time, or try to get it merged first to avoid progressive loading regressions.

### DIFF
--- a/lib/jxl/enc_frame.cc
+++ b/lib/jxl/enc_frame.cc
@@ -1768,11 +1768,7 @@ Status PermuteGroups(const CompressParams& cparams,
 bool CanDoStreamingEncoding(const CompressParams& cparams,
                             const FrameInfo& frame_info,
                             const CodecMetadata& metadata,
-                            const JxlEncoderChunkedFrameAdapter& frame_data,
-                            size_t group_size_shift) {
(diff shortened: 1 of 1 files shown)


veluca93:
Please try to integrate #4580 :-)

jonnyawsom3:
Was planning on it, but didn't know if @jonsneyers may want to handle it, as we could do the discussed buffering API change at the same time.
(Levrl 0 buffers the full image, level 1 buffers output and input within 1 group, level 2 only buffers input if it fits in 1 group) 

API edits are a bit more than I'm comfortable with (I'm assuming it's more involved than a normal PR) and I've got no idea why half the builds fail with this PR.

Might be a lot faster and cleaner if an expert spent an hour instead of me spending a week to get it working.

PR #4617, opened 2026-02-10
CI: Replace retired macos-13 runner with macos-15-intel

kleisauke:
macos-13 was retired on December 4, 2025. Switch to macos-15-intel, the last supported Intel runner, available until August 2027.


eustas:
Thanks

kleisauke:
FWIW, the remaining CI failure is likely due to an expired PAT that needs to be refreshed:
https://github.com/libjxl/libjxl/blob/0d2d32fd32bd9cbe3299b051d467e6430024820b/.github/workflows/scorecard.yml#L55
https://github.com/libjxl/libjxl/actions/runs/21864202978/job/63100941707#step:6:245

PR #4619, opened 2026-02-10
(conformance run for ProPhoto)

sboukortt:
Rebased duplicate of #4296 just to try the CI again.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -256,6 +256,7 @@ jobs:
       VCPKG_ROOT: vcpkg
       VCPKG_DISABLE_METRICS: 1
       BUILD_TYPE: ${{ matrix.build_type || 'Release' }}
+      PREFIX: ${{matrix.triplet}}
 
     steps:
     - name: Harden Runner
@@ -315,7 +316,7 @@ jobs:
           -DJPEGXL_ENABLE_TOOLS='ON' \
           -DBUILD_TESTING=OFF \
           -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }} \
-          -DCMAKE_INSTALL_PREFIX=`pwd`/prefix \
+          -DCMAKE_INSTALL_PREFIX=`pwd`/${PREFIX} \
           -DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake \
           -DJPEGXL_ENABLE_JPEGLI_LIBJPEG=OFF \
           -DJPEGXL_ENABLE_OPENEXR=OFF \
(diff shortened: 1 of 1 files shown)


goodusername123:
just in case the logs expire again here's the failing test:
```sh
Testing lossless_pfm
Running: ['/home/runner/work/libjxl/libjxl/djxl', '/home/runner/work/libjxl/libjxl/conformance/testcases/lossless_pfm/input.jxl', '/tmp/lossless_pfmhrhbrcx8/decoded_image.npy', '--metadata_out', '/tmp/lossless_pfmhrhbrcx8/meta.json', '--icc_out', '/tmp/lossless_pfmhrhbrcx8/decoded.icc', '--norender_spotcolors']
JPEG XL decoder v0.12.0 48366e9 [_SSE4_,SSSE3,SSE2] {Clang 18.1.3}
Decoded to pixels.
500 x 500, 10.014 MP/s, 4 threads.
RMSE: [1.6235774e-08, 5.960465e-11, 2.5352236e-08], actual peak: 1.1920928955078125e-07
RMSE too large: 2.535223586619395e-08 > 0.0
```

eustas:
@sboukortt is it know why difference is not 0?

goodusername123:
> @sboukortt is it know why difference is not 0?

It seems like the tests are treating differences in the ICC profile which don't affect rendering as a failure. when doing the test locally and manually checking what's different all I see is that the description and id strings within the ICC profile are different but anything that would affect rendering/pixels like the internal curves or rendering intent are identical. here is both the ICC before and after this PR: [lossless_pfm changed ICC.zip](https://github.com/user-attachments/files/25939346/lossless_pfm.changed.ICC.zip)

There was an [attempt](https://github.com/libjxl/conformance/pull/42) to fix this at the [conformance repo](https://github.com/libjxl/conformance) however the libjxl repo has a local out of sync copy of the testing scripts meaning this change/fix didn't propagate over to the libjxl repo here.


Below is a patch for [conformance.py](https://github.com/libjxl/libjxl/blob/main/tools/conformance/conformance.py) within this repo that seemingly may solve the issue (based on https://github.com/libjxl/conformance/pull/42 and https://github.com/libjxl/conformance/pull/47):
```patch
--- conformance.py
+++ conformance.py
@@ -35,12 +35,14 @@
         return Failure(f'Expected shape {ref.shape} but found {dec.shape}')
     ref_frame = ref[frame_idx]
     dec_frame = dec[frame_idx]
     num_channels = ref_frame.shape[2]
 
-    if ref_icc != dec_icc:
+    if ref_icc != dec_icc and peak_error > 0:
         # Transform colors before comparison.
+        # Skip this if we expect fully lossless, since it introduces tiny errors
+        # (even in the case where ref_icc and dec_icc are equivalent so it's a no-op)
         if num_channels < 3:
             return Failure(f"Only RGB images are supported")
         dec_clr = dec_frame[:, :, 0:3]
         dec_frame[:, :, 0:3] = lcms2.convert_pixels(dec_icc, ref_icc, dec_clr)
```


Alternatively another approach is just changing the test itself (lossless_pfm) within the conformance repo to expect/match the new ICC.

PR #4622, opened 2026-02-12
Upload both 7z and Zip release archives (for Windows)

### LINKED ISSUE #3559: Backslashes in released zip archives paths
**Describe the bug**

In relased zip archives tgere are backslashes in paths. Zip uses forward slash as path separator.

It's a problem add not all un/zip clients assume `\` to be a `/`.

**To Reproduce**

Get the [released](https://github.com/libjxl/libjxl/releases) archives and see the path names, like `cmake\hwy\hwy-config-release.cmake`.

**Expected behavior**

To use forward slashed as path separator in zips.

**Environment**
 - OS: Linux

**Additional context**

[specification](https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.10.TXT) is clear about that:

>   4.4.17.1 The name of the file, with optional relative path.
>  The path stored MUST NOT contain a drive or
>  device letter, or a leading slash.  All slashes
  MUST be forward slashes '/'

eustas:
Based on https://github.com/mo271/libjxl/compare/use_zip...sboukortt:libjxl:use_zip

See also #3852

Fixes #3559

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -256,6 +256,7 @@ jobs:
       VCPKG_ROOT: vcpkg
       VCPKG_DISABLE_METRICS: 1
       BUILD_TYPE: ${{ matrix.build_type || 'Release' }}
+      PREFIX: ${{matrix.triplet}}
 
     steps:
     - name: Harden Runner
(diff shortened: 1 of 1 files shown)


FooIbar:
Is it necessary to use both 7z and ZIP? We won't get much benefit from the smaller 7z file size since they're being uploaded as a single artifact.

eustas:
They are "uploaded" as single artifact - yes, but on "publish" stage they will become separate items (see "Download all artifacts" job).

FooIbar:
Yeah that's why I said we won't get much benefit given the low frequency of releases, compared to the increased size and extra layer in nightly/CI builds.

eustas:
If there is a use case for nightlies, of course we could separate artifacts (that is ~5 lines of copy-paste)

jonnyawsom3:
I think most would agree that separate 7z and zip artifacts would be good for nightlies. I often download static builds for testing, but the 8MB 7z is double-zipped with the original 45MB zip, so it makes downloads slower instead of faster.

eustas:
Got it. Will update CI soon.

PR #4628, opened 2026-02-19
Add "build under qemu"

eustas:
Drive-by: dechipher cron time in scorecard.yml

### DIFF
--- a/lib/extras/compressed_icc.cc
+++ b/lib/extras/compressed_icc.cc
@@ -36,6 +36,9 @@ JXL_BOOL JxlICCProfileEncode(const JxlMemoryManager* memory_manager,
   *compressed_icc_size = bytes.size();
   *compressed_icc = static_cast<uint8_t*>(
       jxl::MemoryManagerAlloc(&local_memory_manager, *compressed_icc_size));
+  if (*compressed_icc == nullptr) {
+    return JXL_FALSE;
+  }
   memcpy(*compressed_icc, bytes.data(), bytes.size());
   return JXL_TRUE;
 }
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -15,7 +15,7 @@ on:
   # To guarantee Maintained check is occasionally updated. See
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
-    - cron: '13 2 * * 3'
+    - cron: '13 2 * * 3' # Daily on 02:37 UTC
```
QrczakMK, line 18: Contents do not match the comment (which I guess was the intended version).

.github/workflows/build_test_emu.yml
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) the JPEG XL Project Authors. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Workflow for building and running tests.
```
QrczakMK: I am sorry, I am not familiar with this language (i.e. YAML semantics for workflows) and I cannot really review this.

In any case, build_test_emu does not seem to be referenced anywhere. Is it picked up by being present in the directory or what?

.github/workflows/scorecard.yml
eustas: Added as drive-by in description.

.github/workflows/build_test_emu.yml
eustas: .github/workflows directory contains descriptors of CI workflows executed by GitHub Actions. 
YML / YAML is JSON extension to "python-like" formatting.
No worries, will add more reviewers.

veluca93: I'd probably change this comment a bit since this is for tests under qemu specifically :-)

eustas: Done

PR #4629, opened 2026-02-20
tools: add overflow and allocation checks in PAM decoder

rootvector2:
### Description

Fix integer overflow and missing NULL check in `DecodePAM()` (`tools/fast_lossless/pam-input.h`).

The multiplication `*w * *h * *nb_chans * bytes_per_sample` can overflow `size_t` when a crafted PAM/PNM file contains large WIDTH/HEIGHT values. This causes an undersized `malloc`, and the subsequent `memcpy` reads out of bounds. Additionally, the `malloc` return value was not checked for NULL.

The fix adds checked multiplication using `SIZE_MAX` division guards before the allocation, and adds a NULL check on the `malloc` result.

No behavior change for valid inputs. Minimal overhead (two integer divisions per decode).


eustas (approved):

eustas:
Thanks!

PR #4631, opened 2026-02-23
fix: heap buffer underflow in jpegli horizontal chroma upsampling

rootvector2:
### Description

Fix heap buffer underflow in jpegli horizontal chroma upsampling.

`Upsample2Horizontal()` writes `scratch_space[-1]` for boundary pixel replication. The `upsample_scratch_` buffer is allocated with 64-float left padding in its size, but the pointer is never advanced past it — so `scratch_space[-1]` writes 4 bytes before the allocation, corrupting `hwy::AllocateAlignedBytes` metadata.

Triggered by default (`do_fancy_upsampling=TRUE`) on any JPEG with horizontal chroma subsampling (4:2:0, 4:2:2).

**Fix:** Advance `upsample_scratch_` by `kPaddingLeft` after allocation, matching the offset pattern used by `RowBuffer::Row()`.

### DIFF
--- a/lib/jpegli/decode.cc
+++ b/lib/jpegli/decode.cc
@@ -534,12 +534,9 @@ void AllocateOutputBuffers(j_decompress_ptr cinfo) {
   }
   m->idct_scratch_ = Allocate<float>(cinfo, 5 * DCTSIZE2, JPOOL_IMAGE_ALIGNED);
   // Padding for horizontal chroma upsampling.
-  constexpr size_t kPaddingLeft = 64;
-  constexpr size_t kPaddingRight = 64;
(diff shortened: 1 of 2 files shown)


eustas:
Nice catch! But I would fix is a little bit differently.

 - remove `kPaddingLeft` / `kPaddingRight` - looks like a local magic number
 - use `2 * HWY_ALIGNMENT / sizeof(float)` instead
 - do not offset value stored in `upsample_scratch_`
 - offset `tmp` derived from it by `HWY_ALIGNMENT / sizeof(float)`

eustas:
Thanks!

PR #4639, opened 2026-02-24
fix upsample in jxl_from_tree

jonsneyers:
Fixes jxl_from_tree producing too large files too slowly when using `Upsample`. It was encoding a frame with double upsampled dimensions (so way too large) which then gets cropped to the image dimensions.


veluca93 (approved):

jonnyawsom3:
I thought it was the already_downsampled logic itself that was broken, not just jxl_from_tree?
Might be worth double checking in case it's causing normal encodes to be larger too.

PR #4643, opened 2026-02-25
Fix OOB read in DequantDC with subsampled chroma and DC context map

rootvector2:
Clamp shifted row index when accessing subsampled channel planes in the num_dc_ctxs > 1 path to prevent heap buffer over-read in release builds.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -226,7 +226,7 @@ jobs:
 
     steps:
     - name: Harden Runner
-      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+      uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
       with:
         egress-policy: audit
 
--- a/.github/workflows/build_test_bazel.yml
+++ b/.github/workflows/build_test_bazel.yml
@@ -32,7 +32,7 @@ jobs:
 
     steps:
       - name: Harden Runner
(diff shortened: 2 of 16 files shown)


eustas:
Hi. Thanks for the discovery and the PR.

You mentioned that it fixes "release" builds, so I assume "debug" builds fail? Likely release builds should reject the input as well (have to check the spec).

eustas:
Also, would be very grateful if you provide repro. Thanks in advance.

rootvector2:
Hi thanks for pointing that out

I went back and looked at it more carefully and I think I need to correct myself After digging into how FrameDimensions Set computes ysize_blocks it looks like when vertical subsampling is enabled r ysize always ends up even Because of that the r ysize minus 1 shift case will not actually exceed the subsampled plane height

So I am not able to produce a real repro that triggers an actual OOB That part in my PR description about release builds was misleading sorry about that

In practice this change is more of a defensive clamp than a fix for something currently reproducible If you think it is unnecessary I am okay to adjust the PR or close it

Thanks for the review and sorry again

eustas:
Let's then turn this PR to "precondition" style. Instead of using `std::min` in loop we can place 3 `JXL_DASSERT` before the loop.

rootvector2:
Updated the branch  replaced the `std::min` clamping with 3 `JXL_DASSERT` preconditions before the loop, one for each channel. The original `Row()` calls are restored as-is. No runtime behavior change in release builds.

eustas (approved):

rootvector2:
@eustas ?

eustas:
Retrying the merge-queue.

rootvector2:
ok

PR #4644, opened 2026-02-28
jpegli: add overflow and OOM checks in memory destination growth

uwezkhan:
This change adds defensive overflow and out-of-memory checks in empty_output_buffer() when growing the in-memory destination buffer.

The buffer size is doubled using buffer_size * 2. On platforms where size_t may overflow, this multiplication could wrap around, resulting in an undersized allocation.

The patch adds:

A guard against buffer_size > SIZE_MAX / 2

A null check for malloc failure

This change does not alter behavior for valid inputs and only improves robustness in extreme edge cases.


eustas:
Hi. Could you please move this PR to https://github.com/google/jpegli. Thanks!

uwezkhan:
Hi. i have raised pr on https://github.com/google/jpegli/pull/205 as suggested. thanks!

PR #4645, opened 2026-02-28
jpegli: guard against size_t overflow in output pixel allocation (32-…

uwezkhan:
This change adds defensive overflow checks in ReadOutputPass() before computing the output pixel buffer size.

On 32-bit platforms, the calculation:

num_samples = output_height * (out_color_components * output_width)

may overflow size_t when processing large JPEG dimensions (e.g., 65535 × 65535 with 4 components). This could result in an undersized allocation.

The patch adds division-based guards to ensure both intermediate and final size computations fit in size_t.

There is no behavior change on 64-bit platforms. This improves robustness and portability for 32-bit builds and other constrained targets.


eustas:
Hi. Could you please move this PR to https://github.com/google/jpegli. Thanks!

PR #4646, opened 2026-03-01
Fix heap buffer overflow in UpsamplingStage temp buffer allocation

rootvector2:
### Description

Fix heap buffer overflow in `UpsamplingStage::PrepareForThreads`.

The temp buffers (`col_min`, `col_max`) used in `PreProcessRowImpl` were allocated as `kChunkSize + 4` floats (1028 elements). However, the SIMD loop that writes to these buffers iterates with bound `len + 4` and stride `Lanes(df)`:

```cpp
for (size_t x = 0; x < len + 4; x += Lanes(df)) {
    Store(min, df, col_min + x);
    Store(max, df, col_max + x);
}
```

When `len == kChunkSize` (1024) and `Lanes(df) >= 8` (AVX2/AVX-512), the last iteration starts at `x = 1024` and writes 8 or 16 floats, reaching up to index 1031 or 1039 — past the end of the 1028-element buffer.

**Fix:** Replace the hardcoded `+ 4` in the allocation with `+ Lanes(df)`, which correctly accounts for the SIMD width on every target. On SSE4 (4 lanes) the allocation size is unchanged. No hot-loop changes, no behavior change.

**Reproduction:** Build with ASAN, encode a ≥2048px-wide image using `cjxl -m 1 -g 3 --resampling=2 -e 1`, decode with `djxl` on an AVX2 machine.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/render_pipeline/stage_upsampling.cc
+++ b/lib/jxl/render_pipeline/stage_upsampling.cc
@@ -127,7 +127,9 @@ class UpsamplingStage : public RenderPipelineStage {
   }
 
   Status PrepareForThreads(size_t num_threads) override {
-    size_t alloc_size = sizeof(float) * (kChunkSize + 4);
+    constexpr HWY_FULL(float) df;
+    size_t alloc_size =
+        sizeof(float) * (kChunkSize + std::max(size_t{4}, Lanes(df)));
     for (size_t i = 0; i < 3; ++i) {
       temp_[i].resize(num_threads);
       for (size_t t = 0; t < num_threads; ++t) {


eustas:
I believe `AlignedMemory::Create` takes care about padding. `+4` here if for different purpose.
Going to try to repro soon.

rootvector2:
Thanks for taking a look!

You're absolutely right that the `+4` corresponds to the 5-tap window used by the filter.

What I observed is that the SIMD loop processes `len + 4` elements with a stride of `Lanes(df)`:

`for (size_t x = 0; x < len + 4; x += Lanes(df))`


When `len == kChunkSize` (1024) and `Lanes(df) >= 8` (AVX2/AVX-512), the final iteration occurs at `x = 1024`.  
The `Store()` then writes `Lanes(df)` floats starting at `col_min + 1024`.

That reaches indices:

- AVX2 → `[1024..1031]`
- AVX-512 → `[1024..1039]`

while the buffer contains `[0..1027]`.

You're correct that `AlignedMemory::Create` allocates `size + kAlias`, so the write likely lands inside the padding region and doesn't corrupt neighboring allocations in practice.

My intention with the change was simply to size the buffer according to the maximum SIMD write width (`kChunkSize + Lanes(df)`), making the allocation match the loop's maximum store extent.

Happy to adjust the fix if there is a preferred approach.

eustas:
There are 2 things to take into account for 100% correctness:
 - for short vectors (in extreme - scalar) we still need 4 floats
 - there is utility function for taking full vector length (in bytes) - `MaxVectorSize` in `simd_util.h`

rootvector2:
Good catch thanks. Updated now using `std::max(size_t{4}, Lanes(df))` so the +4 floor is preserved on scalar/short-vector targets.

Issue #4652, opened 2026-03-06
CMake fails to detect system-installed Highway library in MXE cross-compilation setups


hkunz:
**Describe the bug**
When cross-compiling JPEG XL (`libjxl`) with MXE, CMake fails to find a system-installed Highway library even though `libhwy.a` and headers exist. `JPEGXL_FORCE_SYSTEM_HWY` was set to `ON` to use the MXE-installed Highway library, but CMake failed to detect it unless the CMakeLists.txt is patched to explicitly point to the `HWY_INCLUDE_DIR` and `HWY_LIBRARY`. Problem also happens with `libbrotlicommon`, and the rest of the dependencies

```
[a@AKIKO-SYS]:~/workspace/mxe.git [pkg/openimageio] $ find usr/ -type f | grep -i hwy
usr/x86_64-w64-mingw32.static/include/hwy/print.h
usr/x86_64-w64-mingw32.static/include/hwy/foreach_target.h
usr/x86_64-w64-mingw32.static/include/hwy/base.h
usr/x86_64-w64-mingw32.static/include/hwy/abort.h
usr/x86_64-w64-mingw32.static/include/hwy/auto_tune.h
usr/x86_64-w64-mingw32.static/include/hwy/cache_control.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/set_macros-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/emu128-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/loongarch_lasx-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/x86_128-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/loongarch_lsx-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/ppc_vsx-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/x86_512-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/x86_256-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/arm_neon-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/generic_ops-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/inside-inl.h
usr/x86_64-w64-mingw32.static/include/hwy/ops/shared-inl.h
[...]

PR #4655, opened 2026-03-07
Use CONFIG mode for finding LCMS2 (#4652)

hkunz:
Change `find_package(LCMS2 2.12)` to `find_package(LCMS2 2.12 CONFIG)` to prefer the CMake config file over the system search. This addresses and completes the fix for #4652.

This ensures that the build uses the correct LCMS2 package provided via CMake, avoiding potential “library not found” errors

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -226,7 +226,7 @@ jobs:
 
     steps:
     - name: Harden Runner
-      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+      uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
       with:
         egress-policy: audit
 
--- a/.github/workflows/build_test_bazel.yml
+++ b/.github/workflows/build_test_bazel.yml
@@ -32,7 +32,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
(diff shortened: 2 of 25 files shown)


eustas:
But this does not work without config file then (https://github.com/libjxl/libjxl/actions/runs/22796012724/job/66273639766?pr=4655).
Perhaps there is some middle-ground solution that works both ways...

hkunz:
> But this does not work without config file then (https://github.com/libjxl/libjxl/actions/runs/22796012724/job/66273639766?pr=4655). Perhaps there is some middle-ground solution that works both ways...

Thanks so much for taking a look! Since it seems tricky on your side, I’ll just apply a small patch in our Makefile. It’s a simple change and shouldn’t cause any issues, this approach works fine for us.

```
PKG             := jxl
$(PKG)_WEBSITE  := https://github.com/libjxl/libjxl
...
define $(PKG)_BUILD
	tmpfile=$(mktemp) || exit 1
	sed 's/find_package(LCMS2 2.12)/find_package(LCMS2 2.12 CONFIG)/' \
		"$(SOURCE_DIR)/third_party/CMakeLists.txt" > "$tmpfile" \
		&& mv "$tmpfile" "$(SOURCE_DIR)/third_party/CMakeLists.txt"
	cd "$(BUILD_DIR)" && \
	'$(TARGET)-cmake' "$(SOURCE_DIR)" \
		-DCMAKE_INSTALL_PREFIX='$(PREFIX)/$(TARGET)' \
		....
```

kmilos:
> Perhaps there is some middle-ground solution that works both ways...

Usually one rewrites the local FindLCMS2.cmake module to attempt searching in CONFIG mode first, then fall back to the "old" pkgconf way if unsuccessful. See [this](https://github.com/strukturag/libheif/blob/master/cmake/modules/FindOPENJPH.cmake) for example. See also the (much more involved) FindTIFF shipped by CMake itself.

PR #4656, opened 2026-03-07
Revert inverse Gaborish to original values

jonnyawsom3:
Gives a small density improvement ~0.5%, reduces ringing and slightly improves color accuracy.
Should also help generation loss.

Worth exploring different values in future, but for now the original values from a few years ago give better results.


jonnyawsom3:
Not sure this helps as much as I thought... Think it's just affecting block selection by changing edge cases

PR #4657, opened 2026-03-09
Remove jpegli

eustas:


eustas:
Fuzzer config have to be updated on oss-fuzz side.

Jaldebehot:
In my humble opinion, this is the opposite direction libjxl should be going. The jpegli code seems to share a significant amount of essential code in common with libjxl, and it is a very tricky burden to maintain two separate parallel copies of the same files in different repositories with different git trees.

If you are doing this because of package management on Linux, this is what "split packages" are for. Basically, you need to modify the build system to allow jpegli to be installed in its own separate installation command. So that package maintainers can make separate split packages for libjxl libraries and jpegli libraries.

Just make it more convenient for packagers by explaining in the README and/or BUILDING file, the instructions for how to install/copy the libjxl and jpegli libraries in separate steps or separate "make install" invocations. Then packagers can make split packages while only needing to download and compile the code once. And you guys have much less maintenance burden on your shoulders, instead of parallel git trees causing confusion.

mo271 (approved):
Nice!

Crystal-RainSlide:
This puts an end to any official-ish jpegli binary release so far, since https://github.com/google/jpegli had never made any release yet.

https://github.com/google/jpegli/issues/44

PR #4663, opened 2026-03-09
Fix memory leak in GetContext() in jxl_cms.cc

### LINKED ISSUE #4166: Memory leak in jxl_cms.cc
There is a memory leak in the `GetContext` function in jxl_cms.cc. This function uses a `static thread_local` variable to allocate a CMS "context" object, by calling `cmsCreateContext`. But the code never invokes `cmsDeleteContext`, so one of these objects is being leaked for each thread.

The problematic code looks like this:

```
cmsContext GetContext() {
  static thread_local void* context_;
  if (context_ == nullptr) {
    context_ = cmsCreateContext(nullptr, nullptr);
```

I have tried out a fix using `std::unique_ptr` that seems to work fine.

```
cmsContext GetContext() {
  static thread_local std::unique_ptr<void, void(*)(void*)> context_(nullptr, [](void* ptr)
  {
    if (ptr != nullptr) cmsDeleteContext(static_cast<cmsContext>(ptr));
  });
  if (context_ == nullptr) {
    context_.reset(cmsCreateContext(nullptr, nullptr));
    JXL_DASSERT(context_ != nullptr);

    cmsSetLogErrorHandlerTHR(static_cast<cmsContext>(context_.get()), &ErrorHandler);
  }
  return static_cast<cmsContext>(context_.get());
}
```
If you agree, I would like someone to create a PR to submit this fix.

rootvector2:
### Description

Fix memory leak in `GetContext()` in `jxl_cms.cc` where `cmsCreateContext` allocates a CMS context per thread but `cmsDeleteContext` is never called on thread exit.

Replaced raw `void*` with `std::unique_ptr` using a custom `ContextDeleter`, following the existing RAII pattern (`ProfileDeleter`, `TransformDeleter`, `CurveDeleter`) in the same file.

Fixes #4166

### DIFF
--- a/lib/jxl/cms/jxl_cms.cc
+++ b/lib/jxl/cms/jxl_cms.cc
@@ -805,7 +805,7 @@ cmsContext GetContext() {
     context_.reset(cmsCreateContext(nullptr, nullptr));
     JXL_DASSERT(context_ != nullptr);
 
-    cmsSetLogErrorHandlerTHR(static_cast<cmsContext>(context_.get()),&ErrorHandler);
+    cmsSetLogErrorHandlerTHR(static_cast<cmsContext>(context_.get()), &ErrorHandler);
   }
   return static_cast<cmsContext>(context_.get());
 }


eustas:
BTW, the other place where cmsContext is created would need similar change (`load_jxl.cc`)... But there it is wildly wrong - one context is shared between multiple threads... Creating an issue now, could be addressed later.

rootvector2:
> LGTM, modulo missing space in line 808.

done

eustas:
Thanks!

Issue #4309, opened 2025-06-26
Improve JPEG transcoding error reporting


eustas:
If something goes wrong it just says "Possibly there is too much tail data."

Report should be more certain. Because some problems could be fixed by user (e.g. drop tail data), some are impossible to overcome (e.g. too many reset_points for JXL).

PR #4665, opened 2026-03-10
Improve JPEG transcoding error reporting for JBRD failures

rootvector2:
Improve JPEG transcoding error reporting by returning more specific errors for common JBRD failure cases (e.g., excessive tail data, too many reset points, or markers).

https://github.com/libjxl/libjxl/issues/4309


rootvector2:
@eustas

PR #4668, opened 2026-03-10
Fix stdin evaluation if file "-" exists

stoeckmann:
### Description

If the file `-` exists in file system, wrong information is stored in _size if `-` is a regular file.

Fix this by always using the file descriptor.

You can reproduce it on Linux like this:

1. Create a `-` file in current directory
```
echo > -
```
2. Run `djxl` with one of the example files in main branch
```
djxl - output.png < testdata/jxl/splines.jxl
```
```
JPEG XL decoder v0.12.0 88840bc7 [_AVX2_,SSE4,SSE2] {GNU 15.2.1}
Input file is truncated (total bytes: 1, processed bytes: 0) and --allow_partial_files is not present.
```

You can see that only 1 byte is read although the file is perfectly valid. If you remove the `-` file, the command works without issues:
```
rm ./-
djxl - output.png < testdata/jxl/splines.jxl
```
```
PEG XL decoder v0.12.0 88840bc7 [_AVX2_,SSE4,SSE2] {GNU 15.2.1}
./lib/jxl/splines.cc:676: Large total_estimated_area_reached, expect slower decoding: 244616490
Decoded to pixels.
2048 x 2048, 0.985 MP/s, 14 threads.
```

The issue is not visible on Windows because `_setmode` on stdin seems to always fail. At least on my test system with msys2 (ucrt).

Shoutout to [Let's Read OSS](https://github.com/stoeckmann/lets-read-oss).

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed / clarified
+ - decoder API: timeframes to successfully invoke `JxlDecoderSetImageOutBuffer`
+   and `JxlDecoderSetPreviewOutBuffer` are non-intersecting; it is not possible
+   to accidentally set one buffer when the other is expected
+
+## [0.11.2] - 2026-02-10
+
 ### Fixed
   - Corrupted images when using effort 1 lossless. (#4027)
   - Extremely tall/wide images failed to encode using modular. (#3937)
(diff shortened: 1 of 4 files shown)


eustas:
Thanks for finding this out.
As I can see from code, there is more straightforward way to fix the problem: use `close_on_delete_` (not the most clear name) to decide whether to use `stat` result (i.e. `if (!close_on_delete_ && err == 0 && is_file) {`).

stoeckmann:
I'd recommend the `fileno` approach for four reasons:

1. It's already used partially for the Windows code (weak reason)
2. Calling `fopen` and then `stat` "for the same file" is not necessarily the same file anymore. It's a typical TOCTOU pattern and could allow races. I doubt that this is relevant for the tools here, but it explicitly states that we really want to get information about the opened file, not about a file which hopefully is the one we've meant before as well
3. The best reason to keep it that way is that a file redirection to stdin as shown with the Linux example allows the program to read the whole data at once because the actual file size is returned. Otherwise it proceeds in 16 KB chunks due to unknown size. This leads to fewer system calls with large files
4. Last but not least: If we call `stat`, we should use the results if possible. Or adjust the code to avoid the stat call if not required :)

PR #4675, opened 2026-03-13
Fix potential overflow and null allocation in BitWriter::Allocate

sayedkaif27:
This PR improves memory safety in BitWriter::Allocate(size_t maximum_bit_size) by adding defensive checks for allocation size calculation and memory allocation.

Previously, the allocation size was computed as:
maximum_bit_size / 8 + 64

without validating whether the calculation could overflow size_t. In extreme cases, this could result in an incorrect allocation size and potentially lead to memory safety issues.

This patch introduces the following improvements:

Adds validation to ensure the allocation size computation cannot overflow.

Checks the result of malloc() and throws std::bad_alloc() if allocation fails.

Improves robustness by preventing potential null pointer dereference.

These changes improve defensive programming and memory safety without affecting normal behavior for valid inputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed / clarified
+ - decoder API: timeframes to successfully invoke `JxlDecoderSetImageOutBuffer`
+   and `JxlDecoderSetPreviewOutBuffer` are non-intersecting; it is not possible
+   to accidentally set one buffer when the other is expected
+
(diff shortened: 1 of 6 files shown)


sayedkaif27:
@stoeckmann following up on our discussion from two weeks ago. While the overflow is technically avoided by the division I still think adding the explicit `malloc` failure handling improves the robustness. 
let me know if you would like me to refine the PR or if there are other changes needed to move this forward

eustas:
libjxl is "exception-transparent", which means we do not raise exceptions and do not catch them.
resolved malloc failures in separate PR (that involves error propagation for ~3 levels)

PR #4676, opened 2026-03-15
Add overflow guard in GetMinSize buffer size calculation

rootvector2:
Adds a small overflow check when computing the output buffer size in GetMinSize().

Although earlier dimension checks already limit image sizes, this guard ensures the multiplication and addition used to compute the final buffer size cannot overflow size_t.

This does not change behavior for valid images and simply adds an extra defensive check.


rootvector2:
@eustas

eustas:
Hi. Please consider using `SafeAdd` and `SafeMul` added recently  (#4685).

eustas:
Thanks.

eustas (approved):

PR #4682, opened 2026-03-18
Add overflow checks in GetMinSize buffer size calculation

sayedkaif27:
## Summary
Add safety checks to prevent overflow while calculating the minimum output buffer size.

## What changed
- Added checked multiplication for:
  - width × channels
  - result × bits
- Added checked multiplication for row size and height
- Added checked addition for final buffer size
- Added validation for zero image dimensions

## Why
The buffer size is calculated using multiple arithmetic operations on values derived from image metadata.

For very large or malformed inputs, these operations could overflow and produce an incorrect (smaller) size. This could lead to incorrect buffer allocation.

These checks make sure the calculation fails safely instead of overflowing.

## Impact
- Improves robustness for edge cases and malformed inputs
- No impact on normal image decoding
- Keeps behavior unchanged for valid inputs

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -221,7 +221,7 @@ jobs:
 
     steps:
     - name: Harden Runner
-      uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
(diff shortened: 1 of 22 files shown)


sayedkaif27:
@eustas any update

eustas:
Hello. Please rebase and use `Safe{Add,Mul}` from `base/common.h`

eustas:
Covered by #4676

PR #4683, opened 2026-03-18
Add null check after malloc in JxlFastLossless output allocation

sayedkaif27:
-Adds a null check after malloc for the output buffer.
-Added a check to verify if memory allocation succeeds
-Return failure if malloc returns NULL

If malloc fails, *output becomes NULL and is later used in the write loop, which can lead to a crash. This check prevents that.


eustas (approved):

eustas:
Thanks

PR #4684, opened 2026-03-18
Potential underflow brotli box decoding

sayedkaif27:
Fixes a potential unsigned underflow in the Brotli box decoding path.

After BrotliDecoderDecompressStream, the number of bytes consumed was
subtracted from `remaining_` without validating that `consumed <= remaining_`.

A malformed or crafted Brotli stream could report consuming more bytes than
available, leading to unsigned underflow and inconsistent decoder state.

This patch adds a guard:

  if (consumed > remaining_) return JXL_DEC_ERROR;

so the decoder fails safely instead of continuing with corrupted state.


sayedkaif27:
@eustas any update

eustas:
Hi. Thanks for your PR.

Fix should be a bit more complex. Instead of checking that brotli decoder over-consumed we should restrict its input (similar to `can_read` for non-compressed case below). That way we guarantee `remaining_` won't underflow.
With that in place we should check that `remaining_` is zero before returning `JXL_DEC_BOX_COMPLETE` (success).

sayedkaif27:
> Fix should be a bit more complex. Instead of checking that brotli decoder over-consumed we should restrict its input (similar to `can_read` for non-compressed case below). That way we guarantee `remaining_` won't underflow. With that in place we should check that `remaining_` is zero before returning `JXL_DEC_BOX_COMPLETE` (success).

I implemented the fix in `box_content_decoder.cc` so brotli decoding is now restricted to `remaining_` when `box_until_eof_` is false, instead of only checking after consumption. I also added a final guard that `remaining_ == 0` before returning `JXL_DEC_BOX_COMPLETE`

lib/jxl/decode_test.cc
```diff
@@ -2217,6 +2218,28 @@ TEST(DecodeTest, ExtraBytesAfterCompressedStream) {
...
+TEST(DecodeTest, BrobBoxInputIsRestrictedToBoxSize) {
+  const uint8_t box_contents[] = {
+      'o', 'r', 'i', 'g',
+      0x0f, 0x05, 0x80, 0x68, 0x65, 0x6c, 0x6c, 0x6f,
+      0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64, 0x03};
+  const uint8_t extra_bytes[] = {0xAA, 0xBB, 0xCC};
+  std::vector<uint8_t> input;
+  input.insert(input.end(), std::begin(box_contents), std::end(box_contents));
+  input.insert(input.end(), std::begin(extra_bytes), std::end(extra_bytes));
+  uint8_t out[32] = {};
```
eustas: Let's move that on heap, e.g. use vector.

sayedkaif27: Updated `decode_test.cc` to use a heap-backed buffer: `std::vector<uint8_t> out(32)` instead of `uint8_t out[32]`, and switched the pointer logic to `out.data()`. This removes the fixed stack allocation

lib/jxl/box_content_decoder.cc
```diff
@@ -70,26 +70,32 @@ JxlDecoderStatus JxlBoxContentDecoder::Process(const uint8_t* next_in,
...
     size_t consumed = next_in - next_in_before;
     size_t produced = *next_out - next_out_before;
     if (res == BROTLI_DECODER_RESULT_ERROR) {
       return JXL_DEC_ERROR;
     }
     msan::UnpoisonMemory(next_out_before, produced);
     pos_ += consumed;
-    if (!box_until_eof_) remaining_ -= consumed;
+    if (!box_until_eof_) {
+      if (consumed > remaining_) return JXL_DEC_ERROR;
```
eustas: Brotli decoder will not consume more than provided (`avail_in_brotli`) which in turn limited by `remaining_`. So, we don't need this check.

lib/jxl/box_content_decoder.cc
```diff
@@ -70,26 +70,32 @@ JxlDecoderStatus JxlBoxContentDecoder::Process(const uint8_t* next_in,
...
+      remaining_ -= consumed;
+    }
     if (res == BROTLI_DECODER_RESULT_NEEDS_MORE_INPUT) {
       return JXL_DEC_NEED_MORE_INPUT;
```
eustas, line 94: On success you check that `remaining_` is depleted; thus it will be logical to check on `need more input` that `remaining_` is not zero - otherwise we don't have anything to provide in the next iteration... returning `JXL_DEC_ERROR` seems to be appropriate.

sayedkaif27:
Implemented the review:

- Updated `box_content_decoder.cc` to remove the redundant `consumed > remaining_` check.
- Added the error path for `BROTLI_DECODER_RESULT_NEEDS_MORE_INPUT` when `!box_until_eof_` and `remaining_ == 0`
- Added regression test `DecodeTest.BrobBoxTruncatedStreamWithExactBoxSizeReturnsError` in `decode_test.cc`

eustas (dismissed):

sayedkaif27:
I have addressed the compilation error in `decode_test.cc` by qualifying `JxlBoxContentDecoder` with the `jxl::` namespace. This should resolve the build failure.

eustas (approved):

PR #4687, opened 2026-03-19
Fix ParseUnsigned accepting negative values and missing validation

metsw24-max:
ParseUnsigned currently accepts inputs like "-1" due to strtoull behavior, 
which converts them into large unsigned values. These values can propagate 
into loop bounds and allocation sizes in CLI tools.

This patch:
- Rejects negative inputs (including those with leading whitespace)
- Handles nullptr safely in error reporting
- Checks errno for overflow 
- Ensures full string consumption
- Prevents truncation when converting to size_t

This makes parsing stricter and avoids unintended large values from invalid input.
No behavior change for valid inputs.


metsw24-max:
@yoshiyoshyosh 
@eustas 

I have updated ParseUnsigned to keep only the meaningful checks (ERANGE, full-string consumption, representability in size_t, and explicit negative-value rejection) and removed the redundant null/whitespace-only handling.

PR #4688, opened 2026-03-19
Fix ParseSigned missing overflow and validation checks

metsw24-max:
ParseSigned currently does not validate overflow/underflow and may silently
truncate values when casting to int.

This patch:
- Handles nullptr safely in error reporting
- Rejects whitespace-only inputs
- Checks errno for overflow/underflow (ERANGE)
- Ensures full string consumption (no partial parsing)
- Validates range before casting to int

No behavior change for valid inputs.

### DIFF
--- a/tools/cmdline.h
+++ b/tools/cmdline.h
@@ -368,12 +368,17 @@ class CommandLineParser {
 //
 
 static inline bool ParseSigned(const char* arg, int* out) {
+  // Parse() only passes argv entries while i < argc, so arg is non-null.
   char* end;
-  *out = static_cast<int>(strtol(arg, &end, 0));
-  if (end[0] != '\0') {
+  errno = 0;
+  const long value = strtol(arg, &end, 0);
+  if (errno == ERANGE || end == arg || end[0] != '\0' ||
+      value < std::numeric_limits<int>::min() ||
+      value > std::numeric_limits<int>::max()) {
     fprintf(stderr, "Unable to interpret as signed integer: %s.\n", arg);
     return false;
   }
+  *out = static_cast<int>(value);
   return true;
 }
 


eustas:
Hi @yoshiyoshyosh. Could you explain, why "thumbs down"? Thanks.

yoshiyoshyosh:
@eustas There's a few things that struck me:
- The null check introduced is useless. `ParseSigned()` is only invoked during options parsing, which [already loops until `i` < `argc`](https://github.com/libjxl/libjxl/blob/a3531f8cf336a70fcc8caa549d847bed4f4e7b4d/tools/cmdline.cc#L62). There is no possible way null could be passed to the function in in the program's current state.
- There's no reason for the introduced code to strip leading white space itself. [strtol already does that](https://en.cppreference.com/w/cpp/string/byte/strtol.html). Checking `if (nptr == endptr)` already rejects whitespace-only inputs, as such input is by definition malformed.

The only useful new bits here are:
- clearing `errno` and checking `ERANGE` after the function
- checking the cast from long -> int to ensure it's representable as an int (it's undefined behavior otherwise)
- checking if the full string was consumed (this was already present in the code before, though)

Overall, given the account's history and contributions, it reads like it's using LLMs to scan for "easy footguns" while neglecting the wider context of the code and the intricacies of the language.

I have similar grievances with [the other PR opened from the account](https://github.com/libjxl/libjxl/pull/4687)

eustas:
Thanks for explanations! Yes nowadays repos are flooded with LLM-generated content, lets hope situation will get normalized in the future.
Generally, I'm not standing against overcautious checks (if it is not obvious and not in performance critical path). But here yes, we can have a comment instead of nullptr check; and that eliminates the extra entity: `printable_arg`.

@metsw24-max could you please address the comments and update this PR. Thanks.

metsw24-max:
@eustas 
Removed the null check and the printable_arg helper.

Removed manual leading-whitespace stripping.

Kept the useful checks:
- errno reset + ERANGE check
- full-string consumption (end == arg / end[0] != '\0')
- int range validation before cast

eustas:
Please rebase (conflicts with previous PR that has landed).

metsw24-max:
@eustas i rebased this

metsw24-max:
@eustas any updates on this?

eustas:
Thanks!

metsw24-max:
@eustas any update on this pr?

eustas (approved):

PR #4692, opened 2026-03-19
Fix thread-unsafe cmsContext in viewer's load_jxl.cc

### LINKED ISSUE #4664: BTW, the other place where cmsContext is created would need similar change (`load_jxl.cc`)... But there it is wildly wrong - one context is shared between multiple threads... Creating an issue now, could be addressed later.
BTW, the other place where cmsContext is created would need similar change (`load_jxl.cc`)... But there it is wildly wrong - one context is shared between multiple threads... Creating an issue now, could be addressed later.

_Originally posted by @eustas in https://github.com/libjxl/libjxl/issues/4663#issuecomment-4029433264_

eyupcanakman:
### Description

Same issue as #4663 but in `tools/viewer/load_jxl.cc`. The cmsContext was thread_local but never freed. Wrapped it in a static thread_local unique_ptr with the same RAII deleter pattern.

Fixes #4664

### Pull Request Checklist

- [x] **CLA Signed**
- [ ] **Authors**
- [x] **Code Style**

### DIFF
--- a/tools/cmdline.h
+++ b/tools/cmdline.h
@@ -6,10 +6,13 @@
 #ifndef TOOLS_CMDLINE_H_
 #define TOOLS_CMDLINE_H_
 
+#include <cctype>
+#include <cerrno>
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <limits>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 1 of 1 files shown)


tools/viewer/load_jxl.cc
```diff
@@ -115,22 +124,26 @@ QImage loadJxlImage(const QString& filename, const QByteArray& targetIccProfile,
   EXPECT_EQ(JXL_DEC_FULL_IMAGE, JxlDecoderProcessInput(dec.get()));
 
   auto uint16_pixels = std::make_unique<uint16_t[]>(pixel_count * 4);
-  const thread_local cmsContext context = cmsCreateContext(nullptr, nullptr);
+  static thread_local CmsContextUniquePtr context;
+  if (context == nullptr) {
```
eustas: style: `if (!context) {`

tools/viewer/load_jxl.cc
```diff
@@ -115,22 +124,26 @@ QImage loadJxlImage(const QString& filename, const QByteArray& targetIccProfile,
...
+    context.reset(cmsCreateContext(nullptr, nullptr));
+  }
   EXPECT_TRUE(context != nullptr);
```
eustas: Let's move this one line below and make `EXPECT_TRUE(ctx != nullptr);`

PR #4699, opened 2026-03-27
Improve default progressiveness of -p

jonnyawsom3:
Changing the progressive AC mode results in an extra HF pass while improving bpp slightly (On average 2% smaller) and giving a minor decode speed improvement (3.5% faster).

The new passes are also a smoother transition from the 1:8 LF, while more coarse for before-last HF pass, making it easier to tell when a file is fully loaded. (A topic discussed during jxl-rs work, images from Chrome Canary)

| | LF | HF Pass 1 | HF Pass 2 | HF Final |
|-|---|------------|------------|----------|
Old | image | | image | image
New | image | image | image | image


jonsneyers:
Looking at this sequence of previews, maybe there's something that could be done on the decoder side too. What bothers me is that the LF preview is using nice nonseperable 8x upsampling, but then as soon as there is any HF data, it's just iDCT with many coeffs missing, which produces a blocky/pixelated preview.

What would make conceptual sense is to use the nonseperable 4x upsampling when we have LF available and only the lowest freq HF coefficients, i.e. partial image data that corresponds to a 1:4 image. That would be hard to implement though, but maybe there's some other way to get a similar effect. Maybe something like adjusting Gaborish weights during previews so more blurring is done when the highest freq HF is not yet available.

jonnyawsom3:
I agree, it would be nice if there were a more gentle transition into HF data, but for now this PR avoids the sudden pop from 1:8 to visually completed.

jonnyawsom3:
I just remembered I already had these videos rendered out, so here's a 'live' comparison

https://github.com/user-attachments/assets/579b3f92-ed22-4183-8a65-b4b4d57683e5

https://github.com/user-attachments/assets/98a276eb-7a85-4d0a-bcd0-10306dab48e3

While not as smooth as the LF with it's non-separable upsampling, the earlier pass means a much less jarring transition

jonnyawsom3:
> What would make conceptual sense is to use the nonseperable 4x upsampling when we have LF available and only the lowest freq HF coefficients, i.e. partial image data that corresponds to a 1:4 image. That would be hard to implement though, but maybe there's some other way to get a similar effect. Maybe something like adjusting Gaborish weights during previews so more blurring is done when the highest freq HF is not yet available.

Maybe it could be done alongside the 'reduced resolution' API?
It's long been discussed using the 1:8 LF for thumbnails/previews, but the only output available currently is already upsampled.
If we wanted to add an API for 1:8, 1:4 and 1:2 sized outputs, then we'd already have the 'native' sized buffers to use fancy upsampling on (if it's not too expensive).

jonnyawsom3:
The PR has now been updated to include a few fixes/improvements for Alpha/Lossless too.

Using `progressive_ac` or `qprogressive_ac` now enables squeeze, as images with Alpha wouldn't render the passes.
Lossless will now always prefer `qprogressive_ac` over `progressive_ac`, as loading is identical but makes smaller files.
If the image contains extra channels, `progressive_dc` will be disabled as it's not implemented, saving ~10% on density.
Patches are disabled by default to give a smoother loading experience and avoid a deadzone between LF and HF.
Default `-p` settings can now be overridden.

tools/cjxl_main.cc
```diff
@@ -782,13 +782,22 @@ void ProcessFlags(const jxl::extras::Codec codec,
...
   if (args->progressive) {
-    args->qprogressive_ac = true;
+    // progressive_ac and qprogressive_ac should be made into
+    // a single parameter like progressive_dc to allow overriding.
+    args->progressive_ac = true;
     if (args->progressive_dc == -1) {
       args->progressive_dc = 1;
     }
+    if (args->group_order == jxl::Override::kDefault) {
     args->group_order = jxl::Override::kOn;
```
veluca93: please fix the formatting :-)

veluca93 (dismissed):

veluca93 (approved):

PR #4707, opened 2026-04-02
check allocation failure before memcpy in ICC compression path

metsw24-max:
This change adds a null check after `MemoryManagerAlloc` in the ICC compression path before invoking `memcpy`
This patch ensures that:
- allocation failure is detected
- the function returns JXL_FALSE instead of dereferencing a null pointer


metsw24-max:
@eustas look into this

eustas:
Thanks, sorry for the late response.

eustas (approved):

PR #3332, opened 2024-02-26
Cap the number of reference properties to the # of extra channels.

veluca93:
Fixes #3324.

PR #4729, opened 2026-04-15
allow -E to be used in jpeg transcodes again

Galaxy4594:
This PR (#3332) was meant to cap the number of `max_properties` (`-E`) from going wildly out of bounds during MA tree evaluations. It introduced a clamp on `max_properties` to restrict the number of properties if they exceed the number of channels in the image. 

It was originally written like this:
```cpp
(frame_header.encoding == FrameEncoding::kModular ? 2 : -1));
```
This clamp is too aggressive. It assumes that if a frame isn't encoded in full Modular mode (`kModular`), it must be `VarDCT` (lossy) and thus has 0 Modular color channels available (therefore the `-1` to restrict previous channels). 

Lossless JPEG transcodes still rely on the Modular encoder for compressing the LF. The LF image (I think) contains at most 3 channels. (XYB or YUV)
Because of this flawed logic, the encoder evaluates `(kVarDCT == kModular ? 2 : -1)` which returns `-1`. By forcing `max_properties` to `-1`, it silently kills the `-E` flag for JPEG transcodes (VarDCT can still utilize the `-E` flag somehow).

The fix I've applied checks if the image is grayscale, regardless of whether the system frames it as "Modular" or "VarDCT", substituting the condition with:

```cpp
(frame_header.nonserialized_metadata->m.color_encoding.IsGray() ? 0 : 2));
```

I think this accommodates what the encoder knows about its channels. By ensuring `--modular_nb_prev_channels` isn't aggressively clamped to `-1`, the MA tree once again can take advantage of cross-channel properties for LF compression, bringing it back to how it was since libjxl 0.9.

Edit: Turns out this issue only affected jpeg transcodes. Here are some numbers:

Settings|Main (kB)|PR (kB)|Δ (%)|
--------|---------|--------|--------
Effort 8 -E 2|3539.2|3534.7|-0.127%
Effort 9 -E 2|3537.1|3531.5|-0.158%


Melirius:
Wow, congrats! It helps on my corpus too.

veluca93 (approved):

Galaxy4594:
Waiting on workflows, somebody needs to press the button :)

PR #4731, opened 2026-04-16
incremental decoding by preserving unconsumed input

metsw24-max:
The loader drops bytes returned by JxlDecoderReleaseInput, violating the decoder’s streaming contract. In incremental mode, this can break decoding by losing partially consumed input between calls.

This patch adds a small buffer to retain unconsumed input across load_increment calls and trims it to keep only the remaining bytes after each call.

Non-incremental behavior is unchanged.


metsw24-max:
any update?

PR #4732, opened 2026-04-16
integer overflow in buffer size calculation in GIMP saver

metsw24-max:
The buffer size for pixel data is computed using a single multiplication
of image dimensions, channel count, and sample size.

Since these values come from image input, this multiplication can overflow
size_t and lead to incorrect allocation size.

This patch computes the size step-by-step with bounds checks to prevent
overflow while preserving behavior for valid inputs.


eustas:
GIMP plugin was dropped

metsw24-max:
Right, makes sense now that the plugin's gone. Closing this out, thanks.

PR #4741, opened 2026-04-20
out-of-order jxlp boxes

### LINKED ISSUE #4735: out-of-order jxlp boxes
Currently the order of `jxlp` boxes (i.e. their indices incrementing) is not checked at all, see https://github.com/libjxl/libjxl/blob/6553831b39a542d6450e76ca5afb64520649d838/lib/jxl/decode.cc#L1899

Currently 18181-2 says they shall be in order, which is what libjxl is assuming. However, if it's not the case, then libjxl will not produce a useful error, it will just try to decode a garbled codestream (which will presumably fail but not necessarily, e.g. it could also produce an animation with the frames in the wrong order if the `jxlp` boxes correpond to frames).

In the future there might be a relaxation that allows out-of-order `jxlp` boxes. So in addition to actually checking the order (and refusing to decode if it's out of order), there will likely also be a need to handle the case of out-of-order boxes.

jonsneyers:
Fixes #4735.

Also now checks the `ftyp` minor version number and refuses to decode if the version is not 0 or 1.

If the version is 0 (as in current output of cjxl and the current edition of 18181-2), it now also checks the order of `jxlp` boxes and refuses to decode files with out-of-order `jxlp` boxes.

If the version is 1 (which will likely be defined in a future edition of 18181-2), it allows out-of-order `jxlp` boxes and decodes them in the right order.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -19,6 +19,7 @@
 #include <cstdint>
 #include <cstring>
 #include <functional>
+#include <map>
 #include <memory>
 #include <utility>
 #include <vector>
@@ -213,6 +214,7 @@ enum class BoxStage : uint32_t {
   kSkip,        // Box whose contents are skipped
   kCodestream,  // Handling codestream box contents, or non-container stream
   kPartialCodestream,  // Handling the extra header of partial codestream box
+  kBufferingJxlp,      // Reading an out-of-order jxlp box payload into buffer
   kJpegRecon,          // Handling jpeg reconstruction box
 };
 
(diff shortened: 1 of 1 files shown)


lib/jxl/decode.cc
```diff
@@ -1916,6 +1977,8 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
       if (status == JXL_DEC_NEED_MORE_INPUT) {
         if (dec->file_pos == dec->box_contents_end &&
             !dec->box_contents_unbounded) {
+          // Physical box exhausted; inject the next buffered OOO box if ready.
```
veluca93, line 1980: Shouldn't we inject potentially more than one box?

veluca93 (approved):

lib/jxl/decode.cc
jonsneyers: That's also an option, but this is simpler, just inject the next one and then when that one is exhausted, the code will end up here again anyway and inject the one after that.

PR #4742, opened 2026-04-20
Fix FileDirName returning empty string for bare filenames

### LINKED ISSUE #1872: benchmark_xl crashes with --save_decompressed and PNG input
**Describe the bug**
Consider a PNG input like the following white 8x8 square:
white
`benchmark_xl` crashes with `--save_decompressed` on this input, even on a debug build (the only build where it's supported).

**To Reproduce**
Save `white.png` locally, and run `benchmark_xl --input=white.png --save_decompressed`

**Expected behavior**
benchmark_xl should work as it is documented, without crashing.

**Actual behavior**
```
$ build/tools/benchmark_xl --input=./white.png --save_decompressed
benchmark_xl v0.8.0 997069d1 [AVX2]
4 total threads, 1 tasks, 0 threads, 4 inner threads
./lib/extras/dec/color_hints.cc:54: No color_space/icc_pathname given, assuming sRGB
./lib/extras/codec.cc:172: PNG only supports up to 16 bits per sample
./lib/jxl/dec_external_image.cc:259: JXL_CHECK: float_out ? bits_per_sample == 16 || bits_per_sample == 32 : bits_per_sample > 0 && bits_per_sample <= 16
Illegal instruction (core dumped)
```

**Environment**
 - OS: Arch Linux x86_64 6.0.6-arch1-1
 - Compiler version: GCC g++ 12.2.0
 - cjxl/djxl version string: 997069d111d4ba6c516580c218255c6f46c45c57

**Additional context**
Compiled with `CMAKE_BUILD_TYPE=Debug`

eyupcanakman:
`benchmark_xl --save_decompressed` fails when the input has no directory prefix. `FileDirName` returned `""` in that case, so callers at `tools/benchmark/benchmark_xl.cc:314` and `:540` constructed `"/out"` by appending `"/out"` to an empty dir. `MakeDir("/out")` fails with `EACCES`.

Return `"."` to match POSIX `dirname`. The fix also repairs `MatchFiles` on non-HAS_GLOB platforms: bare glob patterns like `*.png` called `opendir("")` and failed; they now match the current directory.

Fixes #1872


eustas (approved):

eustas (changes requested):
Let's drop the test

eyupcanakman:
Done, dropped it.

eustas (approved):

PR #4745, opened 2026-04-25
encoder output mode using out-of-order jxlp

jonsneyers:
(builds on https://github.com/libjxl/libjxl/pull/4741)

This adds a new encoder option `JXL_ENC_FRAME_SETTING_OUTPUT_MODE` to set the output mode:

- mode 0 (default): buffer the entire output codestream, progressive-decodable
- mode 1 (streaming output using seek): this does what used to be buffering=3: use a permuted TOC to write sections in an order convenient for chunked encode but not progressively decodable; use seeking to update the TOC and jxlp box size after encode
- mode 2 (ooo jxlp, new): uses `ftyp` version 1 which allows `jxlp` boxes to be in arbitrary order. Writes sections in the same order as in mode 1, but wraps each of them in its own `jxlp` box. The final `jxlp` box contains the TOC. No seeking is required during encode, and the 'logical' codestream does not use a permuted TOC, but it is shuffled at the file format level since it is split in `jxlp` boxes that are out of order.

It is not recommended to start using output mode 2 since current decoders do not support this yet. This is mostly just to create test bitstreams; it could be useful in some very specific use cases but generally sticking to mode 0 is recommended (to produce progressive-decodable output).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+ - decoder: support for out-of-order `jxlp` boxes (ftyp minor version 1).
+ - encoder API: `JXL_ENC_FRAME_SETTING_OUTPUT_MODE` frame setting to control
+   how the codestream is written to the output. Mode 0 (default) buffers the
+   output internally and produces a normally ordered, progressively decodable
+   codestream. Mode 1 uses seek-based streaming (reduces peak memory for large
+   images, requires a seekable output stream). Mode 2 uses out-of-order `jxlp`
+   boxes (reduces peak memory without requiring seeking, but requires a decoder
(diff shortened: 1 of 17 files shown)


veluca93 (commented):
Can we add roundtrip tests for this, with and without TOC permutation?

jonsneyers:
> Can we add roundtrip tests for this, with and without TOC permutation?

Initially TOC permutation was being ignored in output mode 2 (just like in output mode 1). Changed it to actually do the requested TOC permutation in mode 2, since even if modes 1 and 2 are inherently non-progressive, mode 2 can be easily converted to mode 0 by sorting and merging the jxlp boxes, so it may be useful to be able to do a center-first permutation even if it will only be usable after doing a conversion from ftyp 1 to ftyp 0.

Added a test that checks that all output modes and TOC permutations produce different files but identical decoded images.

veluca93 (dismissed):

veluca93 (approved):

PR #4746, opened 2026-04-25
Guard AlignedArray::Create against allocation size overflow

uwezkhan:
This patch adds an overflow check in AlignedArray::Create when computing the
allocation size for the backing storage.

Previously the code multiplied `size * sizeof(T)` directly before allocating.
On platforms where size_t can overflow, very large values could wrap and result
in a smaller-than-expected allocation.

The change uses the existing SafeMul helper to validate the multiplication
before passing the size to AlignedMemory::Create.

This does not change behavior for valid inputs and improves robustness of the
shared allocation helper.

### DIFF
--- a/lib/jxl/memory_manager_internal.h
+++ b/lib/jxl/memory_manager_internal.h
@@ -14,6 +14,7 @@
 #include <memory>
 #include <utility>
 
+#include "lib/jxl/base/common.h"
 #include "lib/jxl/base/compiler_specific.h"
 #include "lib/jxl/base/status.h"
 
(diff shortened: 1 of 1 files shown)


eustas (dismissed):

eustas:
Third parameter for `SafeMul` is a reference, not pointer.
For better clarity let's also pin the template type to `size_t`.

uwezkhan:
Updated the branch with the requested SafeMul change. Waiting for CI now, thanks!

eustas (approved):

PR #4747, opened 2026-04-25
Guard BytesPerRow against stride calculation overflow

uwezkhan:
This patch adds overflow checks to BytesPerRow when calculating image row stride.

Previously, stride-related arithmetic in BytesPerRow could overflow on platforms
with smaller size_t ranges when processing very large image dimensions. That
could result in an incorrect row size being returned and potentially lead to an
undersized allocation later in PlaneBase::Allocate.

The change uses checked arithmetic for the intermediate calculations and causes
allocation to fail cleanly if the computed stride would overflow.

PlaneBase::Allocate is updated to detect this case and return an error instead
of continuing with an invalid allocation size.

This does not change behavior for valid image sizes and improves robustness of
the shared image allocation path.


eustas:
Fixed by #4860. Thanks again.

PR #4764, opened 2026-05-12
Harden PaddedBytes growth and allocation arithmetic

uwezkhan:
This patch adds checked arithmetic to a few size calculations inside
PaddedBytes.

The growth logic in reserve(), the padding calculation in
change_capacity(), and the append size calculation were previously using
raw size_t arithmetic. On platforms with smaller size_t ranges, these
calculations could overflow and result in incorrect allocation sizes.

The patch switches these calculations to use the existing SafeAdd helper
and handles overflow cases safely without changing normal behavior for
valid inputs.

This keeps the existing allocation flow while improving robustness of the
shared byte buffer utility.


lib/jxl/padded_bytes.h
```diff
@@ -79,7 +79,11 @@ class PaddedBytes {
   Status reserve(size_t capacity) {
     if (capacity <= capacity_) return true;
 
-    size_t new_capacity = std::max(capacity, 3 * capacity_ / 2);
+    size_t new_capacity;
+    if (!SafeAdd(capacity_, capacity_ / 2, new_capacity)) {
```
eustas, line 83: `reserve` should return failure is `new_capacity` overflows (the contract is - once `reserve` succeeds, we can rely requested `capacity` is available).

uwezkhan:
@eustas review on this too . thanks

eustas (approved):

PR #4765, opened 2026-05-12
Add overflow checks to BytesPerRow calculations

uwezkhan:
### DescriptionThis patch adds checked arithmetic to the stride calculation logic in
BytesPerRow.

Previously, some of the internal size calculations used raw size_t
arithmetic. On platforms with smaller size_t ranges, very large image
dimensions could cause these calculations to overflow and produce an
invalid row stride.

The patch switches these calculations to use the existing SafeMul and
SafeAdd helpers. If an overflow is detected, BytesPerRow now returns 0
as an invalid state, and PlaneBase::Allocate explicitly rejects this
case before any allocation happens.

I also checked all current call sites of BytesPerRow and confirmed that
PlaneBase is the only caller inside lib/jxl, so the validation path is
guaranteed before the value is used.

This keeps the existing API and allocation flow unchanged for valid
inputs while improving safety around large dimension handling.

### Pull Request Checklist

- [ ] **CLA Signed**: Have you signed the [Contributor License Agreement](https://code.google.com/legal/individual-cla-v1.0.html) (individual or corporate, as appropriate)? Only contributions from signed contributors can be accepted.
- [ ] **Code Style**: Have you ensured your code adheres to the project's coding style guidelines? You can use `./ci.sh lint` for automatic code formatting.


eustas:
Let's redesign the fix:
 * `BytesPerRow` should return `StatusOr<size_t>`; overflow, of course is a failure
 * set initial value of `bytes_per_row_` to `0`; the only non-default constructor is followed directly by `Allocate`, thus
 * assign `bytes_per_row_` in `Allocate` or return failure.

uwezkhan:
@eustas can i get a review ?

PR #4768, opened 2026-05-15
Enable "uninitialize field" analysis.

eustas:
Do some suppressions, some initializations. Many warnings still need to be addressed.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,7 @@
 
 Checks: >-
   bugprone-*,
+  cppcoreguidelines-pro-type-member-init,
   clang-*,
   -clang-diagnostic-unused-command-line-argument,
   google-*,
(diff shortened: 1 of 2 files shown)


QrczakMK:
Is the analysis tool able to determine (at least most of the time) that it is OK when a field has been defined uninitialized, then assigned to, without reading it in-between?

If yes, then perhaps this can be avoided at least where the tool does not complain? Forced initialization has downsides. Even if the compiler does not emit the initialization at runtime, if the code has a bug which results in reading the uninitialized value (perhaps after it is changed), then without forced initialization running under msan will catch the bug. With forced initialization, running under msan will get the zero value, which is likely incorrect hiding the bug.

If no, then it is a pity that a tool forces a suboptimal debugging experience with msan. Which analysis tool is that?

eustas:
Alas, check seems to be very deficient. Reverted suppressions / initialization, except the one that is really a finding.
Hopefully, clang-tidy will become better in the future.

QrczakMK (approved):

PR #4770, opened 2026-05-15
Harden ChunkedPNMDecoder against integer overflow in size validation

metsw24-max:
Fixes a real, reachable integer-overflow bypass in `ChunkedPNMDecoder::Init` (`lib/extras/dec/pnm.cc`).

`ChunkedPNMDecoder::Init` parsed unbounded `xsize`/`ysize` values from a PNM header and validated the backing file size using unchecked `size_t` arithmetic:

* `xsize * bytes_per_pixel`
* `ysize * row_size`
* `+ dec.data_start_`

A crafted PNM could overflow these computations and wrap them to small values, causing the file-size validation to succeed incorrectly. After initialization, `GetColorChannelDataAt()` derived pointers from the overflowed geometry and could return out-of-bounds offsets into the mmap’d file.

### Changes

* Added `VerifyDimensions(nullptr, header.xsize, header.ysize)` before any size arithmetic to enforce the standard decoder dimension/pixel limits.
* Replaced unchecked arithmetic with `SafeMul` / `SafeAdd` for:

  * row-size calculation
  * total image-size calculation
  * final file-offset validation

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,7 @@
 
 Checks: >-
   bugprone-*,
+  cppcoreguidelines-pro-type-member-init,
   clang-*,
   -clang-diagnostic-unused-command-line-argument,
   google-*,
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -221,7 +221,7 @@ jobs:
 
     steps:
     - name: Harden Runner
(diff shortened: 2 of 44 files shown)


metsw24-max:
Looks like this got overtaken on main in the meantime. `ChunkedPNMDecoder::Init` already carries the `VerifyDimensions` + `SafeMul`/`SafeAdd` guards (via #4799), and the header-number overflow is covered by #4839, so the net diff here is now empty. Happy to close this as superseded unless you'd rather keep it open.

PR #4771, opened 2026-05-16
simplify jhgm box buffer handling in jxlinfo

jmestwa-coder:
### Description

This change simplifies `jhgm` box accumulation in `tools/jxlinfo.cc` by replacing manual `malloc`/`realloc`/`free` handling with `std::vector<uint8_t>`.

It also adds explicit validation around box buffer accumulation state and checked overflow handling before growing the buffer.

#### Changes

* Replace manual heap allocation with `std::vector<uint8_t>`
* Validate decoder buffer state before updating accumulation offsets
* Add checked `size_t` overflow handling during buffer growth
* Remove manual cleanup paths now handled through RAII

The updated logic keeps the existing behavior while making the accumulation flow easier to reason about and more robust against invalid internal state assumptions.


eustas (changes requested):
Please drop "poc". Thanks

jmestwa-coder:
done, dropped poc.cc. only tools/jxlinfo.cc left now.

tools/jxlinfo.cc
```diff
@@ -405,38 +405,45 @@ static int PrintBasicInfo(FILE* file, int verbose) {
...
-        box_size = 0;
+      if (box_index > box_data.size() ||
+          remaining > box_data.size() - box_index) {
+        box_data.clear();
         box_index = 0;
-        fprintf(stderr, "Memory reallocation failed\n");
+        fprintf(stderr, "Invalid box buffer state\n");
         break;
       }
-      box_data = temp;
```
eustas, line 430: I.e. `box_index = box_data.size() - remaining`?

tools/jxlinfo.cc
```diff
@@ -405,38 +405,45 @@ static int PrintBasicInfo(FILE* file, int verbose) {
...
-        JxlDecoderSetBoxBuffer(dec, box_data, box_size);
+        box_data.resize(chunk_size);
+        JxlDecoderSetBoxBuffer(dec, box_data.data(), box_data.size());
       } else {
         printf("Unknown box:\n  type: \"%.4s\"\n  size: %" PRIu64 "\n",
                box_type, size);
       }
     } else if (status == JXL_DEC_BOX_NEED_MORE_OUTPUT) {
       const size_t remaining = JxlDecoderReleaseBoxBuffer(dec);
-      box_size += chunk_size;
```
eustas, line 421: Both conditions are meant to be impossible.

tools/jxlinfo.cc
```diff
@@ -405,38 +405,45 @@ static int PrintBasicInfo(FILE* file, int verbose) {
...
+        fprintf(stderr, "Box size overflow\n");
+        break;
+      }
+      box_data.resize(box_index + chunk_size);
+      JxlDecoderSetBoxBuffer(dec, box_data.data() + box_index,
+                             box_data.size() - box_index);
     } else if (status == JXL_DEC_BOX_COMPLETE) {
       if (!strncmp(box_type, "jhgm", 4)) {
         size_t remaining = JxlDecoderReleaseBoxBuffer(dec);
-        box_size -= remaining;
```
eustas, line 435: Both conditions are impossible

tools/jxlinfo.cc
jmestwa-coder: yeah, that's cleaner. switched it to box_index = box_data.size() - remaining.

tools/jxlinfo.cc
jmestwa-coder: right, remaining is 0 here so it can't trigger. dropped the check.

tools/jxlinfo.cc
jmestwa-coder: agreed, dropped this one too. the overflow guard went with it since box_index can't get near SIZE_MAX.

tools/jxlinfo.cc
```diff
@@ -405,38 +404,26 @@ static int PrintBasicInfo(FILE* file, int verbose) {
...
-      if (temp == nullptr) {
-        box_size = 0;
-        box_index = 0;
-        fprintf(stderr, "Memory reallocation failed\n");
-        break;
-      }
-      box_data = temp;
-      JxlDecoderSetBoxBuffer(dec, box_data + box_index, box_size - box_index);
+      box_index = box_data.size() - remaining;
+      box_data.resize(box_data.size() + chunk_size);
```
eustas, line 420: Check for `box_data.size() + chunk_size` was actually useful (though unlikely to trigger; given the vector growth strategy, we will get OOM first).

jmestwa-coder: good point, restored it as a check on box_data.size() + chunk_size before the resize.

eustas (approved):

PR #4773, opened 2026-05-16
Fix GIF palette OOB read and PNM size overflow checks

metsw24-max:
This patch fixes two decoder-internal memory safety issues in lib/extras/dec/.

### GIF decoder: off-by-one palette bounds check

lib/extras/dec/gif.cc

The GIF decoder validated palette indices using:

    if (byte > color_map->ColorCount)

This incorrectly allowed `byte == ColorCount`, even though valid palette indices are only:

    0 .. ColorCount - 1

As a result, decoding could read one element past the end of:

    color_map->Colors

The replace path already used the `correct >= check` so this patch aligns both paths and removes the off-by-one OOB read.

### PNM/PAM decoder: overflow-safe required size calculation

lib/extras/dec/pnm.cc

The required pixel-data size calculation previously chained multiple unchecked multiplications:

    header.ysize * header.xsize *
    (num_interleaved_channels + header.ec_types.size()) * twidth

Large crafted image dimensions could overflow `size_t` causing the computed size to wrap and incorrectly pass the subsequent bounds check. Later image-copy operations could then read past the end of the input buffer.

The fix replaces the unchecked arithmetic with chained `SafeMul()` calls, consistent with existing overflow-hardening patterns in the codebase. Decoding now fails cleanly if the computed image size exceeds representable limits.

Together these changes harden two independent decoder paths against malformed untrusted inputs and eliminate reachable out-of-bounds read conditions.

### DIFF
--- a/lib/extras/dec/gif.cc
+++ b/lib/extras/dec/gif.cc
@@ -375,7 +375,7 @@ Status DecodeImageGIF(Span<const uint8_t> bytes, const ColorHints& color_hints,
                          y * sub_frame_image.xsize;
         for (size_t x = 0; x < image_rect.xsize(); ++x, ++byte_index) {
           const GifByteType byte = image.RasterBits[byte_index];
-          if (byte > color_map->ColorCount) {
+          if (byte >= color_map->ColorCount) {
             return JXL_FAILURE("GIF color is out of bounds");
           }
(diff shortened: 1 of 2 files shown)


eustas:
GIF palette index is fixed in separate PR

eustas (approved):

PR #4775, opened 2026-05-17
Harden EXR window validation and allocation arithmetic

metsw24-max:
This patch hardens EXR decoding against multiple integer-overflow and size-validation issues in `lib/extras/dec/exr.cc`.

Changes included:

* Reject EXR `displayWindow` / `dataWindow` coordinates outside ±2^30 and reject inverted windows before downstream signed arithmetic is performed.
* Apply `VerifyDimensions()` to `dataWindow` dimensions in addition to `displayWindow`, since `dataWindow` controls allocation sizes and framebuffer indexing.
* Wrap framebuffer allocation size calculations in `SafeMul()` to prevent unchecked multiplication overflow.
* Recompute framebuffer base offsets using `ptrdiff_t` instead of `int` so expressions like `start_y * row_size` cannot overflow on large but otherwise valid images.

This fixes several reachable overflow conditions originating from untrusted EXR header fields.

The patch is related to #4758 (`caed34c2`), which hardened the inner memcpy size calculation, but did not cover the upstream allocation and framebuffer setup paths fixed here.

### DIFF
--- a/lib/extras/dec/exr.cc
+++ b/lib/extras/dec/exr.cc
@@ -213,12 +213,41 @@ Status DecodeImageEXR(Span<const uint8_t> bytes, const ColorHints& color_hints,
 
   const Imath::Box2i displayWindow = header.displayWindow();
   const Imath::Box2i dataWindow = header.dataWindow();
+  // displayWindow / dataWindow bounds come straight from the EXR file header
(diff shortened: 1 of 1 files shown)


eustas (approved):

eustas:
Thanks

PR #4776, opened 2026-05-17
harden PackedImage allocation arithmetic

jmestwa-coder:
## Summary

Harden `PackedImage` allocation and stride arithmetic by replacing fragile divide-back overflow checks with checked `SafeMul`/`SafeAdd` operations.

## Changes

- Use `SafeMul` for:
  - bits-per-pixel calculation
  - row stride calculation
  - total pixel buffer size calculation

- Use `SafeAdd` during alignment round-up to avoid overflow in aligned stride computation.

- Remove divide-based overflow checks that relied on already-computed arithmetic.

- Preserve zero-width image handling without triggering divide-by-zero behavior.

### DIFF
--- a/lib/extras/codec_test.cc
+++ b/lib/extras/codec_test.cc
@@ -14,6 +14,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <limits>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
any update?

PR #4777, opened 2026-05-17
Fix jpeg recon int overflow

### LINKED ISSUE #4779: Lossless JPEG->JXL->JPEG round-trip silently fails on very large (gigapixel) images
Hi all,

I've been building a small utility that losslessly transcodes JPEGs to JXL and back (using JxlEncoderAddJPEGFrame and the JBRD reconstruction path), and hit a case where some gigapixel images can't be round-tripped. `cjxl --lossless_jpeg=1` happily writes a .jxl with reconstruction data, but decoding it never returns the original JPEG -- `djxl` falls back to `--pixels_to_jpeg`. Smaller images are fine, so it took me a while to pin down that it's size-related.

Tracing it: the `coeffs` buffer is sized correctly in 64-bit (`uint64_t num_blocks; coeffs.resize(num_blocks * kDCTBlockSize)`), but the per-block index in the two scan loops is a 32-bit `int`:

- `lib/jxl/jpeg/enc_jpeg_data_reader.cc`: `int block_idx = ...;` then `&c->coeffs[block_idx * kDCTBlockSize]`
- `lib/jxl/jpeg/dec_jpeg_data_writer.cc`: `int block_idx = ...;` then `&c.coeffs[block_idx << 6]`

Once a component goes past ~33.5M blocks, `block_idx * 64` overflows signed int and the offset goes out of range (often negative), so the reconstruction is silently wrong. Nothing bounds-checks it because the allocation itself is big enough -- it's purely the index arithmetic. In practice ~0.87 Gpx round-trips fine and ~2.47 Gpx (e.g. a 53717x45898 baseline JPEG) fails every time, which matches the INT_MAX/64 threshold.

Repro: a baseline JPEG with a component over ~33.5M blocks, `cjxl --lossless_jpeg=1 in.jpg out.jxl` (jxlinfo shows reconstruction data), then `djxl out.jxl back.jpg` -- it won't reconstruct losslessly.

The fix is a one-liner at each of the two spots: widen `block_idx` to `size_t` (casting the operands so it stays clean under -Werror). I've opened #4777 with that change -- happy to adjust if you'd rather handle the widening differently.

Thanks!

profiteroles1:
Fixes #4779

### Problem
`cjxl --lossless_jpeg=1` accepts a large baseline JPEG and writes a JXL with JPEG bitstream reconstruction data, but `djxl` then cannot reconstruct it ("could not decode losslessly to JPEG, retrying with --pixels_to_jpeg"). No error is reported at encode time. Smaller images round-trip losslessly.

### Reproduction
- A baseline JPEG whose largest component exceeds ~33.5M blocks (e.g. 53717x45898), i.e. roughly >= 2 gigapixel.
- `cjxl --lossless_jpeg=1 big.jpg big.jxl` succeeds; `jxlinfo big.jxl` shows "JPEG bitstream reconstruction data available".
- `djxl big.jxl out.jpg` fails lossless reconstruction and falls back to pixels.
- ~0.87 Gpx (~13.6M blocks) round-trips; ~2.47 Gpx (~38.5M blocks) fails — threshold consistent with signed 32-bit overflow at `block_idx * 64`.

### Cause
The `coeffs` buffer is correctly 64-bit sized (`uint64_t num_blocks; coeffs.resize(num_blocks * kDCTBlockSize)`), but the per-block index is 32-bit `int` in two scan loops:
- `lib/jxl/jpeg/enc_jpeg_data_reader.cc`: `int block_idx = ...;` then `&c->coeffs[block_idx * kDCTBlockSize]`
- `lib/jxl/jpeg/dec_jpeg_data_writer.cc`: `int block_idx = ...;` then `&c.coeffs[block_idx << 6]`

Once `block_idx > INT_MAX / 64` (~33,554,431), `block_idx * kDCTBlockSize` / `block_idx << 6` overflows signed `int`, yielding an out-of-range offset and corrupt reconstruction. No bounds check triggers because the buffer itself is large enough.

### Fix
Widen `block_idx` to `size_t` at both sites; the multiply/shift then promote to 64-bit. No other behavior change. Verified: JPEGs with >2.1B coefficients/component now reconstruct bit-exact; smaller files unchanged.

### DIFF
--- a/lib/jxl/jpeg/dec_jpeg_data_writer.cc
+++ b/lib/jxl/jpeg/dec_jpeg_data_writer.cc
@@ -843,7 +843,8 @@ SerializationStatus JXL_NOINLINE DoEncodeScan(const JPEGData& jpg,
           for (int ix = 0; ix < n_blocks_x; ++ix) {
             int block_y = ss.mcu_y * n_blocks_y + iy;
             int block_x = mcu_x * n_blocks_x + ix;
-            int block_idx = block_y * c.width_in_blocks + block_x;
+            size_t block_idx =
+                static_cast<size_t>(block_y) * c.width_in_blocks + block_x;
             if (ss.block_scan_index == ss.next_reset_point) {
               Flush(coding_state, bw);
               ss.next_reset_point = get_next_reset_point();
--- a/lib/jxl/jpeg/dec_jpeg_data_writer.cc
+++ b/lib/jxl/jpeg/dec_jpeg_data_writer.cc
@@ -844,7 +844,8 @@ SerializationStatus JXL_NOINLINE DoEncodeScan(const JPEGData& jpg,
             int block_y = ss.mcu_y * n_blocks_y + iy;
             int block_x = mcu_x * n_blocks_x + ix;
             size_t block_idx =
-                static_cast<size_t>(block_y) * c.width_in_blocks + block_x;
+                static_cast<size_t>(block_y) * c.width_in_blocks +
+                static_cast<size_t>(block_x);
             if (ss.block_scan_index == ss.next_reset_point) {
               Flush(coding_state, bw);
               ss.next_reset_point = get_next_reset_point();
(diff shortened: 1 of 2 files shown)


eustas:
Nice. Thanks for fixing that.

I would greatly appreciate if you add extra check (before loops) that extreme values (last block) won't overflow (`block_idx * kDCTBlockSize` fits `size_t`). On 64-bit machines that is unlikely to happen, but on 32-bit...

eustas (approved):

Issue #4329, opened 2025-07-08
Github Actions builds are being double zipped


jonnyawsom3:
Upon opening the zip, a second zip is inside. Noticed when downloading `jxl-x64-windows-static` from https://github.com/libjxl/libjxl/actions/runs/16071685267

sboukortt:
Relevant: https://github.com/actions/upload-artifact/issues/702

eustas:
Oh, that is easy to miss on OSX - it double-unzips =)

eustas:
Same problem with .tar.gz artifacts

eustas:
It seems that "upload artifact" is primarily inter-job-communication thing, with UX added just for inspection from developers side. There is not "github actions" API for uploading, and deep inside it sends files to azure blobs.

I'll check if avoiding double-compression is reasonable...

eustas:
In case of `release.yaml` workflow we intentionally upload archives, so that on "publish" step we don't do any processing.
This approach makes adding more release artifacts fool-proof: whatever is uploaded in this workflow will be published.

eustas:
As said - zip inside zip is expected behavior - inner zip is what we want to ship, outer zip is a transport of upload/download artifact system (since it could transfer a bundle of files). We should only care that "release assets" are not double-zipped.

PR #4780, opened 2026-05-18
Remove unnecessary zip layer from release assets

yashikota:
### Description

Thank you for this great product.

I believe the assets currently distributed via Releases are in a “zip-in-zip” format.
Ref: [https://github.com/libjxl/libjxl/issues/4329](https://github.com/libjxl/libjxl/issues/4329)

However, this format is not handled well by some package managers.  
Ref: https://github.com/aquaproj/aqua/issues/4832

Since there does not seem to be any real benefit to using a zip-in-zip format, and the drawbacks outweigh the advantages, I tried removing one layer of zipping.

The `upload-artifact` action provides an `Unzipped` option, so I enabled that and replaced the now unnecessary `compression-level: 0`.
Ref: [https://github.com/actions/upload-artifact#upload-an-individual-file-unzipped](https://github.com/actions/upload-artifact#upload-an-individual-file-unzipped)

You can check the actual build results from the following test PR if needed:
[https://github.com/yashikota/libjxl/pull/1/checks](https://github.com/yashikota/libjxl/pull/1/checks)

This removes the unnecessary extra zipping and makes the release assets available in a more preferable distribution format.

One possible concern is that systems currently relying on the zip-in-zip format may no longer work. However, since this does not modify existing assets and only changes the format of future releases to a better one, I believe the impact should be minimal.

### Pull Request Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -101,7 +101,7 @@ jobs:
       with:
         name: jxl-linux-x86_64-static
         path: jxl-linux-x86_64-static.tar.gz
-        compression-level: 0
+        archive: false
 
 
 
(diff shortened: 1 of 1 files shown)


eustas (approved):

eustas:
Thanks!

We've been waiting for "archive: false" option for years... No excuse not to use us now!

PR #4789, opened 2026-05-21
Account for decode retries in fuzzer

eustas:


QrczakMK (changes requested):
Could you please explain the intent? I don't understand this.

eustas:
`total_pixels` is an approximate meter of performed computations. Before this change we were accounting only for succeeded decodings (i.e. when pixels emitted). But in some scenarios we (intentionally) ask decoder to do as much work as possible for given incomplete input. Those computations were not accounted. Because of this UBSAN gets timeouts on some samples.

Currently there are no means for more precise accounting, so we tax retry as complete frame decoding.

QrczakMK (approved):

PR #4796, opened 2026-05-23
guard ParseHeaderPGX size arithmetic against overflow

rootvector2:
`Parser::ParseHeaderPGX` in `lib/extras/dec/pgx.cc` reads xsize/ysize via
`ParseUnsigned`, which does not bound the parsed value, then evaluates
`pos_ + header->xsize * header->ysize * bytes_per_pixel > end_`. A
crafted header with 20+ digit dimensions wraps the multiplication in
size_t and turns the bounds check into UB pointer arithmetic on `pos_`.
`VerifyDimensions` in `DecodeImagePGX` catches the bad dimensions one
level up, but the parser-layer check should not rely on that. Mirrors
the SafeMul guards added for PNM in #4773 and #4774.

### DIFF
--- a/lib/extras/dec/pgx.cc
+++ b/lib/extras/dec/pgx.cc
@@ -15,6 +15,7 @@
 #include "lib/extras/dec/color_hints.h"
 #include "lib/extras/packed_image.h"
 #include "lib/extras/size_constraints.h"
+#include "lib/jxl/base/common.h"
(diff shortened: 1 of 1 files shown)


eustas (approved):

eustas:
Seems to be fixed with #4791

PR #4797, opened 2026-05-23
validate marker bounds before access in Exif/XMP and ICC handling

metsw24-max:
Reorders existing validation checks in JPEG marker handling so bounds/invariant checks occur before buffer access or mutation.

### `lib/jxl/jpeg/dec_jpeg_data.cc`

* In `DecodeJPEGData`, moved the Exif/XMP marker size checks before `marker[0] = 0xE1`.
* Prevents partial mutation of undersized markers on failure paths.
* No behavioral change for valid inputs.

### `lib/jxl/jpeg/enc_jpeg_data.cc`

* In `DetectIccProfile`, moved `JXL_ENSURE(app.size() >= 3)` before the first `app[pos++]` access.
* Makes the size invariant explicit at the point of first read.

### DIFF
--- a/lib/jxl/jpeg/dec_jpeg_data.cc
+++ b/lib/jxl/jpeg/dec_jpeg_data.cc
@@ -72,6 +72,9 @@ Status DecodeJPEGData(Span<const uint8_t> encoded, JPEGData* jpeg_data) {
   size_t num_icc = 0;
   for (size_t i = 0; i < jpeg_data->app_data.size(); i++) {
     auto& marker = jpeg_data->app_data[i];
+    if (marker.size() < 3) {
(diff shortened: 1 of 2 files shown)


eustas:
Touching `marker[1]` and `marker[2]` happens even earlier. Let's check `marker.size() >= 3` after line 74 of `lib/jxl/jpeg/dec_jpeg_data.cc`

metsw24-max:
added a marker.size() < 3 guard right after the marker reference is taken (before the if/else), which covers both the write path (marker[1]/marker[2] at lines 81-82) and the read path in the else branch (line 93). Also fixes the unsigned underflow at marker.size() - 1 when size is 0.

eustas (approved):

PR #4798, opened 2026-05-23
Fix undefined behavior in patch blending foreground pointer arithmetic

metsw24-max:
This PR removes undefined behavior in patch blending caused by foreground pointer arithmetic that could transiently form out-of-bounds pointers through unsigned subtraction.

Previously, foreground pointers were constructed using:

```cpp
ref_pos.x0 + x0 - bx
```

When `bx > x0`, the subtraction could underflow in `size_t` arithmetic before pointer formation, even though the eventual reads resolved to valid memory.

The fix separates foreground (`fg`) addressing from background/output (`bg`/`out`) addressing inside `PerformBlending` and restructures the patch alignment logic so all intermediate arithmetic is non-negative by construction.

No intended behavior changes on valid inputs.

#### `lib/jxl/blending.h`

* Added a dedicated `fg_x0` parameter to `PerformBlending`
* Documented the pointer-safety rationale

#### `lib/jxl/blending.cc`

* Split foreground and background/output offset handling
* Updated all blending paths and helper lambdas to use explicit side-specific offsets
* `bg/out` continue using `x0`
* `fg` now uses `fg_x0`

#### `lib/jxl/dec_patch_dictionary.cc`

* Reworked patch alignment logic using non-negative offsets:

  ```cpp
  patch_in_ref = patch_x0 - bx;
  out_in_seg   = patch_x0 - x0;
  ```
* Pre-shifts foreground pointers directly
* Calls `PerformBlending(..., fg_x0 = 0, ...)`

#### `lib/jxl/render_pipeline/stage_blending.cc`

* Updated call site to pass `fg_x0 = 0`
* No functional behavior change

### Tests

Build/test execution was not completed locally because submodules were not initialized in the checkout.

Recommended validation before merge:

* Full project build
* `jxl_tests`
* Patch dictionary decoding coverage
* Render pipeline blending tests
* UBSan/ASan validation for patch blending paths

### DIFF
--- a/lib/jxl/blending.cc
+++ b/lib/jxl/blending.cc
@@ -41,8 +41,9 @@ bool NeedsBlending(const FrameHeader& frame_header) {
 
 Status PerformBlending(
     JxlMemoryManager* memory_manager, const float* const* bg,
-    const float* const* fg, float* const* out, size_t x0, size_t xsize,
-    const PatchBlending& color_blending, const PatchBlending* ec_blending,
+    const float* const* fg, float* const* out, size_t x0, size_t fg_x0,
+    size_t xsize, const PatchBlending& color_blending,
+    const PatchBlending* ec_blending,
     const std::vector<ExtraChannelInfo>& extra_channel_info) {
   bool has_alpha = false;
   size_t num_ec = extra_channel_info.size();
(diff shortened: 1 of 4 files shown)


metsw24-max:
@eustas any update on this

PR #4807, opened 2026-05-27
icc_codec: make DecodeVarInt return Status and report errors

evilgensec:
## Summary

`DecodeVarInt()` at `lib/jxl/icc_codec.cc:61` silently mishandles three input conditions that the file's own TODOs have flagged for several releases:

- `lib/jxl/icc_codec.cc:55-58` — "DecodeVarInt does not signal the errors, and marks 11 bytes as used even if only 10 are used"
- `lib/jxl/icc_codec.cc:137-141` (pre-patch) — "we need to check that the entire varint decoding never goes out of bounds, not just the first byte. This requires a DecodeVarInt function that returns an error code"

The bugs are:

1. **Mid-varint truncation.** When the input ends before the terminator byte, the function returns a partially-decoded value with no error indication. The `if (pos >= size)` guard each caller inserts only covers the first byte of a multi-byte varint.
2. **Position over-advance.** When the loop exits via the `i < 10` condition (i.e. 10 continuation bytes were seen), `*pos += i + 1` advances 11 bytes for a 10-byte varint, desyncing every later read in the same stream.
3. **Silent overflow.** A 10th byte whose bits 1..6 are set encodes a value above 2^63 - 1; the previous code drops those bits and returns a value that does not match the encoded varint.

Each of these is exercisable from attacker-controlled bytes via the ICC profile decoding path: bug (1) lets a crafted stream make `CheckOutOfBounds(pos, csize, size)` operate on a garbage `pos`; bug (2) lets the same stream shift every subsequent varint by one byte; bug (3) lets the stream return a value that no longer matches its encoding.

## Change

`DecodeVarInt` is replaced with a `Status`-returning variant that:

- reads at most 9 bytes for the 0..62-bit range and an optional 10th byte for bit 63 only;
- returns `JXL_FAILURE` on truncation, on an unterminated 10-byte sequence, and on a 10th-byte value that would overflow `uint64_t`;
- advances `*pos` by exactly the number of bytes consumed.

All 11 call sites in `icc_codec.cc` are migrated to propagate the `Status` via `JXL_RETURN_IF_ERROR`. The per-call `if (pos >= size)` and `if (cpos >= commands_end)` guards in front of each varint read are removed because the helper now reports those conditions itself.

`kPreambleSize` drops from 22 to 20 to match the new maximum of two 10-byte varints, addressing the explicit value in the `TODO(eustas)` at line 55.

## Diff

```
 lib/jxl/icc_codec.cc | 95 +++++++++++++++++++++++++++++-----------------------
 1 file changed, 53 insertions(+), 42 deletions(-)
```

11 call sites changed:

- `CheckPreamble` (lines 115, 118 pre-patch)
- `UnpredictICC` (lines 143, 146, 171, 214, 223, 259, 266, 296, 311 pre-patch)

## Behavioural compatibility

Every legitimate varint produced by a conforming encoder fits in ≤ 10 bytes and ends with a terminator within bounds, so the helper takes the same fast-path and returns the same value for all valid input. Malformed input (truncated, oversize, or `> 2^64`-encoded varints) now returns `JXL_FAILURE` instead of silently returning garbage.

## Build

```
cmake --build build --target jxl_dec
```

succeeds locally on macOS/arm64 (Debug, `-DJPEGXL_ENABLE_BENCHMARK=OFF -DBUILD_TESTING=OFF` …). No new warnings; the pre-existing `-Wframe-larger-than` warnings in `dec_jpeg_data_writer.cc` are unrelated.

## Dedup

`gh pr list --search "DecodeVarInt OR icc_codec.cc"` shows no in-flight PR touching this code path on `main` as of the date of submission.

### DIFF
--- a/lib/jxl/icc_codec.cc
+++ b/lib/jxl/icc_codec.cc
@@ -52,24 +52,41 @@ Status Shuffle(JxlMemoryManager* memory_manager, uint8_t* data, size_t size,
   return true;
 }
 
-// TODO(eustas): should be 20, or even 18, once DecodeVarInt is improved;
-//               currently DecodeVarInt does not signal the errors, and marks
-//               11 bytes as used even if only 10 are used (and 9 is enough for
-//               63-bit values).
-constexpr const size_t kPreambleSize = 22;  // enough for reading 2 VarInts
+// Two base-128 varints at up to 10 bytes each.
+constexpr size_t kPreambleSize = 20;
 
-uint64_t DecodeVarInt(const uint8_t* input, size_t inputSize, size_t* pos) {
-  size_t i;
+// Decodes a base-128 unsigned varint into *out, advancing *pos by the exact
+// number of bytes consumed. Returns an error if the input is truncated, if
+// the terminator (top-bit-clear byte) is not seen within 10 bytes, or if the
+// 10th byte encodes a value that does not fit in a uint64_t.
+Status DecodeVarInt(const uint8_t* input, size_t inputSize, size_t* pos,
+                    uint64_t* out) {
   uint64_t ret = 0;
-  for (i = 0; *pos + i < inputSize && i < 10; ++i) {
-    ret |= static_cast<uint64_t>(input[*pos + i] & 127)
-           << static_cast<uint64_t>(7 * i);
-    // If the next-byte flag is not set, stop
-    if ((input[*pos + i] & 128) == 0) break;
+  // 9 bytes cover bits 0..62; the 10th byte may only contribute bit 63.
+  for (size_t i = 0; i < 9; ++i) {
+    if (*pos >= inputSize) {
+      return JXL_FAILURE("DecodeVarInt: truncated input");
+    }
+    const uint8_t byte = input[(*pos)++];
+    ret |= static_cast<uint64_t>(byte & 0x7F) << (7 * i);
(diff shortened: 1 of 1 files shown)


evilgensec:
Thanks for the review. Force-pushed an update that addresses both points:

1. **`cpos` reads now bounded to `commands_end`** (line 153 / now 153, 195, 203, 239, 246, 275, 290). The seven `cpos`-based `DecodeVarInt` calls inside `UnpredictICC` now pass `commands_end` as the buffer bound instead of `size`. This closes a latent issue in the prior code: a varint that started inside the commands stream could extend past `commands_end` into the data section while still satisfying the whole-buffer bound, silently reading non-command bytes. The four `pos`-based calls (in `CheckPreamble` and at the top of `UnpredictICC`) continue to use `size` since they operate on the full buffer.

2. **`constexpr const`** → `constexpr` on `kPreambleSize` (line 56). Cleaned up.

Commit message updated to describe the new `commands_end` bound as well. Rebuilt locally: `cmake --build build --target jxl_dec` succeeds without new warnings.

evilgensec:
Closing in favor of coordinated disclosure per the project's SECURITY.md. Re-routing to libjxl-security@google.com.

eustas (approved):

PR #4810, opened 2026-05-27
Fix unsigned overflow in decoder buffer size ceiling division

metsw24-max:
## Problem

`DivCeil()` currently uses:

```cpp
(a + b - 1) / b
```

This can overflow before the division occurs when `a` is close to the maximum
value of an unsigned type.

Example:

```cpp
DivCeil(SIZE_MAX, 8)
```

internally evaluates:

```cpp
(SIZE_MAX + 7) / 8
```

which wraps on unsigned overflow and can produce a drastically smaller result
than required.

In `decode.cc`, this affected allocation-related calculations derived from
decoder-controlled metadata (`row_bits`, `format->align`), potentially causing
buffer size underestimation.

---

## Fix

Add a new overflow-safe helper:

```cpp
SafeDivCeil()
```

implemented as:

```cpp
a / divisor + (a % divisor != 0 ? 1 : 0)
```

This formulation is mathematically equivalent to ceiling division while avoiding
the overflowing intermediate addition.

The helper is intentionally restricted to unsigned dividend types using:

```cpp
std::enable_if<std::is_unsigned<T1>::value>
```

because the original issue concerns unsigned wraparound semantics.

---

## Updated Call Sites

Replace the security-relevant sizing calculations in `decode.cc`:

```cpp
DivCeil(row_bits, kBitsPerByte)
DivCeil(row_size, format->align)
```

with:

```cpp
SafeDivCeil(...)
```

Existing `SafeMul` protections remain unchanged.

---


eustas:
All compilers are smart enough to use `div` only once. Some compilers (e.g. GCC) emit good code:
```
"SafeDivCeil(unsigned long, unsigned long)":
        mov     rax, rdi
        xor     edx, edx
        div     rsi
        cmp     rdx, 1
        sbb     rax, -1
        ret
"DivCeil(unsigned long, unsigned long)":
        lea     rax, [rsi-1+rdi]
        xor     edx, edx
        div     rsi
        ret
```

Anyways, `DivCeil` is not used on hotpath. So, let's merge `SafeDivCeil` into `DivCeil` and add comment that it is safe (about overflows).

metsw24-max:
## Updated

### `lib/jxl/base/common.h`

Added `SafeDivCeil()` into `DivCeil()` and added an explanatory comment
documenting the overflow-safe behavior and intended usage.

### `lib/jxl/decode.cc`

Replaced all usages of `SafeDivCeil()` with the updated `DivCeil()`
implementation.

eustas (dismissed):

metsw24-max:
@eustas any update on this

eustas:
I know that will cause a lot of changes, but would be better to:
 * limit `DivCeil` only to unsigned (mostly no problem here)
 * merge T1 and T2 to T (at this point most of invocations will require explicit type specifier)
 * limit `T` to be either `size_t` or `uint64_t` (likely this covers all use cases)

eustas:
Also, perhaps add and use DivCeilPow2, as it is the most frequent case.

metsw24-max:
Thanks for the review. I've addressed the requested changes:

* `DivCeil` now uses a single template parameter and is restricted to `size_t`/`uint64_t`.
* Added `DivCeilPow2` and updated the power-of-two call sites.
* Updated the remaining call sites to use explicit typing where required.

Build and tests are passing cleanly. Please let me know if you see anything else that should be adjusted.

metsw24-max:
CI was red because one call site in tools/djxl_fuzzer_corpus.cc was missed when DivCeil got restricted to size_t/uint64_t. Pushed a fix, builds cleanly now.

PR #4812, opened 2026-05-28
Optimize release artifacts

Galaxy4594:
Done with the help of @goodusername123. 
 - The .deb packages themselves are already xz compressed, so using gzip is redundant and unnecessary. A simple tar works here with no significant file size increase.
 - The Linux static builds now use lzip instead of gzip which results in a major file size reduction from 13 MB to 3.84 MB.
 - The Windows builds are slightly smaller, using more aggressive zip compression and Deflate64 instead of Deflate. The Windows static zips went from 39.6 MB to 38.5 MB.
 - Removed `-ms`, 7z is already solid by default.
 - Removed `-mmt=1` so 7z is no longer constrained to 1 thread, allowing it to compress faster.
 - Added `-mtp=2` for Windows zips, we don't need 100 ns precision for file timestamps.

The rationale to use Deflate64 is that it has been supported natively since Windows Vista. Likewise, lzip has been supported in GNU tar utility since 2010.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -64,6 +64,7 @@ jobs:
           libopenexr-dev \
           libpng-dev \
           libwebp-dev \
+          lzip \
           ninja-build \
           pkg-config \
           xdg-utils \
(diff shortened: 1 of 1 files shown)


eustas:
Awesome, thanks!

eustas (approved):

PR #4820, opened 2026-06-02
Adopt Safe Buffers in Huffman table construction

metsw24-max:
Migrate the Huffman lookup-table builder to the Safe Buffers programming model by replacing raw buffer interfaces and pointer arithmetic with bounds-carrying abstractions.

The change introduces `jxl::Span`-based APIs, removes raw pointer traversal from the table construction logic, and enables compiler enforcement with `-Werror=unsafe-buffer-usage` for the migrated translation unit.

### Safe Buffers infrastructure
- Add `JXL_UNSAFE_BUFFER_USAGE_BEGIN/END` helper macros for Clang.
- Wrap the audited internal pointer arithmetic used by `jxl::Span`.
- Add `Span::subspan()` helpers for creating bounded views.

### Huffman table migration
- Change `BuildHuffmanTable()` to accept:
  - `Span<HuffmanCode>`
  - `Span<const uint8_t>`
  - `Span<uint16_t>`
- Replace raw pointer walking with offset-based table management.
- Replace raw sub-buffer creation with `Span::subspan()`.
- Replace `memcpy`-based table replication with bounds-aware element copying.
- Replace the fixed-size offset array with `std::array`.

### Call site updates
- Update decoder call sites to pass `Span` views instead of raw pointers and sizes.

### Regression prevention
- Enable `-Werror=unsafe-buffer-usage` for `huffman_table.cc` when supported by Clang.
- Keep libc-call diagnostics disabled for this TU because surrounding code has not yet been migrated.


metsw24-max:
@eustas any update on this pr?

eustas:
Are there plans to make it project-wide? Shouldn't the "unsafe" parts be hardened?

metsw24-max:
Fair questions.

**Scope:** this one's a pilot rather than a committed project-wide migration. The idea was to land the pattern on a small, self-contained TU first (Huffman table construction has had OOB issues historically) and see whether the approach is wanted before churning the rest of the tree. If you'd rather not take it piecemeal, I'm happy to hold it until there's a plan to do the whole lib.

**Hardening:** you're right that the exemption macros only mark the pointer arithmetic as audited, they don't add any checks. Span itself stays caller-checked, like std::span. If the goal is to genuinely harden the unsafe internals, I can add JXL_DASSERT bounds checks to operator[]/subspan so the audited spots fault in debug builds instead of just being silenced. That would make the migration buy something concrete rather than only satisfying the warning.

Let me know which direction you'd prefer and I'll adjust.

metsw24-max:
Pushed a fix for the red CI: clang 16-18 know -Wunsafe-buffer-usage but not -Wno-unsafe-buffer-usage-in-libc-call, so the unknown option failed under -Werror. The cmake check now probes that flag separately and only passes it when supported.

PR #4825, opened 2026-06-03
extra-channel input stride handling for differing EXR windows

jmestwa-coder:
## Summary

Fix the EXR extra-channel copy path to use the dataWindow-based input stride when addressing `input_extra_rows`.

When `displayWindow` is wider than `dataWindow`, the input buffer is laid out using `row_size` (dataWindow width), but the extra-channel copy loop used `ec.stride` (displayWindow width) to walk the input buffer. This can advance reads past the end of the allocated buffer for files with differing EXR windows and extra channels.

## Changes

* Compute a dedicated input stride using `ec.pixel_stride() * row_size`
* Use the dataWindow-based stride when:

  * Computing `input_ec_row`
  * Advancing `input_ec_slice` between channels
* Keep output image addressing unchanged (`ec.stride`)

## Testing

* Added a regression test covering an EXR with:

  * `displayWindow` wider than `dataWindow`
  * an extra channel (`Z`)
* Verified the test fails before the fix with an AddressSanitizer heap-buffer-overflow in `DecodeImageEXR`
* Verified the test passes after the fix and decodes the extra-channel values correctly

## Impact

* Fixes out-of-bounds reads in the EXR extra-channel copy path for valid files with differing EXR windows
* Preserves existing behavior when `displayWindow` and `dataWindow` have the same dimensions


jonsneyers:
Thanks for the bugfix!

Not sure if we really need to add an explicit test for this, it's not part of the core libjxl library and I don't think a regression on this specific bug will be likely once it's fixed...

eustas:
Should be fixed after #4840

PR #4826, opened 2026-06-03
heap OOB read in EXR extra-channel copy-out

jmestwa-coder:
## Summary

Fix a heap out-of-bounds read in the OpenEXR decoder's extra-channel copy-out path when `displayWindow` and `dataWindow` have different widths.

The extra-channel input buffer is allocated and populated using the `dataWindow` width, but the copy-out logic addressed that buffer using `ec.stride`, which is based on the `displayWindow` width. When the display window is wider than the data window, this can read past the end of the allocated buffer.

## Fix

* Introduce a data-window-based input stride for extra-channel reads.
* Use the input stride when addressing and advancing within `input_extra_rows`.
* Keep output image addressing unchanged.

This restores the intended invariant:

* Input buffer addressing → `dataWindow` stride
* Output image addressing → `displayWindow` stride


jonsneyers:
Is this a duplicate of https://github.com/libjxl/libjxl/pull/4825?

eustas:
Should be fixed after #4840

PR #4830, opened 2026-06-06
Add missing semicolon in squeeze.cc big-endian branch

### LINKED ISSUE #4627: Missing semi-colon in squeeze.cc
The Big Endian portion of the Highway < 1.2 section of this commit is missing a semi-colon.

https://github.com/libjxl/libjxl/commit/aa7e6d1fa4d372b7ed2ebf67f1b14252db5f098f

I have verified the code works correctly after putting the semi-colon in.

eyupcanakman:
### Description

`squeeze.cc` does not compile on big-endian targets that use Highway older than 1.2. The big-endian line of the pre-1.2 `a3` block is missing its semicolon, so it runs into the next statement. The other three branches all end in a semicolon. This one did not.

Adding the semicolon makes the branch compile. It is only selected on big-endian builds with Highway older than 1.2, so it has no effect on builds that use Highway 1.2 or newer.

Fixes #4627

### Pull Request Checklist

- [x] **CLA Signed**
- [ ] **Authors**
- [x] **Code Style**


eustas:
Hi. Is there an example GitHub action that would demonstrate a build failure? That will help much avoiding such problem in the future.

eyupcanakman:
That branch only compiles on a big-endian target built against Highway older than 1.2, so a leg in `build_test_cross.yml` that targets s390x with `third_party/highway` pinned to a 1.0.x tag would catch it, and it only needs to compile, not run. I can add that to this PR if you want.

eustas:
Yes, please add.  That was one of the purposes of adding s390 CI (unfortunately, unfinished).

eustas (approved):

PR #4836, opened 2026-06-09
make fast lossless specialized dec path correct

jonsneyers:
It was technically incorrect, in case some channels would satisfy the condition but others wouldn't, it would not decode correctly because lz77 state would be corrupt.

### DIFF
--- a/lib/jxl/modular/encoding/encoding.cc
+++ b/lib/jxl/modular/encoding/encoding.cc
@@ -191,6 +191,24 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
     return val * multiplier + offset;
   };
 
+  // True iff every decision node in global_tree splits on a static property
+  // (channel or group_id) and every leaf has Gradient predictor with identity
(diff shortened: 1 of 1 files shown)


lib/jxl/modular/encoding/encoding.cc
```diff
@@ -191,6 +191,24 @@ Status DecodeModularChannelMAANS(BitReader *br, ANSSymbolReader *reader,
     return val * multiplier + offset;
   };
 
+  // True iff every decision node in global_tree splits on a static property
+  // (channel or group_id) and every leaf has Gradient predictor with identity
+  // transform. When this holds, all channels collapse to the same single-leaf
```
veluca93: not "the same" :-)

veluca93 (dismissed):

lib/jxl/modular/encoding/encoding.cc
jonsneyers: right, different ctx still

veluca93 (approved):

PR #4838, opened 2026-06-11
Harden release asset upload workflow

Alb3e3:
This tightens the release asset upload step without changing the release process.

The release tag is now passed through a step environment variable and quoted before being used in the `gh release upload` command. This avoids direct GitHub context interpolation inside the shell command while preserving the existing release permissions and artifact upload behavior.

I verified the workflow YAML still parses, `git diff --check` passes, and `zizmor --min-severity medium --min-confidence medium .github/workflows/release.yaml` reports no findings.


eustas:
Closing in favor #4841.

PR #4856, opened 2026-06-18
Limit EXIF/XMP uncompressed box size

eustas:


lib/jxl/decode.cc
```diff
@@ -1687,6 +1687,8 @@ static JxlDecoderStatus HandleBoxes(JxlDecoder* dec) {
       if (dec->store_exif == 1 || dec->store_xmp == 1) {
         std::vector<uint8_t>& metadata =
             (dec->store_exif == 1) ? dec->exif_metadata : dec->xmp_metadata;
+        // The largest possible JPEG APP_DATA payload size.
+        constexpr size_t kBlockSizeLimit = 65533;
```
kmilos: IMHO it should be a limit/error only when attempting writing to JPEG APP1 (and even there there are some workarounds). No other image format has this Exif/XMP payload limit for it to be enforced at this particular point?

eustas: You are right. I'll update the limits and rework storage a bit.
Unfortunately it is impossible to know at this point of workflow if it is for JPEG or not...

eustas:
Friendly ping @veluca93.

veluca93 (approved):

PR #4857, opened 2026-06-22
Improve spline caching:

eustas:
- skip spans below image_ysize
 - linear sorting of splines

Drive-by:
 - store splines cache in managed memory
 - prepare to store spline data in managed memory (custom spline data is passed as view)


mo271 (approved):
Thanks, LGTM!

PR #4860, opened 2026-06-23
Safer calculations in BytesPerRow

eustas:
See also https://github.com/libjxl/libjxl/pull/4765


veluca93 (dismissed):

eustas:
PTAL: fixed scalar + double test (removed double - we don't need it)

veluca93 (approved):

PR #4862, opened 2026-06-23
Deal with mistakenly set "interleaved" alpha.

eustas:
"HasAlpha" is used in a case when we would want to check if "interleaved" alpha channel is present; but implementation is not selective and reports true if any EC has type "Alpha".

Thanks to Arthur Chan / Anthropic for repotring this.


veluca93 (approved):

eustas:
Also thanks to David Pokora / Trail of Bits, in collaboration with Anthropic who reported that earlier

PR #4871, opened 2026-06-26
fix heap over-read for grayscale input in EncodeWithSJpeg

nabhan06:
### Description

`EncodeWithSJpeg` hardcodes the input stride to `xsize * 3` and always calls `sjpeg::Encode`, which interprets the buffer as packed RGB. The JPEG encoder also accepts 1-channel grayscale images (alpha is rejected, so a plain grayscale frame reaches this path), and their pixel buffer is only `xsize * ysize` bytes. sjpeg then reads three bytes per pixel and runs two-thirds past the end of the allocation. Backing a 64x64 grayscale buffer with a guard page makes the read fault inside `sjpeg::RGB24PackedToPlanar`.

sjpeg provides `EncodeGray` for exactly this case, so grayscale now goes there with a 1-byte stride (matching how the libjpeg path already handles grayscale), RGB stays on `Encode`, and any other channel count is rejected up front.

### Pull Request Checklist

### DIFF
--- a/lib/extras/enc/jpg.cc
+++ b/lib/extras/enc/jpg.cc
@@ -509,11 +509,20 @@ Status EncodeWithSJpeg(const PackedImage& image, const JxlBasicInfo& info,
     hook->first_iter_slope = params.search_first_iter_slope;
     param.search_hook = hook.get();
   }
-  size_t stride = image.xsize * 3;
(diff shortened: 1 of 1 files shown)


eustas:
Thanks!

eustas (approved):

PR #4881, opened 2026-06-30
Default GIF disposal method to DISPOSE_DO_NOT

### LINKED ISSUE #4471: cjxl fails on animated GIF images with the disposal method undefined

FooIbar:
And treat 4 as DISPOSE_PREVIOUS to match browser behavior.
See https://github.com/mozilla-firefox/firefox/blob/c681e91369f59d0efae43bdc465872b855e8b269/image/decoders/nsGIFDecoder2.cpp#L707

Fixes #4471 

I wonder if we need to make this configurable as well. In contrast to the delay case, the spec does not define what to do here.

### DIFF
--- a/lib/extras/dec/gif.cc
+++ b/lib/extras/dec/gif.cc
@@ -281,6 +281,14 @@ Status DecodeImageGIF(Span<const uint8_t> bytes, const ColorHints& color_hints,
     GraphicsControlBlock gcb;
     DGifSavedExtensionToGCB(gif.get(), i, &gcb);
     msan::UnpoisonMemory(&gcb, sizeof(gcb));
+
+    // Some encoders (and apparently some specs) represent
+    // DisposalMethod::RESTORE_PREVIOUS as 4, but 3 is used in the canonical
+    // spec and is more popular, so we normalize to 3.
+    if (gcb.DisposalMode == 4) {
+      gcb.DisposalMode = DISPOSE_PREVIOUS;
+    }
+
     bool is_full_size = total_rect.x0() == 0 && total_rect.y0() == 0 &&
                         total_rect.xsize() == canvas.color.xsize &&
                         total_rect.ysize() == canvas.color.ysize;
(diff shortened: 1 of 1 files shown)


Amiralgaby:
I confirm it works, good job

PR #4893, opened 2026-07-02
Bump version (to 0.13.0)

eustas:


veluca93:
Why not 0.12.1?

goodusername123:
> Why not 0.12.1?

AFAIU that would be for the split off [v0.12.x branch/tree](https://github.com/libjxl/libjxl/tree/v0.12.x), the bump to v0.13.0 in this PR is fine since everything after the release for v0.12 within the `main` branch would be for v0.13 with relevant changes backported to the v0.12.x branch. that's my understanding at least.

eustas:
Patch versions are meant for security fixes.

eustas:
(At least before v1.0)

veluca93 (dismissed):
turns out I completely misunderstood what was going on :D

veluca93 (approved):

jonnyawsom3:
> Patch versions are meant for security fixes.
> (At least before v1.0)

Given how long it's been and still no v1.0, along with how widespread adoption has become recently, I feel like putting important fixes into patch releases along with security ones would be a wise idea. If not, then just release as 1.0 and use normal version semantics.

We had `-d 0 -e 1` not actually being lossless for 2 years because only security fixes went into v0.11.1 and v0.11.2, who knows how many images might have been impacted with no way of knowing. Even with v0.13 targeting 2 months from now, that's still a very long time if anything major but not security related is discovered.

PR #4899, opened 2026-07-03
djxl: output XYB with ICC in original color space

### LINKED ISSUE #2289: djxl PNG Conversion Acting Up

FooIbar:
Fixes #2289 

This only changes the output of djxl.
Changing it in libjxl requires API changes since the pixel format is set when the `JXL_DEC_NEED_IMAGE_OUT_BUFFER` event occurs, which is after the `JXL_DEC_COLOR_ENCODING` event.

### DIFF
--- a/lib/extras/dec/jxl.cc
+++ b/lib/extras/dec/jxl.cc
@@ -402,6 +402,22 @@ bool DecodeImageJXL(const uint8_t* bytes, size_t bytes_size,
         ppf->extra_channels_info.push_back({eci, i, name});
       }
     } else if (status == JXL_DEC_COLOR_ENCODING) {
+      size_t icc_size = 0;
+      JxlColorProfileTarget target = JXL_COLOR_PROFILE_TARGET_ORIGINAL;
+      if (JXL_DEC_SUCCESS !=
+          JxlDecoderGetICCProfileSize(dec, target, &icc_size)) {
+        fprintf(stderr, "JxlDecoderGetICCProfileSize failed\n");
+      }
+      if (icc_size != 0) {
+        ppf->orig_icc.resize(icc_size);
+        if (JXL_DEC_SUCCESS !=
+            JxlDecoderGetColorAsICCProfile(dec, target, ppf->orig_icc.data(),
+                                           icc_size)) {
+          fprintf(stderr, "JxlDecoderGetColorAsICCProfile failed\n");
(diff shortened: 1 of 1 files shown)


jonnyawsom3:
I think (eventually) changing the API would be worth it. There's still an extremely large amount of programs that don't do color management correctly/at all, and even with correct management, 8bit Linear causes severe banding/dithering in shadowed areas.

FooIbar:
> I think (eventually) changing the API would be worth it. There's still an extremely large amount of programs that don't do color management correctly/at all, and even with correct management, 8bit Linear causes severe banding/dithering in shadowed areas.

They need a CMS to display lossless JXL properly anyway.
They should set the output color profile to the original/display profile and let libjxl handle the conversion, or request float output and do the transform with an external CMS themselves.

jonnyawsom3:
I agree, they should... But best practice and the real world are often far apart unfortunately, so I think mitigating it at the source (in libjxl) would be best. Open to discussion though, especially with Jon and Sami.

FooIbar:
> I agree, they should... But best practice and the real world are often far apart unfortunately, so I think mitigating it at the source (in libjxl) would be best. Open to discussion though, especially with Jon and Sami.

Standard libjxl builds come with a default CMS bundled, so they just need to set the output color profile. If someone explicitly disables the CMS, I'd assume they have the knowledge to handle color management themselves.

PR #4913, opened 2026-07-15
lib/jxl: adopt Safe Buffers model in icc_codec and span primitive

ghost:
## Summary

This PR introduces the infrastructure and a first concrete conversion for the **Safe Buffers Programming Model** in libjxl, targeting the [`-Wunsafe-buffer-usage`](https://clang.llvm.org/docs/SafeBuffers.html) Clang diagnostic (available since Clang 14).

### Motivation

`-Wunsafe-buffer-usage` flags raw pointer subscript (`ptr[i]`) and pointer arithmetic (`ptr + n`) so that security-sensitive code can be progressively migrated from C-style `(pointer, length)` pairs to bounds-carrying C++ span types.  Span types eliminate a whole class of buffer-overrun bugs by making the valid range part of the type rather than an invisible convention between caller and callee.

### Changes

**`lib/jxl/base/compiler_specific.h`** — new suppression macros  
`JXL_UNSAFE_BUFFERS_BEGIN` / `JXL_UNSAFE_BUFFERS_END` wrap the Clang diagnostic push/ignore/pop for `-Wunsafe-buffer-usage` on Clang ≥ 14, and expand to nothing on older compilers and MSVC.  These are modelled on the analogous macros in Chromium's `base/compiler_specific.h`.

**`lib/jxl/base/span.h`** — suppress in the span primitive itself  
`operator[]`, `end()`, and `remove_prefix()` contain intentional raw pointer arithmetic that is by construction safe.  Suppressing the diagnostic here means that *call sites* using `span[i]` or a range-for over a `Span` are automatically clean — no annotation needed at the point of use.

**`lib/jxl/icc_codec.cc`** — convert two internal helpers  

| Before | After |
|---|---|
| `DecodeVarInt(const uint8_t* input, size_t inputSize, …)` | `DecodeVarInt(Bytes input, …)` |
| `Shuffle(JxlMemoryManager*, uint8_t* data, size_t size, …)` | `Shuffle(JxlMemoryManager*, Span<uint8_t> data, …)` |

Both functions live in the anonymous namespace and are only called within this translation unit.  The size argument is now encoded in the span type, making it impossible for callers to accidentally pass a mismatched length.

Updated callers:
- `CheckPreamble` creates a `Bytes` view of its `PaddedBytes` argument, eliminating the raw `enc`/`size` locals.
- `UnpredictICC` introduces a `commands` `Bytes` view bounded to `[0, commands_end)`, making the two-stream (commands vs data) read structure explicit and preventing commands-stream varints from reading into the data section.
- `Shuffle` call sites pass `Span<uint8_t>{shuffled.data(), num}`.

### No functional change

The patch is a pure refactor: identical runtime behaviour, identical binary output, no new dependencies.  It lays the groundwork for project-wide `-Wunsafe-buffer-usage` enforcement (follow-up work would add per-file pragma suppression to unconverted files and enable the flag in CI).

### Testing

Existing unit tests cover `UnpredictICC` and `DecodeVarInt` exhaustively via the ICC codec fuzzer and conformance suite; no new tests are needed for a pure rename/type-widening refactor.

### DIFF
--- a/lib/jxl/base/compiler_specific.h
+++ b/lib/jxl/base/compiler_specific.h
@@ -215,4 +215,24 @@
 #define JXL_CRASH() (void)__builtin_trap()
 #endif
 
+// Suppresses -Wunsafe-buffer-usage for intentional low-level buffer operations
+// that are themselves safe but use raw pointer arithmetic internally, such as
+// the implementation of Span::operator[]. Wrap only the minimal scope that
+// requires suppression; do NOT use these macros to silence warnings in
+// application logic.
+//
+// Example usage:
+//   JXL_UNSAFE_BUFFERS_BEGIN
+//   constexpr T& operator[](size_t i) const noexcept { return *(data() + i); }
+//   JXL_UNSAFE_BUFFERS_END
+#if JXL_COMPILER_CLANG >= 1400
+#define JXL_UNSAFE_BUFFERS_BEGIN \
+  _Pragma("clang diagnostic push") \
+  _Pragma("clang diagnostic ignored \"-Wunsafe-buffer-usage\"")
+#define JXL_UNSAFE_BUFFERS_END _Pragma("clang diagnostic pop")
+#else
+#define JXL_UNSAFE_BUFFERS_BEGIN
+#define JXL_UNSAFE_BUFFERS_END
+#endif
+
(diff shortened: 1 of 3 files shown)


CYANO-01:
@googlebot I have signed the CLA already.

PR #4914, opened 2026-07-15
parse color description numbers with the C locale

nabhan06:
### Description

`ParseDouble` in `lib/extras/dec/color_description.cc` parses the fractional values in a color description (custom white point / primaries `x;y` coordinates, and the gamma in forms like `g2.2`) with `strtod`, which follows the process `LC_NUMERIC` locale. When the embedding application has set a locale whose decimal separator is not `.`, `strtod` stops at the dot and silently returns a truncated value, so `0.3127` parses as `0` and `g2.2` as gamma `2`, decoding the wrong color without any error. This parses the value with the classic C locale instead, matching how the extras code already forces `std::locale::classic()` for case folding. The added test fails on the current code under a comma-decimal locale and passes with the change.

### Pull Request Checklist

### DIFF
--- a/lib/extras/dec/color_description.cc
+++ b/lib/extras/dec/color_description.cc
@@ -8,10 +8,10 @@
 #include <jxl/color_encoding.h>
 
 #include <array>
-#include <cerrno>
 #include <cmath>
 #include <cstddef>
-#include <cstdlib>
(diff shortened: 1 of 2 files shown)


nabhan06:
gentle ping

PR #4916, opened 2026-07-16
validate extra channel output bit depth in the decoder API

nabhan06:
### Description

`JxlDecoderSetExtraChannelBuffer` never checks the requested output bit depth against the buffer's pixel type, unlike the color path, which goes through `VerifyOutputBitDepth`. Under `JXL_BIT_DEPTH_FROM_CODESTREAM` (what `djxl` selects by default) a binary32 extra channel reports 32 bits, and that value reaches `StoreUnsignedRow`, where `(1u << out.bits_per_sample_) - 1` shifts a 32-bit unsigned by 32; UBSan reports `shift exponent 32 is too large for 32-bit type 'unsigned int'` at `stage_write.cc:551` when running `djxl` on a grayscale file carrying a float extra channel, and since the multiplier lands on 0 the extra channel is also written out as a PGM with `MAXVAL 0` that our own PNM decoder then rejects. The existing `VerifyOutputBitDepth` template moves up next to `GetBitDepth` and is now called from the extra channel setter, and `JxlDecoderSetImageOutBitDepth` re-checks extra channel outputs that are already registered so that either call order is covered; extra channels whose bit depth fits the output type decode exactly as before.

### Pull Request Checklist

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -188,6 +188,21 @@ uint32_t GetBitDepth(JxlBitDepth bit_depth, const T& metadata,
   return 0;
 }
 
+template <typename T>
+JxlDecoderStatus VerifyOutputBitDepth(JxlBitDepth bit_depth, const T& metadata,
+                                      JxlPixelFormat format) {
+  uint32_t bits_per_sample = GetBitDepth(bit_depth, metadata, format);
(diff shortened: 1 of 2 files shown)


nabhan06:
gentle ping

PR #4928, opened 2026-07-30
Rewrite of LZ77 implementation for lossless image compression

linusvdv:
### Description
Currently most encoder settings disable LZ77 as it is too slow.
This PR rewrites the LZ77 compression, which makes it both faster and gives the option to have many more speed settings.
This LZ77 implementation can be used for fast speed settings which disable predictions or for some few images (mostly screenshots) where the predictions are enabled. If compression would get worse LZ77 disables itself automatically. 

### Technical Details
Instead of using a hash chain, the new implementation uses a custom hash map with a ring buffer as buckets. The size of the ring buffer can be changed using `BucketSize`. The smaller `BucketSize` is the faster the algorithm becomes but fewer collisions get checked, which results in weaker compression.
Additionally, the hash map uses a new hashing function (changed in both LZ77 and optimal) where the `HashWindowSize` 
can change how many values are used in the hash computation.
`RuntimeCostComparison` is a flag which enables discarding LZ77 matches which have higher expected cost than the raw values. Even with `RuntimeCostComparison` disabled the final cost check over the whole stream is done.
Lazy matching has been removed as it was not speed/compression effective.
With a LZ77 match, skipped positions will still update the hash map.

### Benchmark
 LZ77_implementation_comparision

_Speed/Compression comparison with d0:I0:P0 on 14 images_

### Pull Request Checklist

### DIFF
--- a/lib/jxl/enc_lz77.cc
+++ b/lib/jxl/enc_lz77.cc
@@ -169,6 +169,7 @@ std::vector<std::vector<Token>> ApplyLZ77_RLE(
 }
 
 // Hash chain for LZ77 matching
+template <uint32_t MaxChainLength>
 struct HashChain {
   size_t size_;
   std::vector<uint32_t> data_;
@@ -196,7 +197,7 @@ struct HashChain {
   std::unordered_map<int, int> special_dist_table_;
   size_t num_special_distances_ = 0;
 
-  uint32_t maxchainlength = 256;  // window_size_ to allow all
+  uint32_t maxchainlength = MaxChainLength;  // window_size_ to allow all
 
   HashChain(const Token* data, size_t size, size_t window_size,
             size_t min_length, size_t max_length, size_t distance_multiplier)
@@ -234,19 +235,23 @@ struct HashChain {
     }
   }
 
+
   uint32_t GetHash(size_t pos) const {
-    uint32_t result = 0;
-    if (pos + 2 < size_) {
-      // TODO(lode): take the MSB's of the uint32_t values into account as well,
-      // given that the hash code itself is less than 32 bits.
-      result ^= static_cast<uint32_t>(data_[pos + 0] << 0u);
-      result ^= static_cast<uint32_t>(data_[pos + 1] << hash_shift_);
-      result ^= static_cast<uint32_t>(data_[pos + 2] << (hash_shift_ * 2));
+    uint32_t h = 0;
+    if (pos + 2 < data_.size()) {
+      for (size_t i = 0; i < 3; i++) {
+        h ^= data_[pos+i] + 0x9e3779b9 + (h << 6) + (h >> 2);
+      }
+      h ^= h >> 16;
+      h *= 0x85ebca6bu;
+      h ^= h >> 13;
+      h *= 0xc2b2ae35u;
+      h ^= h >> 16;
+
     } else {
-      // No need to compute hash of last 2 bytes, the length 2 is too short.
       return 0;
     }
-    return result & hash_mask_;
+    return h & hash_mask_;
   }
 
   uint32_t CountZeros(size_t pos, uint32_t prevzeros) const {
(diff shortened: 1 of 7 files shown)


lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+      hash_map.FindMatch(pos, len, dist_symbol, min_length);
+      if (len < min_length) continue;
+
+      // Bit cost comparison: raw tokens vs LZ77 pair
+      float cost = sym_cost[pos + len] - sym_cost[pos];
+      size_t lz77_len = len - lz77.min_length;
+      float lz77_cost = LenCost(lz77_len) + DistCost(dist_symbol) +
+                        sce.AddSymbolCost(out.back().context);
+
+      if constexpr (RuntimeCostComparison) {
```
veluca93: That's C++17 - libjxl is still at C++11.

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
   return kCostTable[tok] + nbits;
 }
 
+
+// Fast hash map for LZ77 compression using a fixed-size ring buffer per bucket.
+// Overwrites the oldest candidates when a bucket reaches max capacity.
+template <uint16_t BucketSize, int HashWindowSize>
```
veluca93: I'd call "HashWindowSize" just "HashSize".

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+        hash_table_[h].size = std::min(static_cast<uint16_t>(hash_table_[h].size + 1), BucketSize);
+        hash_table_[h].data[hash_table_[h].idx] = static_cast<uint32_t>(pos);
+        hash_table_[h].idx = hash_table_[h].idx + 1 - static_cast<uint16_t>(hash_table_[h].idx + 1 >= BucketSize) * BucketSize;
+    }
+
+private:
+    // Fixed-capacity bucket stored as a ring buffer.
+    struct HashBucket {
+        std::array<uint32_t, BucketSize> data;
+        uint16_t idx = 0;   // Insertion index
```
veluca93, line 525: with idx and size stored inline, you might want to make BucketSize a power-of-two-minus-one so that HashBucket has power of 2 size (not sure if it matters speed-wise)

lib/jxl/enc_lz77.cc
```diff
@@ -169,6 +169,7 @@ std::vector<std::vector<Token>> ApplyLZ77_RLE(
 }
 
 // Hash chain for LZ77 matching
+template <uint32_t MaxChainLength>
```
veluca93: I'd name these params like a constant (kMaxChainLength)

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+private:
+    // Fixed-capacity bucket stored as a ring buffer.
+    struct HashBucket {
+        std::array<uint32_t, BucketSize> data;
+        uint16_t idx = 0;   // Insertion index
+        uint16_t size = 0;  // Current entry count (<= BucketSize)
+    };
+
+    // Computes a Murmur-style mix hash over HashWindowSize elements.
+    uint32_t GetHash(size_t pos) const {
```
veluca93: This is duplicated with GetHash above, please make a single function

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+            h ^= h >> 16;
+        } else {
+            return 0; // Out-of-bounds safety check
+        }
+        return h & hash_mask_;
+    }
+
+    // Measures matching prefix length between indices 'a' and 'b'.
+    size_t MatchLength(size_t a, size_t b) const {
+        size_t len = 0;
```
veluca93, line 532: Assert a < b?

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+            size_t cur_dist_symbol = (num_special_distances_ + dist - 1);
+            if (dist < special_dist_table_.size()) {
+                int lookup = special_dist_table_[dist];
+                if (lookup >= 0) {
+                    cur_dist_symbol = lookup;
+                }
+            }
+
+            // Reject if same length but requires a worse/longer distance symbol
+            if (cur_length == len && cur_dist_symbol >= dist_symbol) {
```
veluca93, line 499: Have you tried instead comparing by best cost - distance cost? I imagine getting an extra byte at the cost of having distance go from 1 to 1000 is not worth it.

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
+      if (len < min_length) continue;
+
+      // Bit cost comparison: raw tokens vs LZ77 pair
+      float cost = sym_cost[pos + len] - sym_cost[pos];
+      size_t lz77_len = len - lz77.min_length;
+      float lz77_cost = LenCost(lz77_len) + DistCost(dist_symbol) +
+                        sce.AddSymbolCost(out.back().context);
+
+      if constexpr (RuntimeCostComparison) {
+        if (lz77_cost > cost) {
```
veluca93: Does the "runtime cost comparison" still do worse with (a) the change to FindMatch above and (b) a scaling factor applied to `cost`?

lib/jxl/enc_lz77.cc
```diff
@@ -436,113 +441,215 @@ float DistCost(size_t dist) {
...
-            out.push_back(in[i + j]);
+      // Find longest sequence match in current bucket
+      hash_map.FindMatch(pos, len, dist_symbol, min_length);
+      if (len < min_length) continue;
+
+      // Bit cost comparison: raw tokens vs LZ77 pair
+      float cost = sym_cost[pos + len] - sym_cost[pos];
+      size_t lz77_len = len - lz77.min_length;
+      float lz77_cost = LenCost(lz77_len) + DistCost(dist_symbol) +
+                        sce.AddSymbolCost(out.back().context);
```
veluca93: I know this was there before, but I would be moderately surprised if this is the right thing to do - especially because we only really pay the AddSymbolCost once per _context_.
What happens if you remove it?

lib/jxl/enc_lz77.cc
linusvdv: the current implementation of the cost estimation is not strong enough. The compression does not improve.

lib/jxl/enc_lz77.cc
linusvdv: did not currently help. this would probably also need a new cost estimation to work

veluca93 (approved):

PR #4929, opened 2026-08-01
Tune LZ77 cost comparison and match evaluation

Galaxy4594:
In the recent LZ77 rewrite there was a mention that the `t` modes compress worse than `f` modes, in my testing I can confirm their findings. This PR fixes a couple bugs which seems to fix this issue. (I haven't tested on a wide range of images yet)

1. Context cost missing in comparison: `AddSymbolCost` was left out when checking `lz77_cost > cost`, but was subtracted from `bit_decrease` later. Because `lz77_cost` was underestimated during the decision check, it ended up accepting matches that were actually worse for compression than keeping raw literals.

2. Skipping ahead on match rejection: When `kRuntimeCostComparison` rejected a match of length `len` at position `pos`, it was emitting `len - 1` literals and jumping `pos` forward by `len - 1`. Skipping over those intermediate positions meant we missed any potential matches starting at `pos + 1`, `pos + 2`, etc. 

Fixing `lz77_cost` to include `AddSymbolCost` up front and letting the loop naturally advance to `pos + 1` on rejection resolves both regressions.

A quick benchmark:
```
-d 0 -e 7 -g 3 -P 0 -I 0
kLZ77b3w3f
Compressed to 485.4 kB (1.873 bpp).

kLZ77b3w3t Main
Compressed to 487.4 kB (1.881 bpp). +2 kB

kLZ77b3w3t Post PR
Compressed to 482.6 kB (1.862 bpp). -2.8 kB
```

### DIFF
--- a/lib/jxl/enc_lz77.cc
+++ b/lib/jxl/enc_lz77.cc
@@ -607,7 +607,7 @@ std::vector<std::vector<Token>> ApplyLZ77_LZ77(
       if (kRuntimeCostComparison && lz77_cost > cost) {
         // If literal cost is very low (e.g. solid colors), skip the entire match
         // to avoid O(n^2) behavior.
-		if (cost < len) {
+		if (cost < 0.25f * len) {
           for (size_t offset = 1; offset < len; offset++) {
             out.push_back(in[pos+offset]);
             hash_map.Update(pos + offset);
--- a/lib/jxl/enc_lz77.cc
+++ b/lib/jxl/enc_lz77.cc
@@ -638,7 +638,7 @@ std::vector<std::vector<Token>> ApplyLZ77_Optimal(
     const HistogramParams& params, size_t num_contexts,
     const std::vector<std::vector<Token>>& tokens, const LZ77Params& lz77) {
   std::vector<std::vector<Token>> tokens_for_cost_estimate =
-      ApplyLZ77_LZ77(params, num_contexts, tokens, lz77);
+      ApplyLZ77_LZ77<7, 3, true>(params, num_contexts, tokens, lz77);
   // If greedy-LZ77 does not give better compression than no-lz77, no reason to
   // run the optimal matching.
   if (tokens_for_cost_estimate.empty()) return {};
--- a/lib/jxl/enc_lz77.cc
+++ b/lib/jxl/enc_lz77.cc
@@ -605,6 +605,15 @@ std::vector<std::vector<Token>> ApplyLZ77_LZ77(
 						sce.AddSymbolCost(out.back().context);
 
       if (kRuntimeCostComparison && lz77_cost > cost) {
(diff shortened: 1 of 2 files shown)


lib/jxl/enc_lz77.cc
```diff
@@ -601,19 +601,15 @@ std::vector<std::vector<Token>> ApplyLZ77_LZ77(
       // Bit cost comparison: raw tokens vs LZ77 pair
       float cost = sym_cost[pos + len] - sym_cost[pos];
       size_t lz77_len = len - lz77.min_length;
-      float lz77_cost = LenCost(lz77_len) + DistCost(dist_symbol);
+      float lz77_cost = LenCost(lz77_len) + DistCost(dist_symbol) +
+						sce.AddSymbolCost(out.back().context);
```
veluca93, line 605: I am not necessarily convinced that adding sce.AddSymbolCost here helps - does getting rid of it entirely work well?

linusvdv:
This was also previously tried, but it is really slow on single colored images. (Worst case `O(n^2)`). Because single coloured images can be compressed really efficiently with entropy encoding, even replacing the whole section can have a worse cost estimation. This results in matching the same pattern n times.

linusvdv:
Using the AddSymbolCost not in the rejection condition was from our testing stronger. This could be small sample size and should be checked on more images

veluca93:
> This was also previously tried, but it is really slow on single colored images. (Worst case `O(n^2)`). Because single coloured images can be compressed really efficiently with entropy encoding, even replacing the whole section can have a worse cost estimation. This results in matching the same pattern n times.

Good point, but this can also be dealt with by keeping the "skip everything" path, but only triggering it if the literal cost is low.

lib/jxl/enc_lz77.cc
Galaxy4594: It's a mixed bag, the results depend on which lz77 level is used. On my test image `kOptc256` is slightly worse, but using `klz77b(x)w3(x)` is slightly better. More testing is needed.

lib/jxl/enc_lz77.cc
```diff
@@ -601,19 +601,24 @@ std::vector<std::vector<Token>> ApplyLZ77_LZ77(
...
 
       if (kRuntimeCostComparison && lz77_cost > cost) {
-        for (size_t offset = 1; offset < len; offset++) {
-          out.push_back(in[pos+offset]);
-          hash_map.Update(pos + offset);
+        // If literal cost is very low (e.g. solid colors), skip the entire match
+        // to avoid O(n^2) behavior.
+		if (cost < len) {
```
veluca93: This could go into the if above, and I'm not sure cost < len is the best option - maybe cost < 0.2 * len or so? (or some tweakable constant in general - if this branch triggers and `len` is high, `cost` should be quite low anyway)

PR #4932, opened 2026-08-05
Enable LF smoothing for 444 JPEG transcoding by default and allow decoding subsampled smoothing

jonnyawsom3:
Significantly improves banding for low quality JPEGs while reducing filesize by ~1 byte. Requires #4931.
While 444 smoothing is allowed, subsampled smoothing requires a clarification/change to the spec, so only 444 is smoothed by default until applications have been given time to update. jxl-rs and jxl-oxide already support it.

image
image

### DIFF
--- a/lib/include/jxl/encode.h
+++ b/lib/include/jxl/encode.h
@@ -417,6 +417,11 @@ typedef enum {
    */
   JXL_ENC_FRAME_SETTING_OUTPUT_MODE = 40,
 
+  /** Enable or disable LF Smoothing for lossless JPEG recompression.
+   * -1 = default, 0 = disable smoothing, 1 = enable smoothing.
+   */
+  JXL_ENC_FRAME_SETTING_JPEG_RECON_LFS = 41,
+
   /** Enum value not to be used as an option. This value is added to force the
    * C compiler to have the enum to take a known size.
    */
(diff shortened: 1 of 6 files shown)


goodusername123:
IMO encoder side default enablement should really be staggered by at least one release/version.

jonnyawsom3:
When I get round to it I'll add a flag to enable/disable it, with it being disabled by default for a release before being switched to enabled by default once decoder updates have rolled out.

jonnyawsom3:
Moved it behind a flag that's disabled by default, to allow time for decoders to update before creating any non-444 files.

goodusername123:
code comments (and the CLI printout/help) containing "LF Smoothing" should probably be changed to say something like "Adaptive LF Smoothing" instead so that it's a bit more clear/verbose.
Also optionally the acronym used in code could be changed from something like "LFS"/"lfs" to "ALFS"/"alfs", although I'm less decided on this since it does look a bit silly.

jonnyawsom3:
There's no other kind so I think it's fine. It's already under `Options for experimentation / benchmarking` in cjxl too.

jonnyawsom3:
Changed the flag to have a default -1 value, which applies LF smoothing to 444 JPEGs by default, with the option to disable with 0 or enable for non-444 with 1. Also did a flyby fix for `--output_mode` in the cjxl help, as it was missing a newline.

goodusername123:
> Changed the flag to have a default -1 value, which applies LF smoothing to 444 JPEGs by default, with the option to disable with 0 or enable for non-444 with 1. Also did a flyby fix for `--output_mode` in the cjxl help, as it was missing a newline.

would it make sense to have/add `2` as a way to explicitly opt into 444 only?

Issue #4670, opened 2026-03-12
Decoder should switch to "unhealthy" state on API misuse


eustas:
When caller does invalid calls, e.g. set output callback after setting output buffer, then any further decoding should not happen.

Motivation: such situation means programming error; while it might not fail immediately (due to good luck), results will definitely be unexpected eventually. If decoding does not happen, then it is more likely that problem in code will be detected and fixed way earlier.

Issue #4921, opened 2026-07-20
jxl::Span bounds are unchecked in release, which makes Safe Buffers migrations a no-op


anonsurf004:
While looking at the Safe Buffers work in #4820, I went to check what `jxl::Span`
actually guarantees at runtime, and I think there's a gap worth addressing before
more call sites are migrated.

## The gap

`Span::operator[]` performs no bounds check at all — not in release, and not in
debug either:

```cpp
// lib/jxl/base/span.h
constexpr T& operator[](size_t i) const noexcept {
  // MSVC 2015 accepts this as constexpr, but not ptr_[i]
  return *(data() + i);
}
```

This is inconsistent with the rest of the type: `remove_prefix` does validate via
`JXL_ENSURE`. So `Span` looks like a checked abstraction at some entry points and
is a raw pointer at the one that matters most.

The consequence is that migrating `ptr[i]` to `span[i]` produces **identical code**
and identical behaviour on an out-of-bounds index. A migration diff can look like a
memory-safety improvement while changing nothing that a fuzzer or an attacker would
notice.

This is, I think, the substance of @eustas's question on #4820 ("Shouldn't the
'unsafe' parts be hardened?"), and I wanted to make it concrete and actionable
rather than leave it as a review comment.

## Why debug asserts don't close it

The natural fix is `JXL_DASSERT` in `operator[]`, but that is compiled out of every
build that ships:

```cpp
// lib/jxl/base/status.h
#else
#define JXL_DASSERT(condition)
#endif
```

The comment above it is explicit that it does not run "in opt or release". So a
DASSERT-based fix protects the test suite and leaves the shipped decoder exactly as
it is today.

`JXL_ENSURE` does survive release, but it expands to a `return JXL_FAILURE(...)`,
so it cannot be used inside `operator[]`, which has to produce a `T&`.

That means there is currently no release-surviving check in the codebase that is
usable in an expression context — which is why I think this needs a small addition
rather than just a call-site change.

## Proposal

Add a `JXL_SECURITY_CHECK` that (a) survives release builds, (b) is a single
expression so it works inside the existing C++11-compatible `constexpr` accessors,
and (c) traps rather than reading out of bounds. It can be built on the existing
`JXL_CRASH()`, which is already defined unconditionally (`__builtin_trap()`, or
`__debugbreak(), abort()` on MSVC) — unlike `jxl::Abort()`, which only exists in
debug builds.

```cpp
#ifndef JXL_HARDENED
[...]

Issue #4872, opened 2026-06-27
Proposal: a small bounds-checked byte reader to consolidate hand-rolled cursor logic in lib/extras/dec


slegarraga:
Hi, and thanks for libjxl.

While reading the format decoders under `lib/extras/dec/`, I noticed several of them independently re-implement the same "advance a cursor, bail out if it would pass the end" logic:

- `pgx.cc` has a `Parser` with `pos_`/`end_` and `ParseUnsigned` / `SkipSpace` / `SkipLineBreak` / `SkipSingleWhitespace`, each guarding `pos_ == end_` by hand.
- `pnm.cc` has a structurally identical `Parser` with its own `pos_`/`end_`, `ParseUnsigned`, `ParseSigned`, keyword scanning, etc.
- The ICC path (`icc_codec.cc`) similarly threads a raw `const uint8_t* enc, size_t size, size_t pos` tuple with manual `if (pos >= size)` / `CheckOutOfBounds` checks.

Each instance is careful today. My observation is only that the "never read past `end`" invariant is re-established by hand in every decoder, so it lives in the reviewer's head rather than in the type, and every new decoder (or edit) has to re-derive it.

### Proposal (small, opt-in, behavior-preserving)

Introduce a minimal bounds-checked byte cursor — e.g. a `ByteReader` over `Span<const uint8_t>` — exposing the few operations these parsers actually need:

- `ReadU8() -> StatusOr<uint8_t>` (fails at end)
- `Read(n) -> StatusOr<Bytes>` (fails if fewer than `n` bytes remain)
- `Peek()` / `Skip(n)` / `remaining()` / `AtEnd()`

with each accessor returning a `Status` / `StatusOr` on overrun instead of relying on a preceding manual check, keeping the existing `Status`-based error flow.

### Suggested scope for a first PR (to keep it reviewable)

- Add the reader plus unit tests.
- Migrate exactly one decoder (`pgx.cc`, the smallest self-contained one) to use it, with **no behavioral change** — identical accepted/rejected inputs, verified against `pgx_test` and the decoder fuzz targets.
- Only consider `pnm.cc` / ICC as later follow-ups if the direction proves worthwhile.

Before writing it up: **is this a direction you'd welcome**, or would you rather keep the per-decoder hand-rolled cursors? I'm happy to adjust the shape of the abstraction (or drop it) based on your preference — and if there's already a preferred internal pattern for this that I missed, please point me at it and I'll follow that instead.

Thanks!

PR #4933, opened 2026-08-06
decode: stop decoding after mixed image-output API misuse

endorphin13:
Addresses the output-API case in #4670.

Mixing the buffer and callback image-output APIs is already rejected with `JXL_API_ERROR`:

- `JxlDecoderSetImageOutBuffer` when a callback is installed
- `JxlDecoderSetMultithreadedImageOutCallback` when a buffer is installed

But the decoder stays usable afterwards, so a caller that does not check the return value keeps decoding while possibly wrong about which output sink is installed. #4670 asks that decoding stop in this situation so the mistake surfaces where it happens.

This latches the existing `DecoderStage::kError` at both detection points. No new state is added — `JxlDecoderProcessInput` already refuses to run in `kError`, and `JxlDecoderReset` already clears it, so the decoder stays reusable.

**Tests** (`lib/jxl/decode_test.cc`): buffer→callback, callback→buffer, and that `JxlDecoderReset` clears the state and a frame still decodes to completion.

**Testing:** `decode_test` 2894/2894 (2891 existing + 3 new) under ASan+UBSan and in a release/NDEBUG build. `./ci.sh lint` clean.

The issue says "invalid calls, e.g. ...", so I kept this narrow to the image-output case — happy to extend it to other API-misuse paths if that's what you meant.

### DIFF
--- a/lib/jxl/decode.cc
+++ b/lib/jxl/decode.cc
@@ -2585,6 +2585,12 @@ JxlDecoderStatus JxlDecoderSetImageOutBuffer(JxlDecoder* dec,
     return JXL_API_ERROR("No image out buffer needed at this time");
   }
   if (dec->image_out_buffer_set && !!dec->image_out_run_callback) {
+    // Mixing the callback and buffer output APIs is a programming error rather
+    // than a recoverable input condition: the caller may hold an incorrect
+    // belief about which output sink is installed. Per issue #4670, latch
+    // kError so no further decoding happens and the mistake surfaces at its
+    // origin. JxlDecoderReset returns the decoder to a usable state.
+    dec->stage = DecoderStage::kError;
     return JXL_API_ERROR(
         "Cannot change from image out callback to image out buffer");
   }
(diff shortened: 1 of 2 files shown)


endorphin13:
Unrelated to this PR, but while working in decode.cc I noticed the container-control
parsing reads directly from next_in in several places — LoadBE32(dec->next_in + 4)
for the ftyp version, LoadBE32(dec->next_in) for the jxlp counter,
memcpy(..., dec->next_in + header_size, 4) for the brob type, and
dec->next_in[0]/[1] for the signature. AdvanceInput also guards its bounds with
JXL_DASSERT, so in release builds an over-advance would underflow avail_in rather
than fail.

Would you be open to extracting the container-control parsing into its own
translation unit behind a small checked cursor, returning a validated result before
any decoder state is written? I'd aim for behaviour-preserving — no changes to
accepted inputs or resource limits — and I'd want to fit in with the existing Safe
Buffers work (#4820, #4913, #4921) rather than add a third primitive alongside #4872.

Happy to open a separate issue with the details if that's a direction you'd consider;
I didn't want to write a lot of code before asking.

endorphin13:
Hello Team, any comments on this?

PR #4940, opened 2026-08-11
Fix mingw32 build

eustas:


veluca93 (approved):

jonnyawsom3:
I noticed that while merging, the bullseye job failed with:
```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/libx/libxml2/libxml2_2.9.10%2bdfsg-6.7%2bdeb11u10_amd64.deb  Error reading from server - read (104: Connection reset by peer) [IP: 199.232.154.132 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Fetched 175 MB in 2s (99.3 MB/s)
Error: Process completed with exit code 100.
```
I've seen errors fetching from servers a few times, possibly from all the jobs downloading in parallel and triggering DDOS protection. Do you think making use of https://github.com/actions/cache could help? At minimum, it should shave a few minutes off the actions by not re-downloading all dependencies every time.

eustas:
`setup-msys2` has `cache` parameter, and by default it is `true`, so likely we can't do better (not sure, perhaps disabling `update` could help).

`avif` for clang32 build system have been dropped somewhere after 2023; so now it is turn for mingw32.
Ideally, for such setups we should have pre-baked docker images, but two factors are stopping me:
 - rules of how downstream traffic is accounted are somewhat unclear; don't want one day see that __we__ can't build because someone else has depleted out quota
 - basically, I have no capacity to make that change =(

PR #4942, opened 2026-08-11
icc: adopt Safe Buffers (Span / std::array) in the ICC decoder

DarkLycn1976:
# icc: adopt Safe Buffers (Span / std::array) in the ICC decoder

## What

Migrates the raw-pointer buffer accesses in the ICC-profile decoder to bounds-carrying
views, eliminating every `-Wunsafe-buffer-usage` diagnostic that originates in
`lib/jxl/icc_codec.cc` (14 → 0), with **no behavioural change**:

- `DecodeVarInt(const uint8_t* input, size_t inputSize, …)` →
  `DecodeVarInt(Span<const uint8_t> input, …)`. The pointer and its length now travel
  together, removing the class of bug where the two arguments desync.
- `UnpredictICC` and `CheckPreamble` wrap their `(const uint8_t*, size_t)` input in a
  `Span<const uint8_t>` view and index through it; sub-range varint reads pass an explicit
  `Span<const uint8_t>(buf.data(), commands_end)` so the effective bound is expressed in the
  type, not tracked separately.
- `Shuffle` indexes its in/out buffer through a `Span<uint8_t>` view.
- `kTagStrings` / `kTypeStrings` in `icc_codec_common.h` change from C arrays to
  `std::array` (already-included header), so the bounds-checked table lookups are no longer
  flagged as unsafe pointer subscripts.

## Why

`icc_codec.cc` decodes attacker-controlled ICC data embedded in JXL images — a classic
memory-safety-sensitive parser. The existing code is correct: every access is guarded by an
explicit bounds check (`CheckOutOfBounds`, `if (pos >= size)`, `commands_end` limits). This
change is **secure-by-design hardening**: it converts those *manually verified* bounds into
*structurally expressed* ones (a pointer that carries its length), so the accesses read as
safe to `-Wunsafe-buffer-usage` and a future edit is less able to silently reintroduce an
out-of-bounds pointer walk. It is a step toward compiling the ICC decoder under Safe Buffers.

This aligns with the Safe Buffers programming model
(https://clang.llvm.org/docs/SafeBuffers.html): prefer sized container/view types over raw
`pointer + separate length`.

## Scope / non-goals

- **Behaviour-preserving.** `Span<const uint8_t>` is a transparent `{pointer, length}` view;
  `span[i]` is identical to `ptr[i]` for the same in-range `i`, and `std::array` aggregate
  initialisation is identical to the C array. All existing bounds checks are retained
  unchanged — this patch adds no new runtime check and removes none.
- This does **not** enable `-Wunsafe-buffer-usage` on the file yet: the header dependencies
  (`base/byte_order.h`, `dec_bit_reader.h`, …) still emit the diagnostic, so turning the
  warning into enforcement is a separate, larger effort. This PR removes the file's *own*
  contribution to that surface as an incremental step.

## Testing

- `icc_codec.cc` compiles clean (`-std=c++17`, project flags); with `-Wunsafe-buffer-usage`
  added, the file's own diagnostics drop from **14 to 0**.
- The full `jxl_dec` library builds and links cleanly with the change (the `std::array`
  table migration is also consumed by the encoder side).
- No functional change is intended; the existing `icc_codec_test` / OSS-Fuzz coverage
  exercise the decode paths unchanged. (Reviewers: please confirm against CI, which runs the
  PNG-dependent test matrix I could not build locally.)

### DIFF
--- a/lib/jxl/icc_codec.cc
+++ b/lib/jxl/icc_codec.cc
@@ -12,6 +12,7 @@
 #include <cstdint>
 
 #include "lib/jxl/base/common.h"
+#include "lib/jxl/base/span.h"
 #include "lib/jxl/base/status.h"
 #include "lib/jxl/dec_ans.h"
 #include "lib/jxl/dec_bit_reader.h"
@@ -33,6 +34,7 @@ namespace {
 // no need to skip missing elements as they are past the end of the data).
 Status Shuffle(JxlMemoryManager* memory_manager, uint8_t* data, size_t size,
                size_t width) {
+  const Span<uint8_t> buf(data, size);
   size_t height = (size + width - 1) / width;  // amount of rows of output
   PaddedBytes result(memory_manager);
   JXL_ASSIGN_OR_RETURN(result,
@@ -41,13 +43,13 @@ Status Shuffle(JxlMemoryManager* memory_manager, uint8_t* data, size_t size,
   size_t s = 0;
   size_t j = 0;
   for (size_t i = 0; i < size; i++) {
-    result[i] = data[j];
+    result[i] = buf[j];
     j += height;
     if (j >= size) j = ++s;
   }
 
   for (size_t i = 0; i < size; i++) {
-    data[i] = result[i];
+    buf[i] = result[i];
   }
   return true;
 }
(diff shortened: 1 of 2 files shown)


DarkLycn1976:
@googlebot I signed it!

PR #4877, opened 2026-06-30
Fix MSVC ARM64 build, restrict _sub_overflow_i32 to x86/x64

Snesnopic:
### Description

The MSVC intrinsic `_sub_overflow_i32` is not available on ARM64, causing build failures when compiling with MSVC for Windows ARM64 targets.
This PR restricts its usage to x86/x64 architectures (`_M_AMD64` and `_M_IX86`), allowing ARM64 to fallback to the generic C implementation provided in the else branch. 
The change restores compilation on Windows ARM64 without affecting other platforms.

### Pull Request Checklist

- [x] **CLA Signed**: Have you signed the [Contributor License Agreement](https://code.google.com/legal/individual-cla-v1.0.html) (individual or corporate, as appropriate)? Only contributions from signed contributors can be accepted.
- [x] **Authors**: Have you considered adding your name to the [AUTHORS](AUTHORS) file?
- [x] **Code Style**: Have you ensured your code adheres to the project's coding style guidelines? You can use `./ci.sh lint` for automatic code formatting.

Issue #3511, opened 2024-04-19
32bit Float Density Regression


jonnyawsom3:
**Describe the bug**

Nightly builds result in bpp higher than the uncompressed source. Using 0.10.2 correctly reduces file size.
```
cjxl metro_noord_2k.hdr.pfm metro.hdr.jxl -d 0
JPEG XL encoder v0.10.2 e148959 [AVX2,SSE2]
Encoding [Modular, lossless, effort: 7]
Compressed to 4144.0 kB (15.808 bpp).
2048 x 1024,  0.200 MP/s [0.20, 0.20], , 1 reps, 16 threads.

cjxl metro_noord_2k.hdr.pfm metro.hdr.jxl -d 0
JPEG XL encoder v0.10.2 60dfa1b [AVX2,SSE2]
Encoding [Modular, lossless, effort: 7]
Compressed to 15722.2 kB (59.976 bpp).
2048 x 1024,  0.548 MP/s [0.55, 0.55], , 1 reps, 16 threads.
```

**To Reproduce**
![metro hdr jxl](https://github.com/libjxl/libjxl/assets/51539240/bdec36ce-aa3e-426e-8c5e-64321b8b6fee)
Run `cjxl -d 0 metro.hdr.jxl Test.jxl` on 0.10.2 and nightly

**Environment**
 - OS: Windows
 - CPU type: x86_64
 - cjxl/djxl version string: JPEG XL encoder v0.10.2 60dfa1b [AVX2,SSE2]

mo271:
Thanks for the report. Seems like this was introduced by 27afa5e4c2e66d917746f407674369f19b6fca47
Let's see how to fix that...

mo271:
Perhaps @jonsneyers and @eustas can take a look?

eustas:
Will take a look soon.

jonsneyers:
I assume this issue is caused by this: https://github.com/libjxl/libjxl/commit/27afa5e4c2e66d917746f407674369f19b6fca47

jonnyawsom3:
Likely also reopens #1323

Alb3e3:
Root-caused this. PR: #4955.

The regression is 27afa5e4 (#3450) as @mo271 and @jonsneyers suspected, but the overflow that commit was preventing is not inherent to 32-bit input — it is an encoder/decoder disagreement.

`palette.cc` (decoder) has always clamped the bit depth used to build the **implicit** palette:

```c++
[...]

PR #4948, opened 2026-08-20
modular: wrap residual subtraction for lossless fp32 predictors

### LINKED ISSUE #4902: Lossless fp32 encoding fails when the image mixes negative and positive values
Lossless float32 encoding fails with JXL_ENC_ERR_GENERIC when the image mixes negative and positive values. Likely a regression from 0.11.2.

Minimum example reproducing the issue is an image with following pixel values: `float pixels[2] = { -1e-6f, 0.25f };`, see the full code of minimal example below.

```
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <jxl/encode.h>

int main(void)
{
    const float pixels[2] = { -1e-6f, 0.25f };  /* 2x1 image */
    const uint32_t W = 2, H = 1;

    printf("libjxl runtime version: %u\n", JxlEncoderVersion());

    JxlEncoder *enc = JxlEncoderCreate(NULL);

    JxlBasicInfo info;
    JxlEncoderInitBasicInfo(&info);
    info.xsize                    = W;
    info.ysize                    = H;
    info.bits_per_sample          = 32;
    info.exponent_bits_per_sample = 8;      /* IEEE binary32 */
    info.num_color_channels       = 1;
    info.uses_original_profile    = JXL_TRUE;
    if (JxlEncoderSetBasicInfo(enc, &info) != JXL_ENC_SUCCESS)
        { printf("SetBasicInfo failed\n"); return 2; }

    JxlColorEncoding color;
    JxlColorEncodingSetToLinearSRGB(&color, /*is_gray=*/JXL_TRUE);
    if (JxlEncoderSetColorEncoding(enc, &color) != JXL_ENC_SUCCESS)
        { printf("SetColorEncoding failed\n"); return 2; }

    JxlEncoderFrameSettings *fs = JxlEncoderFrameSettingsCreate(enc, NULL);
    if (JxlEncoderSetFrameLossless(fs, JXL_TRUE) != JXL_ENC_SUCCESS)
        { printf("SetFrameLossless failed\n"); return 2; }

    /* Verified: uncommenting the workaround below makes 0.12.0 succeed.
     * Uncomment to apply the workaround (forces the Zero predictor,
     * whose residuals cannot overflow) — encoding then succeeds:
    JxlEncoderFrameSettingsSetOption(fs, JXL_ENC_FRAME_SETTING_MODULAR_PREDICTOR, 0);
    */


    JxlPixelFormat fmt = { 1, JXL_TYPE_FLOAT, JXL_NATIVE_ENDIAN, 0 };
    if (JxlEncoderAddImageFrame(fs, &fmt, pixels, sizeof pixels)
        != JXL_ENC_SUCCESS)
        { printf("AddImageFrame failed\n"); return 2; }
    JxlEncoderCloseInput(enc);

    uint8_t buf[1 << 16];
    uint8_t *next = buf;
    size_t avail = sizeof buf;
    JxlEncoderStatus st = JxlEncoderProcessOutput(enc, &next, &avail);

    if (st == JXL_ENC_SUCCESS) {
        printf("encode OK (%zu bytes)\n",
               (size_t)(next - buf));
        JxlEncoderDestroy(enc);
        return 0;
    }
    printf("encode FAILED: status=%d, JxlEncoderGetError=%d\n",
           (int)st, (int)JxlEncoderGetError(enc));
    JxlEncoderDestroy(enc);
    return 1;
}
```

**Expected behavior**
Image should be encoded successfully.

**Environment**
 - OS: Windows
 - Compiler version: gcc 16.1.0
 - CPU type: x86_64
 - libjxl runtime version: 12000

**Additional context**
There exists a workaround by forcing zero predictor by "JxlEncoderFrameSettingsSetOption(fs, JXL_ENC_FRAME_SETTING_MODULAR_PREDICTOR, 0);". Works only for effort >= 4, efforts 1-3 still fail to encode.

The failure likely comes from the SubOverflow checks added in lib/jxl/modular/encoding/enc_encoding.cc ("Residual overflow").

Alb3e3:
## Summary
- add a focused regression for lossless float32 grayscale input that mixes negative and positive samples
- compute fast-path modular residuals with defined modulo-2^32 subtraction instead of rejecting wrapped values as overflow
- use the same wrapped subtraction when estimating weighted-predictor cost

## Root cause
Lossless float32 input is encoded through modular int32 channels that carry the raw IEEE-754 bit patterns. For mixed-sign samples, predictor guesses can differ from the source value by more than the signed int32 range while still being valid modulo-2^32 residuals.

The fast encoder paths were using `SubOverflow(...)` and returning `Residual overflow` when that happened, even though the decoder and slower modular paths operate on wrapped 32-bit residuals. That made valid lossless fp32 inputs fail with `JXL_ENC_ERR_GENERIC`.

## Verification
- reproduced the failure first with `EncodeTest.LosslessFloatMixedSignValues` at efforts 1, 4, and 7
- `./tests/encode_test --gtest_filter=EncodeTest.LosslessFloatMixedSignValues`
- `./tests/encode_test --gtest_filter=EncodeTest.FrameEncodingTest:EncodeTest.LosslessFloatMixedSignValues`

## AI disclosure
Prepared with AI assistance, then manually reviewed and verified locally.

Fixes #4902.

### DIFF
--- a/lib/jxl/base/common.h
+++ b/lib/jxl/base/common.h
@@ -18,10 +18,6 @@
 #include <type_traits>
 #include <vector>
 
-#if JXL_COMPILER_MSVC
-#include <intrin.h>
-#endif
-
 #include "lib/jxl/base/compiler_specific.h"
 
 namespace jxl {
@@ -50,21 +46,6 @@ static inline bool SafeMul(size_t a, size_t b, size_t& product) {
   if (b > (std::numeric_limits<size_t>::max() / a)) return false;
   product = a * b;
   return true;
-}
-
-static inline bool SubOverflow(const int32_t a, const int32_t b, int32_t& c) {
-  // Clang 3.8+ / GCC 5.1+
-#if JXL_COMPILER_GCC || JXL_COMPILER_CLANG
-  return __builtin_sub_overflow(a, b, &c);
-#elif JXL_COMPILER_MSVC >= 1937 && (defined(_M_AMD64) || defined(_M_IX86))
-  return _sub_overflow_i32(/*carry*/ 0, a, b, &c);
-#else
-  uint32_t ua = static_cast<uint32_t>(a);
-  uint32_t ub = static_cast<uint32_t>(b);
-  uint32_t uc = ua - ub;
-  c = static_cast<int32_t>(uc);
-  return !!(((ua ^ ub) & (ua ^ uc)) >> 31);
-#endif
 }
 
 template <typename T1, typename T2>
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -250,7 +250,6 @@ float EstimateWPCost(const Image& img, size_t i) {
     const ptrdiff_t onerow = ch.plane.PixelsPerRow();
     weighted::State wp_state(wp_header, ch.w, ch.h);
     Properties properties(1);
-    bool unhealthy = false;
     for (size_t y = 0; y < ch.h; y++) {
       const pixel_type* JXL_RESTRICT r = ch.Row(y);
       for (size_t x = 0; x < ch.w; x++) {
@@ -268,8 +267,7 @@ float EstimateWPCost(const Image& img, size_t i) {
         for (int c : cutoffs) {
           ctx += (c >= properties[0]) ? 1 : 0;
         }
-        pixel_type res;
-        unhealthy |= SubOverflow(r[x], guess, res);
+        const pixel_type res = PixelSub(r[x], guess);
         uint32_t token;
         uint32_t nbits;
         uint32_t bits;
@@ -278,10 +276,6 @@ float EstimateWPCost(const Image& img, size_t i) {
         extra_bits += nbits;
         wp_state.UpdateErrors(r[x], x, y, ch.w);
       }
-    }
-    if (unhealthy) {
-      // Force this predictor option to be rejected by the cost selector.
-      return std::numeric_limits<float>::max();
     }
     for (auto& h : histo) {
       histo_cost += h.ShannonEntropy();
--- a/lib/jxl/encode_test.cc
+++ b/lib/jxl/encode_test.cc
@@ -318,6 +318,7 @@ TEST(EncodeTest, LosslessFloatMixedSignValues) {
 }
 
 
+
 TEST(EncodeTest, EncoderResetTest) {
   JxlEncoderPtr enc = JxlEncoderMake(nullptr);
   EXPECT_NE(nullptr, enc.get());
--- a/lib/jxl/encode_test.cc
+++ b/lib/jxl/encode_test.cc
@@ -260,6 +260,64 @@ TEST(EncodeTest, FrameEncodingTest) {
   VerifyFrameEncoding(enc.get(), frame_settings);
 }
 
+TEST(EncodeTest, LosslessFloatMixedSignValues) {
+  constexpr float pixels[2] = {-1e-6f, 0.25f};
+  JxlPixelFormat pixel_format = {1, JXL_TYPE_FLOAT, JXL_NATIVE_ENDIAN, 0};
+
+  for (int effort : {1, 4, 7}) {
+    SCOPED_TRACE(testing::Message() << "effort=" << effort);
+
+    JxlEncoderPtr enc = JxlEncoderMake(nullptr);
+    ASSERT_NE(nullptr, enc.get());
+
+    JxlBasicInfo basic_info;
+    JxlEncoderInitBasicInfo(&basic_info);
+    basic_info.xsize = 2;
+    basic_info.ysize = 1;
+    basic_info.bits_per_sample = 32;
+    basic_info.exponent_bits_per_sample = 8;
+    basic_info.num_color_channels = 1;
+    basic_info.uses_original_profile = JXL_TRUE;
+    ASSERT_EQ(JXL_ENC_SUCCESS, JxlEncoderSetBasicInfo(enc.get(), &basic_info));
+
+    JxlColorEncoding color_encoding;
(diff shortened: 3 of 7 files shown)


jonnyawsom3:
Not directly related, but any chance you could take a look at #3511 too? It was also caused by an overflow mitigation (https://github.com/libjxl/libjxl/commit/27afa5e4c2e66d917746f407674369f19b6fca47), but in the palette code instead of predictors. If you can figure it out, density could significantly improve.

Alb3e3:
Follow-up for #3511 is pushed in 0607064.

This keeps the earlier predictor fix intact and narrows the palette regression fix to the exact safe case: allow the lossless single-channel palette path for 32-bit inputs again, while leaving the 32-bit bailout in place for the wider implicit/multi-channel palette code that 27afa5e was protecting.

Focused proof:
- red on unpatched `enc_palette.cc`: `ModularTest.Float32SingleChannelPaletteRegression` fails because `FwdPalette(...)` returns false
- green after the change: the same test passes and inserts a palette meta-channel
- `EncodeTest.LosslessFloatMixedSignValues` still passes on this branch

Commands rerun locally after the push:
- `cmake --build /home/beneatgoon2/libjxl-3511-build --target modular_test encode_test -j2`
- `/home/beneatgoon2/libjxl-3511-build/lib/tests/modular_test --gtest_filter='ModularTest.Float32SingleChannelPaletteRegression\`\n- `/home/beneatgoon2/libjxl-3511-build/lib/tests/encode_test --gtest_filter=\EncodeTest.LosslessFloatMixedSignValues'

kleisauke:
I assume the `SubOverflow()` function introduced in commit 87bee190538313b649e9a1eaaa580657ce30ba87 (plus the MSVC-specific `intrin.h` include) can be removed now as well?

FWIW, I wonder if it might make more sense to revert commit 9f72feaf3036d5db7aa80f81ec2c29445681eabf, 87bee190538313b649e9a1eaaa580657ce30ba87 and 27afa5e4c2e66d917746f407674369f19b6fca47 first and then apply your fixes on top of that. That might make the changes easier to review.

Alb3e3:
Yes, good catch — removed in `5c0503bb`. `SubOverflow()` had no callers left after the residual fix, so that also drops the MSVC-only `<intrin.h>` include it needed, and with it the `_M_AMD64 || _M_IX86` guard that #4877 had to add for the MSVC ARM64 build. `SubOverflow`/`_sub_overflow_i32` now appear nowhere under `lib/`.

On the restructure: I'd rather not fold the reverts into this PR. 9f72feaf, 87bee190 and 27afa5e4 each fixed a real problem, and a revert-then-reapply would put the tree through a state where those are unfixed — bad for bisecting, and it makes the diff look like "undo three fixes" rather than "here is what was wrong with them". What's here now is three commits that read in order:

1. `a5107b2e` — wrap the residual subtraction for lossless fp32 predictors (the 87bee190 overflow check was too strict)
2. `06070643` — re-allow the single-channel palette path for 32-bit lossless (the 27afa5e4 bailout was too broad), narrowly, leaving the multi-channel bailout in place
3. `5c0503bb` — drop the now-dead helper

That said, it's your call and I don't feel strongly. If you and the other reviewers would rather see revert-first, say the word and I'll restructure it.

Verified on this branch after the removal:
- `modular_test` — 79/79 pass, including `ModularTest.Float32SingleChannelPaletteRegression`
- `encode_test --gtest_filter=EncodeTest.LosslessFloat*` — `LosslessFloatMixedSignValues` passes

I'll rebase onto current main as well since the branch is showing behind.

Alb3e3:
@jonnyawsom3 Took a look at #3511 — you were right that it is the same class of bug, and the answer turned out to have two halves.

**This PR already fixes most of it.** Moving the `bitdepth >= 32` bail-out below the single-channel channel-palette path re-enables palette for float input. Measured with `cjxl -d 0` on a float32 PFM built the way the reporter's file was (`.hdr` is Radiance RGBE, so `.hdr` → `.pfm` gives float32 samples drawn from a small discrete set):

| | bpp |
| --- | ---: |
| main `aea3a06e` | 43.143 |
| this PR | **8.592** |

**The rest needed a separate root-cause fix**, which is now #4955. The overflow that 27afa5e4 (#3450) was reacting to is not inherent to 32-bit input — it is an encoder/decoder disagreement. `palette.cc` has always clamped `const int bit_depth = std::min(input.bitdepth, 24);`, but `enc_palette.cc` declared the same clamp *below* the implicit-palette loop, which called `GetPaletteValue()` with the raw `input.bitdepth`. At 32 bits that overflows `pixel_type`:

```
palette.h:102:46: runtime error: signed integer overflow:
    536870912 + 2147483647 cannot be represented in type 'int'
    #0 GetPaletteValue palette.h:102
    #1 jxl::FwdPaletteIteration(...) enc_palette.cc:351
```

(gcc `-fsanitize=signed-integer-overflow -fno-sanitize-recover=all`, reproduced by removing only the bail-out — so a naive "just delete the line" fix really would reintroduce #3450.)

Passing the clamped `bit_depth` makes the encoder agree with the decoder and keeps the arithmetic in range, so the bail-out can go entirely and the **multi-channel** palette works for float too. That is where the remaining gain is:

| input | main | this PR | +#4955 |
| --- | ---: | ---: | ---: |
| RGBE-derived HDR | 43.143 | 8.592 | 8.592 |
| 16 distinct RGB colors | 13.008 | 1.472 | **0.542** |
| 64 distinct RGB colors | 40.243 | 7.920 | **2.652** |
| 256 distinct RGB colors | 69.245 | 23.103 | **7.923** |

All bit-exact lossless on `djxl` round-trip. #4955 is based on `main` and is **independent of this PR** — I built it on a clean `main` worktree with none of the predictor commits here and got identical numbers, so the two can land in either order.

On #1323: plausible that it helps there too, but I have not measured that file, so I have not claimed it.

Issue #4947, opened 2026-08-19
Possible integer overflow in ‎TreeSamples::PreQuantizeProperties


kleisauke:
Hello, the following PFM file, discovered via fuzz testing, triggers an integer overflow when encoded losslessly to JXL (effort level 9) using the latest code on the `main` branch.

**OSS-Fuzz reproducer:**

Minimized testcase: [clusterfuzz-testcase-minimized-vips_fuzzer-5643351927095296.txt](https://github.com/user-attachments/files/31217053/clusterfuzz-testcase-minimized-vips_fuzzer-5643351927095296.txt)

Steps to reproduce via OSS-Fuzz:
```console
$ git clone --depth=1 https://github.com/google/oss-fuzz.git
$ cd oss-fuzz
$ python3 infra/helper.py build_image libvips
$ python3 infra/helper.py build_fuzzers --sanitizer undefined libvips
$ python3 infra/helper.py reproduce libvips vips_fuzzer ~/Downloads/clusterfuzz-testcase-minimized-vips_fuzzer-5643351927095296.txt
```

**Standalone reproducer:**

Test program: https://gist.github.com/kleisauke/510eadcb8a84f613e8a1f3318bfa4cbf

Minimized testcase: https://gist.github.com/kleisauke/510eadcb8a84f613e8a1f3318bfa4cbf/raw/clusterfuzz-testcase-minimized-vips_fuzzer-5643351927095296.raw.txt

```console
$ git clone https://github.com/libjxl/libjxl.git
$ cd libjxl
$ export CC="clang"
$ export CXX="clang++"
$ export CPPFLAGS="-g -fsanitize=signed-integer-overflow"
$ cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=$HOME/libjxl-ub -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=OFF -DJPEGXL_ENABLE_{FUZZERS,SJPEG,SKCMS}=OFF -DJPEGXL_FORCE_SYSTEM_{BROTLI,HWY,LCMS2}=ON -DCMAKE_C{,XX}_FLAGS="$CPPFLAGS"
$ cmake --build build -- -j$(nproc)
$ cmake --install build
$ export PKG_CONFIG_PATH=$HOME/libjxl-ub/lib64/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}
[...]

PR #4953, opened 2026-08-21
Fix absolute-property overflow at integer minimum

carrerasdarren-cell:
### Description

Avoid undefined behavior when modular tree pre-quantization takes the absolute value of the minimum signed pixel value. The new helper saturates that one unrepresentable magnitude to the maximum signed value, keeping absolute-property samples nonnegative for both pixel and difference properties.

A regression test exercises both sample vectors with the minimum `pixel_type` value.

### Testing

- `modular_test` (79 tests)
- Focused `ModularTest.PreQuantizeAbsolutePropertiesAtIntegerMinimum` regression
- `git diff --check`

### Pull Request Checklist

- [ ] **CLA Signed**: Automated CLA status pending.
- [x] **Authors**: Considered; no `AUTHORS` change is needed for this scoped fix.
- [x] **Code Style**: Changed ranges are clang-formatted and the diff passes whitespace checks.

### DIFF
--- a/lib/jxl/modular/encoding/enc_ma.cc
+++ b/lib/jxl/modular/encoding/enc_ma.cc
@@ -799,6 +799,12 @@ std::vector<int32_t> QuantizeSamples(const std::vector<int32_t> &samples,
   return thresholds;
 }
 
+pixel_type SaturatingAbs(pixel_type value) {
+  return value == std::numeric_limits<pixel_type>::min()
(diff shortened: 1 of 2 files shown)


carrerasdarren-cell:
Closing in favor of #4950, which already contains the same fix and regression coverage for #4947. This duplicate was opened during a continuity handoff; keeping the original PR as the single review target.

PR #4955, opened 2026-08-25
modular: palettize 32-bit input again, clamping the implicit palette

### LINKED ISSUE #3511: 32bit Float Density Regression

Alb3e3:
Fixes #3511.

## What regressed

`cjxl -d 0` on float32 input got much worse after 27afa5e4c2e66d917746f407674369f19b6fca47 ("Prevent integer overflow", #3450), which added:

```c++
if (input.bitdepth >= 32) return false;
```

to `FwdPaletteIteration()`. That refuses the palette transform for *any* 32-bit input, so lossless HDR/float images lose palette entirely.

## Why the overflow was there

The bail-out treated the symptom. The actual problem is that the encoder and the decoder disagreed about the bit depth used to build the **implicit** palette.

`palette.cc` (decoder) has always clamped:

```c++
const int bit_depth = std::min(input.bitdepth, 24);
```

`enc_palette.cc` declared the same clamped `bit_depth`, but *below* the implicit-palette construction, which called `GetPaletteValue()` with the raw `input.bitdepth`:

```c++
color[i] = palette_internal::GetPaletteValue(nullptr, k, i, 0, 0,
                                             input.bitdepth);
```

At 32 bits that overflows `pixel_type` inside `GetPaletteValue()`. `Scale<kSmallCube>(2, 32)` is `(2 * (2^32 - 1)) >> 2` = `2147483647`, and `palette.h:102` then adds `1 << (32 - 3)`:

```
lib/jxl/modular/transform/palette.h:102:46: runtime error: signed integer overflow:
    536870912 + 2147483647 cannot be represented in type 'int'
    #0 GetPaletteValue palette.h:102
    #1 jxl::FwdPaletteIteration(...) enc_palette.cc:351
```

(gcc `-fsanitize=signed-integer-overflow -fno-sanitize-recover=all`, reproduced by removing only the bail-out.)

## The change

* Hoist `const int bit_depth = std::min(input.bitdepth, 24);` above the implicit-palette loop and use it there, so the encoder computes the same implicit palette the decoder does and the arithmetic stays in range.
* Drop the `input.bitdepth >= 32` bail-out.
* Widen the lossy clamp from `1l << input.bitdepth` to `int64_t{1} << input.bitdepth` — `long` is 32-bit on Windows, so that shift is UB at 32 bits once this path becomes reachable again.

## Density

`cjxl -d 0`, float32 PFM input, all four **bit-exact lossless** on `djxl` round-trip (compared sample-by-sample after normalising PFM endianness):

| input | before | after |
| --- | ---: | ---: |
| RGBE-derived HDR, 512x256 | 43.143 bpp | **8.592 bpp** |
| 16 distinct RGB colors, 256x256 | 13.008 bpp | **0.542 bpp** |
| 64 distinct RGB colors, 256x256 | 40.243 bpp | **2.652 bpp** |
| 256 distinct RGB colors, 256x256 | 69.245 bpp | **7.923 bpp** |

The first case mirrors the reporter's `metro_noord_2k.hdr.pfm`: `.hdr` is Radiance RGBE, so converting to `.pfm` yields float32 samples drawn from a small discrete set, which is exactly what palette exploits.

## Tests

New `ModularTest.Float32MultiChannelPaletteRegression` drives `FwdPalette()` on a 3-channel 32-bit image built from 8 distinct HDR colors and asserts the transform happens and every emitted index addresses a real palette entry. Without the fix it fails on the untransformed data:

```
Expected: (static_cast<uint32_t>(row[x])) < (nb_colors), actual: 1094713344 vs 1024
```

Suites run locally on this branch, all green: `modular_test` **79/79**, `codec_test` **8/8**, `decode_test` **2891/2891**, `encode_test` **2330/2330**, `jxl_test` **150/150** — 5458 tests, 0 failures. `cjxl` on all four images above is UBSan-clean with the fix, and trips the overflow above without it.

## AI disclosure

Prepared with AI assistance, then reviewed and verified locally; every number above is reproducible from the commands in this description.

### DIFF
--- a/lib/jxl/modular/transform/enc_palette.cc
+++ b/lib/jxl/modular/transform/enc_palette.cc
@@ -186,7 +186,6 @@ Status FwdPaletteIteration(Image &input, uint32_t begin_c, uint32_t end_c,
 
   size_t w = input.channel[begin_c].w;
   size_t h = input.channel[begin_c].h;
-  if (input.bitdepth >= 32) return false;
   if (!lossy && nb_colors < 2) return false;
 
   if (!lossy && nb == 1) {
@@ -339,13 +338,18 @@ Status FwdPaletteIteration(Image &input, uint32_t begin_c, uint32_t end_c,
     }
   }
 
+  // Must match the decoder, which clamps to 24 bits (see palette.cc).  Passing
+  // input.bitdepth unclamped disagrees with the decoder and overflows
+  // pixel_type inside GetPaletteValue() once the bit depth reaches 32.
+  const int bit_depth = std::min(input.bitdepth, 24);
+
   std::map<std::vector<pixel_type>, bool> implicit_color;
   std::vector<std::vector<pixel_type>> implicit_colors;
   implicit_colors.reserve(palette_internal::kImplicitPaletteSize);
   for (size_t k = 0; k < palette_internal::kImplicitPaletteSize; k++) {
     for (size_t i = 0; i < nb; i++) {
-      color[i] = palette_internal::GetPaletteValue(nullptr, k, i, 0, 0,
-                                                   input.bitdepth);
+      color[i] =
+          palette_internal::GetPaletteValue(nullptr, k, i, 0, 0, bit_depth);
     }
     implicit_color[color] = true;
     implicit_colors.push_back(color);
@@ -407,7 +411,6 @@ Status FwdPaletteIteration(Image &input, uint32_t begin_c, uint32_t end_c,
   pixel_type *JXL_RESTRICT p_palette = pch.Row(0);
   ptrdiff_t onerow = pch.plane.PixelsPerRow();
   ptrdiff_t onerow_image = input.channel[begin_c].plane.PixelsPerRow();
-  const int bit_depth = std::min(input.bitdepth, 24);
 
   if (lossy) {
     for (uint32_t i = 0; i < nb_deltas; i++) {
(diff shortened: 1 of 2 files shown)


Alb3e3:
Flagging something that affects a lot more than this PR: no outside-contributor PR in the queue is running CI right now.

Of the 16 open PRs, 13 have only `cla/google` plus the readthedocs build and no build/test at all. The three that do run the full 119-check matrix are all @eustas's. The oldest affected one is #4937, open since 6 August.

That looks like the "approve workflows for outside collaborators" gate rather than anything wrong with the PRs themselves. The practical effect is that a reviewer has no signal to go on, so the queue stalls regardless of patch quality.

If someone with the bit can approve the pending workflow runs, it would unblock the whole set in one go rather than PR by PR.

For this one specifically: it is rebased on current main and independent of #4948, so it can be reviewed and landed on its own. #4948 already recovers most of #3511 by itself; this removes the remaining cause, which is the encoder building the implicit palette with the raw bit depth while the decoder has always clamped to 24.

PR #4958, opened 2026-08-26
Use latest gcc instead of default one

eustas:
Hopefully fixed hwy 1.2+ build


eustas:
@QrczakMK friendly ping

QrczakMK (approved):

PR #4959, opened 2026-08-27
ans: read AliasTable Entry from its base, not a member pointer

### LINKED ISSUE #4951: Intra-object overflow on `memcpy` at `/src/libjxl/lib/jxl/ans_common.h:110:5`
**Description**
A memcpy uses a pointer to the `cutoff` field of `struct Entry` as source pointer in file `ans_common.h` , line `110:5`.

Faulty line:
`memcpy(&entry, &table[i].cutoff, sizeof(entry)); `
Definition of struct Entry:
```
#pragma pack(push, 1)
  struct Entry {
    uint8_t cutoff;       // < kEntrySizeMinus1 when used by ANS.
    uint8_t right_value;  // < alphabet size.
    uint16_t freq0;

    // Only used if `greater` (see Lookup)
    uint16_t offsets1;         // <= ANS_TAB_SIZE
    uint16_t freq1_xor_freq0;  // for branchless ternary in Lookup
  };
```

**To Reproduce**
I found the issue fuzzing `imagemagick`, that depends on `libjxl`.

Run the attached input against the `encoeder_jxl_fuzzer` in `imagemagick` and break at ans_common.h:110 to verify that the memcpy overreads `cutoff`. 
[jxl.zip](https://github.com/user-attachments/files/31275783/jxl.zip)
imagemagick commit hash : 8d1fdcb46b8a29ad1297b0925c915fd8efcb1a60
If you build KImageFormats with ASan (for instance), you can reproduce the issue running against the attached input.

**Expected behavior**
The code is copying a struct Entry into another. If this is the expected behavior, the pattern is probably safe because the struct is packed and `cutoff` is the **first** element. Nonetheless, if the position of cutoff in the declaration ever changed, the code would break. 

**Environment**

OS: Ubuntu 20.04, oss-fuzz base builder image
Compiler version:clang 21.1.0, with modifications to implement intra-object overflow sanitization
CPU type: x86_64
libjxl commit hash d089091afeb7b00b3d0fec6f019d35eaa3b2b410

**Additional context**
It seems a safe pattern, but I am obliged to report any findings of my tool. I understand how frustrating it might be, still this is policy for academic researchers. Thanks for your time.

Alb3e3:
Fixes #4951.

`AliasTable::Lookup()` (`lib/jxl/ans_common.h`) reads a whole packed `Entry` as a single `uint64_t`:

```cpp
uint64_t entry;
memcpy(&entry, &table[i].cutoff, sizeof(entry));
```

`&table[i].cutoff` points at the 1-byte first member, so copying `sizeof(entry)` (8) bytes from it is an intra-object out-of-bounds access. It is undefined behavior and is flagged by sanitizers (the referent object is the `cutoff` subobject, not the whole `Entry`).

The intent is to load the entire `Entry`. `cutoff` is at offset 0 and `sizeof(Entry) == 8 == sizeof(uint64_t)`, so copying from `&table[i]` reads exactly the same bytes while giving the copy correct provenance over the whole object. The single-load optimization is preserved.

I confirmed the copy is byte-identical (same start address, same length) and added a `static_assert(sizeof(Entry) == sizeof(entry))` to document the layout the load relies on, so any future change to `Entry` that breaks the assumption fails at compile time.

### DIFF
--- a/lib/jxl/ans_common.h
+++ b/lib/jxl/ans_common.h
@@ -107,7 +107,12 @@ struct AliasTable {
 
 #if JXL_BYTE_ORDER_LITTLE
     uint64_t entry;
-    memcpy(&entry, &table[i].cutoff, sizeof(entry));
+    // Read the whole Entry, not a pointer into its first member: copying
+    // sizeof(entry) bytes from &table[i].cutoff (a 1-byte subobject) is an
+    // intra-object out-of-bounds access. table[i] is exactly this size.
+    static_assert(sizeof(Entry) == sizeof(entry),
+                  "Lookup reads an Entry as a single uint64_t load");
+    memcpy(&entry, &table[i], sizeof(entry));
     const size_t cutoff = entry & 0xFF;              // = MOVZX
     const size_t right_value = (entry >> 8) & 0xFF;  // = MOVZX
     const size_t freq0 = (entry >> 16) & 0xFFFF;


Alb3e3:
Closing this as a duplicate. #4952 makes the same change and was opened a week earlier, so it should be the one that lands. Apologies for the noise, I should have checked the open queue before filing.

For whoever picks up #4952: the change is correct. `&table[i].cutoff` names the one-byte first member, so an eight-byte `memcpy` from that pointer is an intra-object out-of-bounds read even though the bytes are within the containing `Entry`. `&table[i]` is the right source.

Issue #3720, opened 2024-07-23
Not better than WEBP, sometimes worse than PNG


Jellby:
**Describe the bug**
Losslessly compressing some PNGs does not beat WEBP. Lossy compression with med-high quality gets even worse than PNG.

I don't think it's really a bug, but maybe something can be improved?

**To Reproduce**
Just compressing all files in a directory with `cjxl -d 0 -e 9 $i $(basename $i png)jxl` results in:
```
30M     jxl
34M     png
26M     webp
```
(the WEBP files are lossless compressions too)

`cjxl -d 1 $i $(basename $i png)jxl` gives:
```
67M     jxl
34M     png
26M     webp
```

and only `cjxl -q 20 $i $(basename $i png)jxl` gets down to:
```
25M     jxl
34M     png
26M     webp
```

**Expected behavior**
I hoped all would be smaller than WEBP :)

**Screenshots**
Of course, it all depends on what the files look like. They're all grayscale comic strips like this:
![kk_160423](https://github.com/user-attachments/assets/81f74775-3347-4bd6-b937-1ee834d99fcf)

**Environment**
 - OS: Kubuntu 22.04
 - Compiler version: (installed from the .deb packages here)
 - CPU type: x86_64
 - cjxl/djxl version string: cjxl v0.10.3 0.10.3 [AVX2,SSE4,SSE2]

monad0:
Grayscale or just 2-color/1-bit? That is one rare situation where PNG is still competitive. Passing the bitdepth may help guide cjxl if it doesn't figure it out. Can also try `-P 0 -I 0 -g 3 --patches 0`. But you can expect WebP encoding to be more mature for this content.

yota-code:
For the picture showed, it is a grayscale file with mainly black and white pixels and some (but not all) grey values, the histogram (logarithmic) is the following:

![hist](https://github.com/user-attachments/assets/de486ef5-8551-4599-a5aa-f4090249eac8)

webp is still very competitive on palettized colors. The good news is that the author of the lossless webp is also an author of jxl, the bad news is that it will take some more times before they port all the tricks used

yota-code:
I don't know how you tried, but I get a better result with cjxl for the image you shared:

* `cjxl -d 0 -e 10 -I 100` gives 487kB
* `cwebp -lossless -z 9` gives 674kB

Jellby:
[...]

PR #3420, opened 2024-03-13
tweak palette heuristics

jonsneyers:
Some tweaks to the modular palette heuristics:

- Refactored some code (there was some near-duplicate code between global and group encoding)
- Change the max_colors heuristic to make it less likely to use palette when the image/group has low entropy already anyway (e.g. the group is only a smooth gradient, which has few enough colors for doing a palette but also palette doesn't help)
- Actually use the implicit palette if these colors happen to occur in the image (but still make them explicit if they are frequent)
- Tweak the palette ordering: still luma sorted, but first the 'common' colors and then the 'rare' colors
- Don't do palette if the whole group is just a single color (it's slightly cheaper not to)

# Results

## QOI test set

Before:
benchmark_xl v0.10.0 453f1127 [NEON]
12 total threads, 17088 tasks, 6 threads, 2 inner threads
```
Encoding      kPixels    Bytes          BPP  E MP/s  D MP/s     Max norm  SSIMULACRA2   PSNR        pnorm       BPP*pnorm   QABPP   Bugs
----------------------------------------------------------------------------------------------------------------------------------------
jxl:d0:4      1321975 813352177    4.9220412   3.663  18.466          nan 100.00000000  99.99   0.00000000  0.000000000000   4.922      0
jxl:d0:5      1321975 786705494    4.7607875   2.556  18.754          nan 100.00000000  99.22   0.00000000  0.000000000000   4.761      0
jxl:d0:6      1321975 774050078    4.6842026   1.801  17.320          nan 100.00000000  99.99   0.00000000  0.000000000000   4.684      0
jxl:d0:7      1321975 747782641    4.5252439   1.236  16.101          nan 100.00000000  99.24   0.00000000  0.000000000000   4.525      0
jxl:d0:8      1321975 732645434    4.4336403   0.403  15.206          nan 100.00000000  99.19   0.00000000  0.000000000000   4.434      0
jxl:d0:9      1321975 724420391    4.3838661   0.212  14.877          nan 100.00000000  99.09   0.00000000  0.000000000000   4.384      0
Aggregate:    1321975 762523609    4.6144496   1.101  16.720   0.00000000 100.00000000  99.45   0.00000000  0.000000000000   4.614      0
```

After:
benchmark_xl v0.10.0 8c7f3833 [NEON]
12 total threads, 17088 tasks, 6 threads, 2 inner threads
```
[...]

PR #4970, opened 2026-09-08
Respect user palette and channel color flags

Galaxy4594:
### Description

Fixes regressions introduced in #3420, preserving option overrides for paletted images:

Previously, user-specified options for palette colors and channel color percentages (`--modular_palette_colors`, `-X` / `--pre-compact`, and `-Y` / `--post-compact`) were not consistently respected:

- Heuristic Clamping in `try_palettes`: In `lib/jxl/enc_modular.cc`, `palette_colors` was unconditionally clamped by an entropy/dimension heuristic (`std::min(cost_before * 0.0005f + nb_pixels / 128 + 128, ...)`). This prevented users and benchmark scripts from forcing or experimenting with larger palette sizes on smaller or lower-entropy images.

- Hardcoded Predictor in Lossy Delta Palette:
   PR #627 originally introduced the ability to select the predictor used for lossy delta palettes. When PR #3420 refactored palette code into `try_palettes()`, it hardcoded `Predictor::Average4`, ignoring any requested predictor. The requested predictor in `lossy_palette_predictor` is preserved before setting the palette index tree to predictor zero.

~~- Override in `kTectonicPlate` (Efforts 10/11): In `lib/jxl/enc_frame.cc`, the candidate presets generated by `TectonicPlateSettingsLessPalette` and `TectonicPlateSettingsMorePalette` (as well as the initial test trial) hardcoded `palette_colors` (to `0`, `1024`, or `70000`) and channel color percentages, completely discarding any flags set.~~

Default encoder behavior is preserved when options are not set by the CLI or API.

### Test Case
Image from: https://github.com/libjxl/libjxl/issues/3720#issuecomment-4761391791
| CLI Settings | Main | PR | Δ (%) |
| --- | --- | --- | --- |
| `-d 0 -e 9` | 182.9 kB | 182.9 kB | = |
| `-d 0 -e 9 --modular_palette_colors=60000` | 182.9 kB | 159.5 kB | **-12.794%** |

### DIFF
--- a/lib/jxl/enc_modular.cc
+++ b/lib/jxl/enc_modular.cc
@@ -333,6 +333,7 @@ StatusOr<bool> maybe_do_transform(Image& image, const Transform& tr,
 Status try_palettes(Image& gi, int& max_bitdepth, int& maxval,
                     const CompressParams& cparams_,
                     float channel_colors_percent,
+                    bool custom_channel_colors = false,
                     jxl::ThreadPool* pool = nullptr) {
   float cost_before = 0.f;
   size_t did_palette = 0;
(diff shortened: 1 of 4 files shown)


monad0:
Conceptually, overriding Tectonic Plate parameters with user settings is nonsensical. Those combinations were proven useful and complementary, and the test combinations highly discriminatory for accessing the viable set. Overriding violates all assertions.

Galaxy4594:
Yeah, thinking about it, you're right. Overwriting those flags ruins the trials in effort 11, which defeats the search and hurts compression. I'll drop the `enc_frame.cc` changes so effort 11 keeps its intended behavior.

Regarding UX, do you think it's worth having the CLI notify the user when effort 11 silently discards manual flags (e.g., --modular_palette_colors, -Y, etc.) or falls back to effort 10? When flags are silently overridden, it's easy to assume the option is either broken or unsupported without diving into the source code to verify why. Giving some visibility in the CLI when explicit user flags are bypassed would save some guesswork when benchmarking.

Something like `Warning: Effort 11 is unsupported for VarDCT, falling back to effort 10` or `Note: Effort 11 uses a multi-trial search and ignores manual tuning flags (palette, predictors, group size, etc.).`

monad0:
> Regarding UX, do you think it's worth having the CLI notify the user [...] When flags are silently overridden, it's easy to assume the option is either broken or unsupported without diving into the source code to verify why.

It's fair to notify where practical. I don't believe it's useful for the average user. Even if this information is exposed, there is still a lot of mystery regarding encoder behavior. Experimental user choices may have no effect whatsoever, and the same assumptions about broken options could still be made. Rather than printing on encode, I believe it more user-friendly to clarify any important details in the help printout. If a setting is only a suggestion, it should be made clear there.

I believe the particular implementation details of e11 (as with any effort) are less interesting and more noisy. The claim that e11 or e1 are not supported for VarDCT is misleading when they work perfectly well. More accurately, e1 is the same as e2. Only nerds need to know that, and the code is the only law. That all said, I support any effort to meaningfully improve documentation.

monad0:
Btw, carrying around these "custom value" booleans feels heavy and tedious. I would expect the parameters to have an unset value, and where the encoder would pick a value, it can check if the value is unset before choosing its own.

PR #4979, opened 2026-09-16
Fix intra-object overflow in ANS table Lookup memcpy

### LINKED ISSUE #4951: Intra-object overflow on `memcpy` at `/src/libjxl/lib/jxl/ans_common.h:110:5`

iliasabk:
Fixes #4951.

`Symbol::Lookup` in `lib/jxl/ans_common.h` bulk-loads an ANS table
`Entry` via `memcpy(&entry, &table[i].cutoff, sizeof(entry))`. The
source pointer is derived from the `cutoff` field, so the copy reads
past that field's own extent — flagged by intra-object overflow
instrumentation (clang with the intra-object overflow sanitizer).

Since `cutoff` is the first member of the packed `Entry` struct,
`&table[i].cutoff` and `&table[i]` are the same address. Sourcing the
copy from the struct pointer yields identical bytes with no
intra-object overread and stays correct even if the field order ever
changes.


iliasabk:
Friendly ping — could a maintainer approve the CI run for this PR when convenient? The fix addresses the linked OSS-Fuzz issue; happy to adjust anything on review feedback. Thanks!

Issue #4848, opened 2026-06-17
cjxl: Automatically discard alpha channel if it's fully opaque


jakearchibald:
Right now, if you encode a PNG-with-alpha image with cjxl, it produces a JXL file with an alpha channel even if the alpha channel is fully opaque.

The presence of an alpha channel in the final image has a large negative impact on progressive rendering (progressive_dc is disabled for images that have an alpha channel).

In order to produce better-optimised output, cjxl should detect if the alpha channel is unnecessary, and discard it.

jonnyawsom3:
Fixed by #4981

Issue #4844, opened 2026-06-14
Feature request: a way to disable the alpha channel in the output image


SamusAranX:
It'd be great if there was an option to skip the creation of an alpha channel in any files cjxl generates.
Apps like Lightroom Classic sometimes generate images with fully opaque alpha channels with no way to *not* have those, and encoding those further just wastes space.

(Filing this as an issue as per [a suggestion on the Discord](https://discord.com/channels/794206087879852103/794206170445119489/1515575140828708934).)

jonnyawsom3:
Fixed by #4981

PR #4981, opened 2026-09-18
Ignore empty alpha channels

Galaxy4594:
Many images contain an unnecessary empty alpha channel, meaning it's fully opaque across every pixel. Currently, encoding these images creates an extra channel in the JPEG XL bitstream, which adds a notable bitrate and encode/decode cost.

This change introduces automatic detection and stripping of empty (fully opaque) alpha channels in libjxl. Prior to writing codestream headers, the encoder inspects the alpha buffer across all supported bit depths and data formats. When the alpha channel is identified as entirely opaque, the encoder strips the channel from the metadata and frame header before emission, avoiding unnecessary extra channel encoding.

You can control this new behavior via (JXL_ENC_FRAME_SETTING_STRIP_ALPHA) and via `--strip_alpha` in the CLI. By default, the encoder automatically optimizes lossy compression while strictly respecting transparency and lossless workflows:

-1 (Default): The encoder chooses—empty (opaque) alpha is stripped during lossy encoding, while alpha is always preserved during lossless encoding.
0: Always keep alpha, preserving the channel regardless of whether it is opaque or lossy.
1: Strip alpha if empty (fully opaque), applying the optimization to both lossy and lossless modes.
2: Always strip alpha, completely dropping the channel even if semi-transparent pixels exist.

To support JXLs with containers, the alpha input inspection resolves the first image frame rather than assuming the head of the queue is a frame, ensuring metadata boxes (such as Exif or XMP) preceding frames do not interfere with detection. For fast lossless paths (CanDoFastLossless), requests to strip alpha gracefully fall back to the standard encoder. In my testing it also handles chunked frame input and streaming encoding without requiring an intermediate de-interleaving stage.

Tests have been added under EncodeTest.StripAlphaSetting, verifying all four setting modes (-1, 0, 1, 2) across lossy and lossless modes, semi-transparent inputs, container metadata box ordering, and chunked input sources.


Galaxy4594:
Seems to pass all tests, mingw32 fails for a different reason. It tried to download a package (mingw-w64-i686-gtest) that's been [dropped](https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages).

veluca93 (approved):

jonnyawsom3:
Fixes #4848
Fixes #4844

jakearchibald:
Yay! Thanks for this

PR #4984, opened 2026-09-19
Align strip_alpha with boolean expectations and improve float checks

Galaxy4594:
Follow-up to #4981

Aligns `--strip_alpha` with boolean logic:
   - Swapped modes 1 and 2 to map to boolean logic, allows for future special modes.
     - `-1`: Auto
     - `0`: Never strip
     - `1`: Always strip
     - `2`: Strip if empty (special mode 1).
     - `3`: Hypothetical special mode 2, etc.

Also this PR has better opacity detection. `IsAlphaBufferOpaque` now supports `UINT8`, `UINT16`, `FLOAT` (exact `1.0f`), and `FLOAT16`. There are now various buffer validity, channel count, and bytes-per-pixel safety guards. It also handles endianness when needed. It verifies channel initialization before checking chunk buffers. There are more tests mostly for float32 alpha and interleaved RGBA buffers passed without declaring extra channels in `basic_info`. There is better documentation in `jxl/encode.h`, and both the CLI and API now reject values outside `[-1..2]`.


veluca93 (dismissed):

Galaxy4594:
Sorry, had some last minute linting changes. Also the new typedef enum was not needed, just some AI fluff I had to remove. 😅

Galaxy4594:
Okay, I should not make such changes late at night. Kind of odd how Windows build pass but MacOS and Ubuntu don't.

Galaxy4594:
Okay, it worked. It's a bit strange that Windows builds don't have `-DJPEGXL_WARNINGS_AS_ERRORS=true` unlike Linux and MacOS builds.

veluca93 (approved):

